PR #7, opened 2017-05-31
Layout & Nav (FAD-4964)

jonambas:
- Adds layout and nav components
- Adds enzyme

Left out icons for now. Eventually i want to have an icon library included in matchbox - FA isnt really cutting it.

Also commented out old webui styles - will style that sign up form whenever form inputs and buttons are done in matchbox


src/components/AuthenticationGate.js
```diff
@@ -35,11 +35,12 @@ export class _AuthenticationGate extends Component {
 
   render () {
     const { loggedIn } = this.props.auth;
```
jasonrhodes: Can remove this line too

src/components/Navigation/Item.js
```diff
@@ -0,0 +1,85 @@
...
+  render () {
+    const {
+      to,
+      icon,
+      label,
+      location,
+      children
+    } = this.props;
+
+    const active = location.pathname.includes(to); // Not sure if this is the best way to do this
```
jasonrhodes: We will probably eventually move to something more reduxy so maybe this is fine for now until we get there?

src/pages/AuthPage.js
```diff
@@ -39,71 +42,46 @@ class AuthPage extends Component {
...
-
-                  </form>
-
-                </div>
+      <Layout.Form>
+        <a href="https://www.sparkpost.com" title="SparkPost">
+          <img alt="SparkPost" height="68" src="/assets/images/sparkpost-logo-color.svg" width="188" />
+        </a>
+
+        <Panel sectioned accent>
```
jasonrhodes: 👏 👏 👏

jasonrhodes:
lgtm -- can think about doing icons with SVG directly instead of an icon font, i've heard Good Things about doing it that way anyway and i'm all about not loading all of FontAwesome

PR #13, opened 2017-06-27
Templates List

jgzamora:


src/pages/TemplatesPage.js
```diff
@@ -0,0 +1,79 @@
...
+              { templateRows }
+            </tbody>
+          </Table>
+        </Panel>
+        <Pagination
+          pages={Math.floor(templateCount / this.state.perPage)}
+          pageRange={5}
+          onChange={(index) => { this.setState({currentPage: index}); }}
+        />
+        <Button.Group>
```
jgzamora: not sure how to do selected styling

jonambas: You can pass in additional classNames or style into button, you can do something like:
```
className={ this.state.perPage === 10 ? 'Matchbox-Pagination__Selected' : '' }
```
but this looks kinda messy

jonambas: or if you import classnames
```
className={classnames(this.state.perPage === 10 && 'Matchbox-Pagination__Selected')}
```

jonambas: Now that i'm thinking about it, we might want to build all this in a reusable component so we can use this stuff elsewhere

src/pages/TemplatesPage.js
```diff
@@ -0,0 +1,91 @@
...
+    return (
+      <Table.Row key={template.id} rowData={ [nameLink, template.id, status, Date(template.last_update_time)]} >
+
+      </Table.Row>
+    );
+  }
+
+  renderTemplateRows (templates, currentPage, perPage) {
+    const currentIndex = currentPage * perPage;
+    return templates.slice(currentIndex, currentIndex + perPage).map(
```
jasonrhodes: This is what you were talking about yesterday I think, and looking at it, it looks :+1: to me. I do want to think about one more level of abstraction for commonly paginated things we use but we'll get there, this is great for now.

src/pages/TemplatesPage.js
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  componentDidMount () {
+    this.props.listTemplates();
+  }
+
+  render () {
+    const templatesCount = this.props.templates.length;
+
+    if (!templatesCount) {
```
jasonrhodes: I think we'll want to distinguish between "templates are currently loading" and "I have no templates yet" eventually, and have a loading state and an empty state that are different. For now I would handle that manually in the templates reducer (I'll comment there)

src/reducers/templates.js
```diff
@@ -0,0 +1,11 @@
...
+
+export default (state = initialState, action) => {
+  switch (action.type) {
+    case 'LIST_TEMPLATES_SUCCESS':
+      return { ...state, list: action.payload };
+
+    default:
+      return state;
+  }
+};
```
jasonrhodes: For loading (for now) I'm thinking something like:
```js
case 'LIST_TEMPLATES_PENDING':
  return { ...state, listLoading: true };

case 'LIST_TEMPLATES_SUCCESS':
  return { ...state, list: action.payload, listLoading: false };

case 'LIST_TEMPLATES_FAIL':
  return { ...state, listLoading: false };
```
and later we'll figure out a more global way to handle that.

Note: we also need to figure out a better, somewhat global way of handling API errors and some kind of alerts, but we can punt on that for right now.

PR #22, opened 2017-07-20
Responsive Nav

jonambas:
- Makes side navigation responsive
- We now have an empty top bar on smaller screens. Have room to put in the SP logo or other actions.
- Also scales root font-size at breakpoints. All the MB components use REMs for sizing so we can scale the UI proportionally whenever we want.

kapture 2017-07-20 at 14 40 39

kapture 2017-07-20 at 14 43 42


jasonrhodes:
This looks A+++

PR #23, opened 2017-07-20
Webhooks create

jgzamora:


src/pages/Webhooks/HomePage.js
```diff
@@ -97,7 +99,7 @@ class WebhooksHome extends Component {
 
 function mapStateToProps ({ webhooks }) {
   return {
-    webhooks: webhooks.list,
+    webhooks: sortBy(webhooks.list, ['name']),
```
jgzamora: sort could be moved to the reducer

src/pages/Webhooks/HomePage.js
```diff
@@ -55,13 +56,14 @@ class WebhooksHome extends Component {
       );
     }
 
+    // TODO: fix paging error
```
jgzamora: Error is: When switching to a different `perPage` while `currentPage` is beyond the number of pages that the new `perPage` would render, the page goes blank and no current page is selected. It should instead figure out what page number it needs to go to (the last one). 
To see it in action: with per page set at 10, go to page 7, then click on per page 50. 

I didn't put too much thought into the paging initially, so there are probably more edge cases like this that need to be figured out.

src/pages/Webhooks/HomePage.js
jasonrhodes: Yeah good call, we should store things in the state tree as "ready to use" as possible.

src/pages/Webhooks/HomePage.js
jasonrhodes: Let's leave this alone, I'd like to take a stab at abstracting the pagination into a couple of reusable components that handle all this for us

src/pages/Webhooks/components/CreateForm.js
```diff
@@ -0,0 +1,148 @@
...
+        placeholder='https://www.example.com/target'
+        type={'url'}
+      />
+
+      <Field
+        name='eventsRadio'
+        component={RadioGroup}
+        title='Events'
+        options={[
+          {value: 'all', label: 'All', checked: true},
```
jgzamora: Checked here isn't doing anything I think, the real decider comes from returning `initialValues` on `connect`

src/pages/Webhooks/components/CreateForm.js
```diff
@@ -0,0 +1,148 @@
...
+  const loading = submitting ? (<div>Loading...</div>) : (<div/>);
+
+  const authFields = auth && auth === 'basic' ? basicAuthFields : oAuth2Fields;
+  const showEvents = eventsRadio === 'select';
+  const eventBoxes = buildCheckBoxes(eventsTree);
+
+  return (
+    <form onSubmit={handleSubmit}>
+      <Field
+        // for redux-form
```
jgzamora: If everyone is ok with this, I'd like to collapse most fields into one line. Or at least a hybrid, where name, component, validate, id, label, placeholder are on one line and more complication stuff like arrays can go on it's own line.

src/pages/Webhooks/components/CreateForm.js
```diff
@@ -0,0 +1,148 @@
...
+const RadioGroup = ({input, options, title}) => (
+  <Radio.Group>
+    {title}
+    {options.map(o => {
+      return (<Radio key={o.label} {...input} id={o.label} label={o.label} value={o.value} checked={o.value === input.value}/>);
+    })}
+  </Radio.Group>
+);
+
+const required = value => value ? undefined : 'Required';
```
jgzamora: Validation overall needs more work in this form. Currently it's doing a lot of it onBlur, but most should be onSubmit validation

src/pages/Webhooks/HomePage.js
```diff
@@ -43,12 +40,12 @@ class WebhooksPage extends Component {
...
 
     // This should probably be a universal page-loading component
-    if (loading) {
+    if (loading && !webhookCount) {
       return (
         <Layout.App>
           <Page
-            primaryAction={{content: 'Create Template', onClick: () => { console.log('create template'); }}}
-            title={'Templates'}
+            primaryAction={{content: 'Create Webhook', onClick: () => { this.props.history.push('/account/webhooks/create'); }}}
```
jasonrhodes: We need to make `primaryAction` accept a `to` sub-prop that the `Page` component will interpret to mean "hey, use a `<Link>` component" ... weird to have to use `history.push` for that ( @jonambas do you think that'd be easy?)

src/pages/Webhooks/components/CheckboxGroup.js
```diff
@@ -0,0 +1,63 @@
+import React from 'react';
+import { Checkbox } from '@sparkpost/matchbox';
+
+/* Stateless component that returns a Checkbox.Group that is controled by a parent
```
jasonrhodes: I'm fine with leaving this as a functional component but I don't think I'd call it "stateless" tbh -- it has lots of internal state, we're just tracking it with closure-scoped variables instead of on `this`.

src/pages/Webhooks/components/CreateForm.js
jasonrhodes: I need to get our linting in better shape but we'll eventually make the "no parens" arrow function syntax illegal.

src/pages/Webhooks/components/CreateForm.js
jasonrhodes: What's the reason for this? The "prop-per-line" stuff here looks really readable to me…

src/pages/index.js
```diff
@@ -1,6 +1,14 @@
+import WebhooksHome from './Webhooks/HomePage';
+import WebhooksCreate from './Webhooks/CreatePage';
+
+const Webhooks = {
+  homePage: WebhooksHome,
+  createPage: WebhooksCreate
+};
+
+export { Webhooks };
```
jasonrhodes: casing should be Pascal-case only for the classes themselves, not the directories 
```js
import WebhooksHome from './webhooks/HomePage';

const webhooks = {
  HomePage: WebhooksHome
};

export { webhooks };
```

src/pages/Webhooks/HomePage.js
jonambas: Yeah, these "actions" render out <button>s so you can throw in the `Component` and `to` props or whatever else you need

jonambas:
Should we throw the webhooks stuff under the settings folder?

src/pages/Webhooks/components/CreateForm.js
```diff
@@ -0,0 +1,148 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import { Field, reduxForm, formValueSelector } from 'redux-form';
+import { TextField, Button, Select, Radio, Grid } from '@sparkpost/matchbox';
+import CheckboxGroup from './CheckboxGroup';
+
+const TextFieldWrapper = ({input, meta: {touched, error}, ...rest}) => (
```
jonambas: we should move these into ./src/components if we are going to reuse them on other pages

jonambas: Can also rename them to just `TextField` and `Select`.

src/pages/Webhooks/components/CreateForm.js
```diff
@@ -0,0 +1,148 @@
...
+
+const SelectWrapper = ({input, meta: {error}, ...rest}) => (
+  <Select {...input} error={error} {...rest}/>
+);
+
+const RadioGroup = ({input, options, title}) => (
+  <Radio.Group>
+    {title}
+    {options.map(o => {
+      return (<Radio key={o.label} {...input} id={o.label} label={o.label} value={o.value} checked={o.value === input.value}/>);
```
jonambas: checked here doesn't do anything if we are planning on using redux-form everywhere

src/pages/Webhooks/components/CreateForm.js
jgzamora: Don't feel too strongly about it. But I think that 
* most props on one line isn't unreadable (except for arrays / large objects)
* every single field 1-3 of these props that are boilerplate-y (id, name, label) which blows up the file size a bit

Something like this
```
      <Field name='name' id='name'
          label='Webhook Name'
          component={TextFieldWrapper}
          validate={required}
          placeholder='Opens and Clicks'
      />
```

Just a thought

src/pages/Webhooks/CreatePage.js
```diff
@@ -0,0 +1,119 @@
...
+        this.allEvents.push(eventKey);
+        treeItem.events.push({
+          key: eventKey,
+          label: eventDoc.display_name,
+          description: eventDoc.description // TODO: make tooltips from this?
+        });
+      });
+      this.eventsTree.push(treeItem);
+    });
+  }
```
jgzamora: I like the map, because I changed this to return an array anyways. But I think we should still just push the events into allEvents from the inner .map

src/pages/Webhooks/components/CreateForm.js
jonambas: I'm fine with it either way, it'll get minified anyway (right?). 

We could also make the wrapper use name for the id to slim it down a little. MB only uses the id to tie label & field together

src/pages/Webhooks/components/CheckboxGroup.js
jgzamora: The word is pointlessly descriptive in this comment anyways so I'm just going to remove. But I think this component is actually stateless / will discuss in slack.

src/pages/Webhooks/components/CreateForm.js
jgzamora: I think I need the checked to map the Field's initial value to the the matchbox component

src/pages/Webhooks/CreatePage.js
```diff
@@ -60,39 +58,35 @@ class WebhooksCreate extends Component {
...
-        description: groupDoc.description, // TODO: make tooltipes from this?
-        events: []
+  buildEventsData (eventGroups) {
+    return _.map(eventGroups, ({display_name, description, events}, key) => {
+      return {
+        key: key,
+        label: display_name,
+        description: description, // TODO: make tooltipes from this?
+        events: _.map(events, ({display_name, description}, eventKey) => {
+          this.allEvents.push(eventKey);
```
jgzamora: still need to move this to it's own line.

src/pages/Webhooks/CreatePage.js
```diff
@@ -15,7 +15,8 @@ class WebhooksCreate extends Component {
     let events;
 
     if (values.eventsRadio === 'select') {
-      events = _.concat(values.message_event, values.track_event, values.gen_event, values.unsubscribe_event, values.relay_event);
+      const checkedEvents = _.concat(values.message_event, values.track_event, values.gen_event, values.unsubscribe_event, values.relay_event);
+      events = _.filter(checkedEvents, (event) => (event)); // filter for truthy
```
jgzamora: didn't realize undefined vals were being concated in as well

src/pages/Webhooks/CreatePage.js
```diff
@@ -78,7 +79,7 @@ class WebhooksCreate extends Component {
     const eventDocs = this.props.eventDocs;
     const eventsTree = Object.keys(eventDocs).length !== 0 ? this.buildEventsData(eventDocs) : [];
 
-    if (eventsLoading) {
+    if (eventsLoading || !eventsTree.length) {
```
jgzamora: The `||` can be removes and the both consts above can be defined further down instead.

src/pages/TemplatesPage.js
```diff
@@ -39,7 +39,7 @@ class TemplatesPage extends Component {
     const loading = this.props.listLoading;
 
     // This should probably be a universal loading component
-    if (loading) {
+    if (loading && !templatesCount) {
```
jgzamora: I can revert this to avoid conflicts

src/pages/Webhooks/EditPage.js
```diff
@@ -0,0 +1,223 @@
...
+    if (!allChecked) {
+      eventsTree.forEach(({key, events}) => {
+        const filtered = _.filter(events, ({key}) => (_.includes(webhook.events, key)));
+        checkedEvents[key] = _.map(filtered, ({key}) => (key));
+      });
+    }
+
+    return (
+      <Layout.App>
+        { updateSuccess && this.state.showBanner &&
```
jgzamora: positioning is weird

src/pages/Webhooks/components/CreateForm.js
```diff
@@ -0,0 +1,175 @@
+import React from 'react';
```
jgzamora: Will rename to `WebhookForm`

src/pages/Webhooks/components/CreateForm.js
```diff
@@ -0,0 +1,175 @@
...
+import { connect } from 'react-redux';
+// import _ from 'lodash';
```
jgzamora: delete

src/pages/Webhooks/components/CreateForm.js
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  return values;
+};
+
+const mapStateToProps = (state, props) => {
+  const { name, target, eventsRadio, auth } = selector(state, 'name', 'taget', 'eventsRadio', 'auth');
+
+  const webhookValues = props.newWebhook ? {} : formatEditValues(state.webhooks.webhook);
+  // console.log('from form: ', webhookValues);
```
jgzamora: delete

src/pages/Webhooks/CreatePage.js
```diff
@@ -0,0 +1,128 @@
...
+
+  render () {
+    const { eventsLoading } = this.props;  // Form doesn't load until we have events
+
+    if (eventsLoading) {
+      return (
+        <Layout.App>
+          <Page
+            title={'Create Webhook'}
+            breadcrumbAction={{content: 'Webhooks', onClick: () => { this.props.history.push('/webhooks/'); }}}
```
jonambas: For these actions you can do:
```
{
       content: 'Create Webhook',
       Component: Link, // from react-router-dom
       to: '/webhooks'
     }
```

jasonrhodes:
looks so unbelievably good to me

jasonrhodes:
also: squash plz?

PR #24, opened 2017-07-24
Template New/Edit Form FAD-5065

jonambas:
This pr includes:
- Draft view (edit page)
- Published view 
- Delete
- Create

What's not included:
- Editor
- Preview & Send
- Test Data
- Duplicate

screen shot 2017-08-01 at 9 49 56 am
screen shot 2017-08-01 at 9 50 01 am
screen shot 2017-08-01 at 9 50 24 am
screen shot 2017-08-01 at 9 51 39 am


src/actions/templates.js
```diff
@@ -8,3 +8,56 @@ export function listTemplates () {
...
+      type: 'GET_TEMPLATE',
+      method: 'GET',
+      url: `/templates/${id}`
+    }
+  };
+}
+
+export function createTemplate () {
+  return (dispatch, getState) => {
+    const data = getState().form.templateEdit.values;
```
jasonrhodes: Might be better to pass this to the action instead of relying on it being in a specific form?

src/actions/templates.js
```diff
@@ -8,3 +8,56 @@ export function listTemplates () {
...
+        data
+      }
+    });
+  };
+}
+
+export function updateTemplate (params = {}) {
+  return (dispatch, getState) => {
+    const data = getState().form.templateEdit.values;
+    const id = data.id;
```
jasonrhodes: same thought here about passing these to the action creator

src/pages/Templates/EditPage.js
```diff
@@ -0,0 +1,130 @@
...
+  componentDidUpdate () {
+    const {
+      id,
+      submitSucceeded,
+      history
+    } = this.props;
+
+    if (this.state.newTemplate && submitSucceeded) {
+      history.push(`/templates/edit/${id}`);
+    }
```
jasonrhodes: at the top of `render()` could we do:
```js
if (this.state.newTemplate && submitSucceeded) {
  return <Redirect to=`/templates/edit/${id}`>;
}
```

jonambas: I tried this but it didn't trigger a re-render, will try again

src/actions/templates.js
```diff
@@ -8,3 +8,75 @@ export function listTemplates () {
...
+    meta: {
+      type: 'PUBLISH_TEMPLATE',
+      method: 'PUT',
+      url: `/templates/${id}`,
+      data: { published: true }
+    }
+  };
+}
+
+export function clear () {
```
jonambas: I named this clear because redux-form has a `reset` function

src/actions/templates.js
```diff
@@ -8,3 +8,75 @@ export function listTemplates () {
...
+
+export function update (data, params = {}) {
+  const id = data.id;
+  return {
+    type: 'SPARKPOST_API_REQUEST',
+    meta: {
+      type: 'UPDATE_TEMPLATE',
+      method: 'PUT',
+      url: `/templates/${id}`,
+      data,
```
jasonrhodes: Does it matter that `id` is still on this data object? I think probably not but not sure

src/pages/Templates/CreatePage.js
```diff
@@ -0,0 +1,85 @@
...
+    );
+  }
+
+  render () {
+    const {
+      match,
+      submitSucceeded
+    } = this.props;
+
+    if (submitSucceeded && this.state.shouldRedirect) {
```
jasonrhodes: what's the case where `submitSucceeded` is true but we don't want to redirect?

src/App.js
```diff
@@ -34,7 +37,13 @@ export default () => (
       <ProtectedRoute path='/dashboard' component={DashboardPage} />
       <Route path='/reports' render={() => <Redirect to='/reports/summary' />} />
       <ProtectedRoute path='/reports/summary' component={SummaryReportPage} />
-      <ProtectedRoute path='/templates' component={TemplatesPage} />
+
+      <ProtectedRoute exact path='/templates' component={TemplatesListPage} />
+      <ProtectedRoute exact path='/templates/create/' component={TemplatesCreatePage} />
+      <ProtectedRoute exact path='/templates/edit/:id' component={TemplatesEditPage} />
+      <ProtectedRoute exact path='/templates/edit/:id/published' component={TemplatesPublishedPage} />
```
jonambas: Should I change this to `/templates/published/:id`? Thoughts?

Current ui uses a query param, but it doesnt change when switching to draft and doesn't push history state

src/pages/Templates/EditPage.js
```diff
@@ -0,0 +1,192 @@
...
+    getDraft(match.params.id);
+    getPublished(match.params.id);
+  }
+
+  handlePublish (values) {
+    const { update, publish, match, getDraft } = this.props;
+    return update(values)
+      .then(() => publish(match.params.id))
+      .then(() => getDraft(match.params.id))
+      .then(() => this.setState({ shouldRedirectToPublished: true }));
```
jasonrhodes: this is really nice/readable

src/App.js
jasonrhodes: Hmm not sure how to best handle this, I think I like `:id/published` for now though

PR #25, opened 2017-07-31
Update readme

jgzamora:


jgzamora:
The docs link doesn't work

ewandennis:
Good background docs. For reference, I was able to start the UI with just the following:

```bash
npm install npm@latest -g
git clone ...
npm install
npm start
```

For the README itself, some info on dev-time dependencies would be good:
 - npm 5
 - node 6+?
 - boxen
 - VPN

Also maybe a little detail about the 3 npm scripts. Something like:

- `npm start` - start a dev server backed by `api.sparkpost.dev` with live reload on http://localhost:3100/
- `npm run build` - create an optimised production build under `build/`
- `npm test` - run the test suite (Caveat: currently borked)

PR #26, opened 2017-08-01
FAD-4921 Summary chart and date picker

jasonrhodes:
Went over this with @jgzamora and @jonambas yesterday, things I still need to do in this ticket:

* Already added `date-fns` module, going to replace `moment` everywhere with this
* Add minute-management
* Move metric definitions into config, including calculated metrics, and make metrics selectable using a simple input for now
* Separate our custom `DatePicker` (and any other components I can think of) out of the SummaryPage component
* Have @jonambas tweak styles


jonambas:
Added some quick styles so we have something for demo today

PR #30, opened 2017-08-04
Webhooks test

jgzamora:


src/pages/Webhooks/components/WebhookTest/WebhookTest.js
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  render () {
+    if (this.props.samplesLoading) {
+      return (<div>Loading...</div>);
+    }
+
+    const { webhook, samples, testResponse, testLoading } = this.props;
+    const buttonText = this.state.testSent ? 'Re-send batch' : 'Set Test Batch';
+    const testRequest = this.buildTestRequest(webhook, samples);
```
jgzamora: Is there a way to not make this every render?

src/pages/Webhooks/components/WebhookTest/WebhookTest.js
```diff
@@ -0,0 +1,82 @@
...
+
+    return (
+      <div>
+        <Button primary disabled={testLoading} onClick={() => { this.testWebhook(webhook.id, testRequest); }}>{buttonText}</Button>
+        <br/>
+        <br/>
+        { testLoading && <div>Sending test...</div> }
+        { this.state.testSent && testResponse && <ResponseBlock testResponse={testResponse}/> }
+        { !this.state.testSent && !testLoading && <RequestBlock testRequest={testRequest}/> }
```
jgzamora: Need to make sure only one of these three states is showing at a time. There is a weird state in which the past response shows up during loading.

src/pages/Webhooks/components/WebhookTest/WebhookTest.js
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  componentDidMount () {
+    // Only do this the first time
+    if (!this.props.samples) {
+      this.props.getEventSamples(this.props.webhook.events);
+    }
+  }
+
+  buildTestRequest (webhook, payload) {
```
jgzamora: Straight taken from old webui, point out improvements pls

src/pages/Webhooks/components/WebhookTest/WebhookTest.js
```diff
@@ -0,0 +1,82 @@
...
+import ResponseBlock from './TestResponseBlock';
+import RequestBlock from './TestRequestBlock';
+
+class WebhookTest extends Component {
+  state = {
+    testSent: false
+  }
+
+  componentDidMount () {
+    // Only do this the first time
```
jgzamora: Should use this logic for the create/edit docs as well. You should only need to get the on per login basis

src/pages/Webhooks/components/WebhookTest/TestResponseBlock.js
```diff
@@ -0,0 +1,31 @@
...
+      The test failed (POST to webhook target URL failed with non-2xx response code 404)
+    </Banner>;
+
+  return (
+    <div>
+      { banner }
+      { success &&
+        <pre>
+          <code>
+            <textarea readOnly value={JSON.stringify(testResponse, null, '  ')} rows={10} cols={100}/>
```
jgzamora: need someone cool / responsive for these text areas.

src/pages/Webhooks/components/WebhookTest/TestRequestBlock.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+
+const RequestBlock = (props) => {
+  const { testRequest } = props;
+
+  return (
+    <div>
+      <p>
+        The following request will be sent to your webhook target URL:
```
jgzamora: can fill target val here for convenience

src/pages/Webhooks/EditPage.js
```diff
@@ -213,25 +228,36 @@ class WebhooksEdit extends Component {
...
         { updateSuccess && this.state.showBanner &&
           <Banner title='Update Successful' status='success' onDismiss={this.onDismiss}/>
         }
         <Panel>
           <Panel.Section>
-            <WebhookForm eventsTree={eventsTree} allChecked={allChecked} newWebhook={false} checkedEvents={checkedEvents} onSubmit={(values) => { return this.updateWebhook(values, webhook, allEvents); }}/>
+            { this.state.testTab
+              ? <WebhookTest webhook={webhook}/>
+              : <WebhookForm eventsTree={eventsTree} allChecked={allChecked} newWebhook={false} checkedEvents={checkedEvents} onSubmit={(values) => { return this.updateWebhook(values, webhook, allEvents); }}/>
+            }
```
jgzamora: I decided to load both of these from the edit page so that I didn't have to deal with getting the correct webhook on route access (or getting it at all). I think it's fine for testing a webhook not to be available via url. As far as implementation, I was worried about it being messy but it feels fine so far. I think the edit-specific parts also have room to be simplified.

src/actions/webhooks.js
```diff
@@ -56,13 +56,37 @@ export function deleteWebhook (id) {
...
+      url: 'webhooks/events/documentation'
+    }
+  };
+}
+
+export function getEventSamples (events) {
+  return {
+    type: 'SPARKPOST_API_REQUEST',
+    meta: {
+      type: 'GET_EVENT_SAMPLES',
```
jasonrhodes: Thinking about it, these types need to be globally unique so this may not be unique enough. Is it time to move our types into constants maybe?

src/pages/Webhooks/EditPage.js
```diff
@@ -1,16 +1,17 @@
 import React, { Component } from 'react';
```
jasonrhodes: Should this page be `pages/webhooks/DetailsPage.js` with the various tabs inside of that?

src/pages/Webhooks/EditPage.js
```diff
@@ -163,7 +170,15 @@ class WebhooksEdit extends Component {
...
 
-  handleDelete () {
+  showTest () {
+    this.setState({ testTab: true, settingsTab: false });
+  }
+
+  showSettings () {
+    this.setState({ testTab: false, settingsTab: true });
+  }
+
```
jasonrhodes: Maybe it'd be better/safer to do `{ activeTab: 'test' }` where that value can be 'edit', 'test', or 'statuses'? i'm not crazy about the ability to accidentally set multiple tabs to `true`

... also will probably try to abstract "tabbed pages" to a component so we can reuse this pattern, but later

src/pages/Webhooks/components/WebhookTest/WebhookTest.js
jasonrhodes: We could do these kinds of GETs in a central place the same way we get account/user, but that feels a bit magical ...

We could also build action creators that are exported using `_.once()` so everywhere that needs it calls it, but the HTTP call would only happen on first call

jasonrhodes:
I think the `webhooks/components/WebhookTest/*` structure is a level too deep for now, everything in that directory can just be in `webhook/components/`

src/pages/Webhooks/components/WebhookTest/WebhookTest.js
jasonrhodes: lgtm, hooray for reuse

src/pages/Webhooks/components/WebhookTest/WebhookTest.js
jasonrhodes: do we expect webhook and samples to ever change after mount? if not, can just do something like:

```js
componentDidMount() {
  const { webhook, samples } = this.props;
  this.setState({ testRequest: this.buildTestRequest(webhook, samples) });
}
```

or maybe even skip `setState` and do `this.testRequest = this.buildTestRequest(webhook, samples)` there?

I'm not sure exactly how React works but maybe if the test request was its own component that did that build internally with `webhook` and `samples` passed in as props, it wouldn't re-render that sub-component if webhook and samples never change?

jasonrhodes: And just to add more options, you could use `_.once()` here too in the constructor:
```
this.buildTestRequest = _.once(this.buildTestRequest)
```

src/pages/Webhooks/EditPage.js
jgzamora: Yeah, that would make more sense

src/pages/Webhooks/HomePage.js
```diff
@@ -60,7 +60,7 @@ class WebhooksHome extends Component {
     return (
       <Layout.App>
         <Page
-          primaryAction={{content: 'Create Webhook', onClick: () => { this.props.history.push('/webhooks/create'); }}}
+          primaryAction={{ content: 'Create Webhook', Component: Link, to: '/webhooks/create' }}
```
jonambas: This is a nitpick, but we can move this into a `renderPage` func so we dont have to list out the Page component twice

src/pages/Webhooks/components/WebhookTest/WebhookTest.js
jgzamora: I like the second one, because not that it's expensive but it feels silly to pre-prepare for testing webhooks

src/pages/Webhooks/EditPage.js
jasonrhodes: I think I don't understand the problem/solution here

src/pages/Webhooks/components/WebhookTest/WebhookTest.js
jgzamora: Perfect. I'll try `._once` and maybe the component. I think the `componentDidMount` would be iffy because we can't guarantee the samples request has finished by then.

src/pages/Webhooks/components/WebhookTest/TestResponseBlock.js
jonambas:
```
<TextField multiline  />
```

jonambas: Think you'll have to set `height={}` though if you need responsive height

src/pages/Webhooks/HomePage.js
jgzamora: Nitpicks are the best

src/actions/webhooks.js
jgzamora: Lets make a ticket for that?

src/pages/Webhooks/DetailsPage.js
```diff
@@ -155,52 +180,60 @@ class WebhooksEdit extends Component {
...
-            title={'Webhook Settings'}
-            breadcrumbAction={{content: 'Webhooks', onClick: () => { this.props.history.push('/webhooks/'); }}}
-          />
-          <Panel>
-            <Panel.Section>
-              Loading...
-            </Panel.Section>
-          </Panel>
-        </Layout.App>
+        <WebhooksLoading
```
jgzamora: Moved loading to its own component. Figured I'd minimize its footprint on each file while we figure out the global loading prop.

src/actions/webhooks.js
```diff
@@ -56,13 +56,37 @@ export function deleteWebhook (id) {
...
+export function testWebhook (id, payload) {
+  return {
+    type: 'SPARKPOST_API_REQUEST',
+    meta: {
+      type: 'TEST_WEBHOOK',
+      method: 'POST',
+      url: `/webhooks/${id}/validate`,
+      data: { message: payload }
+    }
+  };
```
jasonrhodes: Every time I think about abandoning redux at all, I see these declarative API call actions and uggggh so beautiful

src/actions/webhooks.js
jasonrhodes: Go for it

src/pages/Webhooks/DetailsPage.js
```diff
@@ -0,0 +1,134 @@
...
+        Component: Link,
+        to: editPath,
+        onClick: this.showSettings
+      },
+      { content: 'Test',
+        Component: Link,
+        to: testPath,
+        onClick: this.showTest
+      }
+    ];
```
jasonrhodes: Would it be ok to set `this.secondaryActions` and `this.tabs` inside of `componentDidMount` so we don't recreate these on every render, and then `render` gets tighter too...

src/pages/Webhooks/DetailsPage.js
```diff
@@ -0,0 +1,134 @@
...
+import TestTab from './components/TestTab';
+import EditTab from './components/EditTab';
+
+class WebhooksDetails extends Component {
+  constructor (props) {
+    super(props);
+
+    this.state = {
+      id: props.match.params.id,
+      selectedTab: 0
```
jasonrhodes: We are going to need to check the route and set the right tab to account for people coming to the routes directly, i think... maybe we should just link selected tab to the route somehow and then we don't even need to handle the tab clicks at all .... 🤔

src/pages/Webhooks/DetailsPage.js
```diff
@@ -0,0 +1,134 @@
...
+        <Page
+          title={webhook.name}
+          secondaryActions={secondaryActions}
+          breadcrumbAction={{ content: 'Webhooks', Component: Link, to: '/webhooks/' }}
+        />
+        <Tabs
+          selected={this.state.selectedTab}
+          tabs={tabs}
+        />
+        <div>
```
jonambas: Are the tabs rendering correctly with this extra `div`? Tabs uses a sibling selector to "merge" it with the panel below it. This should work without the `div` though

jgzamora: Yup, I think they are there from something I copy pasted. Removing

PR #40, opened 2017-08-09
Globa Errors FAD-5252

jonambas:


src/middleware/sparkpostApiMiddleware.js
```diff
@@ -86,11 +87,7 @@ export default function sparkpostApiRequest ({ dispatch, getState }) {
       });
 
       if (response.status >= 500) {
-        dispatch({
-          type: 'API_FAILURE_RECEIVED',
-          payload: { message, response },
-          meta
-        });
+        dispatch(received({ message, response }, meta));
```
jonambas: response object still empty, will need to move `dispatch(received({ message, response }, meta));` outside this `if` if you want to test

PR #42, opened 2017-08-15
FAD-5253 fixes refresh token flow

jasonrhodes:
To test:

* Checkout branch
* Log in with "keep me logged in" checked
* In "Application" panel, find auth cookie and change a few characters of the auth token
* Refresh any page, you should stay logged in
* Back in "Application" panel, find auth cookie and change the auth token _and_ the refresh token, refresh a page, it should log you out

Note: There were a few different issues, let me walk you through the PR before you jump in

tl;dr

* Log out didn't work because we were trying to `JSON.parse` the stored cookie twice, which was blowing up and causing the log out action to fail and be swallowed
* Refresh token didn't work because we often make multiple requests at once, so the first one was getting a new auth token but before it finished using it, the 2nd request also requested a new token using the same refresh token which was then invalid (they can only be used once), resulting in a 400 that caused the console errors.

### DIFF
--- a/src/actions/auth.js
+++ b/src/actions/auth.js
@@ -56,9 +56,6 @@ export function authenticate (username, password, rememberMe = false) {
 
 export function refresh (token, refreshToken) {
   let oldCookie = authCookie.get();
-  if (oldCookie) {
-    oldCookie = JSON.parse(oldCookie);
-  }
   const newCookie = Object.assign({}, oldCookie, { token, refreshToken });
   authCookie.save(newCookie);
   return login(newCookie);
--- a/src/actions/auth.js
+++ b/src/actions/auth.js
@@ -56,7 +56,7 @@ export function authenticate (username, password, rememberMe = false) {
 
 export function refresh (token, refreshToken) {
   let oldCookie = authCookie.get();
-  const newCookie = Object.assign({}, oldCookie, { token, refreshToken });
+  const newCookie = Object.assign({}, oldCookie, { access_token: token, refresh_token: refreshToken });
   authCookie.save(newCookie);
   return login(newCookie);
 }
--- a/src/helpers/authCookie.js
+++ b/src/helpers/authCookie.js
@@ -4,6 +4,7 @@ import config from '../config';
 const { name, options } = config.authentication.cookie;
 
 function save (data) {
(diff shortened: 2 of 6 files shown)


src/helpers/http.js
```diff
@@ -10,12 +11,34 @@ const sparkpostRequest = axios.create({
 
 // TODO handle timeout error better
 
+const tokensUsed = [];
```
jgzamora: Can this ever actually have more than one token? If not then then the boolean should do. If it can, this would be a cool place for a Set 🤓

src/actions/auth.js
```diff
@@ -56,9 +56,6 @@ export function authenticate (username, password, rememberMe = false) {
 
 export function refresh (token, refreshToken) {
   let oldCookie = authCookie.get();
-  if (oldCookie) {
-    oldCookie = JSON.parse(oldCookie);
-  }
   const newCookie = Object.assign({}, oldCookie, { token, refreshToken });
```
jgzamora: This breaks the cookie for me, unless I do:
```
  const newCookie = Object.assign({}, oldCookie, { access_token: token, refresh_token: refreshToken });
```

src/helpers/authCookie.js
```diff
@@ -4,6 +4,7 @@ import config from '../config';
 const { name, options } = config.authentication.cookie;
 
 function save (data) {
+  console.log('saving new cookie', JSON.stringify(data, null, 2));
```
jasonrhodes: remooooove

src/helpers/http.js
```diff
@@ -1,5 +1,6 @@
 import config from '../config';
```
jasonrhodes: Remove commented code here

src/helpers/promise.js
```diff
@@ -0,0 +1,41 @@
+function wait (options) {
+  const {
+    condition,
+    resolve,
+    reject = resolve,
+    ms = 500,
+    retry = 0,
+    maxRetries = 100
```
jasonrhodes: not really, my thinking was if you adjust the ms from the default it's nice to just have this be a high number unless you want to bring it down -- i had originally used `Infinity` as the default but 100 seemed sufficiently high

PR #46, opened 2017-08-20
Paginated collection component (FAD-5241)

jasonrhodes:
To use the `TableCollection`:
```jsx
const rows = [
  { id: 1, name: 'First thing', description: 'This is my first thing', active: true },
  { id: 2, name: 'Second thing', description: `Yep, it's the second thing` },
  { id: 3, name: 'Third', description: `Ta da third thing` }
];
const getRowData = ({ id, name, description, active }) => {
  const status = active ? 'active' : 'inactive';
  return [id, name, status, description];
};
const props = { columns, rows, getRowData, pagination: true };

return <TableCollection {...props} />;
```

Pagination will be handled for you in all cases if you set `pagination={true}`. You can set a `defaultPerPage` here as well as a `perPageButtons` array of numbers, but they both have default values (`25` and `[10, 25, 50]` for now).

To use the `Collection` directly:
```jsx
const rows = [
  { id: 1, name: 'First thing', description: 'This is my first thing', active: true },
  { id: 2, name: 'Second thing', description: `Yep, it's the second thing` },
  { id: 3, name: 'Third', description: `Ta da third thing` }
];
const headerComponent = () => <div><h1>Collection Header</h1></div>;
const rowComponent = ({ id, name, description, active }) => (
  <div>
    {active && <Icon name='checkbox' className='active-check' />}
    <h3>{name} (ID: {id})</h3>
    <p>{description}</p>
  </div>
);
const props = { rows, headerComponent, rowComponent };

return <Collection {...props} />;
```

If you want to wrap your custom collection (`outerWrapper` will wrap the header and body but _not_ the pagination elements, `bodyWrapper` will just wrap the body and not the header, so you can do complex panels etc):

```jsx
const outerWrapper = ({ children }) => <Panel>{children}</Panel>;
const bodyWrapper = ({ children }) => <PanelBody>{children}</PanelBody>;

// add these to props for the raw `Collection` example above:
props.outerWrapper = outerWrapper;
props.bodyWrapper = bodyWrapper;  
```

This ^ will result in something like:
```html
<Panel>
  <div><h1>Collection Header</h1></div>
  <PanelBody>
    <div>
      <Icon name='checkbox' class='active-check' />
      <h3>First thing (ID: 1)</h3>
      <p>This is my first thing</p>
    </div>
    <!-- etc for each item in "rows" -->
  </PanelBody>
</Panel>

<!-- Pagination stuff would be here if it was on and applied to perPage/length -->
```


jasonrhodes:
Re: pagination:

* Page buttons only show up there would be more than 1 page ... need a matchbox tweak to cover when there are e.g. only 2 pages of data but Matchbox thinks it should show 5 pages... SparkPost/matchbox#35
* Per page buttons will always show up right now if you set pagination to true ... but I think if `rows.length < Math.min(perPageButtons)` we could hide it too
* If you come to the url with `?page=n` it will try to load the correct page -- and then also keep track of page/perPage in the query string for you ... otherwise for now it doesn't do anything with the query params unless you include the prop `updateQueryParam={true}`

jasonrhodes:
Also, I made an error banner for API errors, I'm so sorry I couldn't help myself:

```
const { error, listTemplates } = this.props;
return (
  <ApiErrorBanner
    message={'Sorry, we seem to have had some trouble loading your templates.'}
    errorDetails={error.message}
    reload={listTemplates}
  />
);
```

results in:
screen shot 2017-08-21 at 12 51 49 pm

src/components/Collection/Pagination.js
```diff
@@ -0,0 +1,31 @@
...
+    <div>
+      {data.length > perPage && <Pagination
+        pages={Math.ceil(data.length / perPage)}
+        pageRange={pageRange}
+        initialIndex={(currentPage - 1)}
+        onChange={onPageChange}
+      />}
+      <Button.Group>Show Per Page:&nbsp;
+        {perPageButtons.map((perPage) => (
+          <Button key={perPage} onClick={() => onPerPageChange(perPage)}>{perPage}</Button>
```
jonambas:
```

<Button className={classnames(isSelected && styles.Selected)} ... />
```

Pagination.module.scss
```

button.Selected {
  font-weight: 700;
  color: color(orange);
}
```

jonambas: the css wont actually work, just noticed the global css is being pulled in after module css so local/matchbox styles get overridden. 

Easy to fix though

PR #47, opened 2017-08-22
Date range management FAD-5334

jonambas:
- Selecting "Custom" opens Datepicker
- Datepicker & Date fields stay in sync
  - Date Fields are validated on blur
  - Dates must be valid dates
  - From must be before To
  - To must be before now
  - Must match `MM/DD/YY` and `h:mma`
- Reverts to previous dates if invalid

kapture 2017-08-22 at 16 09 16

### DIFF
--- a/src/actions/reportFilters.js
+++ b/src/actions/reportFilters.js
@@ -1,7 +1,14 @@
+/* eslint-disable */
+import { getRelativeDates } from 'helpers/metrics';
+
 export function setRelativeTime(payload) {
-  return {
-    type: 'SET_RELATIVE_TIME',
-    payload
+  const dates = { ...getRelativeDates(payload) }
+  return (dispatch) => {
+    dispatch({
+      type: 'SET_RELATIVE_TIME',
+      payload: { ...dates, range: payload }
+    });
+    return Promise.resolve();
   };
 }
 
--- a/src/actions/reportFilters.js
+++ b/src/actions/reportFilters.js
@@ -1,23 +1,17 @@
 import { getRelativeDates } from 'helpers/metrics';
 
-export function setRelativeTime(payload) {
-  return (dispatch) => {
-    dispatch({
-      type: 'SET_RELATIVE_TIME',
-      payload: { ...getRelativeDates(payload), range: payload }
-    });
-    return Promise.resolve();
-  };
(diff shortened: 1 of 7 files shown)


src/actions/reportFilters.js
```diff
@@ -1,7 +1,12 @@
+import { getRelativeDates } from 'helpers/metrics';
+
 export function setRelativeTime(payload) {
```
jonambas: Think we can technically use one action for exact and relative time, but figured I'd separate them out for when we get to auto-refreshing

jasonrhodes:
First initial thought is: we can't hardcode the `MM/DD/YY` format ... internationally that format is bizarre and confusing.

It would be nice if we could use a format like `15-Aug 2017`, but if not that, then i'd vote for `YYYY-MM-DD` ... i wish there was an easy way to do this with locale formatting in the browser but i was looking a few weeks ago and it seemed _almost_ there but not quite

jasonrhodes:
We need to switch up the x-axis for "last hour" -- less ticks and showing every quarter hour somehow. 

Current app x-axis for last hour:
screen shot 2017-08-23 at 2 49 50 pm

jasonrhodes:
Can we add a "cancel" button on the date picker pop up that resets everything to the way it was before it was opened? so reset datepicker internal state, reset relative drop down if it was previously selected, close datepicker popup (pressing "esc" should do all these things too)

jasonrhodes:
Pressing "enter" while in the inputs should do the same as "Apply" -- a nice to have would be having enter pressed at any time while that picker is open will apply the settings ... maybe with the exception of while you're in between date selections because during that time "enter" locks in the "to" date ...

src/pages/reports/components/DateFilter.js
```diff
@@ -45,6 +27,15 @@ class DateFilter extends Component {
     window.removeEventListener('keydown', this.handleEsc);
   }
 
+  componentWillReceiveProps(nextProps) {
+    this.syncStateToProps(nextProps);
+  }
+
+  // Sets local state from reportFilters redux state - need to separate to handle pre-apply state
+  syncStateToProps = ({ filter }) => {
```
jonambas: should be `syncPropstoState`

PR #48, opened 2017-08-24
Metrics selection FAD-5331

jonambas:


src/pages/reports/SummaryPage/SummaryPage.js
```diff
@@ -45,18 +52,29 @@ class SummaryReportPage extends Component {
...
+      label: _.find(METRICS_LIST, { key: metric }).label
+    }));
 
     this.props.fetchMetrics('deliverability/time-series', query)
-      .then(() => this.setState({ chartOptions: {
-        ...this.state.options,
-        precision: query.precision
-      }}));
+      .then(() => this.setState({
+        chartOptions: {
```
jonambas: Wondering if we should move all options / chartOptions into redux

jasonrhodes: yeah i think we probably need to do this

PR #55, opened 2017-08-27
Refactored chart refresh logic to use redux

jasonrhodes:
Basic goal of this was to move "chart options" into redux, which made it easier to simplify the chart refresh flow. So instead of doing this in a component:

1. get new query data (date, typeahead filters, new list of metrics)
1. calculate query into format
1. store local state
1. refresh metrics call

We can just do:

1. refresh chart with new query data `this.props.refreshSummaryChart({ from, to })`

That action will calculate the query and do the metrics call attaching some additional info to that call's `meta` key, and the chartSummary and reportFilters reducers are listening for `FETCH_METRICS_SUCCESS` and populating their state accordingly.

I thought about combining those 2 reducers into 1, but chartSummary is stuff specific to the summary chart while reportFilters will be shared across all reports, so I think this works ok for now.

It still feels a little complicated tbh, but I think it's heading in a good redux direction. Thoughts? We can go over in person.

### DIFF
--- a/src/actions/metrics.js
+++ b/src/actions/metrics.js
@@ -6,13 +6,14 @@ const defaultParams = () => ({
   metrics: 'count_targeted'
 });
 
-export function fetch(url, params = {}) {
+export function fetch({ path, params = {}, meta = {}}) {
   return {
     type: 'SPARKPOST_API_REQUEST',
     meta: {
+      ...meta,
       type: 'FETCH_METRICS',
       method: 'GET',
-      url: `/metrics/${url}`,
+      url: `/metrics/${path}`,
       params: {
         ...defaultParams(),
         ...params
@@ -21,10 +22,12 @@ export function fetch(url, params = {}) {
   };
 }
 
-export function fetchDeliverability(params = {}) {
-  return fetch('deliverability', params);
+export function fetchDeliverability(params = {}, meta = {}) {
+  const path = 'deliverability';
+  return fetch({ path, params, meta });
 }
 
-export function getTimeSeries(params = {}) {
-  return fetch('time-series', params);
+export function getTimeSeries(params = {}, meta = {}) {
+  const path = 'deliverability/time-series';
+  return fetch({ path , params, meta });
 }
--- a/src/actions/reportFilters.js
+++ b/src/actions/reportFilters.js
@@ -1,18 +1,18 @@
-import { getRelativeDates } from 'helpers/metrics';
+// import { getRelativeDates } from 'helpers/metrics';
 
-export function setRelativeTime(range) {
-  return (dispatch) => Promise.resolve(dispatch({
-    type: 'SET_RELATIVE_TIME',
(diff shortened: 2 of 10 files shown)


src/actions/metrics.js
```diff
@@ -21,10 +22,12 @@ export function fetch(url, params = {}) {
...
+  const path = 'deliverability';
+  return fetch({ path, params, meta });
 }
 
-export function getTimeSeries(params = {}) {
-  return fetch('time-series', params);
+export function getTimeSeries(params = {}, meta = {}) {
+  const path = 'deliverability/time-series';
+  return fetch({ path , params, meta });
 }
```
jasonrhodes: tbh not even sure we need these, I can remove

src/actions/reportFilters.js
```diff
@@ -1,18 +1,18 @@
...
-    type: 'SET_EXACT_TIME',
-    payload: { ...rangeDates, range: 'custom' }
-  }));
-}
+// export function setExactTime(rangeDates) {
+//   return (dispatch) => Promise.resolve(dispatch({
+//     type: 'SET_EXACT_TIME',
+//     payload: { ...rangeDates, range: 'custom' }
+//   }));
+// }
```
jasonrhodes: So instead of calling these actions, you just refresh the chart with new to/from/range values and then in the reportFilter reducer it sets its state after a successful metrics refresh -- that's the philosophy change here.

Might not be exactly the same with the typeahead filters since we might not want to wait for the metrics call to finish before adding the filter tag into the UI, we can discuss how to best do that.

src/helpers/metrics.js
```diff
@@ -1,6 +1,8 @@
 import moment from 'moment';
```
jasonrhodes: This file is getting big, can probably split out "date helpers"

src/pages/reports/SummaryPage/SummaryPage.js
```diff
@@ -141,22 +109,21 @@ class SummaryReportPage extends Component {
...
-    const { metricsData } = this.props;
-    const { chartOptions = false } = this.state;
+    const { metricsData, chart } = this.props;
 
     return (
       <Layout.App>
         <Page title='Summary Report' />
 
-        <Filters refresh={this.refresh}/>
+        <Filters refresh={this.props.refreshSummaryChart}/>
```
jonambas: Does this still need to be passed through if its in redux?

jonambas: Nvm, probably should be since the refresh would be different on other pages

jonambas:
`<MetricsModal selectedMetrics={chart.metrics}`
and in the metrics modal you'll need to change the isActive line
`const isActive = _.findIndex(selectedMetrics, (m) => m.key === metric.key) > -1;`

jasonrhodes:
This PR is contained in #56

PR #57, opened 2017-08-29
Summary chart: bookmark and share charts FAD-5333

jonambas:
Passes `filters.to`, `filters.from`, and `chart.metrics` key into query string from redux.

Will need to pass through other chart options - scaling, filters, event/injection time


jonambas:
@jasonrhodes we can wait until you have typeahead done - will need to plug those into the share link anyway

src/pages/reports/SummaryPage/SummaryPage.js
```diff
@@ -5,44 +5,63 @@ import { withRouter } from 'react-router-dom';
...
+  parseSearch() {
+    const { location, addFilter } = this.props;
+    let options = {};
+
+    if (location.search) {
+      const { from, to, metrics = [], filters = [] } = qs.parse(location.search);
+
+      // Checks if there is only one metric/filter
+      const metricsList = typeof metrics === 'string' ? [metrics] : [...metrics];
+      const filtersList = typeof filters === 'string' ? [filters] : [...filters];
```
jasonrhodes: Do we need to spread these here at the end if they are already arrays?

src/pages/reports/SummaryPage/SummaryPage.js
```diff
@@ -5,44 +5,63 @@ import { withRouter } from 'react-router-dom';
...
+    this.handleRefresh(this.parseSearch());
     this.props.refreshTypeaheadCache();
   }
 
+  parseSearch() {
+    const { location, addFilter } = this.props;
+    let options = {};
+
+    if (location.search) {
+      const { from, to, metrics = [], filters = [] } = qs.parse(location.search);
```
jasonrhodes: grab `delimiter` here too

src/pages/reports/SummaryPage/SummaryPage.js
```diff
@@ -5,44 +5,63 @@ import { withRouter } from 'react-router-dom';
...
+
+      // Checks if there is only one metric/filter
+      const metricsList = typeof metrics === 'string' ? [metrics] : [...metrics];
+      const filtersList = typeof filters === 'string' ? [filters] : [...filters];
+
+      filtersList.forEach((filter) => {
+        const parts = filter.split(',');
```
jasonrhodes: split on `delimiter` pulled from `location.search`

src/pages/reports/SummaryPage/SummaryPage.js
```diff
@@ -5,44 +5,63 @@ import { withRouter } from 'react-router-dom';
...
+        addFilter({ value: parts[0], type: parts[1] });
```
jasonrhodes: instead of parts [0] and [1] here, to count for the case where the comma might be in the value, we should do something like:
```js
const parts = filter.split(',');
const type = parts.pop();
const value = parts.join(',');
addFilter({ value, type });
```

src/pages/reports/summary/SummaryPage.js
```diff
@@ -50,85 +70,73 @@ class SummaryReportPage extends Component {
...
 
-  renderTimeMode() {
-    const { eventTime } = this.state;
+  updateLink = () => {
+    const { filters, chart, history } = this.props;
+    const options = {
+      from: moment(filters.from).utc().format(),
+      to: moment(filters.to).utc().format(),
+      metrics: chart.metrics.map((metric) => metric.key),
+      filters: filters.activeList.map((filter) => `${filter.value},${filter.type}`)
```
jasonrhodes: nitpick but for optics alone for those who inspect the query string, can we do `${filter.type}:${filter.value}` and then use `shift()` above instead of `pop()`? that will match up nicely if we do special search keywords in the input later, like searching for `domain:something.com`

jonambas:
Rebased & fixed folder names 👍

jasonrhodes:
This looks good, wondering if we should somehow move this stuff into actions that you pass `location` and/or `history` into, so we can call this from the other charts? Or how do we handle sharing this logic across charts, is my real question

PR #62, opened 2017-08-31
Critical CSS & Production build fixes

jonambas:
- Both global and module css now extract to `.css` files.
- All css is now minified
- CSS in `critical.scss` inlines inside `<head>`
- Adds a loading screen with the SP logo

kapture 2017-08-30 at 22 42 00


jonambas:
Just a thought, but we can rename module classnames with `css-loader`'s `localIdentName` in MB and in the app to reduce size further.

https://medium.freecodecamp.org/reducing-css-bundle-size-70-by-cutting-the-class-names-and-using-scope-isolation-625440de600b

PR #67, opened 2017-09-01
FAD-5332 replaces middleware with api action helper

jasonrhodes:
This was so much easier than I even thought it would be.


jasonrhodes:
tl;dr now you make api calls in action creators like this:
```js
// myAction.js
import sparkpostApiRequest from 'actions/helpers/sparkpostApiRequest';

export function fetch(id) {
  return sparkpostApiRequest({
    type: 'FETCH_THING',
    meta: {
      url: `/things/${id}`
    }
  });
}
```

which means all of our API call action creators will return promises now, so you can also do:
```js
export function fetchAndSortThreeThings() {
  return (dispatch) => Promise.all([fetch('a'), fetch('b'), fetch('c')])
    .then(([a, b, c]) => dispatch({ type: 'SORTED_THINGS', payload: [...a, ...b, ...c] });
}
```

src/middleware/sparkpostApiMiddleware.js
```diff
@@ -28,6 +28,8 @@ export default function sparkpostApiRequest({ dispatch, getState }) {
       return;
     }
 
+    dispatch({ type: 'YOU_HAVE_FAILED', meta });
```
jonambas: lol

jgzamora:
Looks fine to me. I'd like to `generalApiRequest` ported over and `createBilling` made into a promise chain, to see the effects of `onSuccess` no longer being a thing.

jasonrhodes:
Ok so a few notes: this wasn't strictly _necessary_ for this ticket after all, because instead of processing the cache for the typeahead and storing it in redux, we're using a selector to format the data so we only store it in the redux store once.

That said, I still think this is a better model, so I'm using it in metrics, but I'm not going to switch everything over in this ticket. Let's let it marinate a little while longer and then we can pick one and convert everything one way or the other -- I'm leaning toward the helper over the middleware.

Closing this PR but we can refer back to it later. Will open a new one to focus on the typeahead work.

PR #69, opened 2017-09-06
Templates Editor FAD-5345

jonambas:
- Some global styles to override ace editor styles
- Cleanup, uses `loading` layout prop, shared redux textfield wrapper, etc
- Started some setup for the test data ticket

### DIFF
--- a/src/pages/Templates/components/Editor.js
+++ b/src/pages/Templates/components/Editor.js
@@ -9,6 +9,7 @@ import 'brace/mode/html';
 import 'brace/theme/tomorrow';
 import { Panel, Button } from '@sparkpost/matchbox';
 
+import './Editor.scss';
 import styles from './FormEditor.module.scss';
 
 const AceWrapper = ({ input, ...rest }) => (
@@ -16,15 +17,18 @@ const AceWrapper = ({ input, ...rest }) => (
     mode='html'
     theme='tomorrow'
     name='emailContent'
+    className='TemplateEditor'
     height='900px'
     width='auto'
     style={{ marginBottom: '-18px' }}
     tabSize={2}
-    fontSize={11}
-    markers={[]}
+    fontSize={12}
     cursorStart={1}
     highlightActiveLine
     showPrintMargin={false}
+    setOptions={{
+      displayIndentGuides: false
+    }}
     {...rest}
     {...input}
   />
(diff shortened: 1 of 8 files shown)


src/pages/Templates/components/Editor.js
```diff
@@ -6,44 +6,63 @@ import { Field, reduxForm } from 'redux-form';
 // Components
```
jasonrhodes: `AceWrapper` in its own file so we can eventually test?

src/pages/Templates/components/Editor.js
```diff
@@ -6,44 +6,63 @@ import { Field, reduxForm } from 'redux-form';
...
-          name='content.html'
-          component={AceWrapper}
-          readOnly={published}
-        />
-      </Panel>
+      <div className={styles.EditorSection}>
+        <Tabs selected={this.state.selectedTab} tabs={tabs}/>
+        <Panel className={styles.EditorPanel}>
+          <Field
+            name={this.fieldNames[this.state.selectedTab]}
```
jonambas: This won't actually work - test data and text content would also get pushed with the rest of the redux form to template update. Would probably need to separate them out. Was just setting up the tabs.

src/pages/Templates/components/Editor.js
```diff
@@ -6,44 +6,63 @@ import { Field, reduxForm } from 'redux-form';
...
 );
 
 class Editor extends Component {
+  state = {
+    selectedTab: 0
+  }
+
+  fieldNames = ['content.html', 'text', 'test']
+
+  handleTab = (i) => {
```
jasonrhodes: not for this ticket but is it worth it to have a "tabs and panels" component that does all this? or is it too hard because we often want the tabs and the panels positioned far from each other?

src/pages/Templates/components/Editor.js
jonambas: 👍

src/pages/Templates/components/Editor.js
jonambas: As something reusable? Might be difficult cause there are different use cases - could change route, trigger actions, or just change content.

jasonrhodes: Fair, sounds good

PR #70, opened 2017-09-06
Baseline to the Baseline tests

jgzamora:
* Structure
	* `/tests` folders holding tests for everything at that level
	* `Filename.test.js` What’s a spec? 
	* Only structural thing I’m against is test files just laying next to the actual file.
	* Other than that idrc about the structure. This feels fine, but I don’t mind if we opt for a larger `/tests/` folder outside `/src` 
* General Strategy
	* As much snapshot testing whenever possible
	* Use enzyme - https://github.com/airbnb/enzyme
* Strategy for simple SFCs
	* Parent components 
		* Shallow wrapper
		* Snapshot tests 
		* Simulate events
		* unit / snapshot test helpers if extracted
		* example: `src/components/Collection/tests/Pagination.test.js`
	* Base / Non-matchbox components
		* Use a render wrapper (get all the html) 
		* Snapshot tests
		* Simulate events
		* unit / snapshot test helpers if extracted
		* example: `src/components/SparkPost/SparkPost.js`
* Classes
	* I opted to for exporting the Classes themselves for testing. This could be avoided by a tool like rewire, I punted on this.
	* Class with no lifecycle methods:
		* use a shallow wrapper
		* Snapshot test the class
		* simulate events with `.find().sumulate()`
		* example: `src/components/ApiErrorBanner/tests/index.test.js`
	* Lifecycle methods:
		* Mount wrapper
		* Sinon Spies for lifecycle methods
		* `.setProps` / the rest of the enzyme dom manipulation API
		* `.find().simulate` still works
		* Shallow + Snapshot test also
		* example: `src/components/GlobalError/tests/GlobalError.test.js`
* Connected components
	* Use all the Class strategies that apply but also:
	* Test mapStateToProps if it has logic (export or tool problem applies here)
	* Trust the exported  `connect()`  HOC to work without tests (suggested by core team)
	* example: `src/components/GlobalError/tests/GlobalError.test.js`
* `withRouter()`
	* Try testing like a Connected component if possible
	* OR: Figure out Mount wrapping the (I didn’t)
	* Snapshot test would be wrapped with  ``` <MemoryRouter>[…]<MemoryRoute/>```
* Actually implemented
	*  Most components in `src/components`
		* Missed `/Collection/Collection.js`, skipped it initially because of `withRouter()` and forgot to circle back
* Not implemented but probably should done here:
	* Reducer tests, looks simple enough. Snapshots can be easily used
* Not implemented but should be separate tickets
	* Page tests, lol. Sorry if this was meant to be the main part of the ticket, but I spent too much time:
		* reading about testing, have fun: [react-redux-links/react-redux-testing.md at master · markerikson/react-redux-links · GitHub](https://github.com/markerikson/react-redux-links/blob/master/react-redux-testing.md)
		* trying out types of testing I later abandoned
	* I’d make tickets for testing each page, so we can get good some good nuanced implementation and spread out the testing know-how
	* In depth `withRouter()` tests, wasn’t able to figure it out in time. We might not need to worry about this other than snapshotting inside a MemoryRoutes component
	* We should set up a way of grabbing any function we want even if it’s not exported
	* I punted on actions until middleware vs. promise chain is resolved
* Miscellaneous 
	* Set some globals in `config/jest/jestSetup.js`
	* jest-in-case for running lots of cases. What’s an “m-sys test suite”??

Wish I would’ve started with pages and gotten some of the meatier stuff implement. Fwiw I think the the components folder covered a good range of general scenarios that should cover testing most Pages, with some caveats like mocking actions.


jonambas:
I'm down for `/tests/` folders next to the components. In tools we had everything in one `__tests__` folder and was a little difficult to keep track of things.

Also looks like jest will check either test | spec. spec sounds kooler
```
"testMatch": [
      "<rootDir>/src/**/__tests__/**/*.js?(x)",
      "<rootDir>/src/**/?(*.)(spec|test).js?(x)"
    ],
```

For connected components we renamed the class so there was no confusion. Not sure if I like that though:
```
export class DetachedButton extends Component {}
export default connect(null, {})(DetachedButton)
```

jasonrhodes:
I like most of this. Some thoughts:

* I'm cool with `.test.js`
* I think I want to separate snapshot tests out, so maybe just use `.snapshot.js` and tell jest about those maybe? If it's a pain then maybe `.snapshot.test.js` idk, I just feel like they clutter up the files for "regular" tests
* If `mapStateToProps` has logic, pull logic into a selector and test the selector, that gets us around that issue
* I like exporting classes for testing, personally ... seems explicit and simple, the rewire thing looked pretty 😖  at first glance anyway

jasonrhodes:
>hofix

src/App.js
```diff
@@ -1,8 +1,9 @@
 import React from 'react';
```
jasonrhodes: This is all stuff that was added in another branch and has since been removed ... not sure the easiest way to just say "goodbye" to these changes...

jasonrhodes:
I rebased this branch and the conflicts went away FWIW.

src/App.js
jasonrhodes: rebasing solved this

jasonrhodes:
I just added a commit with some examples of straight-up unit tests for helpers, which use jest's manual mocks. 

Last thing I would like to see here is an example of how to do enzyme tests that aren't snapshot, as well as deciding on how to organize snapshot vs other tests. Then I think we're set here.

jasonrhodes:
Not sure why this PR is so jacked up with included commits... gonna close and try to open a new one and see what happens :/

PR #74, opened 2017-09-08
Report Filtering Typeahead (FAD-5332)

jasonrhodes:
Lots going on here, highlights:

1. in `actions/summaryChart` we refresh the typeahead cache whenever the date range changes (which will include the first time when it changes from nothing to initial values) -- we don't need to update this cache at any other time (more on this later)

1. in `actions/reportFilters` there is the new `refreshTypeaheadCache` action creator that will dispatch a bunch of actions. This is where I thought we would need to chain, so the dispatches are wrapped in `Promise.all()`, the idea was I would but a `.then()` at the end there and then process all that data and dispatch another action to store the processed data. But on further consideration, a selector was the better choice to prevent duplicating large lists in the store.

1. `selectors/reportFilterTypeaheadCache` is a new folder and file, with a selector to pull from the various parts of the redux store where the typeahead cache values should live, process them to put them in the right format for the typeahead component, and return them. It gets used in the `mapStateToProps` function in `pages/reports/components/Filter.js`

1. The Typeahead component in `pages/reports/components/Typeahead.js` is using Downshift, a react lib that does all of the keyboard and click events to make the UI of the typeahead work. @jonambas worked on this in parallel to my data work here

Still to do: 
* [x] use the filters in the summary chart metrics request so we get the correct filtered data in the chart
* [x] figure out why the chart isn't actually working right now :sob:
* [x] fix the charts (once we figure out why they aren't working)


src/pages/reports/components/Typeahead.js
```diff
@@ -1,75 +1,89 @@
...
         </div>
-        <TextField onChange={this.handleOnChange} />
-      </div>
-    );
+      );
+    }
+
+    // Downshift automatically sets inputValue to selected value (which is an object and it converts it to a string lol)
+    // So instead we manage input value manually.
+    const handleStageChange = ({ inputValue }) => {
```
jonambas: `handleSta(t)eChange`

typo :persevere:

jasonrhodes: hmm but how it work !

jonambas: Oh also here: `onStateChange={handleStageChange}`

src/pages/reports/components/Typeahead.js
```diff
@@ -1,75 +1,89 @@
...
+      clearSelection
+    }) => {
 
-  render() {
-    const { options = []} = this.props;
+      const mappedMatches = matches.map((item, index) => ({
+        content: <Item value={item.value} helpText={item.type} />,
+        ...getItemProps({ item, index }),
+        highlighted: highlightedIndex === index,
+        className: classnames(selectedItem === item && styles.selected) // Styles does nothing, was testing className pass through
```
jonambas: Can delete this. Should we also filter out selected items from the list? I can use this class to do a `display: none` if thats easier

jasonrhodes: Oh good call, yeah display: none is probably a lot easier than removing/adding from that list

jonambas: actually `selectedItem` probably only points to the most recent selected item

src/pages/reports/components/Typeahead.js
jasonrhodes: hahahahahaha 👌

src/pages/reports/components/Typeahead.js
jasonrhodes: we can probably pass in a prop with selected filters and compare against that list? might be cool to do a style like "greyed out with a check mark" or something

jasonrhodes: or just hide idc

PR #75, opened 2017-09-11
Baseline Tests (FAD-5258)

jasonrhodes:
See #70 for original discussion.


jasonrhodes:
Copied from #70 @jgzamora 

* Structure
	* `/tests` folders holding tests for everything at that level
	* `Filename.test.js` What’s a spec? 
	* Only structural thing I’m against is test files just laying next to the actual file.
	* Other than that idrc about the structure. This feels fine, but I don’t mind if we opt for a larger `/tests/` folder outside `/src` 
* General Strategy
	* As much snapshot testing whenever possible
	* Use enzyme - https://github.com/airbnb/enzyme
* Strategy for simple SFCs
	* Parent components 
		* Shallow wrapper
		* Snapshot tests 
		* Simulate events
		* unit / snapshot test helpers if extracted
		* example: `src/components/Collection/tests/Pagination.test.js`
	* Base / Non-matchbox components
		* Use a render wrapper (get all the html) 
		* Snapshot tests
		* Simulate events
		* unit / snapshot test helpers if extracted
		* example: `src/components/SparkPost/SparkPost.js`
* Classes
	* I opted to for exporting the Classes themselves for testing. This could be avoided by a tool like rewire, I punted on this.
	* Class with no lifecycle methods:
		* use a shallow wrapper
		* Snapshot test the class
		* simulate events with `.find().sumulate()`
		* example: `src/components/ApiErrorBanner/tests/index.test.js`
	* Lifecycle methods:
		* Mount wrapper
		* Sinon Spies for lifecycle methods
		* `.setProps` / the rest of the enzyme dom manipulation API
		* `.find().simulate` still works
		* Shallow + Snapshot test also
		* example: `src/components/GlobalError/tests/GlobalError.test.js`
* Connected components
	* Use all the Class strategies that apply but also:
	* Test mapStateToProps if it has logic (export or tool problem applies here)
	* Trust the exported  `connect()`  HOC to work without tests (suggested by core team)
	* example: `src/components/GlobalError/tests/GlobalError.test.js`
* `withRouter()`
	* Try testing like a Connected component if possible
	* OR: Figure out Mount wrapping the (I didn’t)
	* Snapshot test would be wrapped with  ``` <MemoryRouter>[…]<MemoryRoute/>```
* Actually implemented
	*  Most components in `src/components`
		* Missed `/Collection/Collection.js`, skipped it initially because of `withRouter()` and forgot to circle back
* Not implemented but probably should done here:
	* Reducer tests, looks simple enough. Snapshots can be easily used
* Not implemented but should be separate tickets
	* Page tests, lol. Sorry if this was meant to be the main part of the ticket, but I spent too much time:
		* reading about testing, have fun: [react-redux-links/react-redux-testing.md at master · markerikson/react-redux-links · GitHub](https://github.com/markerikson/react-redux-links/blob/master/react-redux-testing.md)
		* trying out types of testing I later abandoned
	* I’d make tickets for testing each page, so we can get good some good nuanced implementation and spread out the testing know-how
	* In depth `withRouter()` tests, wasn’t able to figure it out in time. We might not need to worry about this other than snapshotting inside a MemoryRoutes component
	* We should set up a way of grabbing any function we want even if it’s not exported
	* I punted on actions until middleware vs. promise chain is resolved
* Miscellaneous 
	* Set some globals in `config/jest/jestSetup.js`
	* jest-in-case for running lots of cases. What’s an “m-sys test suite”??

Wish I would’ve started with pages and gotten some of the meatier stuff implement. Fwiw I think the the components folder covered a good range of general scenarios that should cover testing most Pages, with some caveats like mocking actions.

jasonrhodes:
I added a commit with some examples of straight-up unit tests for helpers, which use jest's manual mocks. (The `__mocks__` folders described here: https://facebook.github.io/jest/docs/en/manual-mocks.html)

Last thing I would like to see here is an example of how to do enzyme tests that aren't snapshot, as well as deciding on how to organize snapshot vs other tests. Then I think we're set here.

PR #77, opened 2017-09-11
Template Form Validation FAD-5288

### LINKED ISSUE #71: Templates - empty form when draft does not exist
If published exists, we should copy that template data into draft data.

jonambas:
- Template single GETs now get stored under `templates.byId`
- Closes #71 
- Adds form validation:
  - Name
    - required
  - ID
    - required
    - valid ID
  - From email
    - required (this one is split up into two fields in old ui)
    - valid email or substitution
  - Reply to
    - valid email or substitution
  - Subject
    - required
  - Content HTML/Text
    - At least one required

Not included:
- from email uses a verified sending domain
- content html syntax errors


src/pages/Templates/ListPage.js
```diff
@@ -58,9 +58,10 @@ class ListPage extends Component {
     const { loading, error } = this.props;
 
     // No Templates (not error case)
-    if (!loading && templatesCount === 0) {
-      return <Redirect to="/templates/create" />;
-    }
+    // This is broken on a hard refresh
+    // if (!loading && templatesCount === 0) {
+    //   return <Redirect to="/templates/create" />;
```
jonambas: This redirect goes through on a hard refresh on the `/templates` route, even if templates exist. Think we need to figure out how we want to handle empty list states across the app.

jonambas:
Also, figured out that <Field> components still work even without the connecting to redux-form in the same component (connected through parent page). Redux-form being initialized 3 times: parent page, form, and editor - but now only does it once.

This _might_ fix that ace-editor bug?

src/pages/Templates/ListPage.js
jonambas: Made a new issue for this sparkpost/matchbox/#49

jonambas:
Updated with all the folder names

PR #91, opened 2017-09-20
FAD-5490 Create/Edit API Keys

tough-griff:
Create/Update/Delete functionality!


src/actions/subaccounts.js
```diff
@@ -1,11 +1,19 @@
...
 
-export function list() {
-  return sparkpostApiRequest({
-    type: 'LIST_SUBACCOUNTS',
-    meta: {
-      method: 'GET',
-      url: '/subaccounts'
+export function list(force = false) {
+  return (dispatch, getState) => {
+    if (!force && getState().subaccounts.listLoaded) {
```
tough-griff: What do we think of this pattern?

src/pages/credentials/helpers/validIpList.js
```diff
@@ -0,0 +1,16 @@
+import { trim } from 'lodash';
+
+const ipRegex = /^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])(\/([0-9]|[1-2][0-9]|3[0-2]))?$/;
```
tough-griff: This came from the angular app. Maybe pull in a library for this?

src/selectors/credentials.js
```diff
@@ -0,0 +1,63 @@
+import { find, fromPairs, isEmpty, join, keyBy, map, size } from 'lodash';
+import { createSelector } from 'reselect';
```
tough-griff: It might look like I went a little overboard with selectors, but I think this might give us a pretty helpful perf gain.

We should go over this if anyone wants a `reselect` primer.

src/actions/subaccounts.js
jasonrhodes: I have such mixed feelings on this. I think we're mostly better off just always doing the GET so we never get weird stale data in unexpected places but also when there are a ton this will be nice ... I honestly don't know (but if we do cache then yeah we need a force flag)

src/pages/credentials/helpers/validIpList.js
jasonrhodes: lol ... that is wow

jasonrhodes: library is gonna add them kbs though ... so not sure?

src/selectors/credentials.js
jasonrhodes: Yeah we should get a quick reselect primer from you ... maybe i can sit with you about it today and then share with folks. I'm all about selectors though, you sold me.

src/actions/subaccounts.js
jasonrhodes: No matter what, I prefer `list({ force: true })` instead of `list(true)`

src/pages/credentials/helpers/validIpList.js
jonambas: this is also in `helpers/validation`

src/pages/credentials/components/GrantsCheckboxes.js
```diff
@@ -0,0 +1,38 @@
...
+import { Field } from 'redux-form';
+import { Grid } from '@sparkpost/matchbox';
+
+import { CheckboxWrapper } from 'components/reduxFormWrappers';
+import styles from './GrantsCheckboxes.module.scss';
+
+const cx = classnames.bind(styles);
+
+const GrantsCheckboxes = ({ grants, show }) => {
+  const grantFields = map(grants, (grant) => (
```
jgzamora: May look into the new Fields array component for this.

src/pages/credentials/components/ApiKeyForm.js
```diff
@@ -0,0 +1,109 @@
...
+const valueSelector = formValueSelector(formName);
+
+const mapStateToProps = (state, props) => ({
+  grants: getGrants(state),
+  subaccounts: state.subaccounts.list,
+  showGrants: valueSelector(state, 'grantsRadio') === 'select',
+  isNew: getIsNew(state, props),
+  initialValues: {
+    grantsRadio: getInitialGrantsRadio(state, props),
+    subaccount: getInitialSubaccount(state, props),
```
jgzamora: Is there value in doing this instead of extracting it directly from the `props.apiKey.subaccount_id`

src/pages/credentials/components/ApiKeyForm.js
```diff
@@ -0,0 +1,109 @@
...
+          component={SubaccountTypeaheadWrapper}
+          subaccounts={subaccounts}
+        />
+        <Field
+          name="grantsRadio"
+          component={RadioGroup}
+          title="API Permissions"
+          options={grantsOptions}
+        />
+        <GrantsCheckboxes grants={grants} show={showGrants} />
```
jgzamora: Is it ok if I give this checkbox layout a bit of UI/UX though? This long ass list of checkboxes is hard to digest and I want to try some organization

src/pages/credentials/DetailsPage.js
```diff
@@ -0,0 +1,109 @@
...
+      </Layout.App>
+    );
+  }
+}
+
+const mapStateToProps = (state, props) => {
+  const { error, grants, keys } = state.credentials;
+
+  return {
+    apiKey: getApiKey(state, props),
```
jgzamora: So this is only possible because the list endpoint returns all needed information for updates right? There are resources that won't be able to use this same pattern because they will *have* to make a call to their GET /resource/{id}

src/pages/credentials/DetailsPage.js
```diff
@@ -0,0 +1,109 @@
...
+    return updateApiKey(values).then((res) => {
+      history.push('/account/credentials');
+    });
+  };
+
+  render() {
+    const { apiKey, loading } = this.props;
+
+    return (
+      <Layout.App loading={loading}>
```
jgzamora: Gotta display the  short key on this page as well. Probably a good time to look into secondary / subtitle information on the `<Page>` component if it doesn't already exist.

src/pages/credentials/components/GrantsCheckboxes.js
```diff
@@ -0,0 +1,38 @@
...
+    // TODO wrap me in a tooltip
+    <Field
+      key={grant.key}
+      name={`grants[${grant.key}]`}
+      label={grant.label}
+      component={CheckboxWrapper}
+      type="checkbox"
+      normalize={(value) => (value ? 'true' : '')} // TODO is this necessary? otherwise we get proptype warnings
```
jgzamora: Will look into

src/__snapshots__/App.test.js.snap
```diff
@@ -77,6 +77,15 @@ exports[`renders without crashing 1`] = `
     <withRouter(Connect(_ProtectedRoute))
       component={[Function]}
       exact={true}
+      path="/account/credentials/create"
+    />
+    <withRouter(Connect(_ProtectedRoute))
+      component={[Function]}
+      path="/account/credentials/details/:id"
```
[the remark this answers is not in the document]
jasonrhodes: Yeah i think "whatever-else" in this case would actually be details right?

src/actions/helpers/credentials.js
```diff
@@ -0,0 +1,25 @@
+import { pickBy, trim, keys } from 'lodash';
```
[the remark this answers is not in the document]
jasonrhodes: Yeah i'm not sure if this matters a TON, but I always `import _ from 'lodash'`

src/pages/credentials/DetailsPage.js
```diff
@@ -0,0 +1,109 @@
...
+  Component: Link,
+  to: '/account/credentials'
+};
+
+export class CredentialsDetailsPage extends Component {
+  static defaultProps = {
+    apiKey: {}
+  };
+
+  state = {
```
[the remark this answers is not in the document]
jasonrhodes: meh? this pattern is nice and lets us remove a constructor once we no longer actually need it so i'm ambivalent here

src/pages/credentials/components/GrantsCheckboxes.js
jonambas: I can change this in mb, but i think you'd want value to always be a boolean

`parse={(value) => !!value}`

src/pages/credentials/components/GrantsCheckboxes.js
jasonrhodes: Tell me more about this but my knee-jerk reaction is if it's working we should probably leave it alone for now until we've got more of the app in a better place and can refactor later?

src/pages/credentials/components/ApiKeyForm.js
jasonrhodes: Yeah I'd just work with @jonambas and make sure we consider it holistically instead of getting too far down a road on one single implementation, but there are lots of little things like this throughout the app right now that just need a little clean-up/tightening i think

src/App.js
```diff
@@ -47,9 +47,9 @@ export default () => (
       <ProtectedRoute path='/webhooks/details/:id' component={webhooks.DetailsPage}/>
 
       <ProtectedRoute exact path='/account/billing' component={BillingPage} />
-      <ProtectedRoute exact path='/account/credentials' component={credentials.ListPage} />
-      <ProtectedRoute exact path='/account/credentials/create' component={credentials.CreatePage} />
-      <ProtectedRoute path='/account/credentials/details/:id' component={credentials.DetailsPage} />
+      <ProtectedRoute exact path='/api-keys' component={credentials.ListPage} />
```
jasonrhodes: these all still need to be `/account/api-keys` right?

jgzamora: So the api just `v1/api-keys`, so I was matching that.

src/selectors/accountBillingInfo.js
```diff
@@ -33,8 +36,8 @@ export function selectCurrentPlan(state) {
  */
 export function selectIpPools(state) {
   return flow(
```
jasonrhodes: `_fp.flow`

PR #94, opened 2017-09-21
Billing Forms

jonambas:
Adds:
- PlanPicker
- BillingAddressForm
- BillingContactForm (similar to above but still different)
- PaymentForm
- DedicatedIpsForm (missing max IPs)


src/components/planPicker/PlanPickerWrapper.js
```diff
@@ -0,0 +1,65 @@
...
+  // List of plans in the dropdown
+  plans: PropTypes.arrayOf(PropTypes.shape({
+    code: PropTypes.string.isRequired,
+    volume: PropTypes.number.isRequired,
+    monthly: PropTypes.number.isRequired,
+    overage: PropTypes.number.isRequired
+  })).isRequired,
+
+  // The account.subscription object
+  subscription: PropTypes.shape({
```
jonambas: Probably should make optional for the sign up flow

jonambas: Probably should make optional for the sign up flow

src/components/planPicker/readme.md
```diff
@@ -0,0 +1,35 @@
+## PlanPicker Usage
```
jonambas: Is this okay where it is? Or should it be under /docs?

jasonrhodes: I think either is fine for now--we can move to one or the other pretty easily if we decide one works better.

src/components/planPicker/PlanPickerWrapper.js
```diff
@@ -0,0 +1,65 @@
+import React from 'react';
```
jasonrhodes: I wonder if this particular file makes sense in the `src/components` level, or if it should be down with the billing page since it's sort of a more specific usage of this picker thing?

src/components/planPicker/tests/PlanPicker.spec.js
```diff
@@ -0,0 +1,68 @@
...
+  const wrapper = mount(<PlanPicker {...selectedProps} />);
+  expect(wrapper).toMatchSnapshot();
+});
+
+it('should focus on input when opened', () => {
+  const wrapper = mount(<PlanPicker {...props} />);
+  const openSpy = jest.spyOn(wrapper.instance(), 'handleOpen');
+  expect(openSpy).not.toHaveBeenCalled();
+
+  // This passes and I have no idea why it works
```
jasonrhodes: lol should it not pass? can you make it fail by tweaking it?

jonambas: If you remove one of these, `handleOpen` doesnt fire. Just not sure why I need to simulate the click AND call `openMenu`
```
wrapper.find(Downshift).node.openMenu();
wrapper.find('a').at(3).simulate('click');
```

src/selectors/accountBillingInfo.js
```diff
@@ -1,10 +1,44 @@
...
+export function overviewProps(state) {
+  const { subscription } = state.account;
 
   return {
-    hasBillingAccount: !!billing,
-    ...subscription
-  }
+    loading: !Object.keys(state.account).length || state.billing.plansLoading,
+    hasBillingAccount: !!state.billing,
+    plans: selectPublicPlans(state),
```
jasonrhodes: We'll need to re-write some of this because we have to be able to check for "secret" plans at a minimum (bc they can appear in the UI if a query param is present), and we'll eventually have to account for aws plans but I'm not totally sure how those are going to surface. So ... let's leave it as-is with a stake in it that we'll have to revisit.

src/selectors/accountBillingInfo.js
```diff
@@ -1,10 +1,44 @@
...
+    ? _.find(state.billing.plans, { code: state.account.subscription.code })
+    : {};
+}
+
+/**
+ * Gets IP pools and formats for select options
+ */
+export function selectIpPools(state) {
+  return state.ipPools.list
+    ? _.chain(state.ipPools.list)
```
jasonrhodes: From what I understand from @tough-griff, if we use `_.chain` it'll pull in all of lodash to our bundle instead of just the tree-shaken parts. Not sure if we use this in other places already but if that's true, we'll need to rip it out everywhere :P

src/components/planPicker/tests/PlanPicker.spec.js
jasonrhodes: ahh ok weird

src/selectors/accountBillingInfo.js
jasonrhodes: Just checked and it's true, _.chain methods are not supported by our babel-plugin-lodash tree-shaker.

In this case I think you could do something like:
```
import { reject, sortBy, map, flow } from 'lodash/fp';

return state.ipPools.list
  ? flow(
    reject({ id: 'default' }),
    sortBy(({ name }) => name.toLowerCase()),
    map(({ name, id }) => ({ label: `${name} [${id}]`, value: id }))
  )(state.ipPools.list) : []
```
More on this:
* https://github.com/lodash/babel-plugin-lodash#limitations
* https://medium.com/making-internets/why-using-chain-is-a-mistake-9bc1f80d51ba

src/components/planPicker/PlanPickerWrapper.js
jonambas: Yeah, it was pretty heavy on its own and didn't want to clutter `billing/components`. Could move it back down but probably into its own folder `billing/planpicker`

PR #97, opened 2017-09-27
FAD-5307 Role-based access control for routes and nav

jasonrhodes:


src/components/ProtectedRoute.js
```diff
@@ -1,23 +0,0 @@
-import React, { Component } from 'react';
```
jasonrhodes: This just moved to src/components/auth/ ... not sure why git didn't pick that up

src/components/auth/ProtectedRoute.js
```diff
@@ -0,0 +1,42 @@
...
+
+export class ProtectedRoute extends Component {
+
+  renderComponent(propsFromRoute) {
+    const { component: Component, grants = [], condition } = this.props;
+
+    if (grants.length === 0) {
+      // TODO: make this render the loading page somehow??
+      return null;
+    }
```
jasonrhodes: if we don't do this check on `grants.length`, the component first mounts before we have the grants which means they won't have access to any page that requires any grants check, and then the redirect logic kicks in and sends them to `/dashboard` before we can actually figure out if they're allowed to see this route.

i'm all ears if there's a better way to do this! it was flashing an all white screen for a second or so before so i thought we could load that generic loading screen in that case, but now that i've moved the grants into the cookie it's pretty fast.

jasonrhodes:
So to help with this review:

* The initial discussion is #98 
* I included docs for how most of this works: https://github.com/SparkPost/2web2ui/blob/hotfix/FAD-5307/docs/access-control.md
* Here are some notes about how the grants stuff is working:

We get grants from `/authenticate/grants` on every login, but the `<ProtectedRoute>` component has to wait until we have those grants before it can decide how to render/redirect. That was causing a 1 second flash of white screen, so I opted to store the grants in the auth cookie like we do in the current app. Getting them out of the cookie is very fast so things feel pretty snappy.

One problem I ran into with putting the grants in the cookie: I am storing the grants as an array of objects in the Redux store, each object has a key, a label, and a description. Trying to store that whole thing in the cookie resulted in a silent failure to write that cookie, I'm assuming because it was too large. So for now I'm storing the grants as a list of keys only, but instead of re-writing everything else to assume we only ever have the keys, if I find the grants in the cookie I store them mapped as objects with just the `key` property. 

tl;dr there's lots of map and unmap happening that we could probably clean up one way or another.

src/components/auth/ProtectedRoute.js
jonambas: `return <Layout.Form loading={true} />`

or `<Layout.App>` to show the nav - they're logged and inside the app when this hits right?

src/components/navigation/Navigation.js
```diff
@@ -14,8 +15,13 @@ class Navigation extends Component {
     open: false
   }
 
+  constructor(props) {
+    super(props);
+    this.state.navItems = prepareNavItems(this.props.store);
```
jonambas: So `Layout` (and `Navigation`)  is re-mounted on every route, so this is going to run on every route change. Wonder if theres a way to do this once on log in.

Or do you think this isn't expensive?

src/config/default.js
```diff
@@ -30,7 +30,10 @@ const config = {
     { time: 60 * 24 * 190, value: 'week', format: 'MMM Do' },
     { time: Infinity, value: 'month', format: 'MMM YY' }
   ],
+  showThingOnDash: true,
+  showSummaryReport: true,
```
jonambas: Are these just example flags?

src/App.js
```diff
@@ -1,58 +1,31 @@
...
-    </div>
-  </Router>
-);
+export default function App() {
+  return (
+    <Router>
+      <div>
+        <AuthenticationGate />
+
+        {routes.map((route) => {
```
tough-griff: If we're not using nested routes in the `routes` config, we could wrap this collection with a [`<Switch>`](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/Switch.md) to prevent a bunch of `<!-- react-emtpy: * -->`s.

src/App.js
```diff
@@ -1,58 +1,31 @@
...
+          const MyRoute = route.public ? Route : ProtectedRoute;
+          route.exact = !(route.exact === false); // this makes exact default to true
+
+          if (route.redirect) {
+            return <Route key={route.path} {...route} render={() => <Redirect to={route.redirect} />} />;
```
tough-griff: `<Switch>` also allows us to use a nice shorthand here:
```jsx
<Redirect from={route.path} to={route.redirect} />
```

docs/access-control.md
```diff
@@ -0,0 +1,106 @@
...
+  }
+]
+```
+
+_Note: Routes default to exact: true because [it makes more sense](https://github.com/ReactTraining/react-router/issues/4958)._
+
+
+## Condition helpers
+
+Condition helpers are functions that return a condition function. Any time you need to use a condition, create a helper for that condition scenario. Whenever that returned function is called, it will be passed an object with keys `store` (the entire Redux store) and `config` (the current loaded tenant config), and it should return a boolean that reflects the access decision.
```
tough-griff: `store` here might better be referred to as `state`, since we are getting the current state tree, not the Redux `store` object.

It's picky, but the `store` object itself contains more than just the state, and would require `getState()` calls.

src/components/auth/AccessControl.js
```diff
@@ -0,0 +1,20 @@
...
+    const show = condition({ store, config });
+
+    if (redirect && !show) {
+      return <Redirect to={redirect} />;
+    }
+    return show ? children : null;
+  }
+}
+
+export default connect((store) => ({ store }))(AccessControl);
```
tough-griff: The component doesn't need to be aware of all state here:
```jsx
export const AccessControl = ({ children, redirect, show }) => {
  if (redirect && !show) return <Redirect to={redirect} />;
  return show ? children : null;
}

const mapStateToProps = (state, { condition = accept }) => ({
  show: condition({ state, config }),
});

export default connect(mapStateToProps)(AccessControl);
```

This could help perf by preventing rerenders on prop changes that the route doesn't care about.

src/components/auth/ProtectedRoute.js
```diff
@@ -0,0 +1,42 @@
...
+      )}/>
+    );
+  }
+}
+
+const mapStateToProps = ({ auth, currentUser = {}}) => ({
+  auth,
+  grants: currentUser.grants || []
+});
+export default withRouter(connect(mapStateToProps)(ProtectedRoute));
```
tough-griff: I think `withRouter` here is redundant, the `Route` component will pass the relevant props to it's children.

src/components/auth/ProtectedRoute.js
```diff
@@ -0,0 +1,42 @@
...
+        <Component {...propsFromRoute} />
+      </AccessControl>
+    );
+  }
+
+  render() {
+    const { component, auth, condition, ...rest } = this.props;
+    return (
+      <Route {...rest} render={(props) => (
+        auth.loggedIn ? this.renderComponent(props) : (
```
tough-griff: does `AuthenticationGate` take care of this logic for us? is it necessary to check logged in status in each route?

src/components/auth/ProtectedRoute.js
jasonrhodes: Yes they're logged in, that sounds good to me (although will it be nav + blank page?)

src/components/navigation/Navigation.js
jasonrhodes: We could memoize some of it but that could get tricky… see my reply to Griffin's comment about moving Nav outside

src/config/default.js
jasonrhodes: Yeah, I'll remove before merge.

jasonrhodes:
@tough-griff since GitHub doesn't let me directly reply to general comments 😠 , this is re:
>I think wrapping the router in Layout, which could provide the navigation components outside of each individual route could help with loading/race condition issues, as well as prevent recomputing the nav component on each page. It might take some refactoring, but since that content is present on all pages it shouldn't need to be re-rendered all the time.

If we can figure out how to do this it seems fine to me. One thing I've been thinking about is how to handle dynamic conditions in all of this access control logic, so if a condition is something like:
```js
// helpers/conditions/accountStatus.js
export default function(status) {
  return ({ state }) => state.account.status === status
}

// config/routes.js
{
  path: /new/cool/page,
  condition: accountStatus('active')
}
```
What happens if an account is suspended while they're in the app? There are places where we refresh the account but if their account status updates, it'd be nice to re-render the routes and nav so that the conditions are re-checked. But having those run on every state change is, I agree, not a great solution to that.

It's not a huge deal that we solve this right now because we don't have conditions that need to be this dynamic, but I'd hate to build a system that can't handle dynamic conditions. If anyone has thoughts on how we could get this behavior, let me know!

tl;dr @jonambas if you think we can get the nav stuff to render outside of all the regular layouts, I think that'd probably be good for rendering/perf etc.

docs/access-control.md
jasonrhodes: This makes sense. I'm weirded out by calling it `state` bc then you technically have `this.props.state` which is super wacky to me ... could maybe use `globalState` but if `state` is a convention I can be convinced to get over it.

jonambas:
re: @jasonrhodes 
Yeah, I think it'll take a bit of rework.

- We'd lose the `<Layout loading>` prop on all the pages, and would have to replace with the `<Loading>` component. 
- Would need to figure out how to load the right layout by route. Most protected routes would use `App` but there are a few that use `Form` - sign up flow, and I think a TOU page. Maybe we could pass it through routes.js:
```
{
  path: '/templates',
  layout: Layout.App
},
```

src/components/auth/ProtectedRoute.js
jasonrhodes: Good call, I must've been doing some kind of `history.push` here at one point that's no longer around. Removing.

src/components/auth/ProtectedRoute.js
jasonrhodes: Yeah not every route is protected, so we can't do that check in `<AuthenticationGate>` ... tbh I _think_ the only thing we even do in `<AuthenticationGate>` is initiate the check for the cookie and issue a login action if we find it, otherwise we could _probably_ handle the login/logout redirect stuff inside various chained actions, but I lost 4 or 5 days on that when I first set it up so I am a bit scared of messing with it at this point lol

src/components/auth/AccessControl.js
jasonrhodes: Interesting, I'm not sure if I follow how this will lead to less re-renders? It's arguably cleaner so I still kind of like it ... I guess `condition` will be run the same number of times but unless it produces a different bool value, the component won't be re-rendered?

tough-griff: Yup! Passing the entire `state` object will trigger a lot of prop changes when the output would not otherwise change.

docs/access-control.md
tough-griff: As mentioned below, we can and probably should avoid passing the entire `state` object as a prop. `mapStateToProps` exists specifically to limit the scope of props we're passing to a connected component, and these condition helpers can run there.

tough-griff:
re: re: @jonambas @jasonrhodes regarding `<Layout loading>` vs `<Loading>`, I think a `Loading` component that behaves similar to a modal (using React 16 `Portals` even!?) is not a bad pattern, but we could also trigger the `loading` prop similarly to how we handle global errors, with specific actions being fired to handle loading state.

docs/access-control.md
jasonrhodes: Yeah the only reason I needed to do it was so I could provide wholly to condition functions. I'll keep an eye on it, it's still not totally clear what the difference is between some of the ways I've seen us doing things.

src/components/auth/AccessControl.js
jasonrhodes: Gotcha, I think my head is getting around this, good call.

PR #99, opened 2017-10-02
FAD-5464 Users List Page + Set Role, Delete

bkemper:


bkemper:
@jgzamora Let me know if you need anything else.

src/pages/users/ListPage.js
```diff
@@ -0,0 +1,77 @@
...
+export class ListPage extends Component {
+  componentDidMount() {
+    this.props.listUsers();
+  }
+
+  // Do not allow current user to change their access/role or delete their account
+  getRowData = (user) => [
+    user.name,
+    <AccessSelect
+      disabled={user.isCurrentUser}
```
jgzamora: Later on I think a lot of this can be avoided by extracting the current user and displaying it in a different way at the top of the page, if that makes sense. It would be annoying to look for your user in the list if you have a ton of users.

src/pages/users/ListPage.js
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  render() {
+    const { loading, users } = this.props;
+
+    return (
+      <Layout.App loading={loading}>
+        <Page title="Users" />
+        {this.renderError()}
+        <TableCollection
```
jgzamora: I think I agree with loading the table as well as the error, but the table deff needs an empty state. If no error, that would be a nice place to put a centered CTA for creating the first of that resource. If error, it could just have text, or even load the banner inside the table. Maybe @jonambas has some ideas
screen shot 2017-10-03 at 8 16 06 pm

jgzamora: not to be done in this ticket ofc

src/pages/users/ListPage.js
```diff
@@ -0,0 +1,77 @@
...
+
+  renderError() {
+    const { error, listUsers } = this.props;
+
+    if (!error) { return null; }
+
+    return (
+      <ApiErrorBanner
+        errorDetails={error.message}
+        message="Sorry, we seem to have had some trouble loading your API keys."
```
jgzamora: api keys error. 

aside: I think that some of this messaging can be taken care of in the ApiErrorBanner component. Maybe at some point later we can just pass what action we were trying to do and on what resource. 
```
action="PUT"
resource="webhook"
```
"Oh no, looks like there was an error updating this webhook"

src/pages/users/ListPage.js
```diff
@@ -0,0 +1,77 @@
...
+import { Page } from '@sparkpost/matchbox';
+
+import * as usersActions from 'src/actions/users';
+import { selectUsers } from 'src/selectors/users';
+
+import ApiErrorBanner from 'src/components/apiErrorBanner/ApiErrorBanner';
+import TableCollection from 'src/components/collection/TableCollection';
+import Layout from 'src/components/layout/Layout';
+
+import AccessSelect from './AccessSelect';
```
jgzamora: Please move these two into a `/components` folder

src/pages/users/DeleteButton.js
```diff
@@ -0,0 +1,53 @@
...
+
+    // Display nothing when disabled to match current UI style
+    if (disabled) { return null; }
+
+    return (
+      <span>
+        <Button plain onClick={this.onOpen}>
+          <Icon name="Delete"/>
+        </Button>
+        <Modal open={this.state.open}>
```
jgzamora: Does this mean we are rendering a modal per user on the page? If so, is that something we should try to avoid?

src/pages/users/AccessSelect.js
```diff
@@ -0,0 +1,29 @@
...
+
+const OPTIONS = [
+  { label: 'Administrator', value: 'admin' },
+  { label: 'Reporting', value: 'reporting' }
+];
+
+export default function AccessSelect({ disabled = false, user, ...props }) {
+  // Only return text to match the current style
+  // TODO: won't need the span tags with v16
+  if (disabled) {
```
jgzamora:
```
if (user.isCurrentUser) {
```
and avoid the extra prop?

bkemper: This was suppose to be a dumb component that does not need to know about `user`.  What do you think if I refactor the interface to this?

```
/**
 * @param {boolean} disabled
 * @param {Function} onChange - event handler passes new
 * @param {null, string} value
 */
export default function AccessSelect({ disabled = false, onChange = _.noop, value = null }) {
```

src/pages/users/DeleteButton.js
bkemper: I didn't feel good about this either.  I think it was the `open` prop that was throwing me off.  How about I refactor this to only render `Modal` if `this.state.open === true`?

```
renderModal() {
  if (!this.state.open) return null;
  return <Modal open />;
}

render() {
  return (
    <span>
      <Button />
      {this.renderModal()}
    </span>
  );
}
```

src/pages/users/ListPage.js
bkemper: That is a great suggestion.  Who should we include to make sure this gets considered?

src/pages/users/ListPage.js
jonambas: yup https://github.com/SparkPost/matchbox/issues/49

src/pages/users/ListPage.js
bkemper: Yikes!  Nice catch.  I'll change this message.

What would you think about using an internationalization library (e.g. [react-i19next](https://github.com/i18next/react-i18next), [react-intl](https://github.com/yahoo/react-intl), etc.)?

src/pages/users/ListPage.js
bkemper: I dig it!

src/pages/users/DeleteButton.js
jonambas: Yeah this should be on the ListPage

jonambas: There is also a delete modal in `src/components` that you can use

bkemper: Oh, I think I miss understood @jgzamora's request.  The reason I included the `Modal` with the `Button` is to avoid bloating the `ListPage` and to isolate this functionality.

src/pages/users/tests/DeleteButton.test.js
```diff
@@ -0,0 +1,24 @@
...
+import fp from 'lodash/fp';
+import React from 'react';
+
+import DeleteButton from '../DeleteButton';
+
+const defaultProps = {
+  user: {},
+};
+
+const TestComponent = (props = {}) => {
```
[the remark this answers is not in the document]
bkemper: I tinkered with writing a test helper.  I stopped after you suggested jest-in-case.  This is test needs to be refactored to use jest-in-case.

src/pages/users/ListPage.js
```diff
@@ -0,0 +1,77 @@
...
+    <AccessSelect
+      disabled={user.isCurrentUser}
+      onChange={this.props.updateUser}
+      user={user}
+    />,
+    user.email,
+    user.last_login,
+    <DeleteButton
+      disabled={user.isCurrentUser}
+      onSubmit={this.props.deleteUser}
```
jonambas: Could use the GlobalAlert action here

`this.props.deleteUser().then(() => showAlert({ type: 'success', message: 'User Deleted' }))

src/reducers/tests/users.test.js
```diff
@@ -0,0 +1,22 @@
...
+  const state = {
+    ...initialState,
+    entities: {
+      'test-user': {
+        username: 'test-user',
+      }
+    }
+  }
+
+  expect(usersReducer(state, action)).not.toHaveProperty('entities.test-user');
```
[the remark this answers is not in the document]
bkemper: I originally wrote this as a snapshot test and then got the feeling that not everything should be a snapshot test.  I fear we are going to get into a major refactor that requires several snapshots to change and it is going to be hard to resolve.  Do you have the same concern or am I being ridiculous?

src/pages/users/ListPage.js
jonambas: Oh nevermind, I see you've got them in the action 👍

src/selectors/users.js
```diff
@@ -0,0 +1,8 @@
+import fp from 'lodash/fp';
+
+// Get, reduce, enrich, and sort list of users
+export const selectUsers = ({ currentUser, users }) => fp.flow(
```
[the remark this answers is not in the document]
bkemper: I'll add a todo comment.  I want this to use reselect, but I held off for now because of this comment.

> we can perfect selectors and other refactors later if necessary
> https://msys.slack.com/archives/C5AEHSQUA/p1506431627000150

src/pages/users/DeleteButton.js
jasonrhodes: So IIUC this has been a regular thing with modal rendering ... if we only render the modal when it's supposed to be open, then we don't get the animation in that the modal ships with. Not a huge deal but I think we should try to make it work as intended, meaning the modal would be on the page and only be open if the `open` prop is true.

I think because of this we try to push modals into the component where they appear so they only get rendered once which is why Jon mentioned moving this to the List page. Eventually we'll probably want to play with the new portals API.

src/selectors/users.js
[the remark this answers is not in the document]
jasonrhodes: FWIW I think it's fine to install reselect (I installed in my access control branch too) and throw it in here unless that blows up the ticket by causing some unexpected problems.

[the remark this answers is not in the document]
jgzamora: +1 to do it here

src/reducers/tests/users.test.js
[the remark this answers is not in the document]
jgzamora: all snapshots all the time, every time, forever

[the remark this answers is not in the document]
jgzamora: I'm generally friends towards the idea of snapshots everywhere. I think they should among unit tests whenever they make things easier. I also wouldn't worry about a big refactor yet. 
That said, the reason I suggested a snapshot here was because I thought adding a couple more entities then verifying the complete output of the state is a better test than checking if one property is no longer there. But now I'm worried about property print order, which I don't think is guaranteed, so a snapshot might not actually be a good idea here.

src/pages/users/ListPage.js
jgzamora: I'm most concerned about clear consistent messaging in English in this specific case. As in, engineers shouldn't be writing these sentences every time they make a page. It's error prone, and I'd rather control the tone of messages from a central place. They should just give some info to a component that figures the text out for them.

*If* we were want translations I would totally use one of those. And if the messaging was all handled in components it would be even easier to add that.

src/pages/users/AccessSelect.js
jgzamora: Yup, that works!

src/pages/users/ListPage.js
bkemper: Done... 9b17eaa

src/pages/users/AccessSelect.js
bkemper: Done... a0a29e5

src/pages/users/tests/DeleteButton.test.js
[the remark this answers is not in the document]
bkemper: Done... a0a29e5

src/pages/users/ListPage.js
bkemper: Done... 337883c

src/selectors/users.js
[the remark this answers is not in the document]
bkemper: Done... 932af41

src/reducers/tests/users.test.js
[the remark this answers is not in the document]
bkemper: Done... c282ace

src/pages/users/DeleteButton.js
bkemper: Done... e76e929

bkemper:
@jgzamora This is ready for final review.

PR #100, opened 2017-10-02
Change Plan Page FAD-5557

### LINKED ISSUE #11: Clear state when logging out
I imagine we'll redo all the actions/reducers at some point soon, but right now user and account data still exists even after logout. When logging in with a second account you'll see a flash of the first accounts data before new data comes in.

jonambas:
- Adds a new route at `/account/billing/plan` for `<ChangePlanPage />`
- Three actions can be called from this page
  - `updateSubscription` - plan changes only without updating CC
  - `billingUpdate` - plan change with new CC
    - free accounts with account.billing will go through this, but will not see their saved CC
  - `billingCreate` - free accounts with no account.billing
- Adds three banners that we can use elsewhere
  - These users shouldnt be able to get to this page, but in case they do these are displayed in place of the form
  - `<PendingPlanBanner />`
  - `<SuspendedBanner />`
  - `<ManuallyBilledBanner />`
- Adds this [payment lib](https://github.com/jessepollak/payment)
  - Splits CC expiry date from one field to two: `month` and `year`
  - Generates card type from card number
- **Resets redux state on `LOGOUT`** closes #11 

Also, completely ignored `BillingContactForm` for now until we figure out what fields we need from that.


src/config/default.js
```diff
@@ -14,11 +14,18 @@ const config = {
       'Content-Type': 'application/x-www-form-urlencoded'
     }
   },
+  cardTypes: [
+    { paymentFormat: 'visa', apiFormat: 'Visa' },
+    { paymentFormat: 'mastercard', apiFormat: 'MasterCard' },
+    { paymentFormat: 'amex', apiFormat: 'AmericanExpress' },
+    { paymentFormat: 'discover', apiFormat: 'Discover' }
```
jonambas: The strings the payment lib provides don't match with what our API/Zuora looks for. Using this to get the right strings. We can also use this map to validate card type - these 4 are the types we accept.

jasonrhodes:
This looks really good. Let me take a look at tests more later and we can probably get this merged tomorrow morning, will let you know.

PR #101, opened 2017-10-02
Subaccounts List Page

jgzamora:


src/pages/subaccounts/helpers/tests/getRowData.test.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import { shallow } from 'enzyme';
```
jgzamora: don't need shallow or React I think

src/pages/subaccounts/helpers/getRowData.js
```diff
@@ -0,0 +1,30 @@
...
+
+/*
+ Subaccounts getRowData passed to TableCollection in ListPage.
+*/
+const getRowData = (subaccount) => {
+  const { status, id, name } = subaccount;
+  let className;
+  switch (status) {
+    case 'active':
+      className = styles.StatusActive;
```
jonambas: You can do this instead with the `classnames` lib if your classnames match up with `status`

`className={classnames(status && styles[status])}`

src/pages/subaccounts/helpers/getRowData.js
```diff
@@ -0,0 +1,30 @@
...
+      break;
+    case 'terminated':
+      className = styles.StatusTerminated;
+      break;
+    case 'suspended':
+      className = styles.StatusSuspended;
+      break;
+  }
```
bkemper: What do you think about this?

```
const STATUS_CLASSES = {
  active: styles.StatusActive,
  suspended: styles.StatusSuspended,
  terminated: styles.StatusTerminated,
};

<p className={STATUS_CLASSES[subaccount.status]}>{status}</p>
```

src/pages/subaccounts/helpers/getRowData.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import styles from './rowData.module.scss';
+
+import { Link } from 'react-router-dom';
+
+/*
+ Subaccounts getRowData passed to TableCollection in ListPage.
+*/
+const getRowData = (subaccount) => {
+  const { status, id, name } = subaccount;
```
bkemper: Since `subaccount` isn't used, this destructuring could replace the arguments declaration.

`const getRowData = ({ status, id, name }) => {`

src/pages/subaccounts/helpers/tests/__snapshots__/getRowData.test.js.snap
```diff
@@ -0,0 +1,46 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Get row data active 1`] = `
+Array [
+  <Link
+    replace={false}
+    to="/account/subaccounts/undefined"
```
bkemper: Is this suppose to be `undefined`?

src/pages/subaccounts/helpers/tests/getRowData.test.js
bkemper: I don't think so either.

src/pages/subaccounts/ListPage.js
```diff
@@ -0,0 +1,70 @@
...
+        reload={this.onReloadApiBanner}
+      />
+    );
+  }
+
+  render() {
+    const { error, loading } = this.props;
+    return (
+      <Layout.App loading={loading}>
+        <Page title='Subaccounts' primaryAction={primaryAction}/>
```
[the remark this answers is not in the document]
jgzamora: 🤷🏻‍♂️I don't think that's been thought out in that way, it's just the current API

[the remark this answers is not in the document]
jgzamora: no idea wtf is up that that female emoji with a male sign next to it...

src/pages/subaccounts/ListPage.js
```diff
@@ -0,0 +1,70 @@
+import React, { Component } from 'react';
+import { connect } from 'react-redux';
+import { Link } from 'react-router-dom';
+
+import Layout from 'src/components/layout/Layout';
+import { Page } from '@sparkpost/matchbox';
+import TableCollection from 'src/components/collection/TableCollection';
+import ApiErrorBanner from 'src/components/apiErrorBanner/ApiErrorBanner';
+
+import { list as listSubaccounts } from 'src/actions/subaccounts';
```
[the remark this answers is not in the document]
jgzamora: Not doing cause api keys uses it like 7 times

src/pages/subaccounts/ListPage.js
```diff
@@ -0,0 +1,89 @@
...
+      />
+    );
+  }
+
+  render() {
+    const { error, loading, subaccounts } = this.props;
+
+    return (
+      <Layout.App loading={loading}>
+        { error && this.renderError() }
```
bkemper: Please don't change unless you want to, but I recommend putting this logic in the `renderError` function as a guard clause.

src/pages/subaccounts/helpers/tests/getRowData.test.js
```diff
@@ -0,0 +1,34 @@
...
+  compliance_status: 'terminated'
+};
+
+cases('Get row data', (opts) => {
+  const rowData = getRowData(opts.subaccount);
+  expect(rowData).toMatchSnapshot();
+}, [
+  { name: 'active', subaccount: active },
+  { name: 'suspended', subaccount: suspended },
+  { name: 'terminated', subaccount: terminated }
```
bkemper: Please don't fix unless you want to, but I recommend using the other format as
[an object of objects with the names as the keys](https://github.com/Thinkmill/jest-in-case#usage).  It is a little easier to read.

src/pages/subaccounts/tests/ListPage.test.js
```diff
@@ -0,0 +1,50 @@
...
+beforeEach(() => {
+  wrapper = shallow(<ListPage {...props} />);
+});
+
+it('renders correctly', () => {
+  expect(wrapper).toMatchSnapshot();
+});
+
+it('renders empty state', () => {
+  wrapper = shallow(<ListPage loading={false} error={null} subaccounts={[]} />);
```
bkemper: You may have found another case for why we need a test helper instead of reassigning `wrapper` in `beforeEach`.

jgzamora: I'm somewhere in-between. I don't like the beforeEach, it only works because we have like 2 test cases per page right now. I'm also not trying to write this test helper myself anytime soon. The earlier it gets done the better though.

PR #102, opened 2017-10-05
Install eslint-plugin-jest plugin

bkemper:
While poking around for a quick fix for #103, I noticed `eslint-plugin-jest` wasn't installed.  This is a quick win could catch some errors in our tests.

```
./2web2ui/src/actions/tests/account.test.js
   3:1  warning  Skipped test  jest/no-disabled-tests

./2web2ui/src/pages/webhooks/components/tests/Fields.test.js
  35:1  error  Test title is used multiple times in the same test suite  jest/no-identical-title

./2web2ui/src/actions/helpers/tests/zuoraRequest.test.js
  38:28  error  "toBeDefined" was not called  jest/valid-expect
```


bkemper:
My only concern with this are the "jest/no-disabled-tests" warnings.  I think I saw that warnings on CI are treated as failures which means this would fail every build if there are any skipped tests.

jasonrhodes:
For hotfixes we keep the base "master" and then merge the branch into master and develop locally (using `git finish` is the easiest way via git+) ... when we start doing feature branches then we'd make the base `develop` for those.

bkemper:
@jasonrhodes Okay.  I rebase it on to master.  What is the preferred prefix for a branch like this?

jasonrhodes:
OK i see this branch was created off of develop so I set the base back to develop like you had it @bkemper -- we would usually call that `feature/FAD-xxxx` but since we aren't making much of a distinction right now, no worries.

You can merge this to develop by doing `git finish feature`, then once it's in tst, you can merge develop -> master with `git promote` (adding `push` to the end of each of those commands will auto push after merge for you, if you want)

PR #104, opened 2017-10-05
Analyzing bundle size and testing require.ensure splitting

jasonrhodes:


jasonrhodes:
Swapped out `require.ensure` for `import()` using a small `asyncComponent` helper/HoC thing so that `Editor.js` just wraps and loads `./_Editor.js` and then none of the async handling is actually written into the Editor component at all.

https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/template/README.md#code-splitting
https://serverless-stack.com/chapters/code-splitting-in-create-react-app.html

Also open to using other analyze tools, just grabbed the one recommended by CRA: https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/template/README.md#analyzing-the-bundle-size

PR #106, opened 2017-10-07
Updated deps for react 16, lots of tests failed, leaving it here for now

jasonrhodes:
Note: had to remove the babel plugins "lodash" and "recharts" because one of them caused errors about a plugin returning undefined when it should have returned a function. No idea what react 16 had to do with that but we'll need to figure it out and re-enable those plugins as part of fixing this PR.

### DIFF
--- a/config/jest/setup.js
+++ b/config/jest/setup.js
@@ -1,8 +1,15 @@
+import raf from './tempPolyfills'
 import Enzyme from 'enzyme';
 import Adapter from 'enzyme-adapter-react-16';
+import toJson from 'enzyme-to-json';
 
 // Provides enzyme assertions.
 // See https://github.com/blainekasten/enzyme-matchers#assertions
 import 'jest-enzyme';
 
 Enzyme.configure({ adapter: new Adapter() });
+
+// Fail tests on any warning
+console.error = message => {
+   throw new Error(message);
+};
--- a/config/jest/setup.js
+++ b/config/jest/setup.js
@@ -1,3 +1,8 @@
+import Enzyme from 'enzyme';
+import Adapter from 'enzyme-adapter-react-16';
+
 // Provides enzyme assertions.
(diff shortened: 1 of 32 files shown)


src/pages/profile/tests/__snapshots__/NameForm.snapshot.test.js.snap
```diff
@@ -1048,6 +1087,10 @@ exports[`renders correctly 1`] = `
               >
                 <button
                   className="Matchbox-Button"
+                  disabled={undefined}
+                  onBlur={undefined}
+                  onClick={undefined}
+                  onFocus={undefined}
                   onMouseUp={[Function]}
                   type="submit"
                 >
```
jonambas: Why is this snapshot this long lol

config/jest/setup.js
```diff
@@ -1,3 +1,22 @@
...
 // See https://github.com/blainekasten/enzyme-matchers#assertions
 import 'jest-enzyme';
+
+Enzyme.configure({ adapter: new Adapter() });
+
+// Make Enzyme functions available in all test files without importing
+global.shallow = shallow;
+global.render = render;
+global.mount = mount;
+global.toJson = toJson;
```
jasonrhodes: Let's get rid of these four lines ... I like seeing where they come from in tests where they're used.

jasonrhodes: Oh shit just noticed you already got rid of all the imports. :/ We had actually already removed these globals before ... I think it's nice to know that the only globals in test files are from jest ... hopefully it's easy to add those imports back but if not we can punt.

jasonrhodes:
@jonambas do you mind fixing lint errors seen with `npm run lint -- --fix` and then rebasing?

Otherwise LGTM and good to merge

jasonrhodes:
Actually @jonambas if you rebase first the lint errors may go away

PR #107, opened 2017-10-09
FAD-5653 Fixes Access Control state

jasonrhodes:
Changed the way access control works slightly so that it only passes a limited set of state keys to the access conditions. This makes it possible for the components that rely on those condition helpers in any way to connect using the `accessConditionState` selector, so that they can re-compute the access control logic but only on changes to the things that could affect that logic.

See the changes in `src/components/navigation/Navigation` and the `prepareNavItems` for an example.

This makes access control completely dynamic. If you use Redux Dev Tools to dispatch an event like this:
```js
{ type: 'GET_GRANTS_SUCCESS', payload: [] }
```
you will see the nav update itself in real time without a refresh or route change etc.

Last: I made this dynamic access control robust enough to handle async state loading. In other words, if you rely on something like `account.status` to determine if a route should load and otherwise it should be redirected, the first check might happen before the account has been loaded from the API, so the route will be redirected before we know if the status is valid.

To solve this, I created an `accessControlReady` reducer and action so that the login action now dispatches this, which fetches account, user, and grants, and then dispatches an action of type `ACCESS_CONTROL_READY` -- the `AccessControl` component waits for that flag to be set before redirecting (it only applies to the redirect, so that during the time when we don't know if you're allowed to see something, you won't see it until the state is ready, but you won't be redirected away from the page in that time either).

I'll update the docs to reflect all this. Ping me with questions!


src/config/routes.js
```diff
@@ -93,7 +93,7 @@ export default [
   {
     path: '/templates',
     component: templates.ListPage,
-    condition: hasGrants('templates/modify')
+    condition: composeConditions(hasGrants('templates/modify'), ({ account }) => account.status === 'active')
```
jasonrhodes: This is here as an example of async state -- doing this before the async handling would result in being redirected to `/dashboard` if you tried to load /templates directly. Now it works fine, and also works if you change "active" to "suspended" here (you'll be redirected to /dashboard in that case).

src/selectors/accessConditionState.js
```diff
@@ -0,0 +1,10 @@
+import { createSelector } from 'reselect';
+
+const getAccount = (state) => state.account;
+const getUser = (state) => state.currentUser;
+
+export default createSelector(
+  [getAccount, getUser],
+  (account, currentUser) => ({ account, currentUser })
+);
+
```
jasonrhodes: Not 100% sure if we gain anything at all from this since we're not doing anything besides passing through account and currentUser from the Redux state, but I put it here so it's easy to grab this in one place and if we add keys later, we're set up to do it "correctly" here in a memoized selector.

src/components/navigation/Navigation.js
```diff
@@ -43,26 +41,26 @@ class Navigation extends Component {
...
+              {this.renderItems()}
             </ul>
             <Footer />
           </div>
           <Icon
             name='Close'
             className={styles.close}
             size={33}
-            onClick={() => this.handleClick()} />
+            onClick={this.handleClick} />
```
bkemper: Should we install https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md?

src/actions/accessControl.js
```diff
@@ -0,0 +1,12 @@
...
+
+// initialize some state used for access control
+export function initializeAccessControl() {
+  return (dispatch) => Promise.all([
+    dispatch(fetchAccount({ include: 'usage,billing' })),
+    dispatch(getCurrentUser())
+      .then(({ access_level }) => dispatch(getGrants({ role: access_level })))
+  ])
+  .then(() => dispatch({ type: 'ACCESS_CONTROL_READY' }));
+}
```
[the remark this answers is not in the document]
jasonrhodes: Good call, we need a catch here to avoid unhandled promise rejection. Not sure if we should dispatch an action there because if we've never initialized this stuff, the default value for `reduxState.accessControlReady` will be `false`, and if we've gotten it before and for some reason run this action again, we should just use the state we have and not fail everything.

I'll add the catch, and I think this will be a good place to use sentry to log something for us to be aware of if this is failing a lot.

src/components/navigation/Navigation.js
jasonrhodes: Yeah good idea 👍

src/selectors/navItems.js
```diff
@@ -0,0 +1,47 @@
...
+        return false;
+      }
+    }
+
+    return true;
+  });
+}
+
+export function prepareNavItems(accessConditionState) {
+  console.log('preparing nav items'); // eslint-disable-line
```
bkemper: @jasonrhodes Please remove.

src/selectors/navItems.js
```diff
@@ -0,0 +1,47 @@
...
+  return filterNavByAccess(mapNavToRoutes(navItems), accessConditionState);
+}
+
+export default createSelector([accessConditionState], prepareNavItems);
```
bkemper: 👍 Much better!

src/selectors/navItems.js
jasonrhodes: Why didn't linting catch this HMMM 🤔

PR #108, opened 2017-10-10
Billing Summary FAD-5559

jonambas:
Split up billing forms into 3 layers:

**Pages**
Decides what banners & forms should be shown
- ChangePlanPage
- SummaryPage

**Form Containers**
Wrapped with redux-form. Handles submit actions & props required by the fields below.
- ChangePlan
- UpdatePayment
- UpdateContact
- AddIps ( Not completed in this ticket )

**Form Fields**
Sets of redux-form fields. Handles logic and validation at the field level.
- BillingAddressForm
- BillingContactForm 
  - only difference with above is this one does not hide firstname/lastname if they exist, and the email field
- PaymentForm
- PlanPicker ( in src/components but should probably be moved down into billing )
- DedicatedIpsForm ( Not completed in this ticket )

--- 

**States/Flows we should check:**
- Free account -> Upgrade
- Paid account ->
  - Plan change with saved CC
  - Plan change with new CC
  - Update payment info
  - Update billing contact
- Manually billed -> 
  - New zuora account on same plan ( if on an existing zuora plan )
  - New zuora account with new plan
- Pending subscription change ( downgrade )
  - For both manually billed & self serve
- Suspended due to billing ->
  - Update payment info

Should also verify that each of these states show or hide the right components.

---
**TODO:**
- [x] Finish tests

Not done in this ticket?
- Field level validation
- Dedicated IPs


src/components/layout/Layout.module.scss
```diff
@@ -48,6 +48,7 @@
 
 .container {
   max-width: rem(1020);
+  height: 100%;
```
jonambas: Remove

src/pages/billing/formContainers/UpdateContact.js
```diff
@@ -0,0 +1,59 @@
...
+  componentDidMount() {
+    this.props.getBillingCountries();
+  }
+
+  onSubmit = (values) => {
+    const { updateBillingContact, onCancel, reset, showAlert } = this.props;
+    return updateBillingContact(values).then(() => {
+      showAlert({ type: 'success', message: 'Billing Contact Updated' });
+      onCancel();
+      reset();
```
jonambas: This closes the modal and resets the redux form on successful submit.

jasonrhodes:
Some general comments as I use the app, before I dip into code specifics etc

**On the `appteam` account which is manually-billed:**

1.  I clicked on "Enable Automatic Billing" and the next view doesn't give me much context, it would be nice to have messaging appear only if you're manually billed? The button text here says "Upgrade" but should probably also reflect the "Enable Automatic Billing" language.

1. Not sure if we should offer the free plan here. Is it easy to default to the highest plan instead of the lowest in this case?

1. When you select a paid plan, the credit card form appears and says "Pay With New Credit Card" -- maybe if that said "Add a Credit Card" it would work for the regular case _and_ this one?

**For a regular auto-billed customer on the free plan:**

1. We should probably call out that this plan is free from the overview page (the paid plans say the cost but this one doesn't).

Otherwise everything looks ****fantastic**** and I'm super excited for people to see this. 

OK looking through the code now.

src/actions/billing.js
```diff
@@ -13,30 +14,44 @@ export function syncSubscription() {
...
+    type: 'UPDATE_BILLING_CONTACT',
     meta: {
       method: 'PUT',
-      url: '/account/subscription',
-      data
+      url: '/account/billing',
+      data: {
+        ...formatContactData(data),
+        address1: 'N/A', // TODO remove when accusers is updated FAD-5660
+        city: 'N/A'
```
jonambas: Here are those hardcoded required values

src/helpers/billing.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable max-lines */
```
jasonrhodes: Instead of disable let's do:
```
/* eslint max-lines: ["error", 200] */ 
```
where "200" is a number a few lines more than the number in the file, we can keep bumping if necessary but each one will require PR review :)

src/actions/billing.js
```diff
@@ -1,4 +1,5 @@
-import { formatDataForCors, formatCreateData, formatUpdateData } from 'src/helpers/billing';
+/* eslint-disable max-lines */
```
jasonrhodes: Same comment as another place, adjust this to be a value slightly larger than the lines in the file, instead of disabling.

src/pages/billing/forms/UpdateContact.js
```diff
@@ -0,0 +1,59 @@
...
+  componentDidMount() {
+    this.props.getBillingCountries();
+  }
+
+  onSubmit = (values) => {
+    const { updateBillingContact, onCancel, reset, showAlert } = this.props;
+    return updateBillingContact(values).then(() => {
+      showAlert({ type: 'success', message: 'Billing Contact Updated' });
+      onCancel();
+      reset();
```
jonambas: Just realized that `onCancel` closes the modal, which also unmounts the form. So this might not be neccessary. Should double-check previous form values are removed from the redux-form store

PR #109, opened 2017-10-10
Tracking Domains List

jgzamora:


src/components/apiErrorBanner/ApiErrorBanner.js
```diff
@@ -1,6 +1,22 @@
 import React, { Component } from 'react';
 import { Banner, Button } from '@sparkpost/matchbox';
 
+const actionToText = {
+  GET: 'loading',
+  PUT: 'updating',
+  DELETE: 'deleting',
+  POST: 'creating'
+};
```
jonambas: Hmmm, so I've been using `showAlert()` / `GlobalAlert` for most errors. Should probably discuss where we want to use this banner vs. alerts

src/pages/trackingDomains/helpers/getRowData.js
```diff
@@ -0,0 +1,41 @@
...
+      return compliance_status;
+    }
+  } else {
+    return 'unverified';
+  }
+};
+
+const linkColumn = (domain, isDefault) => (
+    <div>
+      <Link to={`/account/trackingDomains/${domain}`}>{domain}</Link>
```
jgzamora: route needs to be tracking-domains

src/reducers/trackingDomains.js
```diff
@@ -0,0 +1,22 @@
...
+export default (state = initialState, { type, payload, meta }) => {
+  switch (type) {
+    case 'LIST_TRACKING_DOMAINS_PENDING':
+      return { ...state, listLoading: true, error: null };
+
+    case 'LIST_TRACKING_DOMAINS_SUCCESS':
+      return { ...state, listLoading: false, listLoaded: true, list: payload };
+
+    case 'LIST_TRACKING_DOMAINS_FAIL':
+      return { ...state, listLoading: false, listLoaded: true, error: { payload, meta }};
```
jasonrhodes: Can you spread `...payload` here so we don't have to reference payload at all?

src/pages/trackingDomains/ListPage.js
```diff
@@ -0,0 +1,65 @@
...
+  componentDidMount() {
+    this.props.listTrackingDomains();
+    this.props.listSubaccounts();
+  }
+
+  onReloadApiBanner = () => {
+    this.props.listTrackingDomains({ force: true }); // force a refresh
+  };
+
+  renderError({ payload, meta }) {
```
jasonrhodes: If we spread `...payload` in the reducer we can reference `{ message, meta }` here 

Alternatively I think I prefer to just pass `error` as a whole into the `ApiErrorBanner` so it can calculate everything from there?

src/pages/trackingDomains/ListPage.js
```diff
@@ -0,0 +1,71 @@
...
+        />
+      </div>
+    );
+  }
+
+  render() {
+    const { error, loading, trackingDomains, hasSubaccounts } = this.props;
+
+    return (
+      <Layout.App loading={loading}>
```
jasonrhodes: I think we need to as a team figure out how to get `<Layout>` out of the components, and this will clean things up a lot.

Issue #112, opened 2017-10-11
Splitting error handling and action creators


jgzamora:
At some point we will want to chain different actions, while controlling error handling and alerts that come from each action. To do this with an action that has error handling built in, you'd have to re-write the whole action creator. Take `deleteUser` for example:
```javascript
export function deleteUser(username) {
  const action = {
    type: 'DELETE_USER',
    meta: {
      data: { username }, // need in reducer, no user reference in response
      method: 'DELETE',
      url: `/users/${username}`
    }
  };

  return (dispatch) => dispatch(sparkpostApiRequest(action))
    .then(() => dispatch(showAlert({
      type: 'success',
      message: `Successfully deleted ${username}`
    })))
    .catch(() => dispatch(showAlert({
      type: 'error',
      message: `Unable to delete ${username}.`
    })));
}
```
It dispatches alerts on success and fail. If I needed to delete a user as a part of a chain of actions but didn't want to dispatch the alerts, or I wanted to throw an error instead, I would have to write a new action creator. 

Proposed solution: extract the action object and dispatch call from current action creators into separate functions.

`/actions/someFolder/users.js`
```javascript
function deleteUserAction(dispatch, username) {
  const action = {
    type: 'DELETE_USER',
    meta: {
      data: { username }, // need in reducer, no user reference in response
      method: 'DELETE',
      url: `/users/${username}`
    }
  };

  return dispatch(sparkpostApiRequest(action));
}

//this can go anywhere else, just another example
function showAlertAction(dispatch, alert) {
  return dispatch(showAlert(alert));
}
```

`/actions/users.js`
```javascript
import { deleteUserAction,  showAlertAction} from './someFolder/users.js'
export function deleteUser(username) {
  return (dispatch) => deleteUserAction(dispatch, username)
    .then(() => showAlertAction(dispatch, {
      type: 'success',
      message: `Successfully deleted ${username}`
    }))
    .catch(() => showAlertAction(dispatch, {
      type: 'error',
      message: `Unable to delete ${username}.`
    }));
}
```

This way, if I want to delete a user while doing a bunch of other stuff to I can just:
```javascript
// async await cause it looks pretty
export function notJustDeleteUser(username, otherStuff) {
  return async (dispatch) => {
    try {
      const userDelete = await deleteUserAction(dispatch, username);
      return await otherThingAction(otherStuff);
    } catch(error) {
      return showAlertAction(dispatch, {
        type: 'error',
[...]

PR #113, opened 2017-10-12
FAD-5493: Create User

ewandennis:
# To Do
 - [x] Handle `/users/invite` 409 - `User with that email already exists`
 - [x] Handle `/users/invite` 400 - `Email address is suppressed`
 - [x] Replace button text during submission
 - [ ] ~Form-centric error messages?~
 - [ ] ~Tests - could use some guidance here~

# Questions
 - ~Do we need a separate UserForm component? There is no edit case for users.~ Removed.
 - Is there a better way to capture the default value of a `<select>` than setting `initialValues`? [This issue suggests no](https://github.com/erikras/redux-form/issues/3226).

# Follow-on Work
 - SSO support
 - Handle invitation tokens


src/pages/users/CreatePage.js
```diff
@@ -0,0 +1,42 @@
...
+  to: '/account/users'
+};
+
+
+class CreatePage extends React.Component {
+  onSubmit = (values) => {
+    const { email, access } = values;
+    const { inviteUser, history } = this.props;
+    return inviteUser(email, access).then((res) => {
+      history.push('/account/users');
```
jonambas: This will push even on a create fail, which may not be what we want. Added my thoughts to that issue (https://github.com/SparkPost/2web2ui/issues/112).

src/pages/users/CreatePage.js
```diff
@@ -0,0 +1,42 @@
...
+    });
+  };
+
+  render() {
+    return <Layout.App loading={this.props.loading}>
```
jonambas: Can remove `loading` prop here as this page or form do not require anything from the store

src/pages/users/components/UserForm.js
```diff
@@ -0,0 +1,37 @@
...
+      Add user
+    </Button>
+    {submitting && !submitSucceeded && <div>Loading&hellip;</div>}
+  </form>;
+};
+
+const UserReduxForm = reduxForm({ form: formName })(UserForm);
+export default connect(() => ({
+  initialValues: {
+    access: 'admin' // Sadly redux-form does not reflect a select's initial value
```
jonambas: Not sure what you mean here, seems to be working as intended?

I changed initial value to `reporting`:
screen shot 2017-10-12 at 10 24 20 am

If you want to set a default placeholder you can do this:
```
<Field ... component={AccessSelect} placeholder='Select a role' />

initialValues: {
    access: '_placeholder' // this is something specific to MB
}

// in AccessSelect pass through rest
<Select
      {...input}
      {...rest}
    />
```

src/pages/users/components/UserForm.js
```diff
@@ -0,0 +1,37 @@
...
+    <Field name="access" label="Role" component={AccessSelect} />
+    <Button submit primary disabled={submitting || pristine}>
+      Add user
+    </Button>
+    {submitting && !submitSucceeded && <div>Loading&hellip;</div>}
+  </form>;
+};
+
+const UserReduxForm = reduxForm({ form: formName })(UserForm);
+export default connect(() => ({
```
jonambas: Nitpicking here and not required but I generally like to separately define `mapStateToProps` and (if lenghty) `mapDispatchToProps` as well, then string them together within the export

src/pages/users/CreatePage.js
```diff
@@ -0,0 +1,42 @@
...
+
+const breadcrumbAction = {
+  content: 'Users',
+  Component: Link,
+  to: '/account/users'
+};
+
+
+class CreatePage extends React.Component {
+  onSubmit = (values) => {
```
jonambas: This function should be in the redux-form connected component (if you decided to keep them separate). This page might be small enough not to require splitting them up.

Honestly very confused as to how the form is submitting. You pass `onSubmit` as a prop to `UserForm`, but aren't referencing or pulling in `this.props.onSubmit` anywhere in the form. redux-form magic??

src/pages/users/components/UserForm.js
```diff
@@ -0,0 +1,37 @@
...
+  initialValues: {
+    access: 'admin' // Sadly redux-form does not reflect a select's initial value
+  }
+}), { formChange: change })(UserReduxForm);
```
jonambas: Can remove this action as its not used anywhere

src/pages/users/components/UserForm.js
```diff
@@ -0,0 +1,37 @@
...
+      name="email"
+      validate={[required, email]}
+      label="Email address"
+      component={TextFieldWrapper}
+    />
+    <Field name="access" label="Role" component={AccessSelect} />
+    <Button submit primary disabled={submitting || pristine}>
+      Add user
+    </Button>
+    {submitting && !submitSucceeded && <div>Loading&hellip;</div>}
```
jonambas: Maybe we could replace button text instead?

src/pages/users/CreatePage.js
jonambas: form should look like (to my understanding):
```js
<form onSubmit={this.props.handleSubmit(this.yourSubmitFunc)}>
```

ewandennis: `reduxForm()` uses the `onSubmit` prop from the component its wrapping [according to the docs](https://redux-form.com/7.1.1/docs/api/reduxform.md/#-code-onsubmit-function-code-optional-). I naively didn't think about it until you mentioning though, heh.

src/pages/users/components/UserForm.js
ewandennis: Yeah, it works bc I set an initial value for the `access` dropdown. If I don't do that, the underlying DOM control has a value but redux-form doesn't know about it. I say "bug", `redux-form` says "Eh, just initialise it".

src/pages/users/CreatePage.js
jonambas: Ah I see, magic confirmed

src/pages/users/CreatePage.js
ewandennis: Resolved in a hopefully workable way.

src/pages/users/components/UserForm.js
ewandennis: This is a nice convention. We should probably apply it everywhere though. The API keys form  does the same shonky 'Loading' text message.

PR #114, opened 2017-10-12
Pins moment, fixes route bug

jasonrhodes:
Moment has a bug: https://github.com/moment/moment/issues/4216

ProtectedRoute started passing all props through to RR Route, causing
an issue with component vs render, fixed that here too.


chamini2:
This is already fixed in 2.19.1

jasonrhodes:
Yeah I noticed right after I merged and deployed this lol, thanks! Will get it unpinned asap.

PR #118, opened 2017-10-13
FAD-5655 Template and API Keys filters

jonambas:
- Adds FilterDropdown component. Added a usage guide as well.
- Implements filtering into Templates and API Keys list pages
- Adds EmptyState to Templates list page
- Adds custom header component for Template/Api Keys tables to set widths.

Lists are filtered through a selector BEFORE they hit the table collection component. Each list page will probably need its own filtering selector and filtering (redux-form wrapped) component.

TODO
- [x] Implement into API Keys list page
- [x] Tests
- [ ] Create ticket - handle `hasSubaccounts`


src/config/default.js
```diff
@@ -27,6 +27,9 @@ const config = {
     supportEmail: 'support@sparkpost.com',
     billingEmail: 'billing@sparkpost.com'
   },
+  filters: {
+    minCount: 25
```
jonambas: New config key. Filters are show if list length is greater than this number.

src/components/reduxFormWrappers.js
```diff
@@ -47,7 +47,7 @@ export const RadioGroup = ({ input, options, title }) => (
   </Radio.Group>
 );
 
-export const CheckboxWrapper = ({ input, ...rest }) => (
+export const CheckboxWrapper = ({ input, meta, ...rest }) => (
```
jonambas: Small bug I found, just stripping out meta before passing through to the Checkbox

src/components/collection/Collection.js
```diff
@@ -74,7 +74,7 @@ class Collection extends Component {
         <OuterWrapper>
           <HeaderComponent />
           <BodyWrapper>
-            {this.getVisibleRows().map((row, i) => <RowComponent key={row[rowKeyName] || i} {...row} />)}
+            {this.getVisibleRows().map((row, i) => <RowComponent key={row[rowKeyName] ? `${row[rowKeyName]}-${i}` : i} {...row} />)}
```
jonambas: Under the appteam account, there are a couple templates with the same `id` (idk). Should replace with `key={'${row[rowKeyName]}-${i}'}`

src/components/subaccountTypeahead/SubaccountTypeahead.js
```diff
@@ -34,8 +34,7 @@ export class SubaccountTypeahead extends Component {
       (item) => `${item.name} ID: ${item.id}`
     ).map((item, index) => ({
       ...getItemProps({ item, index }),
-      content: <Item name={item.name} id={item.id} />,
-      highlighted: highlightedIndex === index
```
jonambas: Removed `highlighted` prop from MB, replaced with `selected` which adds an `<Icon name='Check'>`

src/pages/api-keys/components/TableComponents.js
```diff
@@ -0,0 +1,23 @@
...
+const columns = [
+  { label: 'Name', width: '40%' },
+  { label: 'Key', width: '20%' },
+  { label: 'Subaccount', width: '20%' }
+];
+
+export const TableHeader = () => (
+  <thead>
+    <Table.Row>
+      {columns.map(({ label, ...rest }) => <Table.HeaderCell key={label} {...rest}>{label}</Table.HeaderCell>)}
```
jonambas: Should probably just make this the default TableHeader for the TableCollection

src/components/collection/Collection.js
```diff
@@ -64,17 +64,17 @@ class Collection extends Component {
...
     } = this.props;
 
     return (
       <div>
         <OuterWrapper>
-          <HeaderComponent />
+          {headerComponent}
           <BodyWrapper>
-            {this.getVisibleRows().map((row, i) => <RowComponent key={row[rowKeyName] || i} {...row} />)}
+            {this.getVisibleRows().map((row, i) => <RowComponent key={`${row[rowKeyName]}-${i}`} {...row} />)}
```
jasonrhodes: why this change? if the key is unique why does it need the `i` appended?

src/components/collection/Collection.js
```diff
@@ -64,17 +64,17 @@ class Collection extends Component {
...
+      headerComponent,
       outerWrapper: OuterWrapper = PassThroughWrapper,
       bodyWrapper: BodyWrapper = PassThroughWrapper
     } = this.props;
 
     return (
       <div>
         <OuterWrapper>
-          <HeaderComponent />
+          {headerComponent}
```
jasonrhodes: not sure I fully understand this change ... it _seems_ better to pass components around and not JSX strings?

src/components/filterDropdown/FilterDropdown.js
```diff
@@ -0,0 +1,92 @@
...
+          trigger={<TextField prefix={prefix} value={displayValue} readOnly suffix={<Icon name='CaretDown'/>} />}>
+          <ActionList actions={actions} />
+        </Popover>
+        {this.renderCheckboxes()}
+      </div>
+    );
+  }
+}
+
+FilterDropdown.propTypes = {
```
jasonrhodes: Are we using propTypes in this codebase? I don't know much about them so I don't actually have an opinion beyond "so long as they aren't included in the prod bundle at all" -- if we're using them should we cut a ticket to apply them everywhere?

src/components/reduxFormWrappers.js
jasonrhodes: Can you use `_.omit` for this? I was doing this exact thing somewhere else but it got a bit sticky and once we fix our linting this is going to get nailed for unused variables.

Something like:
```js
({ input, ...rest }) => {
  const checkBoxProps = _.omit(rest, 'meta');
  // etc
```

jasonrhodes:
This is good. I'd like to see if we can put the filters inside of the TableCollection as some sort of declarative prop somehow, so we don't have to pre-filter the list given to TableCollection. It could render the filters (if necessary) and then apply the filtering to its own set of rows, which all seems very natural.

src/components/collection/Collection.js
jonambas: On the appteam account, there are templates with the same ID - not sure where they came from: 
screen shot 2017-10-17 at 2 31 14 pm

src/components/collection/Collection.js
jonambas: I removed the custom header components from the list pages, so I can revert. But, IMO it looks a little cleaner.

```js
<TableCollection headerComponent={<Component {...props}/>}

// or

<TableCollection headerComponent={() => <Component {...props}/>}

// or

<TableCollection headerComponent={Component} /> // but how to add props?
```

src/components/filterDropdown/FilterDropdown.js
jonambas: We don't have to apply them everywhere. I think for these complicated components its just useful to say "hey you're not using this correctly"

jonambas:
I wasn't really sure how to get it integrated into the TableCollection. To get filters working you need:
1. a redux-form filters component
2. a selector to filter the list (with a selector to get form values)

Would still have to pass those to the collection component, which would only really be doing the `list.length > config.filters.minCount` check

src/components/subaccountTypeahead/SubaccountTypeahead.js
jonambas: Actually - should add back in. Key events don't show up (oops)

src/components/reduxFormWrappers.js
jonambas: Realized this component doesn't pass through `error`, will use meta like the other wrapper components:
```js
export const CheckboxWrapper = ({ input, meta, ...rest }) => {
  const { active, error, touched } = meta;
  return (
    <Checkbox
      id={input.name}
      {...input}
      error={!active && touched && error ? error : undefined}
      {...rest}
    />
  );
};
```

src/components/collection/Collection.js
```diff
@@ -1,18 +1,23 @@
...
-import Pagination from './Pagination';
+import Pagination, { defaultPerPageButtons } from './Pagination';
+import CollectionFilter from './Filter';
+import { objectSortMatch } from 'src/helpers/sortMatch';
 
 const PassThroughWrapper = (props) => props.children;
+const NullComponent = () => null;
 
-class Collection extends Component {
+export class _Collection extends Component {
```
jonambas: Why underscore?

src/components/collection/Pagination.js
```diff
@@ -23,7 +25,7 @@ export default class CollectionPagination extends Component {
   }
 
   renderPerPageButtons() {
-    const { data, perPage, perPageButtons = [10, 25, 50, 100], onPerPageChange } = this.props;
+    const { data, perPage, perPageButtons = defaultPerPageButtons, onPerPageChange } = this.props;
```
jonambas: Nitpick but we should try to use `defaultProps` instead of defaulting in a spread

src/components/filterDropdown/tests/__snapshots__/FilterDropdown.test.js.snap
```diff
@@ -0,0 +1,139 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`renders correctly 1`] = `
+<div>
+  <t
```
jonambas: Will probably need to update MB and these tests

src/config/default.js
jonambas: Can remove i think?

src/pages/api-keys/ListPage.js
```diff
@@ -89,13 +78,14 @@ export class ListPage extends Component {
   }
 
   render() {
-    const { error, loading, newKey } = this.props;
+    const { error, loading, newKey, count } = this.props;
 
     return (
       <Layout.App loading={loading}>
```
jonambas: Will need to update with layout & loading changes

src/pages/api-keys/components/Filters.js
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
```
jonambas: Can delete file

src/pages/templates/tableConfig.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import { Link } from 'react-router-dom';
+import { format } from 'date-fns';
+
+export const getRowData = ({ published, id, name, last_update_time, subaccount_id, shared_with_subaccounts }) => [
+  <Link to={`/templates/edit/${id}`}>{name}</Link>,
+  id,
+  published ? 'Published' : 'Draft',
+  format(last_update_time, 'MMM D, YYYY [at] h:mma'),
```
jonambas: if theres a test that renders these rows, might need to mock `format` so these pass in different timezones. Should be okay with `shallow`

src/selectors/templates.js
```diff
@@ -1,7 +1,60 @@
...
+import sortMatch from 'src/helpers/sortMatch';
 
-export {
-  templateById
+export const getTemplates = (state) => state.templates.list;
+export const getTemplateById = (state, props) => state.templates.byId[props.match.params.id] || { draft: {}, published: {}};
+
+const getFilters = (state) => state.form.templateFilters && state.form.templateFilters.values;
+
+export const getfilterTemplates = createSelector(
```
jonambas: Can remove as well if we aren't using filter dropdowns?

src/components/collection/Collection.js
jasonrhodes: I know we aren't doing this everywhere but it always makes me a little nervous when you can do both of these:

```js
import Collection from 'src/components';
import { Collection } from 'src/components';
```

So the `_prefix` seemed like maybe a good way to signify an export that's sort of meant to be internal-only for testing etc... 

Could be convinced to not care.

PR #119, opened 2017-10-16
FAD-5465 Implements ip pools list view

rajumsys:


src/pages/ipPools/ListPage.js
```diff
@@ -0,0 +1,71 @@
...
+
+function mapStateToProps({ ipPools }) {
+  return {
+    ipPools: ipPools.list,
+    loading: ipPools.listLoading,
+    error: ipPools.listError
+  };
+}
+
+export default withRouter(connect(mapStateToProps, { listPools })(IpPoolsList));
```
jasonrhodes: I don't think you need `withRouter()` here

rajumsys:
Ok, I just realized, this way we are not showing `Default Pool` at all. we'll need to figure out how to show it.

src/pages/ipPools/ListPage.js
rajumsys: Removed

rajumsys:
@jasonrhodes rebased and fixed tests

PR #123, opened 2017-10-18
WIP for Feature/fad 5496

rajumsys:
This is still in progress. 

This is branched off FAD-5465 (#119). 

Basic list, create, update and delete works.  lot to do...puuf! 

- [x] fix messaging (after create, delete, edit)
- [x] initial create page looks like validation error even before any input. 
- [x] edit page shows a list of all ips and it can be reassigned to another pool (Update: @ewandennis  we can probably use the ip pools typeahead that @jgzamora created in [his pr](https://github.com/SparkPost/2web2ui/pull/133))
- [ ] cases for no ips, no paid plans etc.


package.json
```diff
@@ -98,7 +98,8 @@
     "build": "node scripts/build.js",
     "analyze": "source-map-explorer build/static/js/main.*",
     "test": "node scripts/test.js --env=jsdom",
-    "lint": "eslint src"
+    "lint": "eslint src",
+    "autolint": "eslint src -- --fix"
```
jasonrhodes: Can we call this `lintfix`?

src/pages/ipPools/CreatePage.js
```diff
@@ -0,0 +1,41 @@
...
+          <Panel.Section>
+            <PoolForm onSubmit={this.onSubmit} isNew={true} />
+          </Panel.Section>
+        </Panel>
+      </Layout.App>
+    );
+  }
+}
+
+const mapStateToProps = (state, props) => ({});
```
jasonrhodes: You need to connect to a loading key so that the loading state works.

src/pages/ipPools/EditPage.js
```diff
@@ -0,0 +1,82 @@
...
+          text='Are you sure you want to delete this IP Pool?'
+          handleToggle={this.toggleDelete}
+          handleDelete={this.deletePool}
+        />
+      </Layout.App>
+    );
+  }
+}
+
+const mapStateToProps = (state, props) => ({});
```
jasonrhodes: Need to connect to a loading key while it's doing the GET so the loading state works (right now it flashed with empty values)

src/pages/ipPools/ListPage.js
```diff
@@ -0,0 +1,71 @@
+import React, { Component } from 'react';
+import { connect } from 'react-redux';
+import { Link } from 'react-router-dom';
+
+// Actions
+import { listPools } from '../../actions/ipPools';
```
jasonrhodes: `from src/actions/ipPools`

src/pages/ipPools/PoolForm.js
```diff
@@ -0,0 +1,90 @@
+import React, { Component } from 'react';
```
jasonrhodes: This file should move into a `components` folder here

src/pages/ipPools/PoolForm.js
```diff
@@ -0,0 +1,90 @@
...
+import { connect } from 'react-redux';
+import { Field, reduxForm } from 'redux-form';
+
+import { TextField, Button, Select } from '@sparkpost/matchbox';
+
+import { TableCollection } from 'src/components';
+import { getPool } from '../../actions/ipPools';
+
+const required = (value) => (value || value === null ? undefined : 'Required');
```
jasonrhodes: Use 

```js
import { required } from 'src/helpers/validation';
```

src/pages/ipPools/PoolForm.js
```diff
@@ -0,0 +1,90 @@
+import React, { Component } from 'react';
+import { connect } from 'react-redux';
+import { Field, reduxForm } from 'redux-form';
+
+import { TextField, Button, Select } from '@sparkpost/matchbox';
+
+import { TableCollection } from 'src/components';
+import { getPool } from '../../actions/ipPools';
```
jasonrhodes:
```js
from 'src/actions/ipPools';
```

src/pages/ipPools/PoolForm.js
```diff
@@ -0,0 +1,90 @@
...
+
+const required = (value) => (value || value === null ? undefined : 'Required');
+
+const TextFieldWrapper = ({ input, meta: { error }, ...rest }) => (
+  <TextField {...rest} {...input} error={error} />
+);
```
jasonrhodes: Use
```js
import { TextFieldWrapper } from 'src/components';
```

which will correctly handle all of the error cases for you (right now it shows error state immediately before blur)

src/pages/ipPools/PoolForm.js
```diff
@@ -0,0 +1,90 @@
...
+
+const columns = ['Sending IP', 'Hostname', 'IP Pool'];
+
+export class PoolForm extends Component {
+  render() {
```
jasonrhodes: by convention we usually try to have render be the last function defined in the class FWIW

src/pages/ipPools/PoolForm.js
```diff
@@ -0,0 +1,90 @@
...
+          validate={required}
+          label="Pool Name"
+        />
+
+        { this.renderCollection() }
+
+
+        <Button submit primary disabled={submitting || pristine}>
+          {submitText}
+        </Button>
```
jasonrhodes: `<Button>` component needs to be wrapped in a `<Panel>` or something, I think, to display and float correctly 

Ping: @jonambas

src/pages/ipPools/PoolForm.js
```diff
@@ -0,0 +1,90 @@
...
+    );
+  }
+}
+
+const mapStateToProps = ({ form, ipPools }) => {
+  const pool = ipPools.pool || {};
+  return {
+    theForm: form,
+    pool: pool, //TODO need to figure out how to access from initialValues and get rid of this
+    initialValues: pool
```
jasonrhodes: You don't want initial values if this is a new form. You can access non-redux props in `mapStateToProps` via the 2nd arg, like this:
```js
const mapStateToProps = ({ form, ipPools }, { isNew }) => {
  return {
    initialValues: isNew ? {} : pool
  }
}
```

src/pages/ipPools/PoolForm.js
```diff
@@ -0,0 +1,90 @@
...
+      />
+    );
+  }
+}
+
+const mapStateToProps = ({ form, ipPools }) => {
+  const pool = ipPools.pool || {};
+  return {
+    theForm: form,
+    pool: pool, //TODO need to figure out how to access from initialValues and get rid of this
```
jasonrhodes: What are you wanting to do here?

jonambas: I suggest using `formValueSelector` if you need access to values
https://redux-form.com/7.0.0/docs/api/formvalueselector.md/

src/reducers/ipPools.js
```diff
@@ -12,6 +13,46 @@ export default (state = initialState, { type, payload }) => {
     case 'LIST_IP_POOLS_FAIL':
       return { ...state, listError: payload, listLoading: false };
 
+    //create pool
+    case 'CREATE_IP_POOL_PENDING':
+      return { ...state, currentPool: { }, createError: null };
```
jasonrhodes: why `currentPool` ?

src/pages/ipPools/ListPage.js
```diff
@@ -0,0 +1,71 @@
...
+    return (
+      <ApiErrorBanner
+        message={'Sorry, we seem to have had some trouble loading your ip pools.'}
+        errorDetails={error.message}
+        reload={listPools}
+      />
+    );
+  }
+
+  renderCollection() {
```
jasonrhodes: When creating a new pool we don't want to show this, so maybe `if (isNew) { return null; }`

src/pages/ipPools/PoolForm.js
jasonrhodes: Never mind this is because we're rendering pagination in the table collection and we are always including a `<Panel>` in there too. I have an idea we can talk out.

src/pages/ipPools/PoolForm.js
```diff
@@ -0,0 +1,90 @@
...
+  renderCollection() {
+    const { ips = []} = this.props.pool;
+    const getRowDataFunc = this.getRowData.bind(this);
+
+    return (
+      <TableCollection
+        columns={columns}
+        rows={ips}
+        getRowData={getRowDataFunc}
+        pagination={true}
```
jasonrhodes: Let's not include pagination for this, set it to `{false}` to be explicit

src/pages/ipPools/PoolForm.js
jgzamora: Pool isn't a form value (assigned through a Field) afaict, so `formValueSelector` shouldn't work either. And the use of `initialValues` below will only work if the attributes of pool match the `name` prop on the fields

jgzamora: Oh @jonambas, I get that you probably meant to grab each field from `pool` using `formValueSelector`

src/pages/ipPools/PoolForm.js
jgzamora: and `pool`'s attributes have to match Field `name`'s for the values to actually be mapped. If you want to pass something that isn't part of field back to the form, just pass it above `initialValues`

src/pages/ipPools/PoolForm.js
jonambas: oh I read this as
>//TODO need to figure out how to access FORM initialValues and get rid of this

rajumsys:
Thanks everyone for the reviews. If i get time after finishing my current works, I'll handle some or all of them before someone else take this over; but don't wait for it though.

src/pages/ipPools/PoolForm.js
rajumsys: > You don't want initial values if this is a new form. 

This same form is used for editing pool. So what I tried to set, during creation of pool, it'll pass empty object but for editing it'll pass current pool.

ewandennis:
Closing to make room for moar feedback

PR #126, opened 2017-10-20
Layout, Portals, DisplayName

### LINKED ISSUE #122: Move Layout out of page components and into route config
To simplify render functions in the Page components, we should just specify the type of layout (app vs form) in the route config and wrap the given component in the Layout inside of our custom Route components. To do this we have to:

1. Move loading stuff out of Layout and handle that in render methods another way
1. Create a `<PublicRoute>` component that wraps given components in `<Layout>`
1. Edit `<ProtectedRoute>` to wrap in `<Layout>` similarly to how it currently wraps in `<AccessControl>`

This may be more work than it's worth right now, tbh. 🤔

jonambas:
- Updates MB - component displayNames now render in snapshot tests
- Adds a Modal portal
  - `{ Modal } from 'src/components'` should now be used instead of the matchbox modal
  - `DeleteModal` now uses the portal
```js
// src/components/modal/Modal
import { Portal, Modal as MBModal } from '@sparkpost/matchbox';

const Modal = ({ children, ...rest }) => (
  <Portal containerId='modal-portal'>
    <MBModal {...rest}>{ children }</MBModal>
  </Portal>
);

// public/index.html
<div id="root"></div>
<div id='modal-portal'></div>
```

- Refactors `Layout`
  - Layout now wraps routes - and will only render once
  - `Loading` added to all pages where `<Layout loading>` was present
  - Closes #122 

MB's `EmptyState` is now an optional prop through `Page` and will share `primaryAction` supplied to Page:
```js
<Page empty ={{
  test: bool,
  content: <p/>,
  // ...rest of EmptyState props
}}/>
```
But I did not add it anywhere - will do in a follow on ticket


public/index.html
```diff
@@ -32,6 +32,7 @@
       You need to enable JavaScript to run this app.
     </noscript>
     <div id="root"></div>
+    <div id='modal-portal'></div>
```
jonambas: I found creating a portal on a React component to be unreliable. (maybe I just needed to pay with it more)

Sometimes the target container wouldn't be mounted at the time the initiator was rendered - resulting in something like `DOM node not found`.

Adding this div just makes sure the container will always be there.

src/components/layout/Layout.js
```diff
@@ -1,9 +1,24 @@
...
 import Form from './Form';
+import routes from 'src/config/routes';
+
+/**
+ * Returns layout component from routes config
+ */
+const Layout = ({ children, location }) => {
+
+  // matchPath uses the same matching that <Route> uses
+  const route = _.find(routes, (route) => matchPath(location.pathname, {
```
jonambas: Using react-routers own route matching function to find the right layout.

src/components/modal/Modal.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import { Portal, Modal as MBModal } from '@sparkpost/matchbox';
+
+const Modal = ({ children, ...rest }) => (
+  <Portal containerId='modal-portal'>
+    <MBModal {...rest}>{ children }</MBModal>
+  </Portal>
+);
```
jasonrhodes: Is the idea here going forward that you would include this `<Modal>` component alongside every element that might trigger it, instead of in a more central place on the page like before?

PR #128, opened 2017-10-20
Fixes recursive mutation bug in nav render

jasonrhodes:
This is the bug where account mysteriously disappears from the nav after logout/login ... when the grants aren't found in the cookie (so first login or logout/login), it would compute access for nav items before it had grants and replace their children with empty arrays because none of the children had access yet. By the time we re-calculated it was too late and we had already mutated the children so they were gone forever, so we remove the parent too since none of the children were there.

Whole fix is just this:
```diff
 export function filterNavByAccess(items, accessConditionState) {
-  return items.filter((item) => {
+  return items.filter((reference) => {
+    const item = { ...reference }; // prevent weird mutation bugs
```

I also added a memoize around the mapping of the nav items so we don't have to re-map them over and over for no reason.

### DIFF
--- a/src/components/auth/AccessControl.js
+++ b/src/components/auth/AccessControl.js
@@ -1,18 +1,23 @@
 import React from 'react';
 import { connect } from 'react-redux';
 import { Redirect } from 'react-router-dom';
+import { Loading } from 'src/components/loading/Loading';
 import accessConditionState from 'src/selectors/accessConditionState';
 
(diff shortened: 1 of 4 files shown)


jonambas:
LGTM.

note:
When logging into a page with its own loading state, <Loading> will double render - animation looks like it stutters. Only fix, i think, is to hook loading up to redux, but whatever.

PR #133, opened 2017-10-25
Subaccount Create

jgzamora:


bkemper:
@jgzamora Three things I noticed during a quick smoke test.

1. If I create a subaccount with a new API key without selecting any permissions, I am unable to view it in on the api keys list page.

screen shot 2017-10-25 at 1 49 48 pm

2. Instead of "You can do this later", should we has a clause like in the webui, "If left blank, the default IP Pool will be used."?

3. Does it make sense to make the "Assign to IP Pool" field a dropdown with the default pool as the selected value?

bkemper:
@jgzamora I have a better understanding of why #1 happened now that I took a look at your code.  Since the API allows you to create an API without grants, I would leave your code and I'll submit a follow-up issue to fix the api keys list page.

rajumsys:
@jgzamora seems the ip pools typeahead is case sensitive. can we make the search case insensitive?

src/actions/helpers/subaccounts.js
```diff
@@ -0,0 +1,43 @@
...
+    const {
+      keyName,
+      grantsRadio,
+      validIps
+    } = values;
+
+    subaccount.key_label = keyName;
+
+    if (grantsRadio === 'all') {
+      subaccount.key_grants = _.keys(getSubaccountGrants(getState()));
```
bkemper: Instead of needing to convert 'all' to a list of grants, what do you think about passing the full list of grants from the component to the action in `values`?

src/actions/helpers/tests/subaccounts.test.js
```diff
@@ -0,0 +1,99 @@
...
+    'message_events/write': false,
+    'relay_webhooks/view': undefined,
+    'relay_webhooks/write': null,
+    'smtp/inject': 'true'
+  }
+};
+
+test('name only', () => {
+  const result = formatSubaccount({ name: 'mySubAccount' }, getState);
+  expect(result).toEqual({ name: 'mySubAccount', setup_api_key: false });
```
bkemper: Would a snapshot test be easier/better?

src/actions/helpers/tests/subaccounts.test.js
```diff
@@ -0,0 +1,99 @@
...
+  expect(result).not.toHaveProperty('key_label');
```
bkemper: If the full object is tested above, do we need the `toHaveProperty` assertions?

src/actions/helpers/tests/subaccounts.test.js
```diff
@@ -0,0 +1,99 @@
...
+      assingToPool: true,
+      ipPool: {
+        id: 'marketing_ip_pool',
+        name: 'Marketing IP Pool'
+      }
+    },
+    getState
+  );
+  expect(result.ip_pool).toEqual('marketing_ip_pool');
+});
```
bkemper: Is there a reason to shy away from snapshot testing?

bkemper: Should we shy away from multiple assertions in a single test?  Instead, we could group each assertion as a single test in a `describe` block.

```
describe('with api key - select grants', () => {
  const subaccount = { name: 'selectPerms', ...keyFieldsSelect };
  const result = formatSubaccount(subaccount, getState);

  test('returns label set with name')
  test('returns list of grant's)
});
```

src/actions/subaccounts.js
```diff
@@ -1,19 +1,27 @@
 import sparkpostApiRequest from 'src/actions/helpers/sparkpostApiRequest';
+import { formatSubaccount } from './helpers/subaccounts';
 
-export function list({ force } = {}) {
-  return (dispatch, getState) => {
-    if (!force && getState().subaccounts.listLoaded) {
-      return;
-    }
+export function list() {
+  return (dispatch) => dispatch(
```
bkemper: Does this need to be dispatched?

src/components/grantBoxes/GrantsCheckboxes.js
```diff
@@ -0,0 +1,39 @@
...
+      normalize={(value) => (value ? 'true' : '')} // TODO is this necessary? otherwise we get proptype warnings
+    />
+  ));
+
+  const grantFieldChunks = _.chunk(grantFields, Math.ceil(_.size(grants) / 3));
+
+  const grantCols = _.map(grantFieldChunks, (grantFields, i) => (
+    <Grid.Column xs={12} md={4} key={i}>
+      {grantFields}
+    </Grid.Column>
```
bkemper: Do we have a multi column list in matchbox that allows you to define list like a <ul> or <ol> and the css handles the columns based on available space?

src/components/reduxFormWrappers.js
```diff
@@ -68,3 +69,13 @@ export const SubaccountTypeaheadWrapper = ({ input, subaccounts }) => (
     subaccounts={subaccounts}
   />
 );
+
+
+export const PoolTypeaheadWrapper = ({ input, pools }) => (
```
bkemper: Is there a way to make this more generic to merge the subaccount and pool typeaheads together?

src/pages/api-keys/tests/CreatePage.test.js
```diff
@@ -16,8 +16,8 @@ const props = {
 let wrapper;
 
 beforeEach(() => {
-  wrapper = shallow(<CreatePage {...props} />)
-})
+  wrapper = shallow(<CreatePage {...props} />);
+});
```
bkemper: Oops.  This looks like it is from me when I ran autofix.  I'm sorry for adding this to your branch.

src/pages/subaccounts/CreatePage.js
```diff
@@ -0,0 +1,59 @@
...
+export class CreatePage extends React.Component {
+  componentDidMount() {
+    this.props.listSubaccountGrants();
+    this.props.listPools();
+  }
+
+  onSubmit = (values) => {
+    const { createSubaccount, history } = this.props;
+
+    return createSubaccount(values).then((res) => {
```
bkemper: Does this promise need to be returned?

src/pages/subaccounts/CreatePage.js
```diff
@@ -0,0 +1,59 @@
...
+        <Panel>
+          <SubaccountForm onSubmit={this.onSubmit}/>
+        </Panel>
+      </div>
+    );
+  }
+}
+
+const mapStateToProps = (state, props) => ({
+  loading: state.apiKeys.subaccountGrantsLoading || state.ipPools.listLoading
```
bkemper: Should we move this to the individual field and leave the fields disabled until the available values are loaded?

src/selectors/tests/api-keys.test.js
```diff
@@ -69,7 +69,7 @@ describe('ApiKey Selectors', () => {
         grants: ['grant'],
         valid_ips: ['ip']
       }
-    }
+    };
```
bkemper: I'm so sorry about these.

src/pages/subaccounts/CreatePage.js
jonambas: I believe it does in order for redux-form `submitted`, `submitSucceded`, etc to work. Not 100% sure.

src/pages/subaccounts/components/SubaccountForm.js
```diff
@@ -0,0 +1,137 @@
...
+              type="checkbox"
+              label="Assign to IP Pool"
+              helpText={ 'You can do this later' }
+            />
+            { assingToPool && ipPoolFields(assingToPool, ipPools) }
+          </Panel.Section>
+        }
+        <Panel.Section>
+          <Button submit primary disabled={submitting || pristine}>
+            Create
```
jonambas: `{ submitting ? 'Loading...' : 'Create' }` maybe?

src/pages/subaccounts/components/SubaccountForm.js
```diff
@@ -0,0 +1,137 @@
...
+      ipPools,
+      assingToPool
+    } = this.props;
+
+    const hasIpPools = !!ipPools.length;
+
+    return (
+      <form onSubmit={handleSubmit}>
+        <Panel.Section>
+          <Field
```
jonambas: Not super necessary, but we should add `disabled={submitting}` to all fields

src/actions/helpers/subaccounts.js
jgzamora: If I can sneak the whole list of grants as the value of the select options I'll do this. Matchbox might yell. Using a hidden Field or other round about ways wouldn't be worth it.

src/actions/helpers/tests/subaccounts.test.js
jgzamora: No reason, I was mimicking  from another test file, will change to snapshots.

src/actions/subaccounts.js
jgzamora: Why wouldn't it be?

jgzamora: I thought they all had to be dispatched. Only chains need to be dispatched?

src/components/reduxFormWrappers.js
jgzamora: Yeah, that should happen. Should be it's own ticket, and maybe after we implement it a couple more times.

src/pages/subaccounts/CreatePage.js
jgzamora: hmm, I'll try it out. If we get rid of the loading component and just have disabled fields with no loading animation or indication that might be confusing. I like the direction though

src/selectors/tests/api-keys.test.js
jgzamora: lol, np, I shoulda ran it myself.

src/actions/helpers/subaccounts.js
jgzamora: I think it functionally works, but not going to do cause: 
```
Warning: Failed prop type: Invalid prop `value` supplied to `Radio`.
```

bkemper: After rereading my initial comment, I should have provided an example.  Don't worry about changing it, but for future discussion, this is what I was thinking.

```
// in src/pages/subaccounts/components/SubaccountForm.js, create function to handle the conversion
handleSubmit = ({ grantsRadio, ...values }) => {
  const grants = grantsRadio === 'all'
    ? this.props.grants  // full list of grants 
    : values.grants;       // checked list of grants

  this.props.handleSubmit({ ...values, grants });
}
```

This would keep component specific state from being passed to actions and asking actions to know and do more.

src/actions/helpers/tests/subaccounts.test.js
```diff
@@ -33,10 +33,7 @@ const keyFieldsSelect = {
 
 test('name only', () => {
   const result = formatSubaccount({ name: 'mySubAccount' }, getState);
-  expect(result).toEqual({ name: 'mySubAccount', setup_api_key: false });
-  expect(result).not.toHaveProperty('key_label');
-  expect(result).not.toHaveProperty('ip_pool');
-  expect(getState).not.toHaveBeenCalled();
+  expect(result).toMatchSnapshot();
```
bkemper: 👍

src/actions/subaccounts.js
bkemper: Yep, only if you were to call it inside a thunk.  If you return it, the [request is a thunk that dispatches for you](https://github.com/SparkPost/2web2ui/blob/4f9c9feff3ed2d41d8adc2d5b5b08277325ceada/src/actions/helpers/sparkpostApiRequest.js#L106).

PR #139, opened 2017-10-26
Added filter box to remaining pages and fixed small collection bug

jasonrhodes:


src/pages/recipientLists/ListPage.js
```diff
@@ -32,6 +32,12 @@ const renderCollection = (rows) => (
       getRowData={getRowData}
       pagination={true}
       rows={rows}
+      filterBox={{
+        show: true,
```
jonambas: Why do we have a `show` flag on here again? couldn't the collection just check if `filterBox` is passed through?

jasonrhodes: Me being over-prepared, I hate having to remove all the config if I want to turn something off, then have to put it all back to turn it back on. People around here _love_ turning things on and off...

jasonrhodes:
Not sure why this went to [closed] instead of [merged] 💩

PR #145, opened 2017-10-30
Integration with Sentry

bkemper:


src/helpers/errorTracker.js
```diff
@@ -0,0 +1,102 @@
...
+ * List of breadcrumbs to ignore to reduce noise
+ * @note Sentry has a 100 breadcrumb per event/error maximum
+ */
+const BLACKLIST = new Set([
+  '@@redux/INIT',
+  '@@redux-form/BLUR',
+  '@@redux-form/CHANGE',
+  '@@redux-form/DESTROY',
+  '@@redux-form/FOCUS',
+  '@@redux-form/REGISTER_FIELD',
```
jonambas: + `@@redux-form/UNREGISTER_FIELD`

package.json
```diff
@@ -94,7 +96,7 @@
     "whatwg-fetch": "2.0.3"
   },
   "scripts": {
-    "start": "node scripts/start.js",
+    "start": "REACT_APP_VERSION=local-$USER-$(git rev-parse HEAD) node scripts/start.js",
```
jasonrhodes: whoa this looks awesome, great code here

bkemper: Special thanks to you.

src/helpers/errorTracker.js
```diff
@@ -0,0 +1,103 @@
...
+export function breadcrumbCallback(crumb) {
+  if (BLACKLIST.has(crumb.message)) {
+    return false;
+  }
+
+  return crumb;
+}
+
+// Closure to safely enrich events with data from Redux store
+export function enrichData(store) {
```
jasonrhodes: can we call this outer function something so it's clear you're getting the real function?

My suggestion: `getEnricherOrDieTryin`

src/helpers/errorTracker.js
```diff
@@ -0,0 +1,103 @@
...
+    const dsn = `https://${sentry.publicKey}@sentry.io/${sentry.projectId}`;
+    const options = {
+      breadcrumbCallback,
+      dataCallback: enrichData(store),
+      environment,
+      release,
+      extra: { tenant }
+    };
+
+    this.service.config(dsn, options).install();
```
jasonrhodes: Kind of nitpicky I admit, but I think I might just reference `Raven` directly here instead of storing off `this.service` ... we aren't actually decoupled from Raven in this file almost anywhere (but we make it much easier to decouple later by only having to replace this one file, so still worth it). But in here, the `this.service` redirection is more mental energy for not much gain IMO.

jasonrhodes:
lgtm besides 1-2 comments :+1:

src/helpers/errorTracker.js
bkemper: I agree.  That is a much better name.

src/helpers/errorTracker.js
bkemper: It didn't feel great for me either.  I only did it for testing to avoid the magic of `jest.mock`.  I swapped it out... d9181f9.

jasonrhodes:
SHIPPP IT 🚢  :it:

PR #147, opened 2017-10-30
Code Splitting

jgzamora:
## Components
Use react-loadable to split heavy components into async imported components. Basic concept of how this works can be found here: [Code Splitting in Create React App | Serverless Stack](https://serverless-stack.com/chapters/code-splitting-in-create-react-app.html). We’ll use [react-loadable](https://github.com/thejameskyle/react-loadable), which gives us a few additional features:
	* Loading: react-loadable accepts a Loading component that can handle loading, time-outs, and errors when importing the split chunk.
	* Preload: components wrapped in react-loadable expose a preload method that makes it easy to load the chunk from other pages. (see templates list)

## Multiple Entry Points
The idea it to split off another entry point for our app that holds `/join`, `/auth`, `/invite`. We’d have to extract these pages into a separate `/other-index.js` location that we can tell webpack to load. I think this would make it so those pages have their own redux store and we’d have to set up some way for the info to be passed to the app’s store. We also discussed making this pages not use react/redux at all.  I think this requires it’s own R&D.

## Vendor Package

Split off our node_modules into a vendor chunk that is loaded separately from our app bundle. This split is beneficial for caching since only one of the two bundles will need to be updated with most changes. This is done using the [CommonsChunkPlugin](https://webpack.js.org/plugins/commons-chunk-plugin/). As we split more of our app by pages etc, this plugin can be leveraged to split out our most commonly shared modules and other cool stuff

There is also a third chunk that holds Webpack boilerplate. Not entirely sure what’s in there but it’s suggested by Webpack.


src/pages/reports/summary/components/AsyncLineChart.js
```diff
@@ -0,0 +1,19 @@
...
+import loadable from 'react-loadable';
+
+const LoadableLineChart = loadable({
+  loader: () => import('./LineChart'),
+  render(loaded, props) {
+    const Component = loaded.default;
+    return <Component {...props}/>;
+  },
+  loading() { //Actual Loading Component goes here
+    return <div>Loading...</div>;
```
jasonrhodes: Bring in our `<Loading />` component?

src/pages/templates/components/Editor.js
```diff
@@ -1,2 +1,19 @@
...
+import loadable from 'react-loadable';
+
+export const LoadableEditor = loadable({
+  loader: () => import('./_Editor'),
+  render(loaded, props) {
+    const Component = loaded.default;
+    return <Component {...props}/>;
+  },
+  loading() { //Actual Loading Component goes here
+    return <div>Loading...</div>;
```
jasonrhodes: `<Loading />`?

src/pages/reports/summary/components/AsyncLineChart.js
```diff
@@ -0,0 +1,19 @@
...
+  }
+});
+
+export default class LineChart extends Component {
+  render() {
+    return <LoadableLineChart {...this.props}/>;
+  }
+}
```
jasonrhodes: Why do we have to do this instead of exporting `LoadableLineChart` directly?

src/pages/reports/summary/components/AsyncLineChart.js
jonambas: Curious what this would look like. Think the summary page already pulls in Loading.

jasonrhodes: I think it will look like the access control double loading "stutter" you noticed.

jasonrhodes: https://github.com/thejameskyle/react-loadable#avoiding-flash-of-loading-component

jasonrhodes:
I did some more reading last night and still can't figure out what that runtime.js chunk is or how it works ... lol. Makes sense to me to leave it there anyway but I wish I knew wtf that was.

src/pages/reports/summary/components/AsyncLineChart.js
jgzamora: Our `<Loading />` component doesn't work for this out of the box for this, last time I tried. Want me to make it work in this ticket?

jonambas: What about it doesn't work? 

Could also just return null here

src/pages/reports/summary/components/AsyncLineChart.js
jgzamora: My bad, I think this is stuff I was doing as I was running into the edit templates error, and I thought I needed it.

jgzamora:
@jasonrhodes  / @jonambas I cleaned up the react-loadable boilerplate and made a basic loading component for it. It doesn't show up unless the it takes more that the default 200 milliseconds. Errors and timeouts just return null, but can be expanded on easily. I couldn't get the `withLoadable` idea to work. There we two issues:

1. webpack warning when trying to import use a path that is a variable (I think??):
```
Critical dependency: the request of a dependency is an expression
```
It's just a warning, but apparently we don't want to do it

2. The component just couldn't find the import when loaded :/.

src/components/loading/LoadableLoading.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { Loading } from 'src/components/loading/Loading';
+
+const LoadbleLoading = ({ pastDelay }) => {
```
jasonrhodes: `LoadableLoading` (wrong in 2 places in this file only)

jasonrhodes:
OK this looks good with 2 caveat comments:

1. If we use this pattern in more places I think we can still abstract it out, using the same pattern that `AsyncComponent` was using, ie `export default loadableHelper(() => import(./Editor))` and then the `LoadableLoading` stuff can move into the helper since that's the only place it'll ever be needed. That gets around the "dependency is an expression" problem but keeps the Loadable stuff in one place. We'll probably need to account for different settings though, also, in case we want different delay/timeout stuff etc.

1. I'm worried about having `Editor` and `AsyncEditor`, i.e. having to remember to import the specially-named one instead of having `Editor` be the async loaded component. That's why I had used `_Editor` to prevent anyone from accidentally importing the sync editor and blowing up our bundle and we don't realize. If we don't like the `_` prefix solution, can we think of another way to solve this?

PR #148, opened 2017-10-30
Templates - Duplicate

rajumsys:
Known issue: Sending domain selection. However, that's [a known issue](https://github.com/SparkPost/2web2ui/blob/master/src/pages/templates/components/Form.js#L74) anyway and will be handled later, separately.

### DIFF
--- a/src/config/routes.js
+++ b/src/config/routes.js
@@ -119,13 +119,7 @@ export default [
     layout: App
   },
   {
-    path: '/templates/create',
-    component: templates.CreatePage,
-    condition: hasGrants('templates/modify'),
-    layout: App
-  },
-  {
-    path: '/templates/create/:cloneFrom/duplicate',
+    path: '/templates/create/:id?',
     component: templates.CreatePage,
     condition: hasGrants('templates/modify'),
     layout: App
--- a/src/config/routes.js
+++ b/src/config/routes.js
@@ -125,6 +125,12 @@ export default [
     layout: App
   },
   {
+    path: '/templates/create/:cloneFrom/duplicate',
+    component: templates.CreatePage,
+    condition: hasGrants('templates/modify'),
+    layout: App
+  },
+  {
     path: '/templates/edit/:id',
     component: templates.EditPage,
     condition: hasGrants('templates/modify'),
--- a/src/pages/templates/CreatePage.js
+++ b/src/pages/templates/CreatePage.js
@@ -2,8 +2,14 @@ import React, { Component } from 'react';
 import { Link, Redirect } from 'react-router-dom';
 import { connect } from 'react-redux';
 import { reduxForm, formValueSelector } from 'redux-form';
+import _ from 'lodash';
+
 // Actions
-import { create } from '../../actions/templates';
+import { create, getDraft } from 'src/actions/templates';
+import { showAlert } from 'src/actions/globalAlert';
+
+// Selectors
+import { getTemplateById, cloneTemplate } from 'src/selectors/templates';
 
 // Components
 import Form from './components/Form';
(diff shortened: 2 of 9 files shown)


src/pages/templates/CreatePage.js
```diff
@@ -61,12 +89,24 @@ class CreatePage extends Component {
...
-  id: selector(state, 'id')
-});
+const mapStateToProps = (state, props) => {
+  const stateToProps = {
+    id: selector(state, 'id'),
+    loading: state.templates.getLoading
+  };
+
+  const template = getTemplateById(state, props.match.params.cloneFrom);
+  if (_.get(props, 'match.params.cloneFrom') && !_.isEmpty(template.draft)) {
```
jonambas: It'd be nice if we could move all of this logic into the `cloneTemplate` selector to keep this clean.

I'd look at some of the other selectors where we use the `reselect` package: https://github.com/reactjs/reselect

src/pages/templates/CreatePage.js
```diff
@@ -61,12 +89,24 @@ class CreatePage extends Component {
...
-const mapStateToProps = (state) => ({
-  id: selector(state, 'id')
-});
+const mapStateToProps = (state, props) => {
+  const stateToProps = {
+    id: selector(state, 'id'),
+    loading: state.templates.getLoading
+  };
+
+  const template = getTemplateById(state, props.match.params.cloneFrom);
```
jonambas: this is a bit of an anti-pattern - we should be passing through `state` and `props` unmolested I think.

Otherwise we would just do
```js
template: state.templates.byId[props.match.params.cloneFrom]
```

src/config/routes.js
```diff
@@ -125,6 +125,12 @@ export default [
     layout: App
   },
   {
+    path: '/templates/create/:cloneFrom/duplicate',
```
jonambas: Do we actually need a route here? Can we use the original create route with an optional param?

`/templates/create/:id?` - I _think_ that will work

src/pages/templates/CreatePage.js
```diff
@@ -2,34 +2,61 @@ import React, { Component } from 'react';
...
+    super(props);
+
+    this.state = {
+      shouldRedirect: false,
+      isDuplicating: !!props.match.params.cloneFrom
+    };
+  }
+
+  componentDidMount() {
+    if (this.state.isDuplicating) {
```
jonambas: this is the only place `state.isDuplicating` is in use - you could just check match.params.cloneFrom directly here

src/selectors/templates.js
```diff
@@ -1,2 +1,3 @@
 export const getTemplates = (state) => state.templates.list;
-export const getTemplateById = (state, props) => state.templates.byId[props.match.params.id] || { draft: {}, published: {}};
+export const getTemplateById = (state, id) => state.templates.byId[id] || { draft: {}, published: {}};
```
jonambas: I know this is not your change but we should use `defaultProps` on the component instead of setting a default object here in the selector

src/config/routes.js
rajumsys: `/templates/create/:id` will certainly work. i just added `/duplicate`  and named `cloneFrom` just to make it more meaningful. i also tried `/templates/create?cloneFrom=` but there is no clean way to parse query params.

src/pages/templates/CreatePage.js
rajumsys: why this will be anti pattern? generally, func should receive just the info it needed to work. see how test is now simplified (you don't need to prep `props`).

src/pages/templates/CreatePage.js
rajumsys: turns out right, i though i would need it in a few places :).

src/pages/templates/CreatePage.js
jonambas: It's an anti-pattern for redux selectors (https://github.com/reactjs/reselect) - don't really understand why tbh

src/config/routes.js
rajumsys: i wasn't feeling strong about my routes, so just changed to what you suggested. let's see what others think.

src/pages/templates/EditPage.js
```diff
@@ -107,7 +107,7 @@ class EditPage extends Component {
         disabled: submitting
       },
       { content: 'Delete', onClick: () => this.handleDeleteModalToggle() },
-      { content: 'Duplicate', disabled: true },
+      { content: 'Duplicate', to: `/templates/create/${match.params.id}` },
```
jonambas: Needs: `Component: Link,`

PR #151, opened 2017-10-31
add XSparky header to "all" sparkpost api calls

rajumsys:
`/authenticate` doesn't support this header. and also it uses a different instance of axios so that's not affected by this change.


src/config/default.js
```diff
@@ -11,7 +11,8 @@ const config = {
     },
     headers: {
       Authorization: 'Basic bXN5c1dlYlVJOmZhODZkNzJlLTYyODctNDUxMy1hZTdmLWVjOGM4ZmEwZDc2Ng==',
-      'Content-Type': 'application/x-www-form-urlencoded'
+      'Content-Type': 'application/x-www-form-urlencoded',
+      'xsparky': '1d24c3473dd52a2f4a53fb6808cf9a73'
```
bkemper: These headers are specific for the [authentication requests](https://github.com/SparkPost/2web2ui/blob/74de7462e85fbc10a24cf8315b155eafeafc17e5/src/helpers/http.js#L18).  I recommend moving this configuration to the top level with `apiRequestTimeout` as `apiSparkyToken`.

```
// src/config/default.js
const config = {
  // ...
  apiRequestTimeout: 15000,
  apiSparkyToken: '1d24c3473dd52a2f4a53fb6808cf9a73',
  // ...
};
```

src/actions/helpers/axiosInstances.js
```diff
@@ -5,7 +5,10 @@ const { apiBase, apiRequestTimeout, zuora: zuoraConfig } = config;
 
 export const sparkpost = axios.create({
   baseURL: apiBase,
-  timeout: apiRequestTimeout
+  timeout: apiRequestTimeout,
+  headers: {
+    'x-sparky': config.authentication.headers.xsparky
+  }
```
bkemper: There is another [Axios instance used for authentication](https://github.com/SparkPost/2web2ui/blob/74de7462e85fbc10a24cf8315b155eafeafc17e5/src/helpers/http.js#L6-L9).  To avoid adding 'X-Spark' header in both places, I recommend replacing the authentication instance with this instance from `axiosInstances`.

src/config/default.js
bkemper: Please ignore.  I see it was already addressed.

src/helpers/http.js
```diff
@@ -15,7 +12,7 @@ function useRefreshToken(refreshToken) {
     method: 'POST',
     url: '/authenticate',
     data: `grant_type=refresh_token&refresh_token=${refreshToken}`,
-    headers: authentication.headers
+    headers: _.merge({}, apiRequestHeaders, authentication.headers)
```
bkemper: This additional merge is not necessary.  The `apiRequestHeaders` have already been set as defaults in the `axiosInstances.sparkpost`.   The request headers, `authentication.headers`, will be merged with the defaults by Axios.

screen shot 2017-11-01 at 4 02 55 pm

PR #153, opened 2017-11-01
FAD-5496: IP pools create, edit, delete

ewandennis:
(This is a new PR on Raju's branch to clear out the outdated feedback).

### Questions
 - The edit page has a relatively complex update procedure to reassign any changed IPs and update the pool name. Should that update logic live somewhere else or be handled in an update-pool-and-ips action maybe?
    - (I personally like dispatching the form submission actions in the controlling component)

### TODOs
- [ ] Handle accounts on free plan
- [ ] Selectors...


src/pages/ipPools/EditPage.js
```diff
@@ -0,0 +1,139 @@
...
+          content: 'Delete',
+          onClick: this.toggleDelete
+        }
+      );
+    }
+  }
+
+  componentDidMount() {
+    this.props.listPools();
+    this.props.getPool(this.id);
```
rajumsys: Should we do something if any or all of these fail?

src/pages/ipPools/components/PoolForm.js
```diff
@@ -0,0 +1,114 @@
...
+export class PoolForm extends Component {
+  ipOptions = (pools) => pools.map((pool) => ({
+    value: pool.id,
+    label: pool.name
+  }));
+
+  poolSelect = (ip, pools, submitting) => (<Field
+    name={encodeIp(ip.external_ip)}
+    component={SelectWrapper}
+    options={this.ipOptions(pools)}
```
rajumsys: it seems to call `isOption` (and hence `map`) runs for each row? not a big deal as number of ips aren't usually many but probably we can calculate this once.

src/pages/ipPools/components/PoolForm.js
```diff
@@ -0,0 +1,114 @@
...
+        pagination={false}
+      />
+    );
+  }
+
+  render() {
+    const { isNew, pool, handleSubmit, submitting, pristine } = this.props;
+    const submitText = isNew ? 'Create IP Pool' : 'Update IP Pool';
+    const editingDefault = isDefaultPool(pool.id);
+    const helpText = editingDefault ? 'Sorry, you can\'t edit the default pool\'s name. Then it wouldn\'t be the default!' : '';
```
rajumsys: :D

PR #154, opened 2017-11-01
Removed environment configuration variable

bkemper:
This is to address [Ewan's code review comment](https://fisheye.int.messagesystems.com/cru/APP-2975#CFR-120512).  The `environment` variable is not needed.  Each Sentry project is setup per environment.


bkemper:
This was merged.

Issue #156, opened 2017-11-02
Relative ranges for reports


jasonrhodes:
Time to discuss how we want to do this! :)

I think the way I'd like this to work is that the relative range buttons to populate the query string with something like `range=last_hour`, and it gets calculated on every refresh. Then in the share widget, we present the option to "pin" the time ranges since when you share a chart, you are probably trying to share the exact window you're currently looking at (probably).

If we do this, I think it'd be nice to 

a) show the current fixed range you're looking at when in "relative" mode (updates on refresh), 
b) provide a refresh button that just refreshes the current data and not the whole page, 
c) provide a "pin" button next to refresh that converts the current relative range to fixed

jasonrhodes:
Doing this ^ work also puts us one step closer to thinking about how we might do "auto-refresh" later...

jonambas:
We already have a `relativeRange` key in `reportFilters`. I think all we have to do is pass this value to `updateLink` / `getShareLink` instead of `from` and `to` if relativeRange is not set to `custom`. And parse that back down on componentDidMount

b) I dont think we need this - cmd+r would work as intended if the url is updated correctly
c) maybe also not needed - choosing custom does this

Also if on a relative range, maybe we should also change the textfields value from dates to "{from date} - now" ?

jasonrhodes:
Re: existing relative range, sounds good. I might want to massage the names we're using for those ranges so the make more sense if someone is looking at the share URL but otherwise :+1:

I'd still like to do (b) because it's much faster to reload the chart only than to rebuild everything from scratch every time you want to refresh the chart. Seems like it'd be easy to do.

jasonrhodes:
I'll cut a ticket tomorrow to do this.

jasonrhodes:
See FAD-5924

Also relates to #202

PR #158, opened 2017-11-02
Bounce Report

jonambas:
Components:

#### BouncePage
- Wraps date filters
- Handles refresh action
- [x] TODO Handles no data state
- [x] TODO Handles share link & modal

#### ChartGroup
- Wraps recharts components and Legend together
- Keeps click and hover events in sync
- Everything below is reusable. This is the only component specific to Bounces.

#### Legend
- The table on the right
- Wraps **LegendItem** component

#### BounceChart
- The recharts pie chart

#### ActiveLabel
- Label in the middle of the pie chart

#### ActiveShape
- Custom recharts `Sector` that replaces generated sectors on hover

note: There is some funkiness when hovering while chart is animating - could try to disable pointer events while the chart is transitioning if this is annoying.

kapture 2017-11-02 at 15 57 32


src/helpers/bounce.js
```diff
@@ -0,0 +1,81 @@
...
+    const index = _.findIndex(categories, (o) => o.name === category);
+
+    if (index === -1) {
+      categories.push({
+        name: category,
+        count: rest.count,
+        children: [rest]
+      });
+    } else {
+      categories[index].count = categories[index].count + rest.count;
```
jonambas: I can technically get these numbers from aggregates but we have them from the classifications call, so why not use them?

src/pages/reports/bounce/BouncePage.js
```diff
@@ -0,0 +1,43 @@
...
+
+  handleModalToggle = (modal) => {
+    // TODO
+  }
+
+  render() {
+    const { loading, aggregates } = this.props;
+
+    const pageContent = !loading && aggregates && aggregates.countBounce === 0
+      ? <Panel sectioned><h6>No bounces to report.</h6></Panel> // TODO
```
jonambas: No `return <Loading/>` here as that would unmount date filters - need to handle loading in the components below on refresh

src/pages/reports/bounce/components/ChartGroup.js
```diff
@@ -0,0 +1,152 @@
...
+export class ChartGroup extends Component {
+  state = {
+    hoveredItem: null,
+    active: null
+  };
+
+  handleMouseOver = (e, hoverSet) => {
+    const { categories, types } = this.props;
+    const { active } = this.state;
+    const { name, count } = e;
```
jonambas: `e` is not actually the native browser event, but is a recharts specific event that passes through the data item being handled

src/pages/reports/bounce/components/ChartGroup.js
```diff
@@ -0,0 +1,152 @@
...
+
+    if (active) {
+      dataSet = active.children;
+    }
+
+    const hoveredItem = {
+      name,
+      count,
+      index: _.findIndex(dataSet, { name }),
+      dataSet: hoverSet
```
jonambas: `index` here is used to find the right sector to replace with ActiveShape. `dataSet` used to tell the chart which pie chart is being hovered.

src/pages/reports/bounce/components/ChartGroup.js
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  handleMouseOut = () => {
+    this.setState({ hoveredItem: null });
+  }
+
+  handleClick = ({ name, children, count }) => {
+    if (children) {
+      this.setState({
+        active: { name, children, count },
```
jonambas: this is used to replace Legend data and chart `primaryData`

src/pages/reports/bounce/components/tests/Legend.test.js
```diff
@@ -0,0 +1,51 @@
...
+  });
+
+  it('should render with secondaryData', () => {
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('should handle click', () => {
+    wrapper.find('LegendItem').at(1).simulate('click');
+    wrapper.find('LegendItem').at(2).simulate('click');
+    expect(props.handleClick).toHaveBeenCalledWith({name: 'primary'});
```
jonambas: found a way to `spyOn` functions in a stateless component 🙌

src/helpers/bounce.js
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * Formats category objects from getBounceClassifications
+ */
+function formatCategory(classification) {
+  const { bounce_category_name, bounce_class_name, bounce_class_description, count_bounce } = classification;
+  return {
+    category: bounce_category_name,
+    name: bounce_class_name,
+    description: bounce_class_description,
```
jonambas: Not currently using description anywhere, could maybe use these as tooltips or help text somewhere?

bkemper:
This looks great!

bkemper:
@jonambas Should the legend be highlighted in grey when hovering over the associated donut section?

screen shot 2017-11-03 at 1 02 55 pm

When hovering legend item:

screen shot 2017-11-03 at 1 02 46 pm

bkemper:
> note: There is some funkiness when hovering while chart is animating - could try to disable pointer events while the chart is transitioning if this is annoying.

I wouldn't worry about it.  I didn't noticed it.

src/actions/bounceReport.js
```diff
@@ -0,0 +1,66 @@
...
+
+        if (!aggregates[0].count_bounce) {
+          return;
+        }
+
+        const bounceParams = { ...aggregateParams, metrics: 'count_bounce' };
+
+        // dispatch(getBounceReasons(bounceParams)) For table data
+
+        dispatch(getBounceClassifications(bounceParams)).then((classifications) => {
```
bkemper: To avoid synchronous calls, what do you think about wrapping these two calls in a `Promise.all`?

src/actions/metrics.js
```diff
@@ -54,3 +54,15 @@ export function getTimeSeries(params = {}) {
   const path = 'deliverability/time-series';
   return fetch({ path , params });
 }
+
+export function getBounceClassifications(params = {}) {
```
bkemper: We should settle on a common naming convention.  There are several other functions in this file that are doing a similar thing, but called 'fetch'.

src/actions/tests/bounceReport.test.js
```diff
@@ -0,0 +1,24 @@
...
+  let dispatchMock;
+  let getStateMock;
+
+  beforeEach(() => {
+    dispatchMock = jest.fn((a) => Promise.resolve(a));
+    getStateMock = jest.fn((a) => Promise.resolve(a));
+  });
+
+  it('should dispatch a refresh action', async () => {
+    metricsActions.fetchDeliverability = jest.fn(() => [{ count_bounce: 1 }]);
```
bkemper: Does this restore after each test?  If not, we could use `jest.spyOn`.

src/actions/bounceReport.js
```diff
@@ -0,0 +1,66 @@
...
+            }
+          });
+
+          // refresh the date range
+          dispatch({
+            type: 'REFRESH_REPORT_RANGE',
+            payload: { ...options }
+          });
+        });
+      });
```
bkemper: What happens if `fetchDeliverability` throws an error?

src/actions/tests/bounceReport.test.js
```diff
@@ -0,0 +1,24 @@
...
+    const thunk = bounceReport.refresh();
+    await thunk(dispatchMock, getStateMock);
+    expect(dispatchMock.mock.calls).toMatchSnapshot();
+  });
+});
```
bkemper: Is it worth testing the other conditions?  (e.g. `relativeRange`, `count_bounce `)

src/helpers/bounce.js
```diff
@@ -0,0 +1,81 @@
+import { snakeToCamel } from 'src/helpers/string';
+import color from 'color';
+import _ from 'lodash';
+
+/**
+ * Camel cases aggregates
+ */
+function formatAggregates(aggregates) {
+  const map = {};
+  _.each(aggregates, (value, key) => map[snakeToCamel(key)] = value);
```
bkemper: I think [_.mapKeys](https://lodash.com/docs/4.17.4#mapKeys) could be used here.

src/helpers/bounce.js
bkemper: Do our users use the instructional tooltips?  I found at my last company from some user testing that our users would never hover the question marks to get help text.

src/helpers/bounce.js
```diff
@@ -0,0 +1,81 @@
...
+    count: count_bounce
+  };
+}
+
+/**
+ * Reshapes results of getBounceClassifications for the bounce chart
+ */
+function reshapeCategories(data) {
+  const categories = [];
+  const formatted = _.flatten(data).map((item) => formatCategory(item));
```
bkemper: I think [_.flatMap](https://lodash.com/docs/4.17.4#flatMap) could be used here.

`const formatted = _.flatMap(data, formatCategory);`

src/actions/bounceReport.js
jonambas: Yep, I had a `Promise.all` in place but removed for this ticket

src/actions/bounceReport.js
jonambas: This wraps `fetch` which wraps `sparkpostApiRequest` so the error should get handled there.

src/helpers/tests/reports.test.js
```diff
@@ -0,0 +1,34 @@
...
+  expect(reports.getSearch(options)).toEqual('?one=two&three=four');
+});
+
+
+
+it('Should get share link', () => {
+  Object.defineProperty(window.location, 'href', {
+    value: 'test.com?not=this',
+    writable: true,
+  });
```
bkemper: Does Jest know how to restore `window.location` after this test?  If not, I recommend using `jest.spyOn`.

src/helpers/bounce.js
jonambas: Yeah, I may just remove for now. We only have descriptions on the children, and not on any of the top level bounce metrics

src/pages/reports/bounce/components/ActiveLabel.module.scss
```diff
@@ -0,0 +1,27 @@
...
+.ActiveLabel {
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
+  text-align: center;
+}
+
+.Label {
+  display: block;
```
bkemper: Why not just use a block element?

src/pages/reports/bounce/components/ChartGroup.js
bkemper: Do you think a jsdoc style comment would help?

```
//**
  * @param {object} event - Rechart synthetic event
  * @param {string} event.name - the name property of element
  */
```

src/actions/tests/bounceReport.test.js
jonambas: I believe so. This is only mocking the function to return `[{ count_bounce: 1 }]`. Not actually checking that it is called

src/actions/bounceReport.js
```diff
@@ -0,0 +1,75 @@
...
+  };
+}
```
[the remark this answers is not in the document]
jonambas: I'm not sure, the summary chart action is pretty lengthy as well. It all needs to tie together to filter updates. It'll end up getting longer with the tables coming in. I could move `REFRESH_BOUNCE_REPORT` and `REFRESH_REPORT_RANGE` into their own actions though

src/actions/tests/bounceReport.test.js
jonambas: Added a mockRestore

src/helpers/tests/reports.test.js
jonambas: Renamed testURL to `http://phoenix.test` in the jest config - no need to overwrite `window`

src/helpers/reports.js
```diff
@@ -0,0 +1,90 @@
...
+  filtersList = typeof filters === 'string' ? [filters] : filters;
+
+  filtersList = filtersList.map((filter) => {
+    const parts = filter.split(':');
+    const type = parts.shift();
+    let value;
+    let id;
+
+    // Subaccount filters include 3 parts
+    // 'Subaccount:1234 (ID 554):554' -> { type: 'Subaccount', value: '1234 (ID 554)', id: '554' }
```
bkemper: Why is `554` needed twice?

src/actions/reportFilters.js
```diff
@@ -40,3 +40,11 @@ export function removeFilter(payload) {
     payload
   };
 }
+
+// refresh the date range
+export function refreshReportRage(options) {
```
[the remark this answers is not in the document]
jonambas: Oh nice catch

src/helpers/reports.js
jonambas: The subaccount filter in the api call only accepts ID, but the filter tag displays the name plus id. I could have done some regex to get it out of `(ID x)`, but I think probably safer to pass on its own.

Either way, were going to rework some of this for #156 and can look at it again. cc @jasonrhodes

src/actions/tests/bounceReport.test.js
```diff
@@ -0,0 +1,48 @@
+import * as bounceReport from '../bounceReport';
+import * as metricsActions from 'src/actions/metrics';
+import * as dateHelpers from 'src/helpers/date';
+import * as filterActions from 'src/actions/reportFilters';
+
+jest.mock('../helpers/sparkpostApiRequest', () => jest.fn((a) => a));
+jest.mock('src/helpers/bounce');
+jest.mock('src/helpers/metrics');
+
+Date.now = jest.fn(() => 1487076708000);
```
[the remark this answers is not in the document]
jonambas: 🤞

src/actions/bounceReport.js
```diff
@@ -0,0 +1,72 @@
...
+        if (!aggregates[0].count_bounce) {
+          return;
+        }
+
+        const bounceParams = { ...aggregateParams, metrics: 'count_bounce' };
+
+        // dispatch(getBounceReasons(bounceParams)) For table data
+        dispatch(fetchBounceClassifications(bounceParams)).then((classifications) => {
+          dispatch(refreshBounceReport({ aggregates, classifications }));
+          dispatch(refreshReportRage(options));
```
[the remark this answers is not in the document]
jonambas: Hmm yeah I noticed, I'll play around with it

PR #162, opened 2017-11-08
Subaccounts Resource View

jgzamora:


src/components/layout/Layout.js
```diff
@@ -12,9 +12,9 @@ import routes from 'src/config/routes';
 const Layout = ({ children, location }) => {
 
   // matchPath uses the same matching that <Route> uses
-  const route = _.find(routes, (route) => matchPath(location.pathname, {
-    path: route.path,
-    exact: true,
+  const route = _.find(routes, ({ path, exact }) => matchPath(location.pathname, {
+    path: path,
+    exact: typeof exact === undefined ? true : exact,
```
rajumsys: it should be `typeof exact === 'undefined`'. typeof returns [string](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof)

src/pages/subaccounts/components/ApiKeysTab.js
```diff
@@ -0,0 +1,67 @@
...
+        filterBox={filterBoxConfig}
+      />
+    );
+  }
+
+  renderEmpty() {
+    return (
+      <Panel.Section>
+        <p>This subaccount will need an API Key in order to send email</p>
+        <Button default to="/account/api-keys/create">Create</Button>
```
jgzamora: `primary` not `default`

src/components/layout/Layout.js
jasonrhodes: I just realized if you just default the arg on the way in, you don't need this ternary:
```js
const route = _.find(routes, ({ path, exact = true }) => matchPath(location.pathname, {
  path,
  exact,
  strict: false
}));
```

src/pages/subaccounts/DetailsPage.js
```diff
@@ -0,0 +1,88 @@
+import React, { Component } from 'react';
+import { connect } from 'react-redux';
+import { Link, withRouter, Route, Switch } from 'react-router-dom';
+import { Page, Tabs } from '@sparkpost/matchbox';
+import _ from 'lodash';
+
+import { getSubaccount } from 'src/actions/subaccounts';
+import { getSubaccount as getSubaccountSelector } from 'src/selectors/subaccounts';
```
jasonrhodes: what do you think about exporting it named as `selectSubaccount` and then using it like that here? I just started doing that and it feels like a nice convention for selectors...

src/pages/subaccounts/DetailsPage.js
```diff
@@ -0,0 +1,88 @@
...
+];
+
+export class DetailsPage extends Component {
+
+  constructor(props) {
+    super(props);
+
+    const { id } = props;
+    tabs[0].to = `/account/subaccounts/${id}`;
+    tabs[1].to = `/account/subaccounts/${id}/api-keys`;
```
jasonrhodes: There's something about using a prop in a constructor to mutate a variable that's scoped outside of the class that feels like it's going to cause headaches one day ... what if we do this in mstp?
```js
  const { subaccounts } = state;
  const { id } = props;
  return {
    id: props.match.params.id,
    loading: subaccounts.getLoading,
    subaccount: getSubaccountSelector(state),
    tabs: [
      {
        content: 'Settings',
        Component: Link,
        to: `/account/subaccounts/${id}`
      },
      {
        content: 'API Keys',
        Component: Link,
        to: `/account/subaccounts/${id}/api-keys`
      }
    ]
  };
};
```

jasonrhodes: OR level up:
```js
// src/selectors/subaccounts.js
function selectId(state, { id }) {
  return id;
}
export const selectDetailTabs = createSelector(
  [selectId],
  (id) => ([
    {
      content: 'Settings',
      Component: Link,
      to: `/account/subaccounts/${id}`
    },
    {
      content: 'API Keys',
      Component: Link,
      to: `/account/subaccounts/${id}/api-keys`
    }
  ])
)
```

src/pages/subaccounts/DetailsPage.js
```diff
@@ -0,0 +1,88 @@
...
+
+  componentDidMount() {
+    this.props.getSubaccount(this.props.id);
+    this.props.listPools();
+    this.props.listApiKeys();
+  }
+
+  render() {
+    const { loading, location } = this.props;
+    const selectedTab = _.endsWith(location.pathname, 'keys') ? 1 : 0;
```
jasonrhodes: Can do `location.pathname.endsWith()` instead of lodash here, and while you're at it can you do the same for the `_.endsWith` use in the webhooks DetailsPage? Every function we remove makes the build a few b smaller and I _just now_ realized `endsWith` is a native JS function :mindblown:

src/pages/subaccounts/DetailsPage.js
```diff
@@ -0,0 +1,88 @@
...
+        <Page title={'loading...'} breadcrumbAction={breadcrumbAction} >
+          <Tabs selected={selectedTab} tabs={tabs} />
+          <PanelLoading />
+        </Page>
+      );
+    }
+    const { match, subaccount } = this.props;
+
+    return (
+      <Page title={`${subaccount.name} (id: ${subaccount.id})`} breadcrumbAction={breadcrumbAction}>
```
jasonrhodes: I like the name/id title but I don't think you need the `id:` prefix

src/pages/subaccounts/components/SubaccountForm.js
```diff
@@ -1,73 +1,11 @@
 import React, { Component } from 'react';
```
jasonrhodes: Can we rename this file `SubaccountCreateForm.js`

src/pages/subaccounts/DetailsPage.js
jgzamora: Sure, I like the selector

src/pages/subaccounts/components/PanelLoading.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
```
jasonrhodes: 2 thoughts about this:

1) If we use this, it can probably move to src/components

2) I don't think I like this more than the regular Loading component -- what were you thinking this would help with?

src/pages/subaccounts/DetailsPage.js
```diff
@@ -0,0 +1,88 @@
...
+
+const breadcrumbAction = {
+  content: 'Subaccounts',
+  Component: Link,
+  to: '/account/subaccounts'
+};
+
+const tabs = [
+  {
+    content: 'Settings',
```
jasonrhodes: Is "Settings" right here? It feels off to me for some reason... I think "Details" feels better

src/pages/subaccounts/components/ApiKeysTab.js
```diff
@@ -0,0 +1,67 @@
...
+      return null;
+    }
+
+    const { keys } = this.props;
+    const count = !!keys.length;
+
+    return (
+      <Panel>
+        { !count && this.renderEmpty() }
+        { this.renderCollection(keys) }
```
jasonrhodes: Could we add some copy above the collection that explains what these are?

"The following API keys are assigned to this subaccount." or something?

src/pages/subaccounts/components/ApiKeysTab.js
```diff
@@ -21,21 +21,26 @@ export class ApiKeysTab extends Component {
...
     );
   }
 
   renderEmpty() {
     return (
-      <Panel.Section>
-        <p>This subaccount will need an API Key in order to send email</p>
-        <Button default to="/account/api-keys/create">Create</Button>
+      <Panel.Section style={{ textAlign: 'center' }}>
+        <p>This subaccount has no API Keys assinged to it. You can assing an existing one, or create a new one.</p>
```
jasonrhodes: assing -> assign

jasonrhodes: also assinged -> assigned

src/pages/subaccounts/helpers/getRowData.js
```diff
@@ -9,10 +8,11 @@ const STATUS_CLASSES = {
   terminated: styles.StatusTerminated
 };
 
+
 /*
  Subaccounts getRowData passed to TableCollection in ListPage.
 */
-const getRowData = ({ status, id, name }) => [
+const getRowData = ({ status, id, name, compliance }) => [
```
jasonrhodes: why `compliance` here?

src/selectors/subaccounts.js
```diff
@@ -1,29 +1,37 @@
...
 
-const resolveStatus = (status, compliance_status) => {
-  // Defensive status check for when it's not there
-  if (status && compliance_status === 'active') {
-    return status;
-  } else {
-    return compliance_status;
-  }
+const formatSubaccount = ({ compliance_status, status, ...rest }) => {
+  const compliance = compliance_status !== 'active' || !status;
```
jasonrhodes: What's this `|| !status` doing?

src/selectors/subaccounts.js
```diff
@@ -1,29 +1,37 @@
...
+export const selectSubaccounts = (state) => state.subaccounts.list.map((subaccount) => (formatSubaccount(subaccount)));
 
-  const formatted = { ...subaccounts.subaccount };
-  formatted.status = resolveStatus(formatted.status, formatted.compliance_status);
-  return formatted;
-};
+export const selectSubaccount = ({ subaccounts }) => (formatSubaccount(subaccounts.subaccount));
 
 export const getSubaccountIdFromProps = (state, props) => props.id;
+export const getSubaccountIdFromParams = (state, props) => props.match.params.id;
```
jasonrhodes: Is there any way that `props.match` or `props.match.params` won't be defined? Or `id` even? If so we can use `_.get` here but then we probably need `selectDetailTabs` to handle the case where `id` is null/undefined or whatever and just return an empty array for those tabs or something.

Not sure if that's a real scenario though?

src/pages/subaccounts/helpers/getRowData.js
jgzamora: accidentally left over from Icon logic

PR #163, opened 2017-11-08
Templates Test Data FAD-5344

jonambas:
Test data is saved in 3 places in this ticket:
- Create
- Update (save as draft)
- Publish (calls update first, then publish)
  - Test data saved under draft and published keys
- On send in "Preview & Send" page (for FAD-5342)
  - If Preview & Send is another route, we'll need to save published test data before route change

```js
// Template test data stored under same keys as old web ui:

const getTestDataKey = ({ id, username, mode }) => {
  return [ 'tpldata', username, id, mode === 'draft' ? 'd' : 'p' ].join('/');
}
```
```js
// Reshapes test data if it does not conform with default JSON structure:

if (!['substitution_data', 'metadata', 'options'].every((key) => key in testData)) {
  testData = { ...config.templates.testData, substitution_data: testData };
}
```
```js
// Default test data stored in config:

templates: {
  testData: {
    substitution_data: {},
    metadata: {},
    options: {}
  }
},
```

Also:
- ~Adds `displayName` to `<Loading/>` and now renders name instead of `<Component>` in tests~
- Updates MB
- Maybe fixes FAD-5814

Some things I found that we need to create tickets for:
- Subaccount assignment in templates edit
- From Email field
  - Uses a typeahead in current UI
  - Validation should fail if sending domain not verified
  - Should disable and default to `sandbox@sparkpostbox.com` if no verified sending domains (I think)


src/pages/templates/EditPage.js
```diff
@@ -1,93 +1,95 @@
...
-      return;
-    }
-
-    const { draft = {}, published = {}} = template;
+  componentDidUpdate(prevProps) {
+    const { template, showAlert, history } = this.props;
 
-    if (!Object.keys(draft).length && !Object.keys(published).length) {
-      this.setState({ redirectTo: '/templates/' }); // Redirect if no draft or published found
+    if (_.isEmpty(prevProps.template) &&  _.isEmpty(template.draft) && _.isEmpty(template.published)) {
```
jonambas: I think this fixes the redirect bug: FAD-5814

jonambas: Also - we aren't handling this 404 case on other pages with url params, but should get those fixed.

bkemper:
Smoke test feedback.

- Works exactly as described by acceptance criteria.
- Should the substitution data be copied when a template is duplicated?
- Not related to this feature, but the `metadata` and `options` keys are confusing to include while editing the template.

src/pages/templates/EditPage.js
bkemper: As we discussed, I don't believe this will actually fix the problem because `template` will never be empty.  Instead, how about the following.

```
// https://github.com/m0ppers/promise-any
import promiseAny from 'promise-any';

promiseAny([
  getDraft(match.params.id),
  getPublished(match.params.id)
]).catch(() => {
  // Redirect if no draft or published found
  history.push('/templates/');
  showAlert({ type: 'error', message: 'Could not find template' });
});
```

src/reducers/templates.js
```diff
@@ -25,7 +25,8 @@ export default (state = initialState, action) => {
         ...state,
         byId: {
           ...state.byId,
-          [action.payload.id]: { ...state.byId[action.payload.id], draft: action.payload }
+          [action.payload.id]: { ...state.byId[action.payload.id], draft: action.payload
+          }
```
bkemper: This may have been accidental.

PR #166, opened 2017-11-09
Feature/fad 5594

bizob2828:


src/pages/reports/message-events/MessageEventsPage.js
```diff
@@ -0,0 +1,72 @@
...
+        { error ? this.renderError() : this.renderCollection() }
+      </Page>
+    );
+
+  }
+
+}
+
+const mapStateToProps = ({ messageEvents }) => {
+  const events = formatMessageEvents(messageEvents.results);
```
jasonrhodes: Call this `selectMessageEvents` to signal it's a selector, and then pass it the entire `state` object. Selectors should always get the full state so they can be memoized.

src/selectors/messageEvents.js
```diff
@@ -0,0 +1,7 @@
+import _ from 'lodash';
+import moment from 'moment';
+
+export const formatMessageEvents = (events) => _.map(events, (event) => {
+  event.formattedDate = moment(event.timestamp).format('YYYY/MM/DD HH:mm');
+  return event;
+});
```
jasonrhodes: So you don't want to pass events to this function, you want the selector to be "selecting" it off of the state.

```js
import { createSelector } from 'reselect';

const getMessageEvents = (state) => _.get(state, 'messageEvents.results', [])

export const selectMessageEventsList = createSelector(
  [getMessageEvents],
  (events) => // map and format here

```

src/pages/reports/message-events/MessageEventsPage.js
bizob2828: cool, makes sense. i thought i saw an example elsewhere that wasn't passing in state but makes sense

src/reducers/messageEvents.js
```diff
@@ -0,0 +1,15 @@
+export default (state = { pending: false, error: null, results: []}, { type, payload }) => {
+  switch (type) {
+    case 'GET_MESSAGE_EVENTS_PENDING':
+      return { ...state, pending: true };
+
+    case 'GET_MESSAGE_EVENTS_SUCCESS':
+      return { ...state, error: null, pending: false, events: payload };
```
jonambas: Probably want `error: null` on the pending action and not here in success

src/pages/reports/messageEvents/tests/MessageEventsPage.test.js
```diff
@@ -0,0 +1,47 @@
...
+  });
+
+  it('should render error when action fails', () => {
+    wrapper.setProps({ error: { message: 'You done f\'ed up now' } });
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('should only render loading component while loading', () => {
+    wrapper.setProps({ loading: true });
+    expect(wrapper).toMatchSnapshot();
```
jonambas: I've been doing `expect(wrapper.find('Loading')).toHaveLength(1);`

Not sure if faster/better, but helps keep snapshot noise down

src/components/loading/Loading.js
```diff
@@ -5,7 +5,9 @@ import styles from './Loading.module.scss';
 /**
  * A centered loading animation
  */
-export const Loading = () => <LoadingSVG className={styles.Center}/>;
+export const Loading = function Loading() {
+  return <LoadingSVG className={styles.Center}/>;
```
jonambas: I fixed this in another branch too with `displayName`, but this is probably the better fix

src/pages/reports/messageEvents/MessageEventsPage.js
```diff
@@ -0,0 +1,72 @@
...
+
+  render() {
+    const { loading, error } = this.props;
+
+    if (loading) {
+      return <Loading />;
+    }
+
+    return (
+      <Page title='Message Events'>
```
jonambas: This page probably needs an empty state with 0 results - but no idea what the primary action or title would be.

Made this issue for some discussion: #167

src/reducers/messageEvents.js
bizob2828: error: null is the default, i'll remove this hear to be more clear

src/pages/reports/messageEvents/tests/MessageEventsPage.test.js
bizob2828: /shrug. I'll do whatever you think is best, just want to be consistent

src/components/loading/Loading.js
bizob2828: yea i noticed this and it bothered me slightly when I was asserting the snapshots.

src/reducers/messageEvents.js
jasonrhodes: Default case only happens on the very initial load -- once an error gets put into the state, it'll stay there forever until it's cleared. The corresponding PENDING event is a good place to clear since the API call is being attempted again.

src/pages/reports/messageEvents/MessageEventsPage.js
jonambas: Can leave out that message here - "There are no message events for your current query", we can do this in the filtering ticket

bizob2828: sure, but then what's the purpose of this attr without a message?

PR #171, opened 2017-11-13
Enforce lodash/chaining never linting rule

bkemper:
This is to enforce our "Don't use lodash _.chain" rule.


jasonrhodes:
Looks good, just linking to this list here so we can see if there are other rules or rulesets ("recommended"?) that we want to pull in here?

https://www.npmjs.com/package/eslint-plugin-lodash

jasonrhodes:
Let's merge this and then discuss other possible rules over time.

Issue #152, opened 2017-10-31
Convert all communication with API from snake to camel case


bkemper:
Our API, like most, uses snake cased properties.  For example, `company_name` property on [/account](https://developers.sparkpost.com/api/account.html).  The goal would be to convert all communication with our API from snake to camel case properties to avoid mixed case in our React application.

bkemper:
I have done enough work to realize that there is not a quick solution, so we should hold off on this for now.

PR #172, opened 2017-11-16
Transform request data, request params, and response data

bkemper:
For #152.  Use [camelcase](https://eslint.org/docs/rules/camelcase) rule to identify and enforce moving forward.


src/selectors/tests/__snapshots__/api-keys.test.js.snap
```diff
@@ -22,11 +22,8 @@ Object {
   "grants": Object {
     "grant": "true",
   },
-  "subaccount_id": "subId",
+  "subaccountId": "subId",
   "validIps": "ip",
-  "valid_ips": Array [
```
bkemper: @jonambas or @jasonrhodes Did we expect and use both keys, `validIps` and `valid_ips`?

bkemper:
This is the majority of the work for this ticket.  I need to smoke test and confirm metrics and grants are working as expected (they are values that won't be converted).

PR #175, opened 2017-11-17
Message Events Details

jonambas:
**Message Details**
- New route at `'/reports/message-events/details/:messageId'`
  - AC asks for `/:messageId/details`, but decided to just follow our other routes
- Adds meta details component & a selectable Table Collection
- Disclaimer about retention window appears if the last event  in the message history table is not an 'injection' event.
- Generates tooltips on meta data keys through the documentation enpoint

> * Need to handle the case where someone revisits a detail link for an event that is outside the retention window (just need a message explaining the case, I think)

Each details page is per message ID, not event ID. Either way, we don't handle this case on any other details page besides templates edit. Need to discuss if we want to do this on each detail page, or try to do it more globally.

Templates edit is handling this by kicking back to the list page and display an error

**Also:**
- Adds custom `rowComponent` prop to TableCollection
- Adds reusable `CopyField` component
  - Added to reports share links
  - With a `hideCopy` prop, it hides the copy button, but is still readOnly and selects all contents on focus
- Adds reusable `LabelledValue` component
  - Added to profile page
  - Added to billing page components
- Lints a lot


src/pages/reports/messageEvents/EventPage.js
```diff
@@ -0,0 +1,88 @@
+import React, { Component } from 'react';
+import { connect } from 'react-redux';
+import { withRouter, Link } from 'react-router-dom';
+
+import { getMessageHistory, getDocumentation } from 'src/actions/messageEvents';
```
rajumsys: What's the usage of documentation endpoint in message events page? doesn't look like existing UI making this API call!

src/actions/messageEvents.js
```diff
@@ -1,12 +1,41 @@
 import sparkpostApiRequest from 'src/actions/helpers/sparkpostApiRequest';
+import { subDays, format } from 'date-fns';
+// import { formatHistory } from 'src/helpers/messageEvents';
```
rajumsys: remove this?

src/actions/messageEvents.js
```diff
@@ -1,12 +1,41 @@
...
     meta: {
       method: 'GET',
-      url: '/message-events'
+      url: '/message-events',
+      params
     }
   });
 }
 
+export function getMessageHistory({ messageId, params = {}}) {
```
rajumsys: any reason not following the convention of triggering three events per action? (pending, success, failure)?

src/components/copyField/CopyField.js
```diff
@@ -0,0 +1,66 @@
...
+    let connectRight = null;
+
+    if (!hideCopy) {
+      const button = <Button onClick={this.handleCopy}><Icon name='Copy' size={14}/> Copy</Button>;
+      connectRight = copied
+        ? <Tooltip content='Copied to clipboard!'>{ button }</Tooltip>
+        : button;
+    }
+
+    return (
```
rajumsys: so how this component will be re-usable specially where we will use redux form? this component is using `TextField` component directly inside it. It would probably easier to re-use if this component's sole job was to render the button and copy the contents of the provided input/state, which we could pass a prop.

src/components/labelledValue/LabelledValue.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
```
rajumsys: what this component does?

src/pages/reports/messageEvents/EventPage.js
jonambas: I'm using it to get descriptions for the tooltips in the MessageDetails component

src/actions/messageEvents.js
jonambas: 👍 
It's still doing the 3 actions for getMessageEvents, but makes an additional action after because I needed a different reducer. Makes a call to the same url, and wasn't sure if I should  duplicate the action. Probably makes more sense to split it into two though so the original list doesnt get rewritten.

src/components/labelledValue/LabelledValue.js
jonambas: Just for styling. A label on the left and value on the right. I used something similar in the billing pages, so I just decided to combine them

src/pages/reports/messageEvents/EventPage.js
rajumsys: Ok, that makes sense. I didn't notice it until you said. Tooltip is nice but can we make it to take little more width? it's narrow width which breaking text in multiple lines.

src/reducers/messageEvents.js
```diff
@@ -1,4 +1,14 @@
-export default (state = { pending: false, error: null, results: []}, { type, payload }) => {
+import { formatDocumentation } from 'src/helpers/messageEvents';
+
+const initialState = {
+  pending: false,
```
rajumsys: i think i've seen usages of both `pending` and `loading` (and `somethingPending`, `somethingLoading`. We should probably stick to one pattern and i've no preference.

src/selectors/messageEvents.js
```diff
@@ -1,13 +1,29 @@
 import _ from 'lodash';
-import moment from 'moment';
```
rajumsys: we should get rid of `moment` if this was the only place it's used

src/pages/reports/messageEvents/components/MessageDetails.js
```diff
@@ -0,0 +1,52 @@
+import React, { Component } from 'react';
+
+import { Panel, Tooltip } from '@sparkpost/matchbox';
+import { LabelledValue, CopyField } from 'src/components';
+
+import _ from 'lodash';
+
+class MessageDetails extends Component {
```
rajumsys: probably a better name will be `EventDetails`.

src/pages/reports/messageEvents/components/MessageDetails.js
```diff
@@ -0,0 +1,52 @@
...
+  static defaultProps = {
+    details: {}
+  }
+
+  renderDetails = () => {
+    const { details, documentation } = this.props;
+    const type = details.type;
+
+    return _.keys(_.omit(details, 'formattedDate')).map((key, i) => {
```
rajumsys: it's hard to know what this part is doing. if you make a small inner function with meaningful name it'll be easier to understand the job the function is doing (without reading all codes each time you come here).

src/components/copyField/CopyField.js
jonambas: I don't think this field would ever be used with redux-form as it is `readOnly` - either way it's pretty easy to make it compatible if we ever needs to. Would only have to assign `value = props.input.value`

src/selectors/messageEvents.js
jonambas: Its used in a lot of other places unfortunately

src/pages/reports/messageEvents/components/MessageDetails.js
jonambas: 👍

src/reducers/messageEvents.js
jonambas: 👍

src/pages/reports/messageEvents/EventPage.js
jonambas: I'll make a note of that in Matchbox

PR #176, opened 2017-11-20
FAD-5062 WIP

bizob2828:


src/actions/api-keys.js
```diff
@@ -85,12 +85,13 @@ export function hideNewApiKey() {
...
 
-export function listApiKeys() {
+export function listApiKeys(key) {
   return sparkpostApiRequest({
     type: 'LIST_API_KEYS',
     meta: {
       method: 'GET',
-      url: '/api-keys'
+      url: '/api-keys',
+      ...formatKeyForRequest(key)
```
jasonrhodes: what's going on here? I can't see how this would do anything for the list GET endpoint

jasonrhodes: If this is for subaccount filtering can we pass in "subaccount" to the function and add the header here instead of re-using that formatKey function?

bizob2828: yea that's fine. i was just going to reuse that because it does that

jasonrhodes:
OK a few last cosmetic details:
* Remove the "upgrade" banner, those stacked banners look bad and I don't think we display that one in the existing app anymore, and if we do, we can figure out a better way to do it in the new app but do it later.
* Suggestd wording change for the suppression banner (I know you didn't come up with this):
>*Coming from another email service?*
>Welcome! Make sure you import your suppression list from your previous provider to avoid sending to people who have previously opted out, consistently bounced, etc. Learn more about migrating from SendGrid, Mailgun, or Mandrill.

The link in the existing app actually goes nowhere, I just now discovered. Sigh. We don't have a generic page for suppression migration, but we have migration guides for those 3 providers, so we should just link each provider name to the migration guide.

* https://www.sparkpost.com/migration-guides/sendgrid
* https://www.sparkpost.com/migration-guides/mandrill
* https://www.sparkpost.com/migration-guides/mailgun

Issue #116, opened 2017-10-13
Style login error


rajumsys:
Currently, login error is not differently styled making it hard to notice. Style it to make it prominent to other texts.

<img width="541" alt="screen shot 2017-10-12 at 10 21 24 pm" src="https://user-images.githubusercontent.com/14877079/31552146-996911ee-b004-11e7-9706-a6b93eb6dbea.png">

rajumsys:
We've handled this on #177

PR #177, opened 2017-11-20
SSO Login

rajumsys:
Todo:
When the review apparently looks good, we'll turn SSO on on tst and verify if it really works. 

- Moved login input fields to it's own component
- added action/reducers for SSO check
- add default config value


src/actions/auth.js
```diff
@@ -81,6 +83,35 @@ export function confirmPassword(username, password) {
...
 }
 
+export function ssoCheck(username) {
+  return (dispatch, getState) => {
+    dispatch({ type: 'SSO_CHECK' });
+
+    return sparkpostRequest({
+      method: 'GET',
+      url: `/users/${username}/saml`
+      // url: 'https://d79ace26-eeea-4129-8811-f51a31449321.mock.pstmn.io/api/v1/users/ddd/saml'
```
rajumsys: delete this!

src/actions/auth.js
```diff
@@ -81,6 +83,35 @@ export function confirmPassword(username, password) {
...
+    })
+    .then((payload) => {
+      dispatch({ type: 'SSO_CHECK_SUCCESS', payload: payload.data });
+    })
+    .catch((err) => {
```
jonambas: What's going on here? Why the need to specify `SUCCESS` and `FAIL`? Doesn't the request helper do this for us?

src/pages/auth/AuthPage.js
```diff
@@ -1,37 +1,32 @@
...
-
-  onChangeUsername = (evt) => {
-    this.setState({ username: evt.target.value });
-  };
-
-  onChangePassword = (evt) => {
-    this.setState({ password: evt.target.value });
-  };
+  redirectToSSO() {
+    window.location.href = `${config.apiBase}/api/v1/users/saml/login`;
```
jonambas: Should use `this.props.history.push(${config.apiBase}/api/v1/users/saml/login)`
if you wrap component `withRouter` you have access to the history object

src/pages/auth/AuthPage.js
```diff
@@ -1,37 +1,32 @@
...
+  }
 
-  onChangeRememberMe = (evt) => {
-    this.setState({ rememberMe: evt.target.checked });
-  };
+  ssoSignIn(username) {
+    return this.props.ssoCheck(username)
+    .then(() => {
```
jonambas: I think this will still run if the sso check `FAIL`, is that intended?

src/pages/auth/components/LoginForm.js
```diff
@@ -0,0 +1,81 @@
...
+import { Button } from '@sparkpost/matchbox';
+
+export class LoginForm extends Component {
+
+  renderLoginButtonText() {
+    const { loginPending } = this.props;
+
+    return loginPending
+          ? <span>
+              <i className="fa fa-spinner fa-spin" /> Logging In
```
jonambas: haha we don't have FA in the app, ca remove this for now

maybe just change text if `submitting`

src/reducers/auth.js
```diff
@@ -1,4 +1,6 @@
-const initialState = { loggedIn: false };
+import config from 'src/config';
+
+const initialState = { loggedIn: false, ssoEnabled: config.sso.enabled === true };
```
jonambas: Can remove `=== true`?

src/reducers/auth.js
```diff
@@ -30,6 +32,19 @@ export default (state = initialState, action) => {
...
+    case 'SSO_CHECK': {
+      return { ...state, errorDescription: null, loginPending: true };
+    }
+
+    case 'SSO_CHECK_SUCCESS': {
+      return { ...state, ssoEnabled: action.payload.saml, loginPending: false };
+    }
+
+    case 'SSO_CHECK_FAIL': {
+      const { errorDescription = 'An unknown error occurred' } = action.payload;
```
jonambas: Not really a fan of handling default error message here. Maybe better in the action or component depending on how we handle the errors

src/pages/auth/AuthPage.js
```diff
@@ -51,6 +46,12 @@ export class AuthPage extends Component {
       : <span>Log In</span>;
```
jonambas: `renderLoginButtonText` can be removed here

src/pages/auth/AuthPage.js
```diff
@@ -1,37 +1,32 @@
...
-  onClickSubmit = (evt) => {
-    const { username, password, rememberMe } = this.state;
-    evt.preventDefault();
-    this.props.authenticate(username, password, rememberMe);
-  };
+  regularSignIn(username, password, rememberMe) {
+    return this.props.authenticate(username, password, rememberMe);
+  }
 
   renderLoginError() {
```
jonambas: For the errors, can we either
a) Use the <Error> component from matchbox
or
2) Use global alerts

jonambas: this will close #116 too 😄

src/pages/auth/AuthPage.js
jonambas: Maybe you intended to do this?
```
componentWillRecieveProps(nextProps) {
  if (nextProps.auth.ssoEnabled) {
    this.redirectToSSO()
  }
}
```

src/pages/auth/AuthPage.js
jonambas: Error is undocumented but usage is `<Error error='message' />`
https://github.com/SparkPost/matchbox/blob/master/src/components/Error/Error.js

src/actions/auth.js
rajumsys: i've (clearly wrongly), used `sparkpostLogin`. I could have used `sparkpostApiRequest` here which will do.

src/pages/auth/AuthPage.js
rajumsys: when api request fails, `ssoEnabled` will be false.

src/pages/auth/AuthPage.js
rajumsys: sweet!

src/pages/auth/components/LoginForm.js
rajumsys: oh yeah, i noticed that and wanted to ask but forgot. thanks for noticing. i'll remove icon.

src/reducers/auth.js
rajumsys: i put it like so that having any truthy value doesn't enable it.

may be i was overthinking. who will do it if not us :P!

src/reducers/auth.js
rajumsys: i totally agree. i just remained compliant with `LOGIN_FAIL` type :).

src/pages/auth/components/LoginForm.js
```diff
@@ -0,0 +1,70 @@
...
+        { !ssoEnabled &&
+        <Field
+            name='rememberMe'
+            id='rememberMe'
+            label='Keep me logged in'
+            component={CheckboxWrapper}
+        />
+        }
+
+        <Button submit disabled={loginPending || pristine}>
```
jonambas: Can you add the `primary` prop here?

PR #178, opened 2017-11-21
FAD-5842 new async redux flow

jasonrhodes:


src/actions/auth.js
```diff
@@ -81,9 +77,29 @@ export function confirmPassword(username, password) {
...
+
+    if (auth.refreshing) {
+      return;
+    }
+
+    dispatch({ type: 'ATTEMPTING_REFRESH', payload: auth.refreshToken });
+
+    return sparkpostRefresh(auth.refreshToken)
+      // save new token results in cookie and store
+      .then(({ data: { access_token, refresh_token }} = {}) => {
```
bkemper: Is the default needed?

src/helpers/http/sparkpostApi.js
```diff
@@ -0,0 +1,82 @@
+import { sparkpost as sparkpostAxios } from 'src/helpers/axiosInstances';
+import { resolveOnCondition } from 'src/helpers/promise';
+import { refresh, logout } from 'src/actions/auth';
+import config from 'src/config';
+import _ from 'lodash';
+
+const maxRetries = _.get(config, 'authentication.maxRefreshRetries', 3);
```
bkemper: Do we need a default or could we handle the default in the configs?

jasonrhodes:
After working on this for a very long time, I finally discovered that even this version that does most of what I want breaks a lot of assumptions we have in our code that assume an action creator will return a promise (it no longer does in this set up). This breaks our ability to use some features of redux-form (like the `submitting` prop).

For now, I can't justify any more time on this, and I'm worried that we will uncover more and more problems. For now, I'm retiring this and going back to what we have, with the following caveats:

1. Global alerts will need to be manually handled in all components or action creators where we want those alerts to happen. We should probably lean toward the action creators but it doesn't really matter.

1. All API calls that fail will result in console errors about "unhandled promise rejection" (which may also flood Sentry, so we may need to ignore those or remove that handler for now). This is ugly but ultimately won't hurt anything, I don't think.

Later when we aren't all heads down on finishing the app, someone else can possibly take another run at solving this issue, or we can decide it's not a big enough problem to worry about after all.

PR #179, opened 2017-11-21
Support for multiple global alerts

### LINKED ISSUE #111: Multiple Global Alerts
Make this work:

kapture 2017-09-19 at 11 29 34

jonambas:
- Adds support for multiple alerts
- Moves alerts to the top right of the screen
- Adds react-transition-group v2 
- Adds a portal for alerts

Layering is now:
1. app
2. modals
3. alerts

Makes sure errors are always above modals for cases where we have forms within modals

Alerts now accept:
- `autoDismiss` boolean - defaults to `true`
- `timeout` auto dismiss timeout - defaults to `15000`

closes #111


src/reducers/globalAlert.js
```diff
@@ -1,10 +1,20 @@
...
-      return { ...state, ...payload };
+      return { ...state, alerts: [ ...state.alerts, payload ]};
 
     case 'CLEAR_GLOBAL_ALERT':
-      return {};
+      return {
+        ...state,
+        alerts: [
+          ...state.alerts.slice(0, payload),
+          ...state.alerts.slice(payload + 1)
```
bkemper: I don't know that users would be able to close alerts fast enough to expose a race condition, but what do you think about trying [`_.uniqueId`](https://lodash.com/docs/4.17.4#uniqueId).  Instead of slicing the array or using date as the identifier, you could let the reducer assign a unique ID on each 'SHOW_GLOBAL_ALERT'.

src/actions/globalAlert.js
```diff
@@ -1,12 +1,18 @@
-export function showAlert({ type = 'default', ...alert }) {
+import _ from 'lodash';
+
+export function showAlert(alert) {
   return {
     type: 'SHOW_GLOBAL_ALERT',
-    payload: { type, ...alert }
+    payload: {
+      id: _.uniqueId('alert_'),
```
bkemper: If you want, the reducer could create this id.

PR #181, opened 2017-11-28
SSO Login

rajumsys:
Handles the part after redirection from onelogin. 

Testing requires several steps to be done. Ping me when you're ready and I'll guide through it and also document those steps.

Testing steps:
https://docs.google.com/document/d/1eJNb4_K3tdSW0_9w1FKIcdCeW7PnkUnwmn4VWUuEK4M/edit

### DIFF
--- a/src/pages/sso/SSOPage.js
+++ b/src/pages/sso/SSOPage.js
@@ -41,7 +41,5 @@ export class SSOPage extends Component {
   }
 }
 
-const mapStateToProps = () => ({});
 
-
-export default withRouter(connect(mapStateToProps, { login })(SSOPage));
+export default withRouter(connect(null, { login })(SSOPage));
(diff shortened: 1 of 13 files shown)


src/pages/sso/SSOPage.js
```diff
@@ -0,0 +1,47 @@
...
+
+  render() {
+    return null;
+  }
+}
+
+const mapStateToProps = () => ({});
+
+
+export default withRouter(connect(mapStateToProps, { login })(SSOPage));
```
jonambas: can remove `const mapStateToProps = () => ({});`
and do `connect(null, { login })`

src/pages/sso/tests/__snapshots__/SSOPage.test.js.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`SSOPage renders correctly 1`] = `""`;
```
jonambas: can maybe just do `expect(wrapper).toEqual('');` or toEqual(null)

rajumsys: well, IMO, it's probably better snapshot take care of matching even if it's empty.

PR #185, opened 2017-11-30
FAD-5882 adds update flow and alerting for tracking domains

jasonrhodes:


src/pages/trackingDomains/components/DeleteButton.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+import { Button } from '@sparkpost/matchbox';
+
+export default function({ status, onClick }) {
+  if (status === 'pending' || status === 'blocked') {
```
rajumsys: not sure what this is doing. if delete button shouldn't be visible for some case, i think that should be outside of this component. the "using component" should have that condition when to include it or not.

rajumsys:
Looks good. Just that small comment.

package.json
```diff
@@ -106,6 +106,8 @@
     "build": "node scripts/build.js",
     "analyze": "source-map-explorer build/static/js/main.*",
     "test": "node scripts/test.js --env=jsdom",
+    "test-coverage": "npm test -- --coverage",
+    "test-ci": "CI=true npm test -- --coverage",
```
jasonrhodes: These are so I can have Bamboo use `npm run test-ci` instead of doing the command itself, so if we want to add more flags we don't have to change the automation scripts. I'm now running tests with coverage in Bamboo so we can fail if coverage drops below x%, and so we have the coverage reports in Bamboo artifacts.

package.json
```diff
@@ -116,10 +118,16 @@
...
       "<rootDir>/src/App.js",
       "<rootDir>/src/index.js",
+      "<rootDir>/src/actions",
       "<rootDir>/src/config",
       "<rootDir>/src/helpers/registerServiceWorker",
       "<rootDir>/src/reducers"
     ],
+    "coverageThreshold": {
+      "global": {
+        "statements": 53
```
jasonrhodes: This is where the value is right now so the idea is we can't let coverage drop below that, and we can keep nudging the number up as we go.

rajumsys: is there a way to tell it not to accept code with reduced coverage? it must go up or at least stay same? IMO, that's the intended behavior. Changing this value will may be difficult (in terms of maintainability)

rajumsys:
Now getting two (different) message after creating tracking domain. 
screen shot 2017-12-05 at 12 43 26 pm

package.json
jasonrhodes: Nope. The coverage run is stateless so it has no way of knowing what it was previously. This is one of those “this is better than nothing” scenarios, eventually it’ll be up to the % we want and we won’t ever have to change it.

jasonrhodes:
Good catches on the duplicate messages and console error, will address this afternoon.

rajumsys:
Works well now 👍

PR #186, opened 2017-11-30
FAD-5876: Bounce report: Added bounce reasons table

ewandennis:
## Notes
It's pretty rudimentary. The bounce reasons data loads separately from the chart b/c its big.

## TODO/Questions
 - ~How should we style the table rows? The reason text is long and ugly when wrapped.~
 - Separate the table out into a component to clean up the page component?
 - ~Alerts~
 - ~Tests~


src/actions/bounceReport.js
```diff
@@ -1,12 +1,36 @@
...
+  // if relativeRange is included, merge in the calculated from/to values
+  if (updates.relativeRange) {
+    Object.assign(updates, getRelativeDates(updates.relativeRange) || {});
+  }
+
+  return {
+    ...reportFilters,
+    ...updates
+  };
+}
```
bkemper: 👍 I really like the abstractions of this function into an easily testable function.  Since we are touching it, I think we should avoid mutating `update` and change `getRelativeDates` to always return an object.  Something as follows.  What do you think?

```
function buildCommonOptions(reportFilters, updates = {}) {
  return (
    ...reportFilters,
    ...updates,
    ...getRelativeDates(updates.relativeRange),
  );
}
```

src/pages/reports/bounce/BouncePage.scss
```diff
@@ -0,0 +1,9 @@
+.BounceTable {
+  .ReasonCell {
+    width: 26rem;
+    font-size: 90%;
```
bkemper: @jonambas Do we want to alter the font size?

src/pages/reports/bounce/BouncePage.js
```diff
@@ -77,15 +123,26 @@ export class BouncePage extends Component {
...
 
-const mapStateToProps = (state) => ({
-  filters: state.reportFilters,
-  loading: state.bounceReport.aggregatesLoading || state.bounceReport.categoriesLoading,
-  aggregates: state.bounceReport.aggregates
-});
+const mapStateToProps = (state) => {
+  const chartLoading = state.bounceReport.aggregatesLoading || state.bounceReport.categoriesLoading;
+  const tableLoading = chartLoading || state.bounceReport.reasonsLoading;
+  const pageLoading = typeof(chartLoading) === 'undefined';
```
bkemper: Could we avoid this check with default store values for `aggregatesLoading`, `categoriesLoading`, and `reasonsLoading`.

src/pages/reports/bounce/BouncePage.js
```diff
@@ -77,15 +123,26 @@ export class BouncePage extends Component {
...
+  const aggregates = state.bounceReport.aggregates;
+  return {
+    loading: pageLoading,
```
bkemper: Is `loading` being used?  I think it can be removed.

src/pages/reports/bounce/BouncePage.js
```diff
@@ -56,18 +62,58 @@ export class BouncePage extends Component {
...
+
+    if (tableLoading) {
+      return <PanelLoading />;
+    }
+
+    if (!reasons) {
+      return <Empty title={'Bounced Messages'} message={'No bounce reasons to report'} />;
+    }
+
+    const getRowData = this.getRowData.bind(this);
```
bkemper: I wish we had the
[jsx-no-bind](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md) eslint rule enabled.  Instead of binding, we can define `getRowData` as a arrow function class property.

```
getRowData = () => {
  // ...
}
```

src/pages/reports/bounce/BouncePage.js
```diff
@@ -56,18 +62,58 @@ export class BouncePage extends Component {
...
+
+  renderCollection() {
+    const { tableLoading, reasons } = this.props;
+
+    if (tableLoading) {
+      return <PanelLoading />;
+    }
+
+    if (!reasons) {
+      return <Empty title={'Bounced Messages'} message={'No bounce reasons to report'} />;
```
bkemper: Strings don't need to be wrapped with curly braces. 

```
return <Empty title='Bounced Messages' message='No bounce reasons to report' />;
```

src/pages/reports/bounce/BouncePage.js
```diff
@@ -56,18 +62,58 @@ export class BouncePage extends Component {
...
 
+  getRowData(rowData) {
+    const { totalBounces } = this.props;
+    const { reason, domain, bounce_category_name, bounce_class_name, count_bounce } = rowData;
+    return [
+      <div className='ReasonCell'>{reason}</div>,
+      domain,
+      bounce_category_name,
+      bounce_class_name,
+      `${count_bounce} (${Number((count_bounce / totalBounces) * 100).toFixed(2)}%)`
```
bkemper: Do we need to protect against a zero denominator?

src/pages/reports/bounce/BouncePage.js
```diff
@@ -1,19 +1,24 @@
...
+import PanelLoading from 'src/components/panelLoading/PanelLoading';
+import { Page, Panel } from '@sparkpost/matchbox';
 import ShareModal from '../components/ShareModal';
 import Filters from '../components/Filters';
 import ChartGroup from './components/ChartGroup';
 import Empty from '../components/Empty';
 
+import './BouncePage.scss';
+
+const columns = ['Reason', 'Domain', 'Category', 'Classification', 'Count (%)'];
```
jonambas: You can specify width here too
```
columns = [{ label: 'Reason', width: 'xx%' }, ... ]
```

src/pages/reports/bounce/BouncePage.scss
jonambas: This is fine, but for consistency - id specify width in the columns array and here: `font-size: rem(15)` maybe

src/actions/bounceReport.js
```diff
@@ -19,55 +43,63 @@ export function refresh(updates = {}) {
...
-      const params = getQueryFromOptions({ from, to });
-      dispatch(refreshTypeaheadCache(params));
-    }
+    const options = buildCommonOptions(state.reportFilters, updates);
 
-    const options = {
-      ...state.reportFilters,
-      ...updates
-    };
+    maybeRefreshTypeaheadCache(dispatch, updates);
```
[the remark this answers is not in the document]
ewandennis: I don't completely follow. Will follow up with you today.

src/pages/reports/bounce/BouncePage.js
ewandennis: This calculation only occurs after loading but I have updated the pre-loading value of `totalBounces` to `1`. _Maybe_ it will get displayed by accident after an error in a future iteration.

src/pages/reports/bounce/BouncePage.js
ewandennis: Removed `pageLoading` entirely.

PR #188, opened 2017-12-01
FAD-5884: coming soon pages

ewandennis:
## Changes
 - Added a ComingSoonPage with only an EmptyState in it.
 - Added the following coming soon pages:
     - reports
         - rejections
         - accepted
         - delayed
         - engagement
      - lists
         - recipient lists -> create
         - suppression lists
      - account
         - security
         - sending domains
         - smtp relay
 - Fixed a lint issue in tracking domains

## Questions
 - Did *not* add account -> usage since its a dup of the dashboard. TBD?
 - ComingSoonPage view and copy could use a sense of humour.
 - A modal might be nicer for recipient lists -> create

## Issues
 - ~My test snapshots are made in GMT so any with dates will fail outside the UK (!).~ Sidestepped with `TZ=` env variable. Better fix in progress on `feature/FAD-5876`.
 - The bounce report snapshot contains an ID attribute not in the snapshot. Just an out of date snapshot?

```
 FAIL  src/pages/reports/bounce/components/tests/BounceChart.test.js
  ● BounceChart:  › should render

    expect(value).toMatchSnapshot()

    Received value does not match stored snapshot 1.

    - Snapshot
    + Received

    @@ -45,11 +45,10 @@
              }
              dataKey="count"
              endAngle={450}
              fill="#808080"
              hide={false}
    -         id="recharts-pie-1"
              innerRadius={100}
              isAnimationActive={true}
              labelLine={true}
              legendType="rect"
              minAngle={2}
    @@ -79,11 +78,10 @@
              }
              dataKey="count"
              endAngle={450}
              fill="#808080"
              hide={false}
    -         id="recharts-pie-1"
              innerRadius={75}
              isAnimationActive={true}
              labelLine={true}
              legendType="rect"
              minAngle={2}

      at Object.it (src/pages/reports/bounce/components/tests/BounceChart.test.js:23:21)
          at Promise (<anonymous>)
          at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:188:7)
```

### DIFF
--- a/src/config/navItems.js
+++ b/src/config/navItems.js
@@ -86,7 +86,7 @@ export default [
         to: '/account/security'
       },
       {
-        label: 'Sending domains',
+        label: 'Sending Domains',
         to: '/account/sending-domains'
       },
       {
--- a/src/config/navItems.js
+++ b/src/config/navItems.js
@@ -18,6 +18,22 @@ export default [
         to: '/reports/bounce'
       },
       {
+        label: 'Rejections',
+        to: '/reports/rejections'
+      },
+      {
+        label: 'Accepted',
+        to: '/reports/accepted'
+      },
+      {
+        label: 'Delayed',
+        to: '/reports/delayed'
+      },
+      {
+        label: 'Engagement',
+        to: '/reports/engagement'
+      },
+      {
         label: 'Message Events',
         to: '/reports/message-events'
       }
@@ -39,7 +55,7 @@ export default [
       },
       {
         label: 'Suppressions',
-        to: '/lits/suppression-list'
+        to: '/lists/suppressions'
       }
     ]
   },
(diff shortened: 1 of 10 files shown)


jonambas:
I encountered that same snapshot issue last week. I believe it was an update to the recharts package

src/config/navItems.js
```diff
@@ -65,6 +81,18 @@ export default [
         label: 'Profile',
         to: '/account/profile'
       },
+      {
+        label: 'Security',
+        to: '/account/security'
+      },
+      {
+        label: 'Sending domains',
```
jonambas: capital D 😄

src/config/routes.js
```diff
@@ -170,6 +191,18 @@ export default [
     condition: hasGrants('recipient_lists/manage'),
     layout: App
   },
+  {
+    path: '/lists/recipient-lists/create',
```
jonambas: Recipient list edit page will need one as well. Or you could remove the links in the that table collection

PR #189, opened 2017-12-04
Fixed failing specs due to matchbox@1.8.6

bkemper:


bkemper:
@jonambas `package.json` is updated.

PR #190, opened 2017-12-04
Summary Table FAD-5335

jonambas:
- **Adds Summary Chart table**

**New Components and helpers**
- Adds reusable formatters `Unit`, `Size`, `Duration`, `Percent`.
- `formatUnit` `formatSize` `formatDuration` `formatPercent` also exported

**Filter Changes**
- Re-words `Domain` filter type to `Recipient Domain`
- Re-words `Sending/Bounce Domain` filter type to `Sending Domain`
- Fixes TypeaheadItem word wrapping
- Displays filter type before filter values in the Tag. ie `Domain: filter.com`.
- Restructures DateFilter grid at certain viewport widths

**Chart Changes**
- Removes time mode toggle
- Fixes DateFilter horizontal sizing with a `Popover` prop `wrapper='div'`
- Added x tick generator for `day` precision - shows ticks every Sunday
- Decreases LineChart height variance when more than 1 chart is shown
- Removes default tooltip transition (felt slow)
- Reduces line transition time 1500 > 400 ms


src/components/formatters/Unit.js
```diff
@@ -0,0 +1,35 @@
...
+const formatUnit = (value, unit) => {
+  if (typeof value === 'undefined') {
+    return '';
+  }
+
+  if (isNaN(parseFloat(value)) === true || isFinite(value) === false) {
+    return value;
+  }
+
+  if (unit === 'bytes') {
```
bkemper: Should the unit value match the component name?

src/components/formatters/tests/Unit.test.js
```diff
@@ -0,0 +1,61 @@
...
+  it('should render when undefined', () => {
+    const wrapper = mount(<Unit />);
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('should render when not a number', () => {
+    const wrapper = mount(<Unit value='not a number' />);
+    expect(wrapper).toMatchSnapshot();
+  });
+});
```
bkemper: To avoid this test file from getting any bigger, unit test`<Unit />` to confirm it renders the expected component and test the unit components individual.

bkemper:
@jonambas I don't know how you want to handle it, but users can unselect all metrics and click apply.

screen shot 2017-12-05 at 1 29 59 pm

src/reducers/summaryChart.js
```diff
@@ -5,21 +5,35 @@ const initialState = {
   loading: false,
```
bkemper: This should be replaced with `chartLoading` and `tableLoading`.

src/pages/reports/summary/components/async/LineChart.js
```diff
@@ -56,26 +57,36 @@ export default class SpLineChart extends React.Component {
...
   }
 
   // Manually generates X axis ticks
   getXTicks() {
     const { data, precision } = this.props;
+    let ticks;
+
+    // Shows ticks every Sunday
+    if (precision === 'day' && data.length > 15) {
+      ticks = data.map((tick) => tick.ts).filter((time) => moment(time).isoWeekday() === 7);
```
bkemper: Instead of iterating over the data set twice with `map` and `filter`, it can be `reduce`'d.

src/pages/reports/summary/components/async/LineChart.js
```diff
@@ -56,26 +57,36 @@ export default class SpLineChart extends React.Component {
...
+    }
 
+    // Show ticks every 15 minutes
     if (precision === '1min') {
-      // Show ticks every 15 minutes
-      return { ticks: data.map((tick) => tick.ts).filter((time) => moment(time).minutes() % 15 === 0) };
-    } else if (precision === '15min') {
-      // Show ticks every 30 minutes
-      return { ticks: data.map((tick) => tick.ts).filter((time) => moment(time).minutes() % 30 === 0) };
+      ticks = data.map((tick) => tick.ts).filter((time) => moment(time).minutes() % 15 === 0);
```
bkemper: Same.

src/pages/reports/summary/components/async/LineChart.js
```diff
@@ -56,26 +57,36 @@ export default class SpLineChart extends React.Component {
...
     }
 
-    return {};
+    // Show ticks every 30 minutes
+    if (precision === '15min') {
+      ticks = data.map((tick) => tick.ts).filter((time) => moment(time).minutes() % 30 === 0);
```
bkemper: Same.

src/pages/reports/summary/components/Table.js
```diff
@@ -0,0 +1,147 @@
...
+  handleRowClick = (item) => {
+    this.props.addFilter(item);
+    this.props.refresh();
+  }
+
+  getColumnHeaders() {
+    const { metrics, groupBy } = this.props;
+
+    const primaryCol = {
+      label: _.find(GROUP_CONFIG, { value: groupBy }).label,
```
bkemper: `GROUP_CONFIG` should probably be reshaped to an object with key equal to the groupBy value.

bkemper:
Should we at least sort by the dimension values?

screen shot 2017-12-05 at 1 41 47 pm

src/components/formatters/Duration.js
```diff
@@ -0,0 +1,37 @@
...
+
+  _.forEach(formatters, (unit, key) => {
+    if (value > unit) {
+      suffix = key;
+      formatted = value / unit;
+      return false;
+    }
+  });
+
+  return `${formatted.toFixed(2)} ${suffix}`;
```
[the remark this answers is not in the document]
jonambas: Should there be? `50s` vs `50 s`?

src/components/formatters/Duration.js
```diff
@@ -0,0 +1,37 @@
...
+  let suffix = 'ms';
+
+  const formatters = {
+    'h': 3.6e+6,
+    'm': 60000,
+    's': 1000
+  };
+
+  _.forEach(formatters, (unit, key) => {
+    if (value > unit) {
```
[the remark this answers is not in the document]
jonambas: I'm not really sure, I copied and pasted this from the old webui. I think the thresholds are one unit below the cutoff (see tests)

src/components/formatters/Percent.js
```diff
@@ -0,0 +1,23 @@
+import React, { Fragment } from 'react';
+
+const formatPercent = (value) => {
+  if (isNaN(parseFloat(value)) === true || isFinite(value) === false) {
+    return value;
+  }
+
+  let formatted = `${value.toFixed(2)}%`;
+
+  if (value < 0.01 && value > 0) {
```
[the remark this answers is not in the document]
jonambas: For rate metrics I dont think its possible to have negative values, but there may be other uses, ie if we every display % change over time

src/components/formatters/Duration.js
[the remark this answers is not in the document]
bkemper: I don't think so. :man_shrugging:

src/components/formatters/Unit.js
jonambas: Probably but, these are how they come from metrics. Feels weird to name these Bytes, Milliseconds

src/actions/metrics.js
```diff
@@ -1,9 +1,9 @@
-import moment from 'moment';
+import { format, subDays } from 'date-fns';
 import sparkpostApiRequest from 'src/actions/helpers/sparkpostApiRequest';
 
 const apiFormat = 'YYYY-MM-DDTHH:MM';
 const defaultParams = () => ({
-  from: moment().subtract(30, 'days').format(apiFormat),
+  from: format(subDays(Date.now(), 30), apiFormat),
```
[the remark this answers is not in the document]
jonambas: Tried doing this but went down a rabbit hole of fixing tests in other metrics pages. Going to hold off

jonambas:
Disabled metrics apply button if none are selected.

Also on sorting, I actually tried to build it out in this ticket. Had `sortable` and `defaultSort` keys on columns. But due to complexity it should probably be its own ticket,

bizob2828:
The old did abbreviations on values.  Is there a plan on showing 35k instead of 35,000, 3m vs 3,000,000?

jonambas:
Just noticed we have format functions in config/metric-units.js

I'll probably remove the format functions from the formatter components and use those isntead

src/components/formatters/Percent.js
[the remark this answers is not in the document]
ewandennis: This test looks fine assuming you intended to check that `value` is between `0` and `0.01`.

PR #191, opened 2017-12-04
Enable package lock to fix broken build

bkemper:
https://bamboo.int.messagesystems.com/browse/CENG-W2UI-1614

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,5 @@ junit.xml
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
-package-lock.json
 .vscode
 .idea/
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+engine-strict=true
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,0 @@
-package-lock = false
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,14734 @@
+{
+  "name": "react-webui",
+  "version": "0.1.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@sparkpost/matchbox": {
+      "version": "1.8.6",
+      "resolved": "https://registry.npmjs.org/@sparkpost/matchbox/-/matchbox-1.8.6.tgz",
+      "integrity": "sha512-kLgPT3iAeYGMxD/E9zRaJiUK+oMUmI9NbyT1xUHjeOO/vHj22So4KAvo+183yEHCNu5CPiQfT0Gjd+q/eqDWNw==",
+      "requires": {
+        "classnames": "2.2.5",
+        "react": "16.2.0",
+        "react-day-picker": "6.2.1",
+        "react-dom": "16.2.0",
+        "react-icon-base": "2.1.1"
+      }
+    },
+    "@timer/detect-port": {
+      "version": "1.1.3",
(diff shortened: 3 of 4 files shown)


jonambas:
So I think I'm okay with this if we can get people to update their global npm package locally - if there's no way to enforce that

bkemper:
It is important to mention that the reason @jonambas has asked to enforce npm version is because v5.3 vs v5.4 produced much different `package-lock.json` files.  v5.4 and v5.6 produced the same file.  This would be a pain and could cause merge conflicts.

bkemper:
@jonambas I think I found a way to enforce it.

> Unless the user has set the engine-strict config flag, this field is advisory only will produce warnings when your package is installed as a dependency.
> -- https://docs.npmjs.com/files/package.json#engines

We can set that config flag in `.npmrc`.  I just tested by setting version constraint to `>5.5` (I'm on 5.4) and it gave me the following.

```
$ npm install
npm ERR! code ENOTSUP
npm ERR! notsup Unsupported engine for react-webui@0.1.0: wanted: {"npm":">5.5"} (current: {"node":"8.5.0","npm":"5.4.2"})
npm ERR! notsup Not compatible with your version of node/npm: react-webui@0.1.0
npm ERR! notsup Not compatible with your version of node/npm: react-webui@0.1.0
npm ERR! notsup Required: {"npm":">5.5"}
npm ERR! notsup Actual:   {"npm":"5.4.2","node":"8.5.0"}

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/bkemper/.npm/_logs/2017-12-05T15_25_52_004Z-debug.log
```

This error would only be thrown when installing.  We could make install a prerequisite of all other scripts, but that feels like overkill to me.

What do you think?

jonambas:
If this prevents our bamboo snapshots from failing every time a package updates, I think I'm okay with it

@jasonrhodes ?

bkemper:
Welp, it is working, but `npm` needs to be upgraded on Bamboo.

https://bamboo.int.messagesystems.com/browse/CENG-W2UI120-2/

bkemper:
I found [npm-install.sh](https://github.com/SparkPost/automation-scripts/blob/master/phoenix-ui/npm-install.sh) uses [exports.sh](https://github.com/SparkPost/automation-scripts/blob/master/shared/exports.sh#L42) to manage the installation/upgrades of npm with `bamboo_NPM_VERSION_PROD`.  I can't find were `bamboo_NPM_VERSION_PROD` is set. 
 It is not on the [build plan](https://bamboo.int.messagesystems.com/chain/admin/config/configureChainVariables.action?buildKey=CENG-W2UI).

bkemper:
After a quick call with @bizob2828, the decision was to relax npm version to v5.3 for Bamboo.

bkemper:
@jonambas and @jasonrhodes Please provide final thoughts and approval.  Thanks!

jasonrhodes:
Wait how does this work? Won’t it cause issues like Jon had for us? Are we saying we will just manually enforce?

bkemper:
> Wait how does this work?

`npm install` will fail if the current machine has a version of npm less than 5.3.

> Won’t it cause issues like Jon had for us?

If an engineer is using v5.3, `npm install` will reformat `package-lock.json` which could be a pain when merging.

> Are we saying we will just manually enforce?

Yes.  For now, we will have to make sure everyone has npm v5.4 or greater.

jasonrhodes:
OK I have no further opinions on it then :)

PR #192, opened 2017-12-05
Add bundle analyzer, Make bounce chart async loaded

jonambas:


src/pages/trackingDomains/components/TrackingDomainRow.js
```diff
@@ -1,3 +1,4 @@
+/* eslint max-lines: ["error", 200] */
```
jasonrhodes: My ticket fixes this so I don't want this to be here later but I know we can't let this fail the build ... if my ticket goes down before yours we can remove and rebase, if not I'll remove this when I merge mine.

.gitignore
```diff
@@ -5,6 +5,7 @@
 
 # testing
 /coverage
+/bundle-report
```
jasonrhodes: Do me a favor and make a top-level folder called "build-reports" and then just ignore that (and remove coverage and bundle-report here) ... then to get that wired up you just need to set webpack-bundle-analyzer's config to put this in `build-reports/bundle/index.html` and add this to package.json inside the "jest" block:
```
"coverageDirectory": "build-reports",
```
https://facebook.github.io/jest/docs/en/configuration.html#coveragedirectory-string

I'm pretty sure it makes the "coverage" folder itself, wherever you put it.

This way I can just set the bamboo artifact once for "build-reports" and we can add whatever we want there ... and then eventually put a nice index page in there.

PR #193, opened 2017-12-05
Adds ci-test script and moves coverage reports into build-reports folder

jasonrhodes:


jasonrhodes:
@jonambas just gave me the OK on this, gonna get it down so we can keep moving :)

PR #194, opened 2017-12-05
Register Page (FAD-5683)

jgzamora:


src/config/routes.js
```diff
@@ -73,6 +74,11 @@ export default [
     public: true,
     component: AuthPage
   },
+  {
+    path: '/#/register',
```
jonambas: Why `#/register`?

src/pages/register/RegisterUserForm.js
```diff
@@ -0,0 +1,47 @@
...
+          component={TextFieldWrapper}
+          label='Password'
+          helpText='You password must have at least 8 characters and include 1 letter, 1 number, and 1 non-alphanumeric character.'
+          validate={[required, minLength(8)]}
+          warn={[hasLetter, hasNumber, specialCharacter, endsWithWhitespace]}
+          disabled={submitting}
+          type='password'
+          autoComplete='new-password'
+        />
+        <p><small>By joining, you agree to SparPost's <UnstyledLink>Terms of Use</UnstyledLink></small></p>
```
jonambas: to `https://www.sparkpost.com/policies/tou/`

Could maybe put this link in config

src/pages/register/RegisterUserForm.js
```diff
@@ -0,0 +1,47 @@
...
+        <Button primary submit disabled={submitting}>{ submitting ? 'Joining' : 'Join SparkPost' }</Button>
```
jonambas: I'd maybe use "Register/Registering" to match the route & panel title

src/pages/register/RegisterPage.js
```diff
@@ -0,0 +1,79 @@
...
+    return (
+      <div>
+        <div className={styles.LogoWrapper}>
+          <a href='https://www.sparkpost.com' title='SparkPost'>
+            <SparkPost.Logo />
+          </a>
+        </div>
+
+        <Panel sectioned accent title='Register'>
+          { this.renderRegisterPanel() }
```
jonambas: Could we add the same copy somewhere
> You have been invited by {email} to join ~the world's most powerful email platform~ SparkPost.

src/pages/register/RegisterUserForm.js
```diff
@@ -0,0 +1,47 @@
...
+          validate={required}
+          disabled={submitting}
+        />
+        <Field
+          name='password'
+          component={TextFieldWrapper}
+          label='Password'
+          helpText='You password must have at least 8 characters and include 1 letter, 1 number, and 1 non-alphanumeric character.'
+          validate={[required, minLength(8)]}
+          warn={[hasLetter, hasNumber, specialCharacter, endsWithWhitespace]}
```
jonambas: Do we need `endsWithWhitespace`? Could we trim or is that bad practice

jonambas:
Can we change this too 
> message: 'Invitation is on its way! Once your user signs up, you\'ll see them ~below~ on this page'

src/pages/register/RegisterUserForm.js
jgzamora: The problem is when they use a whitespace at the end as the special character, which technically works, but the input would then trim it automatically (I think?) and the password would no longer be valid.

src/pages/register/RegisterUserForm.js
```diff
@@ -0,0 +1,47 @@
...
+    } = this.props;
+
+    return (
+      <form onSubmit={handleSubmit}>
+        <Field
+          name='username'
+          component={TextFieldWrapper}
+          label='Username'
+          validate={required}
+          disabled={submitting}
```
jonambas: Current UI has this input perma disabled with the users email prefilled

src/config/routes.js
```diff
@@ -81,6 +82,11 @@ const routes = [
     public: true,
     component: SSOPage
   },
+  {
+    path: '/register',
```
jonambas: Should we rename this to `/set-password`?

PR #195, opened 2017-12-05
FAD-5878 added 404 page on non-existen route

bizob2828:


src/pages/pageNotFound/index.js
```diff
@@ -0,0 +1,16 @@
...
+  render() {
+    return (<EmptyState
+      title='Page Not Found'
+      image='Generic'>
+         <p>We cannot find what you're looking for, use the navigation to find your way.</p>
+     </EmptyState>);
+  }
+}
+
+export default connect(() => ({}), { })(PageNotFound);
```
ewandennis: This component is not using the redux store so we don't need to connect to it.

src/pages/pageNotFound/index.js
```diff
@@ -0,0 +1,16 @@
+import React, { Component } from 'react';
+import { connect } from 'react-redux';
+
+import { EmptyState } from '@sparkpost/matchbox';
+
+export class PageNotFound extends Component {
```
ewandennis: Given that this is a small stateless component with just a render function, I might just use the function form of React component:

```
export default () => <EmptyState ... />;
```

You'd be exporting a default function rather than a named component so your test would then have to import like this:

```
import PageNotFound from '../index';
```

PR #197, opened 2017-12-05
Suppressions search

rajumsys:
**ToDos**
- [x] Fix date picker drop down (doesn't show selected one after selected but search works fine)
- [x] Figure out how to remove label from Subaccount dropdown. If it can be done, remove label from email field too
- [x] Write tests
- [x] There are some small ToDos marked in the code, handle them. 
- [x] Rebase branch (to be merge-able) 
- [x] Refactor suppressions action creator and reducers to work (reuse) with [current one](https://github.com/SparkPost/2web2ui/blob/master/src/reducers/suppressions.js) 
- [x] After reusing suppressions action creator (previous task), now if you go to suppressions page from dashboard, you'll see the one record that dashboard loaded. I've to figure out to reset store on initial load of suppressions page. 

**Other notes**
- Results are to be implemented FAD-5848 but I implemented it here to mainly for personal satisfaction to see a full page :D 
- I've used TableCollection assuming the details will be displayed in a new page (like message events). if we want to show it inline, we'll have to use something else (like trackingDomainRow). We can settle it in FAD-5848 though. 
- I've implemented the deletion functionalities too. But to keep the PR as simpler as possible, I've moved it to a new branch which will be used for FAD-5848


src/pages/reports/bounce/BouncePage.js
```diff
@@ -7,13 +7,13 @@ import { addFilter, refreshTypeaheadCache } from 'src/actions/reportFilters';
 import { getShareLink, getFilterSearchOptions, parseSearch } from 'src/helpers/reports';
 import { showAlert } from 'src/actions/globalAlert';
 
-import { TableCollection } from 'src/components';
+import { TableCollection, Empty } from 'src/components';
```
rajumsys: I moved `Empty` component to top level to re-use it in Suppressions list page. But end up not using it due to the box shadow it's showing when I do so. I can revert this if preferred but if we can fix the box shadow, we can probably re-use it everywhere.

src/pages/suppressions/ListPage.js
```diff
@@ -0,0 +1,97 @@
...
+import EmailSearch from './components/EmailSearch';
+import Results from './components/Results';
+
+const primaryAction = {
+  content: 'Add Suppressions',
+  Component: Link
+  // to: '/lists/suppressions/upload'
+};
+
+const tabs = [
```
rajumsys: Tabs' labels sounds weird to me! Better names expected! 
`Search by Filters` (still this sounds somewhat weird)
`Search by Email`?

src/pages/suppressions/components/Results.js
```diff
@@ -0,0 +1,90 @@
...
+        getRowData={this.getRowData}
+        pagination
+      />
+    );
+  }
+
+  render() {
+    const { results = [], loading } = this.props;
+
+    if (loading) {
```
jgzamora: This loading only works for the filters page. The single recipient get doesn't set `listLoading` in the reducer

rajumsys: This is fixed

rajumsys:
> The search for email doesn't account for the 404 response for an email it can't find because the reducer doesn't set a value for the list on FAIL.

Reducer handles 404 as a non-error case now and UI is reflected correctly with messaging

src/actions/suppressions.js
```diff
@@ -1,14 +1,87 @@
...
+  );
+}
+
+export function searchRecipient({ email, subaccountId } = {}) {
+  const headers = {};
+  if (subaccountId) {
+    headers['x-msys-subaccount'] = subaccountId;
+  }
+
+  return (dispatch, getState) => dispatch(
```
jgzamora: this action needs a catch so that the `404` from not finding the recipient doesn't end up in sentry

src/actions/suppressions.js
```diff
@@ -1,14 +1,87 @@
...
+  }
+
+  if (!_.isEmpty(sources)) {
+    params.sources = sources.join(',');
+  }
+
+  return (dispatch, getState) => {
+    dispatch(refreshReportRange(reportFilters));
+
+    return dispatch(
```
jgzamora: this action should have a `.catch` either here or in the component that call it so that it throws an alert when it fails the search.

src/pages/suppressions/components/EmailSearch.js
```diff
@@ -0,0 +1,76 @@
...
+  handleChange(e, key) {
+    this.setState({ [key]: e.target.value }, this.refresh);
+  }
+
+  handleSubaccountSelect = (subaccount) => {
+    this.setState({ subaccountId: subaccount.id }, this.refresh);
+  }
+
+  componentDidMount() {
+    this.props.resetSearch();
```
jgzamora: I'd consider getting rid of this and keeping the results from the filter page

jgzamora:
So after a discussion with @jasonrhodes, we think the results should stay on the page until another search has been entered. So:

* remove componentDidMount refresh from both email search and filter search
* add a listSuppressions() to the componentDidMount of the ListPage to initialize results. Make sure the query doesn't break the date-picker

src/pages/suppressions/components/EmailSearch.js
jgzamora: See general comment

PR #204, opened 2017-12-07
FAD-5896 abstracted api key success banner, fixed update subby bug wh…

bizob2828:
…en compliance_status != active


src/components/apiKeyBanner/index.js
```diff
@@ -0,0 +1,49 @@
+import React, { Component } from 'react';
```
jasonrhodes: Minor nit but can you name this file ApiKeyBanner.js? we reserve `index.js` files to be true indexes that export other files, like `src/components/index.js` ... if there were more than 2 components in this folder it would get an index.js file that exported them all to make it easier to import later, etc.

src/components/apiKeyBanner/tests/ApiKeyBanner.test.js
```diff
@@ -0,0 +1,43 @@
...
+    expect(props.hideNewApiKey).toHaveBeenCalledTimes(1);
+  });
+
+  it('should set copied state to true on click', () => {
+    const banner = new ApiKeySuccessBanner();
+    banner.props = { newKey: 'foo' };
+    banner.setState = jest.fn();
+    banner.onClickBanner();
+    expect(banner.setState).toHaveBeenCalledWith({ copied: true });
+  });
```
jasonrhodes: This works but if you wanted to test this from a click, you can use enzyme (have to use mount so that you get the whole DOM for the component)
```js
const wrapper = mount(<ApiKeySuccessBanner {...props} />);
expect(wrapper.state('copied')).toEqual(false);
wrapper.find('button').simulate('click');
expect(wrapper.state('copied')).toEqual(true);
```

src/pages/subaccounts/components/formFields.js
```diff
@@ -81,29 +81,16 @@ const ApiKeyFields = ({ show, showGrants = false, grants, disabled }) => {
...
-  } else {
-    return (
+const StatusSelect = ({ disabled, compliance }) => (
       <Field
         name="status"
         component={SelectWrapper}
         options={statusOptions}
         label="Status"
         disabled={disabled}
+        helpText={compliance ? 'Please email compliance@sparkpost.com for help with your subaccount.' : ''}
```
jasonrhodes: Can you include the previous context here in addition to the email note? Would be good to quickly check with Nichelle if that's the right email too.

src/pages/subaccounts/components/SubaccountEditForm.js
```diff
@@ -41,8 +41,7 @@ export class SubaccountEditForm extends Component {
 const formName = 'SubaccountEditForm';
 
 const mapStateToProps = (state, { subaccount }) => {
-  const { compliance } = subaccount;
-  const status = compliance ? `System ${subaccount.status}` : subaccount.status;
+  const { compliance, status } = subaccount;
```
jasonrhodes: I prefer to keep some context here on a sparkpost-set status, can we put that back?

jasonrhodes:
There's a bug with the API key banner where if you create a new subaccount with an API key but don't manually dismiss the banner, the banner will appear on any other subaccount details page you visit. We don't have this exact problem with the regular API keys page because it displays on the list page and not the details page (if it persists on the list page forever until you dismiss, that's not that weird).

We _could_ try to clear it but I think that's gonna be complicated ... I was thinking maybe we should just adjust how we store the "new key" in the redux store, and make it connected to subaccount ID? Then if you are on a subaccount details page that has a "new key" value, it shows up (and gets cleared out on dismiss of the notification)?

src/pages/subaccounts/components/formFields.js
jonambas: This is also in config
```
contact: {
    abuseEmail: 'compliance@sparkpost.com',
    contactEmail: 'hello@sparkpost.com',
    supportEmail: 'support@sparkpost.com',
    billingEmail: 'billing@sparkpost.com'
  },
```

src/components/apiKeyBanner/tests/ApiKeyBanner.test.js
bizob2828: thanks my google fu wasn't showing how to properly do this.

bizob2828:
Good catch.  The bug seems weird in old situation too.  We don't want that persisting if you perform other actions.  I'll figure out how to clear this

src/pages/api-keys/tests/ListPage.test.js
```diff
@@ -6,6 +6,7 @@ import { ListPage } from '../ListPage';
 const props = {
   loading: false,
```
jasonrhodes: Define props inside beforeEach so we never worry about mocks leaking?

bizob2828: let me split out mocks and do the `jest.clearAllMocks()` instead

Issue #174, opened 2017-11-16
Upgrade to Eslint v4


bkemper:
Version 4.0.0 was [released on June 11, 2017](https://eslint.org/blog/2017/06/eslint-v4.0.0-released).  We are [locked on v3.19.0](https://github.com/SparkPost/2web2ui/blob/8db1a3922e7190aac39c7c87d2632c02678c16ff/package.json#L58).  Before long, we should upgrade.

ewandennis:
resolved

PR #208, opened 2017-12-10
Update eslint and fix linting

rajumsys:
I was just looking into what it takes to do #174. 

Seems it wasn't that complicated. A few notes
- I removed all eslint-* packages first and installed each package it complained about. I didn't need to install a few packages. Might be there were not used? 
- Only issue that didn't autofix was [`no-useless-escape`](https://eslint.org/docs/user-guide/migrating-to-4.0.0#-eslintrecommended-changes). So I needed to manually update `src/helpers/regex.js` and `src/helpers/string.js` to remove "useless" escapes it complained about.


src/helpers/regex.js
```diff
@@ -1,6 +1,6 @@
-export const emailRegex = /^[a-z0-9!#$%&'*+\/=?^_`{|}~.-]+@[a-z0-9]([a-z0-9-]*[a-z0-9])?(\.[a-z0-9]([a-z0-9-]*[a-z0-9])?)*$/i;
-export const emailLocalRegex = /^[a-z0-9!#$%&'*+\/=?^_`{|}~.-]+$/i;
-export const usernameRegex = /^[a-z0-9_\-\.]+$/;
+export const emailRegex = /^[a-z0-9!#$%&'*+/=?^_`{|}~.-]+@[a-z0-9]([a-z0-9-]*[a-z0-9])?(\.[a-z0-9]([a-z0-9-]*[a-z0-9])?)*$/i;
+export const emailLocalRegex = /^[a-z0-9!#$%&'*+/=?^_`{|}~.-]+$/i;
+export const usernameRegex = /^[a-z0-9_\-.]+$/;
```
bkemper: This caught my attention, but I confirmed that dot and forward slash are both matched literally when in brackets.

package-lock.json
```diff
@@ -14351,7 +14343,7 @@
       "dependencies": {
         "fs-extra": {
           "version": "0.30.0",
-          "resolved": "https://registry.npmjs.org/fs-extra/-/fs-extra-0.30.0.tgz",
+          "resolved": "http://registry.npmjs.org/fs-extra/-/fs-extra-0.30.0.tgz",
```
bkemper: When I ran `npm install`'ed it changed this to "https".

rajumsys: that's weird! why did it changed to `http`? i've npm `5.5.1`. I'll run npm install again on a clean slate!

src/actions/helpers/tests/sparkpostApiRequest.test.js
```diff
@@ -37,7 +37,7 @@ describe('Helper: SparkPost API Request', () => {
     axiosMocks.sparkpost.mockImplementation(() => Promise.resolve({ data: { results }}));
   });
 
-  it('should successfully call the API', async () => {
+  it('should successfully call the API', async() => {
```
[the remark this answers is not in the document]
rajumsys: > We should always have a space before arrow functions, especially async.

Why is that?

Yes, I'm not changing any policy in this PR. Just upgrading package. Probably we can do it in the ticket we groomed today.

package-lock.json
bkemper: hmm... I am on npm v5.6.0

rajumsys: it should definitely be https. so i would update to 5.6

src/actions/helpers/tests/sparkpostApiRequest.test.js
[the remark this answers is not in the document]
bkemper: I should have prefaced that with "I think", but I have changed my mind.  I think we should leave it as is.  `eslint-config-sparkpost` sets "anonymous" to "never".  That already forces us to do, `function()`.

PR #213, opened 2017-12-12
Add style docs

jonambas:


docs/styles.md
```diff
@@ -0,0 +1,84 @@
...
+    </div>
+  );
+}
+```
+
+## Global Styles
+Global CSS should be avoided, but is necessary in some cases - ie. overwritting package styles like recharts or ace editor that don't support styling through React. CSS class names will no longer be unique, so be sure to encapsulate them properly.
+
+```js
+import './MyGlobal.scss';
```
rajumsys: > CSS class names will no longer be unique, so be sure to encapsulate them properly.

So is it filename (not having `module`) or scoping inside `#someName` makes it "not unique"?

jonambas: It's the `.module` part of the filename that will turn on css-modules. I can make that more clear

PR #218, opened 2017-12-15
[FAD-5936] Pre-commit and build-time linting

jasonrhodes:
To test locally:
* `git fetch && git checkout hotfix/FAD-5936`
* Make a throwaway branch based on this one, so you can test commits `git checkout -b throwaway`
* `npm install`
* Add something valid to a file and do `git add` for that file
* Add a `console.log` to a different file but _don't_ do `git add` on that one
* Commit your valid change, should commit fine
* Now do `git add` on the invalid file with the console.log in it
* Try to commit, notice you get an error and the commit has failed
* Replace the `console.log` with a line missing a semi-colon or some other "fixable" error
* Try to commit, it will go through fine and the file will have been committed as "fixed"
* Delete your test branch `git checkout - && git branch -D throwaway`

You can also try running tests to verify the configuration still works after being moved, and try doing `npm run build` to verify linting happens pre-build.

Last: notice that if you run `npm test -- --coverage` that actions are now covered (or not covered as the case may be)

### DIFF
--- a/jest.config.json
+++ b/jest.config.json
@@ -1,0 +1,50 @@
+{
+  "collectCoverageFrom": [
+    "src/**/*.{js,jsx}"
+  ],
+  "coverageDirectory": "build-reports",
+  "coverageReporters": [
+    "lcov",
+    "text"
+  ],
+  "coveragePathIgnorePatterns": [
+    "<rootDir>/src/App.js",
+    "<rootDir>/src/index.js",
+    "<rootDir>/src/config",
+    "<rootDir>/src/helpers/registerServiceWorker",
+    "<rootDir>/src/reducers"
+  ],
+  "coverageThreshold": {
+    "global": {
+      "statements": 53
+    }
+  },
+  "setupFiles": [
+    "<rootDir>/config/polyfills.js"
+  ],
+  "setupTestFrameworkScriptFile": "<rootDir>/config/jest/setup.js",
+  "testMatch": [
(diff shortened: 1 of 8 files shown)


jest.config.json
```diff
@@ -0,0 +1,50 @@
...
+    "[/\\\\]node_modules[/\\\\].+\\.(js|jsx)$"
+  ],
+  "moduleNameMapper": {
+    "^react-native$": "react-native-web"
+  },
+  "snapshotSerializers": [
+    "<rootDir>/node_modules/enzyme-to-json/serializer"
+  ],
+  "modulePaths": [
+    ""
```
bkemper: Is this valid?

package.json
```diff
@@ -6,6 +6,36 @@
...
   },
+  "scripts": {
+    "start": "REACT_APP_VERSION=local-$USER-$(git rev-parse HEAD) node scripts/start.js",
+    "build": "npm run lint && node scripts/build.js",
+    "analyze": "source-map-explorer build/static/js/main.*",
+    "test": "node scripts/test.js --env=jsdom --config=./jest.config.json",
+    "test-coverage": "npm test -- --coverage",
+    "test-ci": "CI=true npm test -- --coverage",
+    "lint": "eslint src",
+    "lintfix": "eslint src --fix",
```
bkemper: Should we hyphenate `lint-fix` to match test?

PR #219, opened 2017-12-16
Hotfix/fad 5592

bizob2828:


src/actions/tests/__snapshots__/bounceReport.test.js.snap
```diff
@@ -3,14 +3,7 @@
...
   Array [
-    Object {
-      "payload": Object {
-        "from": "2017-02-14T12:02",
-        "metrics": undefined,
-        "to": "2017-05-21T19:05",
-      },
-      "type": "REFRESH_REPORT_RANGE",
-    },
+    undefined,
```
jasonrhodes: Since this snapshot is now useless, I've got some ideas about how to replace it that I'll put in the test file.

src/actions/tests/bounceReport.test.js
```diff
@@ -1,40 +1,34 @@
...
     metricsActions.fetchDeliverability = jest.fn(() => [{ count_bounce: 1 }]);
+    metricsHelpers.getQueryFromOptions.mockImplementation(() => ({ from, metrics }));
 
-    getRelDatesSpy = jest.spyOn(dateHelpers, 'getRelativeDates');
-    dateHelpers.getRelativeDates = jest.fn(() => ({ from, to }));
-
-    metricsHelpers.getQueryFromOptions.mockImplementation(() => ({ from }));
 
     dispatchMock = jest.fn((a) => Promise.resolve(a));
     getStateMock = jest.fn((a) => Promise.resolve(a));
```
jasonrhodes: This should return an object:
```
getStateMock = jest.fn(() => ({}));
```
If we need to add state keys for tests, you can store it in scope as `stateMock` and then assign it to an empty object in this `beforeEach`, then mutate it in any test.

jasonrhodes: PS I know you didn't change this, but not sure how that isn't blowing things up so I'm just calling out the fix.

src/actions/tests/bounceReport.test.js
```diff
@@ -43,36 +37,18 @@ describe('Action Creator: Bounce Report', () => {
     expect(dispatchMock.mock.calls).toMatchSnapshot();
```
jasonrhodes: This has changed under the hood so the snapshot is asserting on `undefined` because the methods are mocked and returning undefined. 

If you import all of the reportFilters via `*` like this:
```js
import * as reportFilters from 'src/actions/reportFilters';
```
Then you can access all of the auto-stubbed methods as spies. 
```diff
  it('should dispatch a chart refresh action', async() => {
    const thunk = bounceReport.refreshBounceChartMetrics();
    await thunk(dispatchMock, getStateMock);
-   expect(dispatchMock.mock.calls).toMatchSnapshot();
+   expect(reportFilters.maybeRefreshTypeaheadCache).toHaveBeenCalledTimes(1);
+   expect(reportFilters.refreshReportRange).toHaveBeenCalledTimes(1);
  });
```

jasonrhodes: Actually I'm going to PR against this PR just to show some test ideas I had, easier to explain that way.

src/helpers/date.js
```diff
@@ -37,7 +37,8 @@ function getStartOfDay(date) {
 }
 
 function getRelativeDates(range) {
-  const to = new Date();
+  const now = moment.utc();
```
jasonrhodes: Glad this works for what we need mock-wise. Since we now have `now` as a moment object, no reason to `moment(to)` over and over below, can just use ```
from: now.clone().subtract()
```
etc.

Comment (jasonrhodes) [reply src/helpers/date.js]: (Hm I wonder if that saves us anything, it probably makes a whole new moment object either way so whatever.)

File: src/helpers/tests/date.test.js [hunk 5]
```diff
@@ -0,0 +1,27 @@
...
+
+  it('should get relative dates', () => {
+    const date = new Date('2017-12-18');
+    Date.now = jest.fn(() => date);
+    expect(dateHelpers.getRelativeDates('hour')).toMatchSnapshot();
+    expect(dateHelpers.getRelativeDates('day')).toMatchSnapshot();
+    expect(dateHelpers.getRelativeDates('7days')).toMatchSnapshot();
+    expect(dateHelpers.getRelativeDates('30days')).toMatchSnapshot();
+    expect(dateHelpers.getRelativeDates('90days')).toMatchSnapshot();
+    expect(dateHelpers.getRelativeDates('invalid-range')).toMatchSnapshot();
```
jasonrhodes: Can you put each of these in its own `it()` so the snapshots have context about what the range _should_ be in their titles? You can use `jest-in-case` if you want, this feels like an okay place to use that lib that someone installed at some point:

```js
cases('relative date calculation', ({ range }) => {
    expect(dateHelpers.getRelativeDates(range)).toMatchSnapshot();
  }, {
    'for an hour ago': { range: 'hour' },
    'for a day ago': { range: 'day' },
    // etc
```

jasonrhodes: docs: https://github.com/Thinkmill/jest-in-case

src/helpers/tests/metrics.test.js
```diff
@@ -0,0 +1,104 @@
...
+
+    to.add(30, 'days');
+    expect(metricsHelpers.getPrecision(from, to)).toEqual('week');
+
+    to.add(200, 'days');
+    expect(metricsHelpers.getPrecision(from, to)).toEqual('month');
+
+  });
+
+  it('should return hours as precision', () => {
```
jasonrhodes: precision type

src/helpers/tests/metrics.test.js
```diff
@@ -0,0 +1,104 @@
...
+    expect(metricsHelpers.getPrecisionType('1min')).toEqual('hours');
+  });
+
+  it('should return days as precision', () => {
```
jasonrhodes: as precision type

src/helpers/tests/metrics.test.js
```diff
@@ -0,0 +1,104 @@
+import * as metricsHelpers from '../metrics';
```
jasonrhodes: I think in general we're still finding a good balance with snapshots and regular assertions. With helpers that do some logic, I think it's often a good idea to assert that the right computations were made etc, either in place of or in addition to the snapshot. Not a clear line and not always necessary, but if there's important logic in any of these, it might be nice to have a few assertions in the tests to make sure we don't accidentally lose something and `u` it away.

src/pages/reports/components/ReasonCell.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
```
jasonrhodes: Do you think it's worth calling this "LongTextContainer" and moving it to `src/components` since it's not really specific to "Reason"? Maybe premature until we need it somewhere else, but I like that it's just styles specific to text that is very long...

src/pages/reports/index.js
```diff
@@ -2,10 +2,12 @@ import BouncePage from './bounce/BouncePage.js';
 import SummaryPage from './summary/SummaryPage.js';
 import MessageEventsPage from './messageEvents/MessageEventsPage.js';
 import EventPage from './messageEvents/EventPage.js';
+import DelayPage from './delays/DelayPage';
```
jasonrhodes: Not sure why we did this index page this way, but would you mind making it look like what we do most everywhere else (see example in `src/components/index.js`)

e.g.
```js
export { default as DelayPage } from './delays/DelayPage';
```

src/pages/reports/delays/DelayPage.js
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  getRowData = (rowData) => {
+    const { totalAccepted } = this.props;
+    const { reason, domain, count_delayed, count_delayed_first } = rowData;
+    return [
+      <ReasonCell reason={reason} />,
+      <UnstyledLink onClick={() => this.handleDomainClick(domain)}>{domain}</UnstyledLink>,
+      count_delayed,
+      <span>{count_delayed_first}(<Percent value={(count_delayed_first / totalAccepted) * 100} />)</span>
```
jasonrhodes: Need a space here:
```jsx
<span>{count_delayed_first} (<Percent
```

jasonrhodes:
See #223 for some test suggestions and a `renderRowData` helper.

src/actions/reportFilters.js
```diff
@@ -8,25 +8,37 @@ import {
...
 import { list as listSendingDomains } from './sendingDomains';
+import { getQueryFromOptions } from 'src/helpers/metrics';
 
-// TODO: make this smarter to not always refresh templates, subaccounts, and
-// sending domains, since they aren't dependent on the params...
-// maybe load (all) vs refresh (4 metrics calls)
-export function refreshTypeaheadCache(params) {
+export function refreshTypeaheadCache(params, timeFiltered = false) {
   return (dispatch) => {
     dispatch({ type: 'REFRESH_TYPEAHEAD_CACHE' });
```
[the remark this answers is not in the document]
bizob2828: good catch

src/pages/reports/components/ReasonCell.js
bizob2828: sure i can move this, didn't really no the use case but makes sense

PR #220, opened 2017-12-18
Initial GA tracking (feature/FAD-5565)

jgzamora:
Tracks only page views. Based on #214


public/index.html
```diff
@@ -25,6 +25,16 @@
     -->
 
     <title>SparkPost</title>
+
+    <!-- Global site tag (gtag.js) - Google Analytics - Keep at top of scripts -->
+    <script async src="https://www.googletagmanager.com/gtag/js?id=UA-111136819-2"></script>
```
jasonrhodes: I think we'll probably need different GA IDs for the different environments. How do we use config out here? We might be able to dynamically add this script to the page so long as we get the one below it loaded up here in the `<head>`...

public/index.html
```diff
@@ -25,6 +25,16 @@
...
+    <script>
+     window.dataLayer = window.dataLayer || [];
+     function gtag(){dataLayer.push(arguments);}
+     gtag('js', new Date());
+     gtag('config', 'UA-111136819-2');
```
jasonrhodes: Hm this line would also have to happen dynamically. But I _think_ the only thing that has to happen before other JS loads is the `gtag` function definition itself...

Alternatively we can do all GA segmentation via hostname, but we would need to make sure we know how that works.

src/components/routeWatch/RouteWatch.js
```diff
@@ -0,0 +1,22 @@
+import { Component } from 'react';
```
jasonrhodes: Can probably add a few simple tests here if we export the class and mimic the few states that can happen, just have to figure out if it's possible to easily mock `window` stuff

PR #215, opened 2017-12-12
FAD-5588: Support panel

ewandennis:
## What's In Here
 - Support panel button and popover on page bottom right
 - Support ticket form -> submission -> success/failure flow

## How To Test
 - Use the `appteam` account or another with a non-free plan and standard service level
 - Submit a ticket using the form
 - You can force a submission error by removing a field from the `data` field of `createTicket()` in `actions/support.js`

## Missing Bits/Questions/Issues
 - [x] File attachment support
 - [x] Validator test coverage
 - [x] Feature test coverage
 - [x] Inline form errors - the current `Error` is a block that causes the elements to move while in use

PR #222, opened 2017-12-20
FAD-5595: Basic filtering of message events

rajumsys:
Should be straightforward, no special things needed to start testing. 

For recipients field, i tried to make it smarter and catch common typing issues like comma, whitespaces etc. so values like `email@domain.com , email2@domain.com` should be fine. let me know if any other things should be caught. values with multiple commas won't work but i thought it's not normal :)


rajumsys:
@bkemper i've updated with pretty decent test coverage.

src/pages/reports/messageEvents/MessageEventsPage.js
```diff
@@ -55,32 +62,49 @@ export class MessageEventsPage extends Component {
...
       ? <Empty message={emptyMesasage} />
       : (
-        <TableCollection
-          columns={columns}
-          rows={events}
-          getRowData={this.getRowData}
-          pagination={true}
-        />
+        <div>
+          { events.length === maxResults &&
```
ewandennis: If we encounter an off-by-one issue in future, this test is a little brittle. Maybe `>=` ?

src/pages/reports/messageEvents/components/BasicFilter.js
```diff
@@ -0,0 +1,112 @@
...
+        relativeRange: relativeRange
+      }
+    }, this.refresh);
+  }
+
+  emailValidator = (value) => {
+    const invalids = this.getInvalidAddresses(this.parseAddresses(value));
+
+    if (invalids.length) {
+      return `${invalids.join(', ')} ${invalids.length > 1 ? 'are' : 'is'} not valid email ${invalids.length > 1 ? 'adresses' : 'address'}`;
```
ewandennis: `s/adresses/addresses/`

If we're going down the path of singular/plural detection, lets make sure we say "not **a** valid email address" in the singular case. I wonder if we could just say `Invalid: ...` though.

src/pages/reports/messageEvents/components/BasicFilter.js
```diff
@@ -0,0 +1,112 @@
...
+  render() {
+    return (
+      <Grid>
+        <Grid.Column xs={12} md={6}>
+          <DateFilter refresh={this.handleDateSelection} />
+        </Grid.Column>
+        <Grid.Column xs={12} md={6}>
+          <Field
+            name="recipients"
+            onBlur={this.handleRecipientsChange}
```
ewandennis: Can we avoid refreshing when we move focus away from the filter field when the field contents have not changed?

I also instinctively hit return after typing into search fields. Can we make that refresh the list?

src/pages/reports/messageEvents/components/BasicFilter.js
```diff
@@ -0,0 +1,112 @@
...
+            validate={this.emailValidator}
+            placeholder='recipient1@domain.com, recipient2@domain.com'
+          />
+        </Grid.Column>
+      </Grid>
+    );
+  }
+}
+
+const formName = 'basicFilter';
```
ewandennis: My understanding is the redux-form names are global. Maybe we should namespace this one a bit more. Something like `msgEventsBasicFilter` ?

src/pages/reports/messageEvents/components/tests/BasicFilter.test.js
```diff
@@ -0,0 +1,133 @@
...
+  describe('getInvalidAddresses', () => {
+    it('detect invalid email addresses correctly', () => {
+      expect(instance.getInvalidAddresses(['email@domain.com', 'email2m'])).toEqual(['email2m']);
+      expect(instance.getInvalidAddresses(['abc'])).toEqual(['abc']);
+      expect(instance.getInvalidAddresses([''])).toEqual([]);
+      expect(instance.getInvalidAddresses(['email@domain.com'])).toEqual([]);
+    });
+  });
+
+  describe('emailValidator', () => {
```
ewandennis: For reference, I have a set of validation helper tests in my support panel PR [here](https://github.com/SparkPost/2web2ui/pull/215/files#diff-9f4eb7044f3b220cde32ebaca62d16cb).

src/pages/reports/messageEvents/components/tests/BasicFilter.test.js
```diff
@@ -0,0 +1,133 @@
...
+      expect(props.onSubmit).toHaveBeenCalledTimes(0);
+    });
+  });
+
+  describe('handleDateSelection', () => {
+    const now = new Date();
+
+    beforeEach(() => {
+      dateMock.getRelativeDates = jest.fn(() => {
+        return {
```
ewandennis: Please `npm run lint`- you can remove the `return { }`: `jest.fn(() => ({ to: ..., from: ... }))`

src/pages/reports/messageEvents/components/BasicFilter.js
```diff
@@ -0,0 +1,112 @@
...
+      reportFilters,
+      recipients: ''
+    };
+  }
+
+  refresh = () => {
+    this.props.onSubmit(this.state);
+  }
+
+  parseAddresses = (value) => {
```
ewandennis: Nice human-safe parsing

src/pages/reports/messageEvents/components/BasicFilter.js
```diff
@@ -0,0 +1,112 @@
...
+          <DateFilter refresh={this.handleDateSelection} />
+        </Grid.Column>
+        <Grid.Column xs={12} md={6}>
+          <Field
+            name="recipients"
+            onBlur={this.handleRecipientsChange}
+            component={TextFieldWrapper}
+            title="Recipient Email(s)"
+            validate={this.emailValidator}
+            placeholder='recipient1@domain.com, recipient2@domain.com'
```
ewandennis: Can we use an active voice call to action in the placeholder text? Something like `Filter by recipient email address`

src/pages/reports/messageEvents/MessageEventsPage.js
rajumsys: that would be breaking change for our API ;). i really wanted it to break in such case as
- api will return more than 1k and we're not limiting in frontend. so it'll say we're showing 1000 but actually showing more
- it's a api change.

src/pages/reports/messageEvents/components/BasicFilter.js
ewandennis: Note sure how we make it clear we can enter lists of addresses there...

src/pages/reports/messageEvents/MessageEventsPage.js
rajumsys: alternative is to limit UI from showing max 1K and then changing to `>=` would make sense.

src/pages/reports/messageEvents/components/BasicFilter.js
rajumsys: yeah, `not a ` sounds better.

src/pages/reports/messageEvents/components/BasicFilter.js
rajumsys: Both are good points, I'll address that.

rajumsys: Seems MB's TextField doesn't support a onEnter (or even onKeyPress) event that'll let me detect when Return key is pressed. I guess, we'll have to implement that in MB and then update that in all filters in UI.

src/pages/reports/messageEvents/components/BasicFilter.js
rajumsys: yeah, that's the reason I put the example and made them like `recipientN` so people know what email is this. i'm happy to change to something that means both things and format (comma separated)

src/pages/reports/messageEvents/components/BasicFilter.js
rajumsys: Actually nvm, seems redux `Field` has support for `onKeyDown`. It'll work for this case.

src/pages/reports/messageEvents/MessageEventsPage.js
ewandennis: I meant we should change the test on the client side to make it resilient to something upstream giving us an array with an off-by-one number of results in it.

bkemper:
@ewandennis It appears all your feedback addressed.  I took a quick look at the code, but didn't want to duplicate your work.  I focused more on smoke testing it before ask for the thumbs up to merge.  I notice this bug in the custom time range.  I'll have to dig in tomorrow.

message-event-filter

ewandennis:
@bkemper Sounds good. A few quick points before we merge:

 - My point about the edge case in `src/pages/reports/messageEvents/MessageEventsPage.js` was not addressed. I don't think I was being clear enough. Could you please skim that note above and address?  
 - If the time range issue is app-wide, we can address separately
 - I'll do a last walk through then sign off after ☝️

bkemper:
I wasn't able to reproduce that bug on UAT.  I'm triaging now to see what could be the cause.

src/pages/reports/messageEvents/MessageEventsPage.js
bkemper: Addressed... cf698e66

src/pages/reports/messageEvents/components/BasicFilter.js
bkemper: Addressed... 57474793

bkemper:
@ewandennis I confirmed that the bug was due to user error.  I was entering a "to" time greater than the present time.  The filter only allows a time range in the past.  Here is the condition.

`if (to.isValid() && from.isValid() && from.isBefore(to) && to.isBefore(now))`

I submitted a follow-up ticket, [FAD-6003](https://jira.int.messagesystems.com/browse/FAD-6003), to improve this experience to avoid the same confusion I had.

With all that said, this feature is ready for deployment pending your final review.

ewandennis:
:shipit:

PR #223, opened 2017-12-20
Report test ideas

jasonrhodes:
It was easier to show some things I was thinking about rather than filling up PR comments. This is against your PR so you can copy/paste over or cherry pick commits, or ignore it all ... whatever you like. :D

The `renderRowData` stuff is especially useful for testing that function all over the place, so I threw that together in a more official way than anything else here.

### DIFF
--- a/src/__testHelpers__/renderRowData.js
+++ b/src/__testHelpers__/renderRowData.js
@@ -1,0 +1,12 @@
+import React from 'react';
+import { mount } from 'enzyme';
+
+/**
+ * Takes a list of items (usually produced by a getRowData function)
+ * and mounts any react elements inside using enzyme's mount method
+ *
+ * @param {Array} row - list of items that may or may not be React elements
+ */
+export default function renderRowData(row) {
+  return row.map((item) => React.isValidElement(item) ? mount(item) : item);
+}
--- a/src/actions/tests/bounceReport.test.js
+++ b/src/actions/tests/bounceReport.test.js
@@ -1,12 +1,14 @@
 import * as bounceReport from '../bounceReport';
 import * as metricsActions from 'src/actions/metrics';
 import * as metricsHelpers from 'src/helpers/metrics';
+import * as reportFilters from 'src/actions/reportFilters';
 
 import moment from 'moment';
 
 jest.mock('../helpers/sparkpostApiRequest', () => jest.fn((a) => a));
 jest.mock('src/helpers/bounce');
 jest.mock('src/helpers/metrics');
+jest.mock('src/actions/metrics');
 jest.mock('src/actions/reportFilters');
 
 describe('Action Creator: Bounce Report', () => {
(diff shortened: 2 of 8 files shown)


src/__testHelpers__/renderRowData.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
```
jasonrhodes: This function should be useful for testing all of our TableCollection instances. We don't have to worry about whether an item in that array is a React element or not, we can just "render" the row as an array of things that can all be snapshotted.

src/actions/tests/bounceReport.test.js
```diff
@@ -1,54 +1,79 @@
...
 import * as metricsActions from 'src/actions/metrics';
 import * as metricsHelpers from 'src/helpers/metrics';
+import * as reportFilters from 'src/actions/reportFilters';
 
 import moment from 'moment';
 
 jest.mock('../helpers/sparkpostApiRequest', () => jest.fn((a) => a));
 jest.mock('src/helpers/bounce');
 jest.mock('src/helpers/metrics');
+jest.mock('src/actions/metrics');
```
jasonrhodes: Forgot to mock this in a few places (before your ticket).

src/actions/tests/bounceReport.test.js
```diff
@@ -1,54 +1,79 @@
 import * as bounceReport from '../bounceReport';
 import * as metricsActions from 'src/actions/metrics';
 import * as metricsHelpers from 'src/helpers/metrics';
+import * as reportFilters from 'src/actions/reportFilters';
```
jasonrhodes: This is how we get a reference to the auto-mocks that jest creates, which are also spies, so we can assert on calls.

src/actions/tests/bounceReport.test.js
```diff
@@ -1,54 +1,79 @@
...
     const metrics = 'count_bounce';
-    fetchSpy = jest.spyOn(metricsActions, 'fetchDeliverability');
     metricsActions.fetchDeliverability = jest.fn(() => [{ count_bounce: 1 }]);
     metricsHelpers.getQueryFromOptions.mockImplementation(() => ({ from, metrics }));
 
+    stateMock = {};
 
     dispatchMock = jest.fn((a) => Promise.resolve(a));
-    getStateMock = jest.fn((a) => Promise.resolve(a));
+    getStateMock = jest.fn(() => stateMock);
```
jasonrhodes: `getState` just returns a state object, not a Promise

src/actions/tests/bounceReport.test.js
```diff
@@ -1,54 +1,79 @@
...
   });
 
   afterEach(() => {
-    fetchSpy.mockRestore();
+    jest.resetAllMocks();
```
jasonrhodes: Wish we could put this in a global `afterEach` tbh, but this is easier than restoring each one individually and it will prevent accidental mock leakage.

src/actions/tests/bounceReport.test.js
```diff
@@ -1,54 +1,79 @@
...
 
   it('should dispatch a chart refresh action', async() => {
     const thunk = bounceReport.refreshBounceChartMetrics();
     await thunk(dispatchMock, getStateMock);
+    expect(reportFilters.maybeRefreshTypeaheadCache).toHaveBeenCalledTimes(1);
+    expect(reportFilters.refreshReportRange).toHaveBeenCalledTimes(1);
+    expect(metricsActions.fetchBounceClassifications).toHaveBeenCalledTimes(1);
+    expect(metricsActions.fetchBounceClassifications.mock.calls).toMatchSnapshot();
+
     expect(dispatchMock.mock.calls).toMatchSnapshot();
```
jasonrhodes: Still have to do this snapshot because calls to functions defined within the same file can't be mocked or spied on for some really aggravating reason, so this captures the final call to `refreshBounceReport` and the action object it returns.

jasonrhodes:
Was just demonstrating some thoughts, closing now

PR #224, opened 2017-12-20
Redirect on register (Hotfix/fad 5949)

jgzamora:


src/pages/register/tests/RegisterPage.test.js
```diff
@@ -1,11 +1,19 @@
...
-// actions
+// mocks
 const checkInviteToken = jest.fn();
 const registerUser = jest.fn(() => Promise.resolve());
+const logout = jest.fn();
+const authenticate = jest.fn(() => Promise.resolve());
+const historyPush = jest.fn();
+const report = jest.fn();
+
+ErrorTracker.default = { report };
```
jgzamora: Found [a bunch of ways to mock imports](https://stackoverflow.com/questions/40465047/how-can-i-mock-an-es6-module-import-using-jest), but none of them are all that clean. This works, but let me know if you know a better way.

src/pages/register/RegisterPage.js
```diff
@@ -11,14 +11,30 @@ import RegisterUserForm from './RegisterUserForm';
...
+    return this.props.registerUser(this.props.token, values)
+      .then(() => {
+        this.props.logout(); // log out of current account
+        this.props.authenticate(username, password)
+          .then(() => {
+            this.props.history.push('/dashboard');
+          })
+          .catch((error) => {
+            // user was created but auth failed, redirect to /auth
+            this.props.history.push('/auth');
```
jgzamora: Redirect to `/auth` here so that it feels natural (inconvenient) instead of just error-ing out

src/pages/register/tests/RegisterPage.test.js
jasonrhodes: Yeahhhhh default exports make `jest.mock` get all borked. This is probably the best way even though it sucks. We should probably write some docs on test mocking with jest for the various problems/solutions.

src/pages/register/RegisterPage.js
```diff
@@ -11,14 +11,30 @@ import RegisterUserForm from './RegisterUserForm';
...
-  onSubmit = (values) => this.props.registerUser(this.props.token, values)
-    .then(() => <Redirect to='/dashboard' />)
-    .catch((error) => {
-      ErrorTracker.report('register-user', error);
-    });
+  onSubmit = (values) => {
+    const { username, password } = values;
+    return this.props.registerUser(this.props.token, values)
+      .then(() => {
+        this.props.logout(); // log out of current account
```
ewandennis: After registering while not already logged in, I was redirected to an empty register page. I was logged in though and able to visit the dashboard. [This sentry error](https://sentry.io/sparkpost/2web2ui-dev/issues/421081174/) suggests an action did not return a promise as expected.

We should at least return the promise in the [authenticate() action here](https://github.com/SparkPost/2web2ui/blob/fc8d5dca077ccec3312b32f904b1fac08653b1a3/src/actions/auth.js#L35) at least and possibly also sequencing the log

Some test coverage on the complex actions in `actions/auth` would help here to verify their behaviour.

jgzamora:
Confirmed old app works without `#`

jgzamora:
This is ready for a final pass @ewandennis. 

The chosen fix ended up being a small change to the `logout` auth action that returns early if not logged in. Also, the register page now logs you out on `componentDidMount` and not during form submit. I also aded an extra variable into loading (`auth.loginPending`) so the auth and redirect transition is smoother. 

Note on tests: I did not write tests for the auth actions because @bizob2828 wrote them in his branch. But I did make small edits to both `authenticate` and `logout` which should be reflected in the tests once both branches are merged. 
* authenticate now returns `sparkPostLogin`
* logout now has an early return if `loggedIn`

PR #225, opened 2017-12-20
FAD-5979: Fix profile updating

rajumsys:
This PR fixes the profile update issues:
- Pristine issue. After saving the name, you can't revert to old one (form things old as pristine). This is because name is not updated in the store (and hence, not re-rendering the component with updated value)
- New name is not reflected before refresh
- Name is updated in the store. That solves both of the above problems. 

For autocomplete attr, check [this doc](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fe-autocomplete-name).


src/pages/profile/ProfilePage.js
```diff
@@ -13,11 +14,13 @@ import { LabelledValue } from 'src/components';
...
+    const { showAlert } = this.props;
     const { username } = this.props.currentUser;
-    return this.props.updateUser(username, { first_name: values.firstName, last_name: values.lastName })
-      .catch(() => {
-        showAlert({ type: 'error', message: 'Unable to update profile' });
-      });
+    const data = { first_name: values.firstName, last_name: values.lastName };
+
+    return this.props.updateUser(username, data)
+      .then(() => this.props.updateCurrentUser(data))
```
jasonrhodes: OK so let me see if I understand what's happening here:
1. We already had an `updateUser` action that performs the real update
1. We added `updateCurrentUser` to push that updated data into the `currentUser` key in the redux store

If that's true ^, this isn't the "redux way". Instead, if possible, we should make the currentUser reducer listen for the `UPDATE_USER_SUCCESS` action, inspect to see if it's the current user, and if so, update the state.

src/pages/profile/components/NameForm.js
```diff
@@ -16,11 +16,11 @@ export class NameForm extends Component {
           // for redux-form
           name='firstName'
           component={TextFieldWrapper}
-
           // for the matchbox component
           id='firstName'
           label='First Name'
           validate={required}
+          autoComplete='given-name'
```
jasonrhodes: :clap: :clap: love this, we should add all of these wherever possible, adding these values to the Angular UI helped with a number of things

src/reducers/currentUser.js
```diff
@@ -10,6 +10,9 @@ export default (state = initialState, action) => {
     case 'GET_GRANTS_SUCCESS':
       return { ...state, grants: action.payload };
 
+    case 'UPDATE_CURRENT_USER':
```
jasonrhodes: This would listen for `UPDATE_USER_SUCCESS` and inspect username to see if it's the current user, and if so, update with new data.

_Note: alternatively if we are worried that the API returns data differently than how we submit it for update, we can just call the "get current user" to refetch it after updating._

src/pages/profile/ProfilePage.js
jasonrhodes: _Note: alternatively if we are worried that the API returns data differently than how we submit it for update, we can just call the "get current user" to refetch it after updating._

rajumsys: those steps are correct. 

why using different actions isn't a redux way while making unnecessary API call will be (yes that api call doesn't return updated data)?

src/reducers/currentUser.js
rajumsys: we won't have such context to know if this is current user as payload won't have those info. i guess only way will be to use `meta` but i'm not convinced yet why we should do it!

src/pages/profile/ProfilePage.js
jasonrhodes: So part of this will involve a little trust in some hand wavyness, but I've been working with Redux a lot and my understanding (and how I'd like us to try to do things) is that a Redux action should always try to describe "what happened" rather than telling the app "please do this". So even our "SHOW_ALERT" actions don't really follow this and there are places it's hard, but we are trying to stick to this.

So for this example, "what happened" is "a user was updated". We should pass the username of the updated user through `meta`, then any reducer that wants to listen for "a user has been successfully updated" can respond accordingly. 

Also, which API call and what data does it not return, out of curiosity?

jasonrhodes: PS the "second API call" strategy isn't so much a "Redux way" thing as it is a general concern of mine. Having two (or more) reducers listening to the same action and updating state accordingly is great from a Redux perspective, but sometimes I get concerned about updating our state manually because the API/backend could be doing data transformations or computations that we'd have to try to simulate, whereas "re-fetching" gives you the pure new response. Case-by-case basis but that's the context there.

rajumsys: so yeah i don't think i still understand the first argument and still interested to understand. but i just replaced with get current user.

src/pages/profile/ProfilePage.js
```diff
@@ -13,11 +14,13 @@ import { LabelledValue } from 'src/components';
...
+    const { showAlert } = this.props;
     const { username } = this.props.currentUser;
-    return this.props.updateUser(username, { first_name: values.firstName, last_name: values.lastName })
-      .catch(() => {
-        showAlert({ type: 'error', message: 'Unable to update profile' });
-      });
+    const data = { first_name: values.firstName, last_name: values.lastName };
+
+    return this.props.updateUser(username, data)
+      .then(this.props.getCurrentUser)
```
rajumsys: i considered putting two `catch` block here (one for update and one for get) but seemed overkill. let me know if you think two `catch`es make sense.

src/pages/profile/ProfilePage.js
jasonrhodes: The first argument is philosophical--actions should describe something that happened and reducers should listen for things that have happened, actions shouldn't tell reducers or the UI to do things directly. 

Good Action: `USER_UPDATED`
Bad Action: `CLEAR_USER_FORM`
(A reducer may listen to the `USER_UPDATE` action and clear values that a form is referencing)

Good Action: `FILTERS_CHANGED`
Bad Action: `REFRESH_CACHE`
(A reducer can listen for when the filters change and update state accordingly)

We are already doing a pretty bad job at this because it's very easy to slip into making actions be imperative directions, but I'm trying to be better at it as we go.

src/pages/profile/ProfilePage.js
jasonrhodes: Might not be a bad idea to catch the `updateUser` action and do the `showAlert`, and maybe just swallow any `getCurrentUser` errors? The update succeeded in that case so it seems bad to say the update failed, not sure we really have to worry about that case that much and can just swallow/ignore...

rajumsys: the messaging for the 2nd catch will be different like unable to reload user or such. it's because, unless we communicate that if you comes back to this page (without refresh), he'll still see old value thinking the update is failed.

jasonrhodes: Yeah I understand, not sure the messaging will be helpful there "We updated your profile but failed to update it in the client, please refresh" just sounds bizarre/weird. This scenario is so unlikely that I'm not worried if the UI looks out of sync, I think, and just ignore it.

jasonrhodes: Or just put a `window.location.reload()` in the 2nd catch :P ;)

rajumsys: lol. ok. i think list we can do move the catch to update. so at least get user error will surface to console.

PR #226, opened 2017-12-20
FAD-5461 Sending Domains List

jonambas:
**helpers/domains.js** - New Helpers
- `resolveReadyFor`
- `resolveStatus`

---

**DomainStatusTag** - New Component
- Returns a correctly styled tag for domain status with tooltips
- [x] TODO: Use in tracking domains
```js
// Props
status: PropTypes.oneOf(['unverified', 'blocked', 'pending', 'verified'])
```

---

**SubaccountTag** - New Component
- Goal of this was to create a common pattern that shows subaccount assignment across our app.
- Returns a correctly styled tag for subaccount assignment (shared, master, single, default)
- [x] TODO: Use in tracking domains
- TODO: Use anywhere we show subaccount assignment. Need another ticket.

UX Note: We use tags in report filters as well, but are formatted differently. `My Subaccount (ID 101)` vs `<Icon> Subaccount 101` in this component and tracking domains page. Should we make these looks the same? Or even try to use the same component?

```js
// Props
id: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
all: PropTypes.bool,
master: PropTypes.bool,
isDefault: PropTypes.bool
```

---

**UnverifiedWarningBanner** - New Component
- Component for that 2 week warning & "clear web presence" requirement
- Used under `hasUnverifiedDomains` selector

---

**ReadyFor** - New Component
- Returns a group of tags for "Ready For" status
- Can use on edit page as well

```js
// Props:
bounce: PropTypes.bool,
sending: PropTypes.bool,
dkim: PropTypes.bool,
bounceDefault: PropTypes.bool
```

---

**MB Fixes - 1.8.14**
- Icon vertical spacing
- Tag sizing, spacing, refactored colors


src/components/tags/StatusTag.js
```diff
@@ -0,0 +1,39 @@
...
+
+  if (status === 'unverified') {
+    content = ' Unverified';
+    yellow = true;
+    icon = <Icon size={15} name='ErrorOutline'/>;
+  }
+
+  if (status === 'blocked') {
+    content = ' Blocked';
+    yellow = true;
```
jonambas: todo for matchbox: make a red tag

rajumsys:
I'll review this tomorrow/next week.

src/pages/sendingDomains/ListPage.js
```diff
@@ -0,0 +1,137 @@
...
+
+    const primaryAction = {
+      content: 'Add a Domain',
+      Component: Link,
+      to: '/account/sending-domains/create'
+    };
+
+    return (
+      <Page
+        title='Domains'
```
rajumsys: 'Sending Domains'?

rajumsys:
- Should we make the UI same as tracking domain? These two sections are inconsistent now. I think if we make it like: All information on left side (that includes verification status, shared with subaccount or not) and actions on right side. 
Status and specially Subaccount column looks so empty  and also overall layout looks little cluttered

src/components/tags/SubaccountTag.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
```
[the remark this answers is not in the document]
rajumsys: Nevermind, I see you're already re-using `Tag` component

src/components/tags/StatusTag.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
```
rajumsys: I see this component is top level component (rather than in sending domain specific). Can this be readily used for other things (like subaccount, tracking domain)?

Tracking domain page also has `StatusTag` component. We should probably consolidate these two into one and make it generic. If it can't be done, we should move this component inside `/pages/sending-domains/components`.

src/pages/sendingDomains/components/UnverifiedWarningBanner.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
```
rajumsys: We could just use `ApiErrorBanner` (rename and refactor to become generic), no? I am not sure whether it's norm in react world to make individual component for each case but I guess, generally speaking, we should re-use any implementation as much as possible.

src/pages/sendingDomains/tests/ListPage.test.js
```diff
@@ -0,0 +1,77 @@
+import { shallow } from 'enzyme';
+import React from 'react';
+
+import { ListPage } from '../ListPage';
+
+describe('Ready For Component', () => {
```
rajumsys: ?

src/components/tags/SubaccountTag.js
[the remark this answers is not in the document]
jonambas: Yeah, I wanted to make tags specific to subaccounts and domain status so we don't have to remember what color & icon to use every time.

src/components/tags/StatusTag.js
jonambas: Yeah we should consolidate with Tracking domains at least. Need to look into other types of statuses to see if this will work. I can rename to `DomainStatusTag` for now maybe

src/pages/sendingDomains/tests/ListPage.test.js
jonambas: :facepalm:

src/pages/sendingDomains/components/UnverifiedWarningBanner.js
jonambas: It would end up being a wrapper around `Banner` tho. This is just a small presentational component, I pulled it out because the list page was getting long. And we can also re-use it on the edit domain page if needed

jonambas:
**RE: tracking domains UI**
@jasonrhodes and I decided that we can go with what's built and later revisit our table vs. collection UI as it would probably take a bit of time to get consistent across different lists

src/components/tags/StatusTag.js
jasonrhodes: When you're in there, can you switch these to a 'color' prop maybe? Worried about the boolean props that can't have multiple true at the same time...

jonambas:
Diff is pretty bloated - fixed a number of issues in matchbox in this PR.

ewandennis:
Comments on the experience first:

 - Tracking domains and sending domains look a little closer together. Definitely an area we could improve on. **Later**. :)
 - The unverified label tooltip reads like the domain is somehow blocked from use:
    screen shot 2018-01-04 at 12 02 10
    - How about something like `This domain must be verified before use.` ?
 - I set up `ewandennis.com` as a blocked domain on `tst`. The nasty red jumps out nicely.

Looking good. On to the code...

src/components/tags/DomainStatusTag.js
```diff
@@ -0,0 +1,49 @@
...
+  const size = 15;
+
+  if (status === 'verified') {
+    return <Tag>Verified</Tag>;
+  }
+
+  if (status === 'unverified') {
+    return (
+      <Tooltip
+        content='This domain is not available for use. For more information, please contact support.'
```
ewandennis: `This domain must be verified before use` - or similar. The current wording makes `unverified` sound like `blocked`.

src/components/tags/SubaccountTag.js
```diff
@@ -0,0 +1,28 @@
...
+import PropTypes from 'prop-types';
+import { Tag, Icon } from '@sparkpost/matchbox';
+
+const SubaccountTag = ({ id, all, master, isDefault }) => {
+  const defaultContent = isDefault ? ' (Default)' : null;
```
ewandennis: Stylistic note and at your discretion: the combination of conditionals and if statements makes this logic a little tough to follow. A group of `let` definitions with initial values, then if statements for special cases might easier on the eye:

```
let color = null;
let content = `Subaccount ${id}`;
let defaultContent = null;

if (isDefault) {
  defaultContent = ' (Default');
  color = 'orange';
}

if (all) { ... }
if (master) { ... }
// ...
```

src/config/routes.js
```diff
@@ -286,6 +287,18 @@ const routes = [
   },
   {
     path: '/account/sending-domains',
+    component: sendingDomains.ListPage,
+    condition: hasGrants('sending_domains/manage'),
+    layout: App
+  },
+  {
+    path: '/account/sending-domains/create',
```
ewandennis: Sadly, the current app uses `/account/sending-domains/add` - can we cover that base too?

src/config/routes.js
```diff
@@ -286,6 +287,18 @@ const routes = [
...
+    component: ComingSoonPage,
+    condition: hasGrants('sending_domains/manage'),
+    layout: App
+  },
+  {
+    path: '/account/sending-domains/:id/edit',
```
ewandennis: The current app uses `/account/sending-domains/edit?domain=:id` :(

src/helpers/tests/domains.test.js
```diff
@@ -0,0 +1,42 @@
+import * as domains from '../domains';
+
+describe('resolveReadyFor', () => {
+  it('should return correct values', () => {
```
ewandennis: Do we need a test case for each variable `bounce` depends upon?

src/pages/sendingDomains/components/ReadyFor.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Tag } from '@sparkpost/matchbox';
+
+import styles from './ReadyFor.module.scss';
+
+const ReadyFor = ({ bounce, dkim, sending, bounceDefault }) => {
```
ewandennis: I'm sad that we don't use green for happy little tags like this one :( I see `Tag` doesn't actually support green so I imagine there's a Reason ™️ ...

src/pages/sendingDomains/components/UnverifiedWarningBanner.js
```diff
@@ -0,0 +1,21 @@
...
+    to: 'https://www.sparkpost.com/docs/getting-started/requirements-for-sending-domains/',
+    external: true
+  };
+
+  return (
+    <Banner
+      status='warning'
+      title='Unverified sending domains will be removed two weeks after creation.'
+      action={action} >
+      <p>All domains will also need a clear web presence to pass verification review.</p>
```
ewandennis: `...to pass verification` ? `verification review` sounds off.

src/pages/sendingDomains/components/tests/__snapshots__/ReadyFor.test.js.snap
```diff
@@ -0,0 +1,41 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Component: Ready For should render bounce as default ready 1`] = `
```
ewandennis: snapshots look 🆗  based on tests

src/pages/sendingDomains/tests/__snapshots__/ListPage.test.js.snap
```diff
@@ -0,0 +1,221 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
ewandennis: 🆗

src/pages/trackingDomains/components/TrackingDomainRow.js
```diff
@@ -110,14 +107,14 @@ export class TrackingDomainRow extends Component {
...
     const { domain, subaccountId, status, isDefault } = this.props;
     return (
-      <Panel.Section className={styles.SpacedSection}>
+      <Panel.Section>
         <Grid>
           <Grid.Column xs={12} md={9}>
-            <h3 className={styles.DomainHeading}>{domain}</h3>
+            <span className={styles.DomainHeading}>{domain}</span>
             <div className={styles.TagRow}>
-              <StatusTag status={status} />
```
ewandennis: Its difficult to validate the changes in logic here against the snapshot changes. I suggest taking a look at each changed snapshot, or the tracking domains page itself and verifying each case is intact.

src/config/routes.js
jonambas: Hmm not sure what to do here. Recreate with url param? a redirect maybe?

src/pages/sendingDomains/components/ReadyFor.js
jonambas: Unfortunately our green doesn't work really well with text on top of it

src/pages/sendingDomains/tests/ListPage.test.js
```diff
@@ -0,0 +1,83 @@
...
+      hasUnverifiedDomains: false,
+      loading: false,
+      listDomains: jest.fn()
+    };
+
+    wrapper = shallow(<ListPage {...props}/>);
+  });
+
+  it('renders correctly', () => {
+    expect(wrapper).toMatchSnapshot();
```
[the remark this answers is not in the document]
jonambas: Yep there is a separate test below this that checks `hasUnverifiedDomains` separately

src/pages/trackingDomains/components/TrackingDomainRow.js
jonambas: I spent an hour yesterday just looking through the snapshot diff, I _think_ everything looks okay. Maybe @jasonrhodes, want to take a peek?

jonambas:
```js
// Status tag is hidden if 'verified'
{status !== 'verified' && <DomainStatusTag className={styles.Tag} status={status} />}

// Default tag only if default to master account
{isDefault && !subaccountId && <IsDefaultTag/>}

// Subaccount tag, default only if not on master account
{subaccountId && <SubaccountTag className={styles.Tag} id={subaccountId} isDefault={!!subaccountId && isDefault}/>}
```

PR #227, opened 2017-12-21
made jons concept more abstract

bizob2828:
I wanted to put this in a PR, not finished but close.  Since I'm out until new year, if someone feels inspired they could button it up and use it for the reports that are missing it.  Shows jon's concept in a more re-usable state.  Still messing with tweaking the natural language around time diff


bizob2828:
I'm creating a new branch and just putting the changes. Something got f'ed here

PR #228, opened 2018-01-02
Hotfix/fad 5976

bizob2828:
I think this is in a good state.  I just have to write necessary tests.  There is weirdness where the required form validator fires before doing anything, I think it may be last pass f'ing with my page.


src/actions/tfa.js
```diff
@@ -0,0 +1,54 @@
...
+      method: 'POST',
+      url: `/users/${authData.username}/two-factor`,
+      headers: {
+        Authorization: authData.token
+      },
+      data: {
+        code
+      }})
+      .then(() => {
+        dispatch(login({
```
bizob2828: wasn't sure how circular we should get with these actions so auth/tfa call each other in certain cases

src/pages/auth/components/TfaForm.js
```diff
@@ -0,0 +1,53 @@
...
+import { connect } from 'react-redux';
+import { Field, reduxForm } from 'redux-form';
+import { required } from 'src/helpers/validation';
+import { TextFieldWrapper } from 'src/components';
+
+import { Button } from '@sparkpost/matchbox';
+
+export class TfaForm extends Component {
+
+  render() {
```
bizob2828: for context i tried to grok redux form and it's working but not sure if this is the right way to do it

src/actions/auth.js
```diff
@@ -1,26 +1,49 @@
...
   }
 
   return (dispatch) => {
-    dispatch({
-      type: 'LOGIN_SUCCESS',
-      payload: authData
-    });
-
+    dispatch(setLoginCredentials(authData));
+    dispatch({ type: 'LOGIN_SUCCESS' });
```
jasonrhodes: I _think_ we could let LOGIN_SUCCESS still send authData as the payload, and just call the one action here. Still have the setLoginCredentials action available to _only_ set auth data when that's necessary.

src/actions/auth.js
```diff
@@ -32,12 +55,26 @@ export function authenticate(username, password, rememberMe = false) {
...
+
     sparkpostLogin(username, password, rememberMe)
       .then(({ data = {}} = {}) => {
-        const payload = { ...data, username };
+        authData = { ...data, username };
 
-        // dispatch login success event
-        dispatch(login(payload, true));
+        dispatch(setLoginCredentials(authData));
+        return tfaGet({ username, token: authData.access_token });
```
jasonrhodes: can we do:
```js
  return Promise.all([authData, tfaGet()]);
})
.then(([authData, { data = {}} = {}]) => {
```

Then we don't have to reach out of scope and mutate a variable

jasonrhodes: Or we can chain the `.then()` to the end of `tfaGet()` so that `authData` stays in scope. Either way.

src/actions/auth.js
```diff
@@ -32,12 +55,26 @@ export function authenticate(username, password, rememberMe = false) {
...
+    let authData;
+
     sparkpostLogin(username, password, rememberMe)
       .then(({ data = {}} = {}) => {
-        const payload = { ...data, username };
+        authData = { ...data, username };
 
-        // dispatch login success event
-        dispatch(login(payload, true));
+        dispatch(setLoginCredentials(authData));
```
jasonrhodes: Ok if we let LOGIN_SUCCESS handle storing authData, the TFA_ENABLED action can also send authData as its payload and store it in the TFA reducer so we can fetch it in the TFA view, and then pass it to LOGIN_SUCCESS ... that way it doesn't get stored in the redux state in the regular place where we usually store it until the user is successfully logged in

src/actions/tfa.js
```diff
@@ -0,0 +1,54 @@
...
+    url: `/users/${username}/two-factor`,
+    headers: {
+      Authorization: token
+    }
+  });
+}
+
+export function verifyAndLogin(code) {
+  return (dispatch, getState) => {
+    const authData = getState().auth;
```
jasonrhodes: So here instead of getting it there, we'd pull it from the TFA block in state...

src/actions/tfa.js
jasonrhodes: That's fine. The same functions don't call each other so I think it's ok.

src/pages/auth/components/TfaForm.js
jasonrhodes: I'd like to get @jonambas and/or @jgzamora to take a look because I'm honestly not super strong on redux-form skillz, looks ok to me at first glance though.

src/pages/auth/AuthPage.js
```diff
@@ -55,8 +61,14 @@ export class AuthPage extends Component {
     this.state.ssoEnabled ? this.ssoSignIn(username) : this.regularSignIn(username, password, rememberMe);
   };
 
+  tfaSubmit = (values) => {
+    const { code } = values;
+    return this.tfaVerify(code);
```
jasonrhodes: Could just grab the stored auth data from `this.props.tfa` and send it into this action creator function, so it can pass it to the `login()` action creator and store it in the right place on success.

src/pages/auth/AuthPage.js
```diff
@@ -29,6 +31,10 @@ export class AuthPage extends Component {
     return this.props.authenticate(username, password, rememberMe);
   }
 
+  tfaVerify(code) {
+    return this.props.verifyAndLogin(code);
+  }
```
jasonrhodes: Don't need this function, can just call `this.props.verifyAndLogin()` from inside `this.tfaSubmit`

src/pages/sso/SSOPage.js
```diff
@@ -19,10 +19,13 @@ export class SSOPage extends Component {
...
-          username: data.username,
-          refresh_token: ''
-        }, true);
+          authData: {
+            access_token: data.accessToken,
+            username: data.username,
+            refresh_token: ''
+          },
+          saveCookie: true
+        });
```
jasonrhodes: Big fan of this change, hate those naked boolean args

src/reducers/tfa.js
```diff
@@ -0,0 +1,26 @@
+const initialState = { tfaEnabled: false, tfaPending: false };
+
+export default (state = initialState, action) => {
+  switch (action.type) {
+    case 'TFA_ENABLED': {
+      return { ...state, tfaEnabled: true };
```
jasonrhodes: Yep just store the auth data here while we're waiting on tfa verification, then grab it from here to pass to the `login()` action creator if tfa is successful

src/pages/auth/components/TfaForm.js
jonambas: I dont see anything out of the ordinary. 

Honestly kind of hate how `<form onSubmit={handleSubmit}>` automatically calls `this.props.onSubmit` behind the scenes tho.

src/pages/sso/SSOPage.js
bizob2828: oh i know, did this one for you JR

src/pages/auth/components/TfaForm.js
bizob2828: @jonambas i agree. i had to read the docs for this and still seems stupid

jasonrhodes: Thirded, it's really stupid.

src/actions/auth.js
```diff
@@ -32,12 +49,24 @@ export function authenticate(username, password, rememberMe = false) {
...
       .then(({ data = {}} = {}) => {
-        const payload = { ...data, username };
+        const authData = { ...data, username };
 
-        // dispatch login success event
-        dispatch(login(payload, true));
+        return Promise.all([ authData, tfaGet({ username, token: authData.access_token })]);
+      })
+      .then(([ authData, { data = {}} = {}]) => {
+        const { results: { enabled }} = data;
```
jasonrhodes: Now that we've added the Promise.all to pass along the authData, the destructuring is probably more confusing than it's worth especially for just one variable. Let's call the 2nd arg here `tfaResponse` like so:
```js
.then(([ authData, tfaResponse ]) => {
  // if tfa enabled must avoid logging in
  if (tfaResponse.data.results.enabled) {
```

jasonrhodes: Not sure if we're worried that those keys might not be there in that response and that's why we were using all the default object values, but if we're really worried we can use `_.get` in that `if`

src/pages/auth/tests/AuthPage.snapshot.test.js
```diff
@@ -8,15 +8,26 @@ const props = {
     loggedIn: false,
```
jasonrhodes: This file has the legacy name of `.snapshot.test.js` ... we can lose the `.snapshot` part of the filename.

src/pages/auth/tests/AuthPage.snapshot.test.js
```diff
@@ -30,3 +41,48 @@ it('redirects when logged in', () => {
   wrapper.setProps({ auth: { loggedIn: true }});
   expect(wrapper).toMatchSnapshot();
 });
+
+it('should display tfa form when TFA is enabled', () => {
+  wrapper.setProps({ tfa: { tfaEnabled: true }});
+  expect(wrapper).toMatchSnapshot();
+});
+
+it('should verify tfa login on submit', () => {
```
jasonrhodes: This test feels really strange to me and hard to understand what's going on/being tested. Is the goal here to test "did we pass the correct function to the tfaForm"?

jasonrhodes: I think I would test this like this:
```js
import TfaForm from '../components/TfaForm';

it('should set up the TFA form correctly', () => {
  wrapper.setProps({ tfa: { tfaEnabled: true }});
  const tfaForm = wrapper.find(TfaForm);
  expect(tfaForm.props().onSubmit).toBe(wrapper.instance().tfaSubmit);
});
```

src/pages/auth/tests/AuthPage.snapshot.test.js
```diff
@@ -30,3 +41,48 @@ it('redirects when logged in', () => {
...
+  const verifySpy = jest.spyOn(wrapper.instance().props, 'verifyAndLogin');
+  wrapper.setProps({ tfa: { tfaEnabled: true }});
+  wrapper.find('Connect(ReduxForm)').simulate('submit', { code: 'code' });
+  expect(verifySpy).toHaveBeenCalled();
+});
+
+it('should sso check on submit when sso is enabled', () => {
```
jasonrhodes: Same thing for this test, would just test that the right callbacks/props were passed to the LoginForm if anything, then test the logic of the `onClickSubmit` handler by calling it directly and passing in `values`. I think it feels off to be testing how Redux Forms handles submit/values.

jasonrhodes: Just as an FYI in all cases, we can do:
```js
wrapper.find(ImportedComponent)
```
anywhere the display name has been hijacked, to avoid the ambiguous use of "Connect(ReduxForm)"

src/pages/auth/tests/AuthPage.snapshot.test.js
bizob2828: since in the form we stub out that method. i want to assert we are calling the right method when we submit the form.  Not sure I understand the point of your test.

src/pages/auth/tests/AuthPage.snapshot.test.js
bizob2828: ah ok, so you are saying just assert the function is registered as the submit handler then test it directly.  Not sure how that's diff from what I currently have though

src/actions/auth.js
bizob2828: Oh I'm all for this.  This f'n code is so confusing for me to read.  Destructuring and defaulting is very hard for me to get used to right now.

bizob2828: that call always returns the right data.

src/pages/auth/tests/AuthPage.snapshot.test.js
jasonrhodes: I think it's slightly different because it seems like your test is testing ReduxForm internals, by relying on the form to actually submit with values. That makes it harder to understand what's going on/being tested when reading the test.

bizob2828: fair enough. i can assert the submit handler has the method and then test it directly

src/actions/auth.js
jasonrhodes: Yeah destructuring assignment stuff takes getting used to for sure. I'm used to it and this was hard for even me to follow, so seems like a good place to be more verbose and explicit. 👍

src/pages/auth/AuthPage.js
```diff
@@ -50,13 +53,26 @@ export class AuthPage extends Component {
...
     this.state.ssoEnabled ? this.ssoSignIn(username) : this.regularSignIn(username, password, rememberMe);
   };
 
+  tfaSubmit = (values) => {
+    const { code } = values;
+    const { tfaEnabled, ...authData } = this.props.tfa;
+    return this.props.verifyAndLogin({ authData, code }).catch((err) => {
+      if (err.response.status === 400) {
+        throw new SubmissionError({
+          _error: 'The code is invalid'
```
jasonrhodes: Change this to `code: 'The code is invalid'` and remove the `{error &&` stuff in the other page and it works even easier :)

src/pages/auth/components/TfaForm.js
```diff
@@ -0,0 +1,54 @@
...
+          <Field
+            autoFocus
+            name='code'
+            id='code'
+            label='Code'
+            placeholder="123456"
+            component={TextFieldWrapper}
+            validate={required}
+          />
+          { error && <Error error={error} /> }
```
jasonrhodes: Can remove this line if you use `code` key in the `SubmissionError`

src/pages/auth/components/TfaForm.js
```diff
@@ -0,0 +1,54 @@
...
+        </div>
+        <form onSubmit={handleSubmit}>
+          <Field
+            autoFocus
+            name='code'
+            id='code'
+            label='Code'
+            placeholder="123456"
+            component={TextFieldWrapper}
+            validate={required}
```
jasonrhodes: Can easily add `maxLength={6}` here to use HTML native max length which prevents them from accidentally typing a 7th number at all.

Can also change validate to `validate={[required, minLength(6), maxLength(6)]}` if you want to save API calls if they put in less than 6 numbers, but up to you if you think that's worth it.

src/pages/auth/AuthPage.js
bizob2828: no shit, that's awesome.  I guess the _error is for generic errors on the entire form

jasonrhodes:
Error handling is still broken for the tfa code, if you hit enter to submit you won't see any error message.

PR #229, opened 2018-01-04
[FAD-5924] Relative date re-calculation on report load/refresh

jasonrhodes:
Adjust the summary report dates until you get a custom range (use last hour for easiest test), then wait a few minutes, refresh the page. Times will update to _current_, re-calculated last hour.

Click share modal when you're on a relative range and you'll see a prompt to pin the range (which only changes the range={relativeRange} to range=custom and back. 

When on a custom range, the share modal shows no note or lock button.

Also, the share button is now disabled while the chart is reloading (prevents weird in-between modal state)

### DIFF
--- a/src/helpers/reports.js
+++ b/src/helpers/reports.js
@@ -1,16 +1,18 @@
 import moment from 'moment';
 import qs from 'query-string';
 import _ from 'lodash';
+import { getRelativeDates } from 'src/helpers/date';
 
 /**
  * Creates search options object from shared report options. Page specific options not included (ie. summary chart selected metrics)
  * @param  {Object} filters - reportFilters state
  * @return {Object} - formatted search options object
  */
-function getFilterSearchOptions(filters) {
+export function getFilterSearchOptions(filters) {
   return {
     from: moment(filters.from).utc().format(),
     to: moment(filters.to).utc().format(),
+    range: filters.relativeRange,
     filters: filters.activeList.map((filter) => {
       const subaccount = filter.type === 'Subaccount' ? `:${filter.id}` : '';
       return `${filter.type}:${filter.value}${subaccount}`;
@@ -27,7 +29,7 @@ function getFilterSearchOptions(filters) {
  *     link - full url
  *   }
  */
-function getShareLink(options) {
+export function getShareLink(options) {
   const search = _.isEmpty(options) ? '' : `?${qs.stringify(options, { encode: false })}`;
   const link = `${window.location.href.split('?')[0]}${search}`;
   return { search, link };
@@ -42,7 +44,7 @@ function getShareLink(options) {
  *     filters - array of objects ready to be called with reportFilters.addFilter action
  *   }
  */
-function parseSearch(search) {
+export function parseSearch(search) {
   let options = {};
   let filtersList;
 
@@ -50,7 +52,7 @@ function parseSearch(search) {
     return { options };
   }
 
-  const { from, to, metrics = [], filters = []} = qs.parse(search);
+  const { from, to, range = 'custom', metrics = [], filters = []} = qs.parse(search);
 
   const metricsList = typeof metrics === 'string' ? [metrics] : metrics;
   filtersList = typeof filters === 'string' ? [filters] : filters;
@@ -76,15 +78,10 @@ function parseSearch(search) {
   options = {
     metrics: metricsList,
     from: new Date(from),
-    to: new Date(to)
+    to: new Date(to),
+    ...getRelativeDates(range) // invalid or custom ranges produce {} here
   };
 
   // Filters are not passed to metrics refresh actions
   return { options, filters: filtersList };
 }
-
-export {
-  getFilterSearchOptions,
-  getShareLink,
-  parseSearch
-};
--- a/src/helpers/reports.js
+++ b/src/helpers/reports.js
@@ -45,19 +45,13 @@ export function getShareLink(options) {
  *   }
  */
 export function parseSearch(search) {
-  let options = {};
-  let filtersList;
-
   if (!search) {
-    return { options };
+    return { options: {}};
   }
 
   const { from, to, range = 'custom', metrics = [], filters = []} = qs.parse(search);
-
   const metricsList = typeof metrics === 'string' ? [metrics] : metrics;
-  filtersList = typeof filters === 'string' ? [filters] : filters;
-
-  filtersList = filtersList.map((filter) => {
+  const filtersList = (typeof filters === 'string' ? [filters] : filters).map((filter) => {
     const parts = filter.split(':');
     const type = parts.shift();
     let value;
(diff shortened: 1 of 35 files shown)


jasonrhodes:
First draft here is seen below, going to experiment with a few other UX ideas soon.
share-modal-relative-toggle

src/pages/reports/components/ShareModal.js
```diff
@@ -1,17 +1,42 @@
...
+
+    return (
+      <Modal open={open}>
+        {open && <WindowEvent event='keydown' handler={handleToggle} />}
+        <Panel title='Share this report'>
+          <Panel.Section>
+            {isRelative && !pinned && <p>Note: This link will always load this report with a relative range ({relativeRange.label}).</p>}
+            {isRelative && pinned && <p>Note: This link will always load this report with specific dates ({from} to {to}).</p>}
+            <CopyField
+              value={`${window.location.href}?${search}`}
```
jonambas: Getting double search here. Think `window.location.href` includes everything.

`http://localhost:3100/reports/summary?from=2017-12-05T20:38:47Z&metrics=count_targeted&metrics=count_rendered&metrics=count_accepted&metrics=count_bounce&range=30days&to=2018-01-04T20:38:47Z?from=2017-12-05T20:38:47Z&metrics=count_targeted&metrics=count_rendered&metrics=count_accepted&metrics=count_bounce&range=30days&to=2018-01-04T20:38:47Z`

src/helpers/reports.js
```diff
@@ -1,16 +1,18 @@
...
  * Creates search options object from shared report options. Page specific options not included (ie. summary chart selected metrics)
  * @param  {Object} filters - reportFilters state
  * @return {Object} - formatted search options object
  */
-function getFilterSearchOptions(filters) {
+export function getFilterSearchOptions(filters) {
   return {
     from: moment(filters.from).utc().format(),
     to: moment(filters.to).utc().format(),
+    range: filters.relativeRange,
```
jonambas: Would it make sense to only include to/from if set to `range` is `custom`, and leave them out if a `range` is set? It would make the share url look a little cleaner

src/pages/reports/summary/SummaryPage.js
```diff
@@ -70,24 +73,22 @@ class SummaryReportPage extends Component {
...
       metrics: chart.metrics.map((metric) => metric.key),
       ...getFilterSearchOptions(filters)
     };
 
-    const { link, search } = getShareLink(options);
-
-    this.setState({ link });
-    history.replace({ pathname: '/reports/summary', search });
+    this.setState({ query });
+    history.replace({ pathname: '/reports/summary', search: qs.stringify(query, { encode: false }) });
```
jonambas: Will need to remove `getShareLink` and do this on every other report page as well.

src/pages/reports/components/ShareModal.js
```diff
@@ -1,17 +1,42 @@
...
+    const { open, query, handleToggle } = this.props;
+    const relativeRange = relativeDateOptions.find((item) => item.value === query.range);
+    const isRelative = relativeRange && query.range !== 'custom';
+    const { pinned } = this.state;
+    const pinnedRange = pinned ? { range: 'custom' } : {};
+
+    const search = qs.stringify({ ...query, ...pinnedRange }, { encode: false });
+    const connectLeft = !isRelative ? null : <Button primary={this.state.pinned} onClick={() => this.setState({ pinned: !pinned })}><Icon name='Lock' /></Button>;
+    const from = moment(query.from).format('d MMM Y h:mma');
+    const to = moment(query.to).format('d MMM Y h:mma');
```
jonambas: I know you're not done here, but this is a little hard to follow

src/helpers/reports.js
jasonrhodes: I removed them from the link inside the share modal when it's relative and not pinned.

jasonrhodes:
screen shot 2018-01-05 at 5 28 44 pm

Simplified a bit for now.

src/helpers/reports.js
```diff
@@ -27,7 +29,7 @@ function getFilterSearchOptions(filters) {
  *     link - full url
  *   }
  */
-function getShareLink(options) {
+export function getShareLink(options) {
```
jasonrhodes: Can remove

jonambas:
I saw @bizob2828 refactored some of the same code a little: https://github.com/SparkPost/2web2ui/pull/231/files#diff-f3da886548ee5040c8ac97df0ed2a055L31

src/pages/reports/components/ShareModal.js
```diff
@@ -1,17 +1,78 @@
...
-        <Button primary onClick={handleToggle}>Done</Button>
-      </Panel.Section>
-    </Panel>
-  </Modal>
-);
-
-export default ShareModal;
+export default class ShareModal extends Component {
+  state = {
+    pinned: true
```
jonambas: I'm still wondering where this assumption that people by default want to share calculated dates. I guess maybe because this is how the old app works, but 🤔 

There are going to be a number of folks who skip the modal completely and just copy the url from the address bar (which defaults to the opposite of this modal if using relative range). Should we make sure the url and default share modal link are always the same? We would need to 1) setting this to false 2) hide to/from from before options get to history.replace

src/pages/reports/components/ShareModal.js
```diff
@@ -1,17 +1,78 @@
...
+    const { pinned } = this.state;
+    const relativeRange = relativeDateOptions.find((item) => item.value === query.range);
+    const isRelative = relativeRange && query.range !== 'custom';
+
+    if (!isRelative) {
+      return null;
+    }
+
+    return (
+      <div style={{ marginBottom: '30px' }}>
```
jonambas: You can avoid this by wrapping the textfield & checkbox in a panel.section, and the done button in its own panel.section

src/pages/reports/components/ShareModal.js
jasonrhodes: My assumption is that the most common use-case for sharing a report is so you can share data you are looking at, some spike or change etc etc. If you accidentally share that as relative, there's no way for the receiver of the link to know what dates you were looking at. 

By defaulting the share link to fixed dates, you have to *decide* to switch back to the relative version. If you accidentally share the fixed version, it's very easy for the receiver of the link to simply pick the relative range they'd rather see. 

In my opinion, if you interact with URL directly, you're on your own and I'm totally ok with that. I think that's actually the reason for the share modal, to provide sugar and better defaults than just copying the URL.

src/pages/reports/components/ShareModal.js
jasonrhodes: Had just added this as a quick fix and forgot to swap out, will do thanks.

src/pages/reports/components/ShareModal.js
jonambas: Hmm okay, makes sense 👍

src/pages/reports/bounce/BouncePage.js
```diff
@@ -105,19 +100,24 @@ export class BouncePage extends Component {
...
+        <Filters
+          refresh={this.handleRefresh}
+          onShare={this.handleModalToggle}
+          shareDisabled={chartLoading}
+        />
         { this.renderChart() }
         <Panel title='Bounced Messages' className='ReasonsTable'>
           { this.renderCollection() }
         </Panel>
         <ShareModal
```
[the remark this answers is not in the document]
jasonrhodes: Good call, I'll take a look at that -- gonna file a ticket now related to the other report refactor/cleanup ticket and to this one.

PR #231, opened 2018-01-05
FAD-6020 added top level metrics to bounce, rejection, delay pages

bizob2828:


src/pages/reports/bounce/BouncePage.js
```diff
@@ -24,27 +26,18 @@ export class BouncePage extends Component {
   }
 
   componentDidMount() {
-    this.handleRefresh(this.parseSearch());
+    this.handleRefresh(parseSearch(this.props.location.search));
     this.props.refreshTypeaheadCache();
   }
 
-  parseSearch() {
```
bizob2828: restore this. I didn't know what I was doing an apparently breaks how redux works

src/pages/reports/delays/DelayPage.js
```diff
@@ -91,12 +92,27 @@ export class DelayPage extends Component {
     />;
   }
 
+  renderTopLevelMetrics() {
+    const { aggregatesLoading, aggregates, filters } = this.props;
+
+    if (aggregatesLoading) {
+      return <PanelLoading />;
```
jonambas: We should make PanelLoading accept a height to make the final render less jarring

src/pages/reports/delays/DelayPage.js
```diff
@@ -91,12 +92,27 @@ export class DelayPage extends Component {
...
+    }
+
+    return <MetricsSummary
+      rateValue={(aggregates.count_delayed_first / aggregates.count_accepted) * 100}
+      rateTitle={'Delayed Rate'}>
+      { aggregates.count_delayed && <span><strong>{aggregates.count_delayed.toLocaleString()}</strong> of your messages were delayed of <strong>{aggregates.count_accepted.toLocaleString()}</strong> messages accepted in the <strong>last {humanizeTimeRange(filters.from, filters.to)}</strong>.</span> }
```
jonambas: Yeah this isn't pretty, but not sure there's a way around it

src/pages/reports/components/MetricsSummary.js
```diff
@@ -0,0 +1,35 @@
+import React, { Component } from 'react';
+import { Panel, Grid } from '@sparkpost/matchbox';
+import { Percent } from 'src/components/formatters';
+
+import styles from './MetricsSummary.module.scss';
+
+class MetricsSummary extends Component {
+  render() {
+    const { rateValue, rateTitle } = this.props;
+    const children = React.Children.toArray(this.props.children);
```
jonambas: Hmm not sure this is necessary, maybe pass in the support text as a separate prop?

```js
const { children, supportText } = this.props;

// in jsx
<p className={styles.Description}>
  {children}
</p>
{supportText}
```

src/pages/reports/components/MetricsSummary.js
```diff
@@ -0,0 +1,35 @@
...
+
+    return (
+      <Panel accent>
```
jonambas: Not sure if I like the accent on this panel or not 🤔

src/pages/reports/components/MetricsSummary.js
```diff
@@ -0,0 +1,35 @@
...
+        <Grid>
+          <Grid.Column lg={3}>
+            <div className={styles.panelvertical}>
```
jonambas: There are a few styles here that I'd like to fix/cleanup. Can do in my ticket if this gets merged first.

src/pages/reports/bounce/BouncePage.js
```diff
@@ -28,23 +30,29 @@ export class BouncePage extends Component {
...
-    this.props.refreshBounceChartMetrics(options)
-      .then(() => this.updateLink())
-      .then(() => this.props.refreshBounceTableMetrics(options))
-      .catch((err) => {
-        this.props.showAlert({ type: 'error', message: 'Unable to refresh bounce report.', details: err.message });
-      });
-  }
+  handleRefresh = (options) => Promise.all([
+    this.props.refreshBounceChartMetrics(options),
+    this.props.refreshBounceTableMetrics(options)
```
jonambas: 👍

src/pages/reports/bounce/BouncePage.js
```diff
@@ -104,12 +112,29 @@ export class BouncePage extends Component {
...
+
+    if (aggregatesLoading) {
+      return <PanelLoading />;
+    }
+
+    if (aggregates) {
+      return <MetricsSummary
+        rateValue={(aggregates.countBounce / aggregates.countTargeted) * 100}
+        rateTitle={'Bounce Rate'}>
+        { aggregates.countBounce && <span><strong>{aggregates.countBounce.toLocaleString()}</strong> of your messages were bounced of <strong>{aggregates.countTargeted.toLocaleString()}</strong> messages targeted in the <strong>last {humanizeTimeRange(filters.from, filters.to)}</strong>.</span> }
```
jonambas: I think we should only humanizeTimeRange if using a relativeRange, and show the full to and from if range is custom.

jonambas: I DM'd you about this, but I'll add here so feedback is all in one place:

- the `&&` check should be done before this return. The component shouldn't render at all if it isn't ready to.
- could maybe destructure `aggregates` to make the message portion a little more readable
- we could also pass through either `filters` or `to`, `from`, `range` separately and handle the date logic inside the component

jonambas:
Made changes to bounce and delay page - just passing through `filters` and adding the date stuff to the end of `children` inside the component.

```js
<MetricsSummary
  rateValue={(countBounce / countTargeted) * 100}
  rateTitle='Bounce Rate'
  {...filters} >
    <strong>{countBounce.toLocaleString()}</strong> of your messages were bounced of <strong>{countTargeted.toLocaleString()}</strong> messages targeted
</MetricsSummary>
```
Also fixed some wonky loading states, and added `minHeight` prop to the loading panel

src/pages/reports/components/MetricsSummary.js
```diff
@@ -0,0 +1,57 @@
...
+import { Panel, Grid } from '@sparkpost/matchbox';
+import { Percent } from 'src/components';
+import { relativeDateOptionsIndexed } from 'src/helpers/date';
+
+import styles from './MetricsSummary.module.scss';
+
+class MetricsSummary extends Component {
+  renderDate() {
+    const { to, from, relativeRange } = this.props;
+    const format = 'MMM D \'YY, h:mma';
```
bizob2828: you're cool with this? most dates i see in app are verbose e.g. 2017 instead of '17

src/pages/reports/components/MetricsSummary.js
```diff
@@ -0,0 +1,57 @@
...
+import PropTypes from 'prop-types';
+import moment from 'moment';
+import { Panel, Grid } from '@sparkpost/matchbox';
+import { Percent } from 'src/components';
+import { relativeDateOptionsIndexed } from 'src/helpers/date';
+
+import styles from './MetricsSummary.module.scss';
+
+class MetricsSummary extends Component {
+  renderDate() {
```
bizob2828: good move on moving this to the component instead of in every page

src/pages/reports/components/MetricsSummary.js
bizob2828: also do we want to put date formats in a central place?

src/pages/reports/components/MetricsSummary.js
```diff
@@ -0,0 +1,57 @@
...
+              {secondaryMessage && <p className={styles.Secondary}>{secondaryMessage}</p>}
+            </div>
+          </Grid.Column>
+        </Grid>
+      </Panel>
+    );
+  }
+}
+
+MetricsSummary.propTypes = {
```
bizob2828: what is this doing?

PR #232, opened 2018-01-05
FAD-5591 Accepted Chart

jonambas:
**Pie Chart Changes**
- Moved `BounceChart` to global components under `PieChart`
  - `LegendItem` now has a chevron icon if it has children. Should help identify which items are clickable.
  - `LegendItem` will now fade if a pie chart sector or another legend item is being hovered. Should help identify which pie sector is being hovered
  - Adjusted outer ring sizing to avoid clipping at certain screen sizes
  - Increased pie chart minimum sector angle to 3 degrees. Should help pie chart visually when massively skewed to one sector

**Bounce Page Changes**
- Fixed bounce reasons table empty state
- Adjusted colors slightly in bounce pie chart
- When in a sub-section, Legend header will now show targeted count

**Accepted Page**
- Attempts are grouped at the top level by **First Attempt** and **2 or More**
- **2 or More** includes:
  - 2-4
  - 5-7
  - 8-10
  - 11 +

**Helper Changes**
- `generateColors` now accepts saturate and rotate values

**TODO**
- [x] Find a good color palette for accepted chart
- [x] Add metrics cards as part of AC? (#231)
- [x] Finish Tests

---
screen shot 2018-01-08 at 5 35 37 pm


src/pages/reports/bounce/components/tests/__snapshots__/ChartGroup.test.js.snap
```diff
@@ -147,57 +147,12 @@ exports[`ChartGroup:  should render 1`] = `
...
-      Object {
-        "fill": "hsl(355.1, 70.5%, 52.2%)",
-        "name": "child",
-      },
-    ]
-  }
-  secondaryData={null}
-/>
-`;
+exports[`ChartGroup:  should render correctly after click 1`] = `null`;
```
jonambas: Fix this test

src/pages/reports/accepted/components/ChartGroup.js
```diff
@@ -0,0 +1,149 @@
...
+import _ from 'lodash';
+import { Grid, Panel } from '@sparkpost/matchbox';
+import { Loading, PieChart } from 'src/components';
+import { generateColors } from 'src/helpers/color';
+import styles from './ChartGroup.module.scss';
+
+// Chart color palette generated from:
+const primaryColor = '#8CCA3A';
+
+export class ChartGroup extends Component {
```
jonambas: There's like a 90% overlap between the accepted ChartGroup and the bounce ChartGroup. Could possibly refactor most of code out into a shared component, just not sure how to handle all the item labels and rates.

src/helpers/color.js
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Generates a color palette for pie chart data
+ * @param  {Array} arr        array of objects
+ * @param  {String} baseColor
+ * @param  {Number} rotate    hue rotation multiplier, divided by array length
+ * @param  {Number} saturate  hue saturation multiplier, divided by array length
+ * @return {Array}            array with 'fill' key added
+ */
+export function generateColors(arr, { color = '#000000', rotate = null, saturate = null } = {}) {
```
jonambas: Can probably put those default `rotate` and `saturate` values in here

src/components/pieChart/PieChart.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import ActiveLabel from './ActiveLabel';
+import Legend from './Legend';
+import Chart from './Chart';
+
+// Overwrites 'LoadableComponent'
+Chart.displayName = 'PieChart.Chart';
```
jasonrhodes: Should we do this before export in `./Chart`?

src/helpers/accepted.js
```diff
@@ -0,0 +1,40 @@
...
+  const first = { name: '1st Attempt', count: 0 };
+  const overTwo = { name: '2 or More Attempts', count: 0 };
+  const twoToFour = { name: '2 to 4 Attempts', count: 0 };
+  const fiveToSeven = { name: '5 to 7 Attempts', count: 0 };
+  const eightToTen = { name: '8 to 10 Attempts', count: 0 };
+  const overEleven = { name: '11 or More Attempts', count: 0 };
+
+  _.each(deliveries, ({ attempt, count_delivered }) => {
+    if (attempt === 1) {
+      first.count = count_delivered;
```
jasonrhodes: I think this still has to be `+=` here or else we end up with just the last "first attempt delivery's" count_delivered value in `first.count`.

jasonrhodes: Oh never mind, I thought this was a list of arbitrary deliveries each with whatever attempt they succeeded on, didn't realize it was broken into 1, 2, 3, 4 etc

src/helpers/accepted.js
```diff
@@ -0,0 +1,40 @@
...
+ */
+export function reshapeAttempts(deliveries) {
+  const first = { name: '1st Attempt', count: 0 };
+  const overTwo = { name: '2 or More Attempts', count: 0 };
+  const twoToFour = { name: '2 to 4 Attempts', count: 0 };
+  const fiveToSeven = { name: '5 to 7 Attempts', count: 0 };
+  const eightToTen = { name: '8 to 10 Attempts', count: 0 };
+  const overEleven = { name: '11 or More Attempts', count: 0 };
+
+  _.each(deliveries, ({ attempt, count_delivered }) => {
```
jasonrhodes: Switch this to `deliveries.forEach(`

I tried to rework this method into some lodash magic and it got considerably worse tbh, so I think this works well.

src/helpers/color.js
```diff
@@ -0,0 +1,21 @@
+import clr from 'color';
+
+/**
+ * Generates a color palette for pie chart data
+ * @param  {Array} arr        array of objects
+ * @param  {String} baseColor
```
jasonrhodes: If this param is really `baseColor` can we call it that below and call the imported lib `color`? Or if not, switch this jsdoc at least

src/pages/reports/accepted/AcceptedPage.js
```diff
@@ -0,0 +1,102 @@
...
+      .then(() => this.updateLink())
+      .catch((err) => {
+        showAlert({ type: 'error', message: 'Unable to refresh report.', details: err.message });
+      });
+  }
+
+  updateLink = () => {
+    const { filters, history } = this.props;
+    const options = getFilterSearchOptions(filters);
+    const { link, search } = getShareLink(options);
```
jasonrhodes: Probably need to rebase and swap this out

src/pages/reports/accepted/components/ChartGroup.js
jasonrhodes: It's ok for now I think, we can circle back around.

src/pages/reports/accepted/tests/AcceptedPage.test.js
```diff
@@ -0,0 +1,76 @@
...
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('should render correctly with no accepted results', () => {
+    wrapper.setProps({ aggregates: null });
+    expect(wrapper.find('Empty')).toMatchSnapshot();
+  });
+
+  it('should handle modal', () => {
+    wrapper.find('Connect(Filters)').simulate('share');
```
jasonrhodes: Can we use a reference to the included component instead of the Redux-flavored display name string?

src/components/pieChart/PieChart.js
jonambas: Yeah probably. I did this so tests would render out 'PieChart.Chart', but probably best to no obscure the loadable component

src/actions/acceptedReport.js
```diff
@@ -1,6 +1,6 @@
 import { fetchDeliverability, fetchDeliveriesByAttempt } from 'src/actions/metrics';
 import { maybeRefreshTypeaheadCache, refreshReportRange } from 'src/actions/reportFilters';
-import { getQueryFromOptions, buildCommonOptions, getMetricsFromKeys } from 'src/helpers/metrics';
+import { getQueryFromOptions, buildCommonOptions, getMetricsFromKeys, transformData } from 'src/helpers/metrics';
```
jasonrhodes: This method isn't new?

src/components/pieChart/async/Chart.js
```diff
@@ -29,7 +29,7 @@ const Chart = ({
 
   return (
     <div className={cx('Chart', hoveredItem && 'Chart--hover')}>
-      <ResponsiveContainer width='100%' height={350}>
+      <ResponsiveContainer width='99%' height={350}>
```
jasonrhodes: 🤔

jonambas: 100% was preventing it from being responsive, idk

src/actions/acceptedReport.js
jonambas: Nope, was built for the summary chart, but works just fine for getting computed values

src/components/pieChart/async/Chart.js
jasonrhodes: A+

jasonrhodes:
This all looks great. I have 3 last comments:

1. Fetching new data still feels like an unnecessarily rough experience. I think maybe we should consider keeping the existing data in place and using the "loading icon up in the top right corner of the chart section" idea, to indicate we're loading new data without having the page become 2 loading icons for however long it takes to get new data.

1. I love the new Metrics cards in general but not a huge fan of how they push the beautiful chart down almost off the screen. Might have to live with that but would that 2nd row of cards look ok below the chart?

1. The tooltip caret sometimes looks slightly separated from the body, maybe by a pixel or two. Maybe a Matchbox tweak? See my screenshot:

screen shot 2018-01-20 at 8 10 44 am

jonambas:
@jasonrhodes 

1. Yeah, not a fan. `loading` is passed to chartgroup, I think I intended to do a fade-chart-on-load thing like the summary chart. But initial load could be better.

2. Could make fonts smaller maybe

3. Yeah has something to do with UI scaling. If you increase viewport width, it'll correct itself. But easy to fix in MB.

PR #234, opened 2018-01-09
FAD-5847 Suppression Detail View

rajumsys:
Todo:

- [x] Figure out if we need to pass `type` during deletion
- [x] Need to revisit the logic in (suppression) reducer to filter out the deleted recipient. Should we also include subaccount (can a recipient be both in master and in subaccount?)

Notes:
- Updated LabelledValue component to support element


src/reducers/suppressions.js
```diff
@@ -37,6 +39,16 @@ export default (state = initialState, action) => {
       }
     }
 
+    case 'DELETE_SUPPRESSION_PENDING':
+      return { ...state, deleting: true };
+
+    case 'DELETE_SUPPRESSION_SUCCESS':
+      return { ...state, deleting: false, list: state.list.filter((s) => s.recipient !== meta.suppression.recipient) };
```
rajumsys: match by subaccount id here

ewandennis:
UX notes on the modal:
    - Should all the values be **bold**? Is that `LabelledValue` styling? If so, it probably looks better in msg events details or somewhere.
    - Lets format the `created` and `updated` fields - this might require a small group conversation
    - We should show the subaccount on the modal

(The AC are pretty specific about implementation but maybe we could vary a little and include a delete button on the modal...)

src/pages/suppressions/components/Detail.js
```diff
@@ -0,0 +1,41 @@
...
+    const { suppression } = this.props;
+
+    const elements = [];
+    _.forEach(suppression, (val, key) => {
+      val = (val || '').toString();
+      if (key === 'description') {
+        val = <LongTextContainer text={val} />;
+      }
+
+      elements.push(<LabelledValue key={key} label={key} value={val}/>);
```
ewandennis: Since we're not changing, adding or removing these elements in a list, I'm not sure they need a `key` prop.

src/pages/suppressions/components/Detail.js
```diff
@@ -0,0 +1,41 @@
...
+    elements.push(<LabelledValue key='raw json' label='raw json' value={<CopyField value={JSON.stringify(suppression)} />}/>);
+    return elements;
+  }
+
+  render() {
+    const { open, deleting, onCancel } = this.props;
+
+    return (
+      <BaseModal open={open}>
+        <Panel title={'Suppression Details'} accent sectioned>
```
ewandennis: Do we need `sectioned` ? I don't think we're using `Panel.Section`.

src/pages/suppressions/components/Detail.js
```diff
@@ -0,0 +1,41 @@
...
+          { deleting ? <div className={styles.Loading}><Loading /></div> : this.renderContents()}
+          <Button onClick={() => { onCancel(); }} className={styles.Cancel}>Close</Button>
```
ewandennis: `onClick={onCancel}` ?

src/pages/suppressions/components/Detail.js
```diff
@@ -0,0 +1,41 @@
...
+        </Panel>
+      </BaseModal>
+    );
+  }
+}
+
+export default connect()(Detail);
```
ewandennis: Does this component need to be connected?

src/pages/suppressions/components/Detail.module.scss
```diff
@@ -0,0 +1,8 @@
+.Cancel {
+  float: right;
```
ewandennis: I'm totally fine with this but I imagine some folks like flexbox. Something like this might do it:

```<div className={styles.Buttons}><button>Click me!</button>```

```css
.Buttons {
   display: flex;
   flex-direction: row;
   justify-content: flex-end;
}
```

src/reducers/suppressions.js
ewandennis: ...and by `type` if we manage to get that working ;)

src/pages/suppressions/components/tests/Results.test.js
```diff
@@ -6,40 +6,140 @@ import { Results } from '../Results';
...
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('renders detail modal correctly', () => {
+    wrapper.setProps({ results, hasSubaccounts: true });
+    wrapper.setState({ detail: { open: false, data: {}}});
     expect(wrapper).toMatchSnapshot();
   });
 
+  it('renders delete confirmatio modal correctly', () => {
```
ewandennis: tiny typo :)

src/pages/suppressions/components/Detail.js
rajumsys: it complained!

src/pages/suppressions/components/Detail.js
rajumsys: not sure, copied from confirmationModal

src/pages/suppressions/components/Detail.js
rajumsys: Then it'll pass the `event` param to `toggleDetailModal` method. But `toggleDetailModal` is expecting a row or has to be `undefined`.

src/pages/suppressions/components/Detail.js
```diff
@@ -0,0 +1,41 @@
...
+    return elements;
+  }
+
+  render() {
+    const { open, deleting, onCancel } = this.props;
+
+    return (
+      <BaseModal open={open}>
+        <Panel title={'Suppression Details'} accent sectioned>
+          { deleting ? <div className={styles.Loading}><Loading /></div> : this.renderContents()}
```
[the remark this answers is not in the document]
rajumsys: Probably we don't needed on detail unless we put a delete button on detail.

src/pages/suppressions/components/Detail.js
rajumsys: it's not connecting to any store. will get rid of.

src/pages/suppressions/components/Detail.module.scss
rajumsys: This comes from ConfirmationModal

src/pages/suppressions/components/Results.js
```diff
@@ -54,21 +92,66 @@ export class Results extends Component {
...
+  }
+
+  renderDeleteModal = () => {
+    const { open, data } = this.state.del;
+    const { deleting } = this.props;
+    if (!data) {
+      return null;
+    }
+
+    return (<ConfirmationModal
```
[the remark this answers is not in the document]
rajumsys: just because I didn't know/notice it exists :(

src/pages/suppressions/components/Detail.js
jasonrhodes: React requires key for all item lists, for better or worse. :|

rajumsys:
> - We should show the subaccount on the modal

It's shown when the field exist. Not all suppression has it.

src/pages/suppressions/components/Detail.js
rajumsys: it looks really bad without it!

src/pages/suppressions/components/Detail.js
ewandennis: Right 🤦‍♂️

src/pages/suppressions/components/Detail.js
ewandennis: Ew yes.

src/pages/suppressions/components/Detail.js
ewandennis: Surprising. Good catch.

ewandennis:
>> Should all the values be bold? Is that LabelledValue styling? If so, it probably looks better in msg events details.

> Update: Looked at message events. looks same to me! they look better because labels are smaller? i can increase the min width of label in LabelledValue component.

Yeah, we haven't given any attention to making message events details readable yet. We could also order the fields here for readability. Show the short summary fields first, then the big ugly one.

>> Lets format the created and updated fields - this might require a small group conversation

> ok. fwiw, seems it's not formatted in msg events ui too. so we need to make them consistent

Agreed, we need to audit date formatting everywhere. We could do more-than-nothing here though since we caught it. A `new Date(...).toLocaleString()` might look good.

>> We should show the subaccount on the modal

> It's shown when the field exist. Not all suppression has it.
Nice.

>> I got into a strange situation where I had no filters active but also no suppressions were listed.

> I've given a fix for it. See if it's sufficient. However, that scenario can occur for even if you search for an email that don't exist.

I'm in a different odd filtering state now.  I'll just file a fixup ticket for suppressions filtering. FAD-6049.

Finally, can we please put explicit markup in place on `Details` modal for each field? It helps readability and when we come to audit and it'll save us having to do it when we do our field rendering audit.
https://github.com/SparkPost/2web2ui/pull/234#discussion_r160982118

src/actions/tests/suppressions.test.js
```diff
@@ -0,0 +1,55 @@
...
+            headers: { 'x-msys-subaccount': 101 },
+            suppression: mockSuppression
+          }
+        }
+      );
+    });
+
+  });
+
+});
```
bkemper: Here is another test pattern that we are using.

```
// Mock sparkpostApiRequest to pass along args
jest.mock('src/actions/helpers/sparkpostApiRequest', (args) => args);

// Snapshot the args to confirm they are the correct values
it('', () => {
  const suppression = {};
  expect(suppressions.deleteSuppression(suppression)).toMatchSnapshot();
});
```

src/pages/suppressions/ListPage.js
```diff
@@ -78,20 +78,21 @@ export class ListPage extends Component {
...
-  const { listLoading, list } = state.suppressions;
+  const { listLoading, list, deleting } = state.suppressions;
   return {
     reportFilters: state.reportFilters,
     loading: listLoading,
     list: list,
     hasSubaccounts: hasSubaccounts(state),
-    subaccounts: state.subaccounts.list
+    subaccounts: state.subaccounts.list,
+    deleting: deleting
```
bkemper: This can be shorthanded to `deleting`.

src/pages/suppressions/components/Detail.js
```diff
@@ -0,0 +1,40 @@
...
+      </div>
+    );
+  }
+
+  render() {
+    const { open, onCancel } = this.props;
+
+    return (
+      <BaseModal open={open}>
+        <Panel title={'Suppression Details'} accent sectioned>
```
bkemper: The curly brackets can be removed from the title value.

`<Panel accent sectioned title="Suppression Details">`

src/pages/suppressions/components/Detail.js
```diff
@@ -0,0 +1,40 @@
...
+    const { suppression } = this.props;
+
+    return (
+      <div>
+        <LabelledValue key='recipient' label='Recipient' value={suppression.recipient} />
+        <LabelledValue key='type' label='Type' value={suppression.type === 'transactional' ? 'Transactional' : 'Non-transactional'} />
+        <LabelledValue key='source' label='Source' value={suppression.source} />
+        <LabelledValue key='updated' label='Updated' value={suppression.updated} type='datetime' />
+        <LabelledValue key='subaccount' label='Subaccount' value={formatSubaccountDisplay(suppression.subaccount_id)} />
+        <LabelledValue key='created' label='Created' value={suppression.created} type='datetime' />
```
bkemper: I am really torn.  I see the value for why `type` was added -- now, everyone that uses `LabelledValue` with `type="datetime"` will get the same formatted value.  However, I don't think this is going to be the only place we need to format a date and time.  I would opt for a dumber version of LabelledValue that doesn't know how to handle types and use your helper `formatDateTime` to format the prop, `value={formatDateTime(suppression.created)}`.

src/pages/suppressions/components/Results.js
```diff
@@ -19,6 +34,29 @@ export class Results extends Component {
...
   }
 
+  deleteSuppression = () => {
+    const { showAlert, deleteSuppression } = this.props;
+    const { data } = this.state.del;
+
+    return deleteSuppression(data)
+      .then(() => {
+        this.toggleDeleteModal();
+        return showAlert({ type: 'success', message: `${data.recipient} was successfully deleted from the suppression list` });
```
bkemper: I don't think this needs to be `return`'ed.

src/pages/suppressions/components/Results.js
```diff
@@ -54,21 +92,65 @@ export class Results extends Component {
...
+    const { open, data } = this.state.del;
+    const { deleting } = this.props;
+    if (!data) {
+      return null;
+    }
+
+    return (<DeleteModal
+      open={open}
+      title={`Are you sure you want to delete ${data.recipient} from suppression list?`}
+      content={<p>This can not be undone!</p>}
```
bkemper: Should this confirmation modal have a similar title and content structure as others (e.g. user delete)?

screen shot 2018-01-16 at 3 47 24 pm

src/actions/tests/suppressions.test.js
```diff
@@ -0,0 +1,55 @@
...
+        }
+      );
+    });
+
+    it('includes subaccount id in header when exists', async() => {
+      mockSuppression.subaccount_id = 101;
+      await suppressions.deleteSuppression(mockSuppression);
+      const args = sparkpostApiRequest.mock.calls[0];
+      expect(sparkpostApiRequest).toHaveBeenCalledTimes(1);
+      expect(args[0]).toEqual(
```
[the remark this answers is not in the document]
rajumsys: that's weird. can't remember why did I do like this while I've definitely used `toHaveBeenCalledWith` in other places.

src/pages/suppressions/components/Detail.js
rajumsys: > everyone that uses LabelledValue with type="datetime" will get the same formatted value.
that's exactly the reason. datetime formatting isn't consistent. goal is to make it. 

actually ewan and i discussed this offline. so sorry for missing context. and i don't have objection against your version too.

src/pages/suppressions/components/Results.js
rajumsys: see @ewandennis's previous suggestion.

src/actions/suppressions.js
```diff
@@ -98,3 +99,19 @@ export function searchSuppressions(options) {
...
+export function deleteSuppression(suppression) {
+  const { recipient, subaccount_id: subaccountId } = suppression;
+
+  return sparkpostApiRequest({
+    type: 'DELETE_SUPPRESSION',
+    meta: {
+      method: 'DELETE',
+      url: `/suppression-list/${recipient}`,
+      headers: setSubaccountHeader(subaccountId),
+      // data: { type }, /* TODO Check if we need to set type as old UI doesn't do it. Also figure out why it's not working if just comment out this line */
```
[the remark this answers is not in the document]
rajumsys: Yeah, that's correct. 

Ok, I think I found why I thought "it's not working" earlier. it's cache. after deleting, it takes a while for the record to actually deleted which made me thinking it's not working.

src/actions/tests/suppressions.test.js
```diff
@@ -0,0 +1,55 @@
...
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  describe('deleteSuppression', () => {
+    it('makes api call with correct parameter', async() => {
+      await suppressions.deleteSuppression(mockSuppression);
+      const args = sparkpostApiRequest.mock.calls[0];
+      expect(sparkpostApiRequest).toHaveBeenCalledTimes(1);
+      expect(args[0]).toEqual(
```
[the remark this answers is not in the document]
rajumsys: i won't combine as i can't assert if it called once or more than once. but i replaced `toEqual` with `toHaveBeenCalledWith`

src/actions/tests/suppressions.test.js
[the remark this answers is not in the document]
bkemper: [Argument Matchers](https://github.com/SparkPost/2web2ui/blob/master/docs/jest.md#argument-matchers) are good for matching objects with `toHaveBeenCalledWith`.

src/pages/suppressions/components/Results.js
bkemper: Ahhh, I missed [his comment](https://github.com/SparkPost/2web2ui/pull/234#discussion_r160978170).

src/reducers/suppressions.js
```diff
@@ -37,6 +39,19 @@ export default (state = initialState, action) => {
       }
     }
 
+    case 'DELETE_SUPPRESSION_PENDING':
+      return { ...state, deleting: true };
+
+    case 'DELETE_SUPPRESSION_SUCCESS':
```
bkemper: Should we add some test coverage for this?

PR #235, opened 2018-01-10
FAD-5593 Report: Engagement

bkemper:
_Refer to [FAD-5593](https://jira.int.messagesystems.com/browse/FAD-5593) for AC_

**Before**

screen shot 2018-01-10 at 2 42 43 pm

**Design**

engagement

**After**

screen shot 2018-01-17 at 2 44 01 pm

**Test Scenarios**

1. With data (happy path, see above)
1. [With no data](https://github.com/SparkPost/2web2ui/pull/235#issuecomment-357344959)
1. With filters
1. With share link (i.e. http://localhost:3100/reports/engagement?range=day)

**Remaining Work**

- [x] Initial Page
- [x] Bar chart (targeted, accepted, unique confirmed opens, unique clicks)
- [x] Filters

### DIFF
--- a/src/actions/reportFilters.js
+++ b/src/actions/reportFilters.js
@@ -8,6 +8,7 @@ import {
 import { listTemplates } from './templates';
 import { list as listSubaccounts } from './subaccounts';
 import { list as listSendingDomains } from './sendingDomains';
+import { getRelativeDates } from 'src/helpers/date';
 import { getQueryFromOptions } from 'src/helpers/metrics';
 
 // array of all lists that need to be re-filtered when time changes
@@ -50,6 +51,13 @@ export function addFilter(payload) {
   };
 }
 
+export function addFilters(payload) {
+  return {
+    type: 'ADD_FILTERS',
+    payload
+  };
+}
+
 export function removeFilter(payload) {
   return {
     type: 'REMOVE_FILTER',
@@ -62,5 +70,14 @@ export function refreshReportRange(options) {
   return {
     type: 'REFRESH_REPORT_RANGE',
     payload: options
+  };
+}
+
+export function refreshRelativeRange(options = {}) {
+  const range = { ...options, ...getRelativeDates(options.relativeRange) };
+
+  return {
+    type: 'REFRESH_REPORT_RANGE',
+    payload: range
   };
 }
--- a/src/actions/reportFilters.js
+++ b/src/actions/reportFilters.js
@@ -44,13 +44,6 @@ export function maybeRefreshTypeaheadCache(dispatch, options, updates = {}) {
   }
 }
 
-export function addFilter(payload) {
-  return {
-    type: 'ADD_FILTER',
-    payload
-  };
-}
-
 export function addFilters(payload) {
   return {
     type: 'ADD_FILTERS',
--- a/src/pages/reports/engagement/EngagementPage.js
+++ b/src/pages/reports/engagement/EngagementPage.js
@@ -1,36 +1,52 @@
-import React from 'react';
+import _ from 'lodash';
+import React, { Component } from 'react';
 import { connect } from 'react-redux';
 import { withRouter } from 'react-router-dom';
 import { Page } from '@sparkpost/matchbox';
 
 import { getChartData } from 'src/actions/engagementReport';
+import { showAlert } from 'src/actions/globalAlert';
 import EngagementChart from './components/EngagementChart';
 import EngagementFilters from './components/EngagementFilters';
 import EngagementSummary from './components/EngagementSummary';
 
-export function EngagementPage({ chart, getChartData }) {
-  return (
-    <Page title='Engagement Report'>
-      <EngagementFilters shareDisabled={chart.loading} onLoad={getChartData} />
-      <EngagementSummary
-        clicks={chart.data.count_unique_clicked_approx}
-        loading={chart.loading}
-        targeted={chart.data.count_targeted}
-      />
-      <EngagementChart
-        accepted={chart.data.count_accepted}
-        clicks={chart.data.count_unique_clicked_approx}
-        loading={chart.loading}
-        opens={chart.data.count_unique_confirmed_opened_approx}
-        targeted={chart.data.count_targeted}
-      />
-    </Page>
-  );
+export class EngagementPage extends Component {
+  onLoad = () => {
+    this.props.getChartData().catch((error) => {
+      this.props.showAlert({
+        details: _.get(error, 'response.data.errors[0].message'),
+        message: 'Unable to load engagement report data.',
+        type: 'error'
+      });
+    });
+  }
+
+  render() {
+    const { chart } = this.props;
+
+    return (
+      <Page title='Engagement Report'>
+        <EngagementFilters shareDisabled={chart.loading} onLoad={this.onLoad} />
+        <EngagementSummary
+          clicks={chart.data.count_unique_clicked_approx}
+          loading={chart.loading}
+          targeted={chart.data.count_targeted}
+        />
+        <EngagementChart
+          accepted={chart.data.count_accepted}
+          clicks={chart.data.count_unique_clicked_approx}
+          loading={chart.loading}
+          opens={chart.data.count_unique_confirmed_opened_approx}
+          targeted={chart.data.count_targeted}
+        />
+      </Page>
+    );
+  }
 }
 
 const mapStateToProps = (state, props) => ({
   chart: state.engagementReport.chart
 });
-const mapDispatchToProps = { getChartData };
+const mapDispatchToProps = { getChartData, showAlert };
 
 export default withRouter(connect(mapStateToProps, mapDispatchToProps)(EngagementPage));
--- a/src/pages/reports/engagement/EngagementPage.js
+++ b/src/pages/reports/engagement/EngagementPage.js
@@ -15,7 +15,7 @@ export class EngagementPage extends Component {
     this.props.getChartData().catch((error) => {
       this.props.showAlert({
         details: _.get(error, 'response.data.errors[0].message'),
-        message: 'Unable to load engagement report data.',
+        message: 'Unable to load engagement data.',
         type: 'error'
       });
     });
--- a/src/pages/reports/engagement/EngagementPage.js
+++ b/src/pages/reports/engagement/EngagementPage.js
@@ -8,20 +8,12 @@ import PanelLoading from 'src/components/panelLoading/PanelLoading';
 import EngagementChart from './components/EngagementChart';
 import EngagementFilters from './components/EngagementFilters';
 
-// Experimental alternative to inline JSX conditions
-// @see https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator
-function If({ children, condition }) {
-  return condition ? children : null;
-}
-
 export function EngagementPage({ chart, getChartData }) {
   return (
     <Page title='Engagement Report'>
       <EngagementFilters disabled={chart.loading} onLoad={getChartData} />
-      <If condition={chart.loading}>
-        <PanelLoading />
-      </If>
-      <If condition={!chart.loading}>
+      {chart.loading && <PanelLoading />}
+      {!chart.loading && (
         <Panel sectioned>
           <EngagementChart
             accepted={chart.data.count_accepted}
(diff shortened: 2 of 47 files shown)


bkemper:
- [ ] Why is Rechart's `ResponsiveContainer` not very responsive if I expand and contract the width of my browser?

screen shot 2018-01-10 at 4 00 11 pm

bkemper:
- [ ] Use [CustomSharpeBarChart](http://recharts.org/#/en-US/examples/CustomShapeBarChart) to round the top corners of each bar.

bkemper:
This is what is expected for brand new accounts or if the filters return no results.

screen shot 2018-01-12 at 3 24 38 pm

bkemper:
This is ready for review.

src/pages/reports/engagement/EngagementPage.js
```diff
@@ -0,0 +1,35 @@
...
+import { Page, Panel } from '@sparkpost/matchbox';
+import { getChartData } from 'src/actions/engagementReport';
+import PanelLoading from 'src/components/panelLoading/PanelLoading';
+import EngagementChart from './components/EngagementChart';
+import EngagementFilters from './components/EngagementFilters';
+
+export function EngagementPage({ chart, getChartData }) {
+  return (
+    <Page title='Engagement Report'>
+      <EngagementFilters disabled={chart.loading} onLoad={getChartData} />
```
rajumsys: To make it consistent with other pages, `disabled` should be renamed to `shareDisabled`.

src/pages/reports/engagement/EngagementPage.js
```diff
@@ -0,0 +1,35 @@
...
+      {chart.loading && <PanelLoading />}
```
rajumsys: this is fine but you could also just `return <PanelLoading />` to avoid `!chart.loading` in next line.

src/pages/reports/engagement/EngagementPage.js
```diff
@@ -0,0 +1,35 @@
...
+      {!chart.loading && (
+        <Panel sectioned>
```
rajumsys: seems other report pages don't use `sectioned`. but i couldn't catch any visual differences in either way!

bkemper: I took a look at the matchbox code.  It is the same as doing...

```
<Panel>
  <Panel.Section>
    //...
  </Panel.Section>
</Panel>
```

src/pages/reports/engagement/components/async/EngagementChart.js
```diff
@@ -0,0 +1,43 @@
...
+    { label: 'Unique Confirmed Opens', value: opens },
+    { label: 'Unique Clicks', value: clicks }
+  ];
+  const completionPercentage = formatPercent(clicks / targeted * 100 || 0);
+
+  return (
+    <div className={styles.EngagementChart}>
+      <div className={styles.EngagementChartCompletionRate}>
+        <span>{completionPercentage}</span>
+        <label>Completion Rate</label>
```
rajumsys: Would be great if we could add little more explanation of what it means. Even a tooltip will be helpful!

src/pages/reports/engagement/components/EngagementChart.js
```diff
@@ -0,0 +1,3 @@
+import asyncComponent from 'src/components/asyncComponent/asyncComponent';
+
+export default asyncComponent(() => import('./async/EngagementChart'));
```
rajumsys: would love to understand well how it works. i'll ping.

src/pages/reports/engagement/EngagementPage.js
rajumsys: those pages not not suing `Panel.Section` either I guess.

src/pages/reports/engagement/components/EngagementFilters.js
```diff
@@ -0,0 +1,67 @@
...
+
+  state = {
+    open: false
+  }
+
+  componentDidMount() {
+    const { filters = [], options: range } = parseSearch(this.props.location.search);
+
+    // hydrate store with filters and range from query string before onLoad
+    this.props.addFilters(filters);
```
rajumsys: these are already (globally) taken care, aren't they? why we've to do it here but not in other reporting like bounce, delay etc.

src/pages/reports/engagement/components/EngagementChart.js
bkemper: I would be more than happy to explain.  Ping me in the morning and I'll share what I learned.

src/pages/reports/engagement/components/async/EngagementChart.js
bkemper: I guess I should just move the rate to a top-level metric like Bob's.  Do you agree?

screen shot 2018-01-16 at 8 10 46 pm

src/pages/reports/engagement/EngagementPage.js
bkemper: I don't believe this could be done inline.  Wouldn't I have to pull it into a `renderX` function to apply that logic?

src/pages/reports/engagement/EngagementPage.js
bkemper: Addressed... 9e7cdab2

src/pages/reports/engagement/components/async/EngagementChart.js
rajumsys: Yeah, Probably will be good. Let's test.

src/pages/reports/engagement/EngagementPage.js
bkemper: Addressed... eb809d64

src/pages/reports/engagement/components/async/EngagementChart.js
bkemper: Addressed... eb809d64

bkemper:
@rajumsys This is ready for final review.

bkemper:
Handle loading data errors:

screen shot 2018-01-17 at 3 27 16 pm

src/actions/reportFilters.js
```diff
@@ -64,3 +72,12 @@ export function refreshReportRange(options) {
     payload: options
   };
 }
+
+export function refreshRelativeRange(options = {}) {
+  const range = { ...options, ...getRelativeDates(options.relativeRange) };
+
+  return {
```
rajumsys: you can probably just call `refreshReportRange(range)` here.

bkemper: I needed to be able to handle a relative range (e.g. `options = { relativeRange: 'day' }`) which will produce `range = { to: Date, from: Date, relativeRange: 'day' }` or a custom range that has all three values.

src/pages/reports/engagement/EngagementPage.js
```diff
@@ -1,36 +1,52 @@
...
-        targeted={chart.data.count_targeted}
-      />
-    </Page>
-  );
+export class EngagementPage extends Component {
+  onLoad = () => {
+    this.props.getChartData().catch((error) => {
+      this.props.showAlert({
+        details: _.get(error, 'response.data.errors[0].message'),
+        message: 'Unable to load engagement report data.',
```
rajumsys: `...engagement report` or `engagement data` should be enough, IMO. both words seems redundant to me

src/pages/reports/engagement/components/EngagementSummary.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { connect } from 'react-redux';
+
+import { PanelLoading } from 'src/components';
+import { formatFullNumber } from 'src/helpers/units';
+import MetricsSummary from '../../components/MetricsSummary';
+
+export function EngagementSummary({ clicks = 0, filters, loading = true, targeted = 0 }) {
+  if (loading) {
```
rajumsys: these loaders don't look good. 
screen shot 2018-01-17 at 5 17 39 pm
screen shot 2018-01-17 at 5 17 40 pm

src/actions/reportFilters.js
rajumsys: you can call `getRelativeDates` from your component if you get rid of `refreshRelativeRange` completely. if not, `range` in this method already has `from, to`. so `refreshReportRange(range)` should just work.

src/pages/reports/engagement/components/EngagementSummary.js
bkemper: So, this is caused by the [`ChunkLoading`](https://github.com/SparkPost/2web2ui/blob/3fe26f5139b840cb3317d854584ea02cde371c37/src/components/asyncComponent/ChunkLoading.js#L7).  I think we should just remove it.

jasonrhodes: This makes sense to me overall, maybe more like this though:
```js

export default function asyncComponent(importComponent, loadableOptions) {
  const LoadableComponent = loadable({
    loader: importComponent,
    loading: ChunkLoading,
    ...loadableOptions
  });

  return LoadableComponent;
}
```

My only concern here is that we'll have to do that in lots of places and then lose the ability to add any abstracted stuff to that common `ChunkLoading` component ... so the other option is to pass through "loading props" that get passed to `ChunkLoading` ...

src/pages/reports/engagement/EngagementPage.js
bkemper: Addressed... 59e3639d

src/pages/reports/engagement/components/EngagementSummary.js
bkemper: Addressed... d016cc90

src/actions/reportFilters.js
bkemper: This would be the fix to replace `refreshRelativeRange` with `refreshReportRange` in `EngagmentFilters` (not including testing). 

```
/src/pages/reports/engagement/components/EngagementFilters.js
@@ -3,6 +3,7 @@ import { connect } from 'react-redux';
 import { withRouter } from 'react-router-dom';

 import * as reportFilterActions from 'src/actions/reportFilters';
+import { getRelativeDates } from 'src/helpers/date';
 import { getFilterSearchOptions, parseSearch } from 'src/helpers/reports';
 import Filters from '../../components/Filters';

 import ShareModal from '../../components/ShareModal';
@@ -21,7 +22,10 @@ export class EngagementFilters extends Component {

     // hydrate store with filters and range from query string before onLoad
     this.props.addFilters(filters);
-    this.props.refreshRelativeRange(range);
+    this.props.refreshReportRangeq({
+      ...range,
+      ...getRelativeDates(range.relativeRange)
+    });
     this.props.onLoad();
     this.props.refreshTypeaheadCache(); // this should wait until after onLoad
   }
@@ -31,7 +35,10 @@ export class EngagementFilters extends Component {
   // filters.  Therefore, this handler only has to update the store with the next date range
   // then call onLoad to update the chart/table data.
   onFilterChange = (nextRange) => {
-    this.props.refreshRelativeRange(nextRange);
+    this.props.refreshReportRange({
+      ...nextRange,
+      ...getRelativeDates(nextRange.relativeRange)
+    });
     this.props.onLoad();
   }
```

I know it is duplication to have both `refreshReportRange` and `refreshRelativeRange`.  Hopefully, in a refactor of all this filter code someone can pick the winner or find a better way.  For now, I want to keep `refreshRelativeRange` because it is easier to test and already tested.

bkemper:
@rajumsys Regarding our offline conversation about duplication with `addFilter` and `addFilters`, I added the `addFilters` action creator to improve the hydration of filters from the request URL.  Here is the usage of both.

```
// src/pages/reports/engagement/components/EngagementFilters.js
componentDidMount() {
  const { filters = [], options: range } = parseSearch(this.props.location.search);
  this.props.addFilters(filters);
  // ...
}
```

Or, ...

```
// src/pages/reports/bounce/BouncePage.js
componentDidMount() {
  this.handleRefresh(this.parseSearch());
  // ...
}

parseSearch() {
  const { options, filters } = parseSearch(this.props.location.search);

  if (filters) {
    filters.forEach(this.props.addFilter);
  }

  // ...
}
```

The reason for introducing `addFilters` is to only update the store once and to avoid x number of rerenders of the `EngagementPage`.

bkemper:
Rebased branch.

bkemper:
@rajumsys Per your request, I removed `addFilter`.  This is ready for final review.

PR #237, opened 2018-01-11
FAD-5912 Make TableCollection sortable

rajumsys:
This PR allows sorting in TableCollection. This PR includes example usage of sorting for suppression list and bounce reports. 

Doc:
 - Pass table configuration to table collection. For a column to be sortable, pass `sortKey` to it. The value should be a key of object. But it can be anything else for manual handling in sort callback. Example:
```
const columns = [
  { label: 'Reason', width: '45%', sortKey: 'reason' },
  { label: 'Domain', sortKey: 'domain' },
  'actions'
];
```
 Note: last column won't be sortable as it doesn't have `sortKey` 
- Add `sortColumn` and `sortDirection` to state in the component using TableCollection. Use the value for default sorting column and direction. `sortColumn` will be similar  as `sortKey` above.
- Pass a callback to TableCollection using `onSort` arg. You can re-use `handleSortChange` from `src/helpers/sort` but it must bound to current component (`handleSortChange.bind(this)`). 
- Pass `sortColumn` and `sortDirection` to TableCollection. 
- Re-use or implement do your own implementation of`getSortedCollection`  to sort collection before feeding it to TableCollection. 

Todo:
- [x]  Link's color for sortable columns
- [x]  Link's color for non-sortable columns
- [x]  Sorting direction icon
cc @jonambas

### DIFF
--- a/src/components/collection/TableCollection.js
+++ b/src/components/collection/TableCollection.js
@@ -44,7 +44,7 @@ class TableCollection extends Component {
         bodyWrapper={TableBody}
         rowComponent={TableRow}
         { ...this.props}
-        rows={getSortedCollection(rows, sortColumn, sortDirection)}
+        rows={sortColumn ? getSortedCollection(rows, sortColumn, sortDirection) : rows }
       />
     );
   }
--- a/src/components/collection/TableCollection.js
+++ b/src/components/collection/TableCollection.js
@@ -1,28 +1,50 @@
-import React from 'react';
+import React, { Component } from 'react';
 import { Panel, Table } from '@sparkpost/matchbox';
 import Collection from './Collection';
 import TableHeader from './TableHeader';
+import { getSortedCollection } from 'src/helpers/sort';
 
 const TableWrapper = (props) => <Panel><Table>{props.children}</Table></Panel>;
 
 const TableBody = (props) => <tbody>{props.children}</tbody>;
 
-const TableCollection = (props) => {
-  const { rowComponent, headerComponent, columns, getRowData, onSort, sortColumn, sortDirection } = props;
-  const HeaderComponent = headerComponent ? headerComponent : () => <TableHeader columns={columns} onSort={onSort} sortColumn={sortColumn} sortDirection={sortDirection}/>;
-  const TableRow = rowComponent
-    ? rowComponent
-    : (props) => <Table.Row rowData={getRowData(props)} />;
+class TableCollection extends Component {
+  state = {
+    sortColumn: null,
+    sortDirection: 'asc'
+  }
 
-  return (
-    <Collection
-      outerWrapper={TableWrapper}
-      headerComponent={HeaderComponent}
-      bodyWrapper={TableBody}
-      rowComponent={TableRow}
-      {...props}
-    />
-  );
-};
+  componentDidMount() {
+    const { sortColumn, sortDirection } = this.props;
+    const { sortColumn: stateSortColumn, sortDirection: stateSortDirection } = this.state;
+    if (sortColumn || sortDirection) {
+      this.setState({ sortColumn: sortColumn || stateSortColumn, sortDirection: sortDirection || stateSortDirection });
+    }
+  }
+
+  handleSortChange = (column, direction) => {
+    this.setState({ sortColumn: column, sortDirection: direction });
+  }
+
+  render() {
+    const { rowComponent, headerComponent, columns, getRowData, rows } = this.props;
+    const { sortColumn, sortDirection } = this.state;
+
+    const HeaderComponent = headerComponent ? headerComponent : () => <TableHeader columns={columns} onSort={this.handleSortChange} sortColumn={sortColumn} sortDirection={sortDirection}/>;
+    const TableRow = rowComponent
+      ? rowComponent
+      : (props) => <Table.Row rowData={getRowData(props)} />;
+
+    return (
+      <Collection
+        outerWrapper={TableWrapper}
+        headerComponent={HeaderComponent}
+        bodyWrapper={TableBody}
+        rowComponent={TableRow}
+        {...{ ...this.props, rows: getSortedCollection(rows, sortColumn, sortDirection) }}
+      />
+    );
+  }
+}
 
 export default TableCollection;
--- a/src/components/collection/TableCollection.js
+++ b/src/components/collection/TableCollection.js
@@ -11,15 +11,17 @@ const TableBody = (props) => <tbody>{props.children}</tbody>;
 class TableCollection extends Component {
   state = {
     sortColumn: null,
-    sortDirection: 'asc'
+    sortDirection: null
+  }
+
+  static defaultProps = {
+    defaultSortColumn: null,
+    defaultSortDirection: 'asc'
   }
 
   componentDidMount() {
-    const { sortColumn, sortDirection } = this.props;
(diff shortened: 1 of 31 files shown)


src/helpers/sort.js
```diff
@@ -0,0 +1,14 @@
...
+  return _.orderBy(collection, sortColumn, sortDirection);
+}
+
+/**
+ * Handles sorting. It sets states. So this function must be invoked with the context of the component (bind `this` to the component being used)
+ * @param {*} column column name like recipient
+ * @param {*} direction direction of sorting like asc/desc
+ */
+export function handleSortChange(column, direction) {
+  this.setState({ sortColumn: column, sortDirection: direction }, this.forceUpdate);
```
jonambas: Is `forceUpdate` required because `setState` is being called outside the component? `render` should automatically run if state or props change.

Also this func is so simple, does it really need to be abstracted?

src/pages/reports/bounce/BouncePage.js
```diff
@@ -1,3 +1,5 @@
+/* eslint max-lines: ["error", 175] */
```
jonambas: Hopefully, can remove this line

src/components/collection/TableHeader.js
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    this.props.onSort(column, direction);
+  }
+
+  renderSortCell = (item) => {
+    const { label, sortKey } = item;
+    const { sortColumn, sortDirection = 'asc' } = this.props;
+
+    const icon = sortDirection === 'asc' ? <Icon name='CaretUp'/> : <Icon name='CaretDown'/>;
```
jonambas: Will see if I can create a better Icon. I might just try to style this myself, will let you know.

src/components/collection/TableHeader.js
```diff
@@ -0,0 +1,50 @@
...
+  render() {
+    const { columns } = this.props;
+
+    const cells = columns.map((item, i) => {
+      if (typeof item === 'string' || item === null) {
+        return <Table.HeaderCell key={`column ${i}: ${item}`}>{item}</Table.HeaderCell>;
+      }
+      const { label, sortKey, ...rest } = item;
+
+      return (<Table.HeaderCell key={label} {...rest}>{ this.renderSortCell(item) }</Table.HeaderCell>);
```
jonambas: outer `( )` not needed if on one line

src/pages/suppressions/components/Results.js
```diff
@@ -1,12 +1,17 @@
 import React, { Component } from 'react';
```
jonambas: This looks simple enough to not need its own table component

src/helpers/sort.js
rajumsys: the abstraction is default. otherwise, you'll have to copy same thing across all components. however, you can implement one in the component itself if it needs something else.

jasonrhodes: Having a helper call this.setState seems like it's probably a bad pattern. Can you explain why this can't be done in the TableCollection?

jasonrhodes: The "get visible rows" logic is all handled inside the Collection component -- can we do the sorting before we slice visible rows?

jonambas: +1 on setState outside the component, feels weird. Also requires component to match the keys.

rajumsys: yeah, it does but did for a reason. i'll remove it and put inside the component.

rajumsys:
@jonambas for your context
@jasonrhodes and I discussed this in DM. So, contrary to my initial understanding, your suggestion of column specific (optional) callback will actually work even for my (may be in future) use-case. So, here's what I'll do
- move default sorting to TableCollection
- This will probably allow us not to set any state in Parent Components. Meaning, we won't need to do separate component that you suggested just for tables, specially in report pages. 
- get rid of "that" helper method!

src/components/collection/TableHeader.js
rajumsys: seems you've already removed it. however, remove other extraneous one

src/pages/reports/bounce/BouncePage.js
rajumsys: why? we need it but different value (200; par another PR)

rajumsys:
@jonambas i've updated this PR addressing all feedbacks.

src/pages/reports/bounce/BouncePage.js
jasonrhodes: the "why" here in general is because we don't really want to use this eslint hack unless we are doing a quick bugfix and need to move super fast, or the file really really needs to be larger than 150 lines (rarely the case).

rajumsys: it was 175 and changed to 200 in  https://github.com/SparkPost/2web2ui/blame/master/src/pages/reports/bounce/BouncePage.js

src/components/collection/TableCollection.js
```diff
@@ -1,43 +1,50 @@
...
-    />
-  );
-};
+    return (
+      <Collection
+        outerWrapper={TableWrapper}
+        headerComponent={HeaderComponent}
+        bodyWrapper={TableBody}
+        rowComponent={TableRow}
+        {...{ ...this.props, rows: getSortedCollection(rows, sortColumn, sortDirection) }}
```
jonambas: Whats going on here?

Could you just do this? if rows is placed after it'll overwrite `this.props.rows`

```js
{...this.props}
rows={getSortedCollection(rows, sortColumn, sortDirection)}
```

src/pages/reports/bounce/BouncePage.js
```diff
@@ -106,6 +113,8 @@ export class BouncePage extends Component {
       columns={columns}
       rows={reasons}
       getRowData={this.getRowData}
+      sortColumn='count_bounce'
+      sortDirection='desc'
```
jonambas: Just so these props are a little more clear, could we use `defaultSortDirection` and `defaultSortKey`?

src/components/collection/TableCollection.js
```diff
@@ -1,43 +1,50 @@
...
+  state = {
+    sortColumn: null,
+    sortDirection: 'asc'
+  }
+
+  componentDidMount() {
+    const { sortColumn, sortDirection } = this.props;
+    const { sortColumn: stateSortColumn, sortDirection: stateSortDirection } = this.state;
+    if (sortColumn || sortDirection) {
+      this.setState({ sortColumn: sortColumn || stateSortColumn, sortDirection: sortDirection || stateSortDirection });
```
jonambas: I don't think the `|| stateSortXXX` is necessary. Initial state is already set at this point, and we can avoid an additional render

```js
if (defaultSortKey) {
  this.setState({ sortColumn: defaultSortKey })
}

if (defaultSortDirection) {
  this.setState({ sortDirection: defaultSortDirection })
}
```

rajumsys: Yeah, your example will work too. But because `setState` is async, I don't know when it'll trigger a re-render. I was thinking it may cause extra render in some cases when both conditions are true and react immediately goes for rendering after first `setState`.

src/components/collection/TableCollection.js
jonambas: Yeah same thing but, to me at least, its more understandable. The double spread looks weird haha

PR #240, opened 2018-01-17
FAD-5885 fixed up ip pools

jasonrhodes:
Was trying to figure out if we needed to do this and ended up accidentally doing it.


src/selectors/ipPools.js
```diff
@@ -3,7 +3,23 @@ import { createSelector } from 'reselect';
...
 );
+
+export const getOrderedIpPools = createSelector(
+  [getIpPools, getDefaultPool, getNonDefaultIpPools],
+  (initialList, defaultPool, others) => {
+    if (!defaultPool) {
+      return initialList;
+    }
+    const sorted = others.sort((a, b) => b.ips.length - a.ips.length);
+    return [ defaultPool, ...sorted ];
```
bizob2828: array spreading FTW!

src/selectors/ipPools.js
```diff
@@ -3,7 +3,23 @@ import { createSelector } from 'reselect';
...
   (ipPools) => ipPools.filter(({ id }) => id !== DEFAULT)
 );
+
+export const getOrderedIpPools = createSelector(
+  [getIpPools, getDefaultPool, getNonDefaultIpPools],
+  (initialList, defaultPool, others) => {
+    if (!defaultPool) {
+      return initialList;
+    }
+    const sorted = others.sort((a, b) => b.ips.length - a.ips.length);
```
bizob2828: we lost alpha sort for same # of ips.

bizob2828: also not sure if people want the ones with most ip's on top

jasonrhodes: I suppose we'll have table collection sorting everywhere soon -- I think by and large the majority case is only 1-2 pools, since most SPC customers can only have max 4 IPs.

src/pages/ipPools/ListPage.js
```diff
@@ -2,21 +2,23 @@ import React, { Component } from 'react';
...
-  return [nameLink, id, ips.length.toString()];
-};
 
 export class IpPoolsList extends Component {
 
   componentDidMount() {
     this.props.listPools();
   }
 
+  getRowData = ({ id, name, ips }) => {
```
jasonrhodes: Did you just move this so you could test it? Could also just export it at the top of the file—it doesn't need access to `this` so doesn't really need to be a class method.

PR #242, opened 2018-01-18
FAD-6046 Redirect enterprise tenants to the summary report on login

bizob2828:
TODO:

- [x] create A-E review to add splashPage to the tenant config and set it to /dashboard for spc


src/config/routes.js
```diff
@@ -140,19 +141,19 @@ const routes = [
   {
     path: '/account/subaccounts',
     component: subaccounts.ListPage,
-    condition: hasGrants('subaccount/manage'),
+    condition: hasGrants('subaccount/manage', 'api_keys/manage', 'sending_domains/manage'),
```
rajumsys: what's going on here?

rajumsys:
This looks good so far.
- Why more capabilities added to some of route (in `hasGrants`)? 
- If we're on dashboard, reloading (after updating `splashPage`) on redirect anywhere. This is not a problem as it won't happen in real life; just wanted to mention as it may happen during test. 
- are you going to do those A-E thing in this ticket? If so, let me know when that's ready for review.

src/config/routes.js
bizob2828: these were missing. you need all 3 grants for this page to properly load

bizob2828:
Added A-D/A-E changes here https://fisheye.int.messagesystems.com/cru/APP-3103

PR #244, opened 2018-01-19
FAD-6026 Created sending domains edit page, action, and reducer

gwalls:


src/pages/sendingDomains/EditPage.js
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  loadDependantData = () => {
+    this.props.getDomain(this.id);
+  };
+
+  renderError() {
+    return <ApiErrorBanner
+      errorDetails={this.props.getError.message}
+      message="Sorry, we seem to have had some trouble loading your Sending Domain."
```
rajumsys: `...some troubles...`

src/pages/sendingDomains/EditPage.js
```diff
@@ -0,0 +1,64 @@
...
+
+export class EditPage extends Component {
+  constructor(props) {
+    super(props);
+
+    this.id = props.match.params.id;
+  }
+
+  componentDidMount() {
+    this.props.getDomain(this.id);
```
rajumsys: we can probably just call `loadDependentData` here too.

src/pages/sendingDomains/EditPage.js
gwalls: Singular `trouble` was the term used by IP Pools. Either is valid, but I think we should be consistent in our error messages.

src/pages/sendingDomains/EditPage.js
```diff
@@ -0,0 +1,64 @@
...
+  content: 'Sending Domains',
+  Component: Link,
+  to: '/account/sending-domains'
+};
+
+export class EditPage extends Component {
+  constructor(props) {
+    super(props);
+
+    this.id = props.match.params.id;
```
jasonrhodes: We should avoid setting values on `this` -- we can remove this constructor and then in `loadDependentData()` just do:

```js
this.props.getDomain(this.props.match.params.id);
```

If that prop value were to ever change, setting it once in the constructor will cause a bug, which is why we avoid that pattern. The only things we usually set on this are methods or static values like `state` to initialize local state and the rare constant like `DEBOUNCE_INTERVAL` etc.

If you copy/pasted this from somewhere else let me know and we can refactor that too. :)

src/pages/sendingDomains/EditPage.js
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  componentDidMount() {
+    this.props.getDomain(this.id);
+  }
+
+  loadDependantData = () => {
```
jasonrhodes: `loadDependentData`

although maybe something more semantic would be fine since all it does is load the current domain?

src/pages/sendingDomains/EditPage.js
jasonrhodes: singular `trouble` would be correct/standard here.

src/pages/sendingDomains/EditPage.js
```diff
@@ -0,0 +1,64 @@
...
+        breadcrumbAction={breadcrumbAction}
+      >
+        {getError ? this.renderError() : 'Coming soon'}
+
+      </Page>
+    );
+  }
+}
+
+const mapStateToProps = (state) => ({
```
jasonrhodes: Feel free to destructure for less repetition below, either:
```js
({ sendingDomains }) => ({
  domain: sendingDomains.domain,
  // etc
```
or even BETTER and WILDER
```js
({ sendingDomains: { domain, getError, getLoading }}) => ({
  domain,
  getError,
  getLoading
})
```

src/pages/sendingDomains/index.js
```diff
@@ -1,5 +1,7 @@
 import ListPage from './ListPage';
+import EditPage from './EditPage';
 
 export default {
-  ListPage
+  ListPage,
+  EditPage
 };
```
jasonrhodes: This is not your fault at all but we should try to prefer this syntax:
```js
export { default as ListPage } from './ListPage';
export { default as EditPage } from './EditPage';
```
A little more succinct and does named exports instead of a default export object, which tend to cause slightly unexpected results. 

I think @bizob2828 ran into some trouble trying to change this around in another file but I don't remember what the problem was. If you get test failures or can't load the pages anymore after changing, ping me and we can ask him what he ran into.

src/pages/sendingDomains/EditPage.js
gwalls: Understood. I'll change this here and in the IP Pools edit page where we copied this from.

src/pages/sendingDomains/index.js
jonambas: I think you'll have to rename since we have other "ListPage"s, ie `SendingDomainListPage`

jasonrhodes: The problem is we expect default exports in the `src/pages/index.js` file so using named exports here breaks that.

src/pages/sendingDomains/EditPage.js
rajumsys: oh, i wanted to tell him same but thought it might be too much for first PR, haha. 
fwiw, the wilder one is little wilder for me now (my processing of that is slow :p )

src/pages/sendingDomains/EditPage.js
rajumsys: I've seen this somewhere (react-native, IIRC) to be normal. but if that's the convention we want to stick, it's fine. @gwalls thanks for updating ip pools.

src/pages/sendingDomains/EditPage.js
rajumsys: yeah, that's fine. in that case, just get rid of `some`?

src/pages/sendingDomains/EditPage.js
jasonrhodes: @rajumsys yeah if you assign values to `this`, others will eventually update those values, not realizing that updating those values does nothing and is totally ignored by the react lifecycle. A pattern we should avoid.

src/pages/sendingDomains/EditPage.js
jasonrhodes: "Sorry, we seem to have had some trouble loading" is totally 👌

rajumsys: Seems correct https://english.stackexchange.com/a/152841. Taking back my suggestion :).

PR #245, opened 2018-01-19
FAD-5593 (follow-up) Change top level metric label

bkemper:


bkemper:
@rajumsys Take a look at this thread -- https://msys.slack.com/archives/C69L53989/p1516380607000509

PR #247, opened 2018-01-22
FAD-5906 finished wiring up batch status tab

bizob2828:


src/pages/webhooks/DetailsPage.js
```diff
@@ -68,31 +39,56 @@ class WebhooksDetails extends Component {
...
+    ];
+    const selectedTab = tabs.findIndex(({ to }) => location.pathname === to);
 
     /*
       Check .events to guard from the create page redirect,
       which sets id on the state but doesn't have the rest of the webhook
     */
-    const isLoading = (webhook.id !== webhookId) || !webhook.events;
-    if (isLoading) {
+    if ((webhook.id !== webhookId) || !webhook.events) {
```
jasonrhodes: This is a smell but I don't have time to fix it now, just saying I know it's here, will try to come back to it later. :)

jasonrhodes:
I tried to write a small TabRouteSet component to abstract the tabs + routes but it causes a small matchbox problem and I pulled the plug on it, punting it to another day.

src/pages/webhooks/DetailsPage.js
bizob2828: yea i just reduced the # of lines here.  i don't get this edge case

PR #251, opened 2018-01-24
FAD-6102 Add sorting capability to all tables

rajumsys:
**Notes:**
- Sorted Message Events table though it's not sorted in old UI. However, bother my eyes with its auto updating 'time ago'. should I not sort this table? 
- Refactored rejection and delay reports to reduce lines per file. 
- For casting subaccount_id in suppression, I feel like it's an API issue. unlike suppressions, subaccount_id is numeric for templates, api keys. So, not sure if we solve it here or log ticket for fixing API! Update: I've added capability to provide custom comparator and casted value there!
- I'll do another round of checking to see if any other tables are needed to be sortable. Update: Added api keys table


src/components/collection/TableHeader.js
```diff
@@ -39,7 +39,7 @@ export default class TableHeader extends Component {
       if (typeof item === 'string' || item === null) {
         return <Table.HeaderCell key={`column ${i}: ${item}`}>{item}</Table.HeaderCell>;
       }
-      const { label, sortKey, ...rest } = item;
+      const { label, sortKey, comparator, ...rest } = item;
```
rajumsys: fyi, `comparator` isn't used anywhere but otherwise `Table.HeaderCell` complains about being incorrect attr.

jonambas: 👍  This is the right way to handle it. React tries to add props as HTML attributes unless it is pulled out.

src/pages/reports/summary/components/Table.js
```diff
@@ -113,6 +115,8 @@ export class Table extends Component {
         defaultPerPage={10}
         rows={tableData}
         filterBox={{ show: false }}
+        defaultSortColumn='count_targeted'
```
jonambas: I dont think this will work. What happens if someone loads the page with different metrics in the url?
Wouldn't you want this to default to either the `groupBy` key or the first metric column key?

src/pages/suppressions/components/Results.js
```diff
@@ -90,7 +90,7 @@ export class Results extends Component {
     ];
 
     if (hasSubaccounts) {
-      columns.push({ label: 'Subaccount', width: '18%', sortKey: 'subaccount_id' });
+      columns.push({ label: 'Subaccount', width: '18%', sortKey: 'subaccount_id', comparator: (row) => parseInt(row.subaccount_id, 10) });
```
jonambas: I think we're going to need this func in a few places, might be good to place in a subaccount helper?

src/components/collection/docs/Collection.md
```diff
@@ -189,13 +189,10 @@ function MyTableCollectionComponent() {
       columns={[ 
         { label: 'Title', sortKey: 'title' },
         { label: 'Director', sortKey: 'director' }
-        { label: 'Released', sortKey: 'released' }
+        { label: 'Released', sortKey: 'year', comparator: (movie) => parseInt(movie.year, 10) }
```
jonambas: Semantically speaking, do you think this would be better as `computeSortValue`?

rajumsys: you mean property/key name? i'm fine with whatever makes obvious sense. i used `comparator` as I guess that's something lodash/underscore used to use.

src/pages/reports/summary/components/Table.js
rajumsys: I'll check again but I guess I've seen all those reports had this field

src/pages/suppressions/components/Results.js
rajumsys: IMO, we can do that when we see other usages.

jonambas:
Webhooks
- Missing `Name` sort

API Keys
- Do we really need to sort the short key?

Sending Domains
- Missing `Domain` name sort
- Could try sorting `Status` and `Subaccount` but those would need comparators

Subaccounts
- Missing `Name`, `ID` sort
- Could try sorting `Status` - `active` > `suspended` > `terminated` works alphabetically 😄 

IP Pools
- Missing `Name` `ID` `Number of IPs Assigned`

Summary Report
- Need to fix this custom table header with the sort icons (can do in another ticket)
screen shot 2018-01-26 at 1 35 35 pm

UX note: need to go through lists and adjust column widths to avoid the widths from changing on sort

jonambas:
You may want to rebase/merge develop, the Templates list page added the `SubaccountTag` component to the lists, so that sort may not work

src/pages/reports/delays/components/tests/DealysDataTable.test.js
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
```
jonambas: Typo in this file name `DealysDataTable.test.js`

src/pages/ipPools/ListPage.js
```diff
@@ -6,7 +6,12 @@ import { getOrderedIpPools } from 'src/selectors/ipPools';
 import { Loading, TableCollection, ApiErrorBanner } from 'src/components';
 import { Page } from '@sparkpost/matchbox';
 
-const columns = ['Name', 'ID', 'Number of IPs Assigned'];
+const columns = [
+  { label: 'Name', sortKey: 'name' },
+  { label: 'ID', sortKey: 'id' },
+  { label: 'Number of IPs Assigned', sortKey: (pool) => pool.ips.length }
```
jonambas: Nice!

src/pages/templates/ListPage.js
```diff
@@ -52,14 +52,14 @@ export default class ListPage extends Component {
     const { hasSubaccounts } = this.props;
 
     const columns = [
```
bizob2828: i don't see default sort.  in the old it was by the label

bizob2828:
Mentioned in slack but for posterity "i noticed when toggling sort on paginated table it doesn't reset or at least show the sorted stuff for that page"

src/pages/templates/ListPage.js
rajumsys: Good catch. fixing...

src/pages/reports/summary/components/Table.js
jasonrhodes: If you come to the summary report with query params you might not have that column. Do collections need a default sort beyond "how they were sorted when the data came back"?

jasonrhodes:
Just saying A+++ for removing `comparator` in favor of implicit function, very nice!

src/pages/reports/summary/components/Table.js
rajumsys: Yeah, I could finally realize issue here. That field was selected for me in either UI which confused me.

No, default sorting is optional but we kind of made it mandatory (by supplying everywhere, he he)

However, I've fixed this by making first metric as default sort. So whatever they select, it'll be sorted on first metric.

rajumsys:
> Just saying A+++ for removing comparator in favor of implicit function, very nice!


Yeah, that was very dumb implementation. Glad I realized it before too late.

jonambas:
Looks good

PR #252, opened 2018-01-24
FAD-5933 fixed series of auth flow bugs and related tests

jasonrhodes:
Things you should test here:

* Visit the auth page, should no longer see a failed request to GET /account in the console and network panel _before logging in_
screen shot 2018-01-24 at 3 12 52 pm

* Log in as a reporting user
  * Grab this same branch of the accusers API and run it locally, switching your local nginx settings so the UI hits the local version of the API)
  * Log out of the app, then visit http://localhost:3100/reports/summary - that should redirect you to auth
  * Log in as an existing reporting user (create one if you don't have one already)
  * App should log you in successfully and take you to the summary report, which should load fine
  * In network panel you should see a request for GET /account that does not have any `include` query param at all

* Log in as a regular user
  * Note: This step doesn't need the accusers API running locally, but won't hurt either way
  * Log out and back in as an admin user
  * Verify that GET /account is performed but without any include query parameter _for the first call_
  * Once you are redirected to the dashboard, a 2nd call to GET /account should be performed, this time with `?include=usage` in order to load the usage report
  * Usage report should load successfully, but display the "PanelLoading" view before it loads
  * You shouldn't see a flash of the suppression banner now that we are checking for strict `false` equality and defaulting to `null` instead of `false`
  * Completed items in the "Getting Started Checklist" should not link to anything and not have their descriptive text
  * New tutorial item should be present "Send an email in the past month"


src/components/usageReport/tests/__snapshots__/UsageReport.test.js.snap
```diff
@@ -1,10 +1,10 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`UsageReport Component render null: no props 1`] = `""`;
-
-exports[`UsageReport Component render null: only subscription 1`] = `""`;
-
-exports[`UsageReport Component render null: only usage 1`] = `""`;
+exports[`UsageReport Component should render null without props 1`] = `
```
jasonrhodes: Change to "should render loading"

bizob2828:
We need to fix redirecting to a page they can see I'm currently getting this

```Warning: You tried to redirect to the same route you're currently on: "/dashboard"```

**Comment by bizob2828** [comment]
Also the access control doesn't seem to be working.  I see all the nav items for things you shouldn't see, when I click i get the same warning as above because I think that's the default route

File: src/pages/dashboard/components/Tutorial.js [hunk 2]
```diff
@@ -61,6 +60,14 @@ export class Tutorial extends Component {
             <p>You'll need to create an API key to use our API or SMTP integration.</p>
           </TutorialItem>
         </Panel.Section>
+
+        <Panel.Section>
```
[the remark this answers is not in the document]
jasonrhodes: Haha I don't remember not wanting it (and don't think I care either way, happy to remove) we just hadn't removed the corresponding "hasSentThisMonth" prop so I was trying to fix. Can also easily remove both.

src/pages/dashboard/DashboardPage.js
```diff
@@ -51,7 +52,7 @@ export class DashboardPage extends Component {
   renderSuppressionBanner() {
     const { accountAgeInWeeks, hasSuppressions } = this.props;
 
-    if (accountAgeInWeeks > 1 && !hasSuppressions) {
+    if (accountAgeInWeeks > 1 && hasSuppressions === false) {
```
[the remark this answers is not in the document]
jasonrhodes: Otherwise you get a flash of this banner on dashboard load, because the "hasSuppressions" default value was false. Now I set default value to null, this way it only shows the banner after we've checked if you have suppressions.

src/components/support/Support.js
```diff
@@ -19,10 +13,6 @@ export class Support extends Component {
     showPanel: false
   };
 
-  componentWillMount() {
```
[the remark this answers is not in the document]
jasonrhodes: yep

src/pages/dashboard/components/Tutorial.js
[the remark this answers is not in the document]
bizob2828: keep it. i spoke to someone when i was fixing up that page and they said ditch it. i think it's ok

src/pages/dashboard/DashboardPage.js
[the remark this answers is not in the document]
bizob2828: oh ok

public/assets/tenant/production.js
```diff
@@ -2,5 +2,10 @@ window.SP = window.SP || {};
 window.SP.productionConfig = {
   apiBase: 'http://api.sparkpost.test/api/v1',
   gaTag: 'UA-111136819-2',
-  splashPage: '/dashboard'
+  splashPage: '/dashboard',
+  entryPoints: {
+    reporting: '/reports/summary',
+    admin: '/dashboard',
+    superuser: '/dashboard'
+  }
```
jasonrhodes: @bizob2828 this was my first instinct but based on what I said at the retro this morning, maybe I can make it even simpler and leave existing config alone

src/components/auth/ProtectedRoute.js
```diff
@@ -3,15 +3,14 @@ import { connect } from 'react-redux';
...
-import config from 'src/config';
 
 export class ProtectedRoute extends Component {
 
   renderComponent(propsFromRoute) {
     const { component: Component, condition } = this.props;
 
     return (
-      <AccessControl condition={condition} redirect={config.splashPage}>
+      <AccessControl condition={condition} redirect='/404'>
```
jasonrhodes: This brings up our 404 because /404 doesn't exist. Meta. I think that's better than trying to figure out a place to send them where they have access.

src/pages/auth/AuthPage.js
```diff
@@ -66,11 +68,14 @@ export class AuthPage extends Component {
...
+    const { ready, location, currentUser } = this.props;
     const { errorDescription, loggedIn } = this.props.auth;
     const { tfaEnabled } = this.props.tfa;
 
-    if (loggedIn) {
-      return <Redirect to={config.splashPage} />;
+    if (ready && loggedIn) {
+      const redirectAfterLogin = _.get(location, 'state.redirectAfterLogin');
+      const redirectTo = redirectAfterLogin || _.get(config, `entryPoints.${currentUser.access_level}`, config.splashPage);
+      return <Redirect to={redirectTo} />;
```
jasonrhodes: Realized it makes more sense to do all of this stuff here to get one more weird thing out of `<AuthenticationGate>`, since we only do this stuff if you are logging in via the Auth page itself.

src/pages/users/components/tests/__snapshots__/RoleSelect.test.js.snap
```diff
@@ -1,19 +1,13 @@
...
-<span>
-  Administrator
-</span>
-`;
-
-exports[`AccessSelect renders select 1`] = `
+exports[`Component: RoleSelect should render the select 1`] = `
 <Select
-  name="renders select"
+  name={null}
```
jasonrhodes: 🤔

src/selectors/tests/__snapshots__/navItems.test.js.snap
```diff
@@ -78,101 +78,11 @@ Array [
   },
   Object {
     "children": Array [
-      Object {
```
jasonrhodes: woops these shouldn't have been here and would've told us that nav filtering wasn't working

src/pages/users/ListPage.js
```diff
@@ -32,14 +33,15 @@ export class ListPage extends Component {
   // Do not allow current user to change their access/role or delete their account
   getRowData = (user) => [
```
bizob2828: i don't see tests for this

src/pages/users/components/tests/__snapshots__/RoleSelect.test.js.snap
bizob2828: you aren't specifying the name in the props

src/pages/users/components/tests/RoleSelect.test.js
```diff
@@ -0,0 +1,56 @@
...
+import { Select } from '@sparkpost/matchbox';
+import RoleSelect from '../RoleSelect';
+
+describe('Component: RoleSelect', () => {
+
+  let props;
+  let wrapper;
+
+  beforeEach(() => {
+    props = { value: 'admin' };
```
bizob2828: missing name, which is why that's null in the snapshot

src/selectors/tests/__snapshots__/navItems.test.js.snap
bizob2828: this test still seems f'ed.  the grants you are given should show more nav items right? also not sure why tracking domains is in here

bizob2828: i think tracking domains is here because the route has `/account/tracking-domains/` and the navItems has `/account/tracking-domains`

bizob2828:
the route conditions aren't 100% fixed.  I think it's because uri's between the route config and navigation aren't aligned.  

old shows:
 * profile
 * templates list
 * template edit but all fields are disabled

new shows:
 * tracking domains list(it shouldn't)

bizob2828:
by fixing the uri for tracking domains, most are showing. templates list isn't.  however the nav needs hard refresh after login for some reason

jasonrhodes:
>by fixing the uri for tracking domains, most are showing.

I don't know what that means. Do you mean fixing the access condition for tracking domains list page? And most of what are showing?

jasonrhodes:
>the route conditions aren't 100% fixed. I think it's because uri's between the route config and navigation aren't aligned.

Profile shows in the new app too, or at least it should. I don't think we have the ability to show list pages for things we don't want people to edit. Why would a reporting user need access to those things anyway?

bizob2828:
this is what i changed

```
diff --git a/src/config/routes.js b/src/config/routes.js
index aa9ec49d..8ff3b9cc 100644
--- a/src/config/routes.js
+++ b/src/config/routes.js
@@ -172,7 +172,7 @@ const routes = [
   {
     path: '/templates',
-    condition: composeConditions(hasGrants('templates/modify'), ({ account }) => account.status === 'active'),
+    condition: composeConditions(hasGrants('templates/view'), ({ account }) => account.status === 'active'),
     layout: App
   },
   {
@@ -267,7 +267,7 @@ const routes = [
     layout: App
   },
   {
-    path: '/account/tracking-domains/',
+    path: '/account/tracking-domains',
     component: trackingDomains.ListPage,
     condition: hasGrants('tracking_domains/view'),
     layout: App
~
```

it seems to fix the showing of things. however in old webui you could edit a template but all values on the edit form would be disabled but you could still test/preview

bizob2828:
there was a requirement that reporting users can view a template but not edit any fields.  it was so they could test/preview

jasonrhodes:
>however in old webui you could edit a template but all values on the edit form would be disabled but you could still test/preview

Don't think we can achieve that easily.

bizob2828:
i think deferring the edit of templates to a diff ticket is ok. we should prob not update the templates list route condition until we fix it all because right now it's weird

jasonrhodes:
>we should prob not update the templates list route condition until we fix it all because right now it's weird

Agreed, for now reporting users won't see templates. I'd love to revisit with product why they would need that ... otherwise we have to build grant-based checks into functionality per page rather than just for the page route, and if we do that I'd like to do it right which will take a little thinking/time.

src/components/auth/ProtectedRoute.js
jasonrhodes: Right, this just triggers that already existing component.

src/pages/users/CreatePage.js
```diff
@@ -1,47 +1,39 @@
...
     const { email, access } = values;
-    return inviteUser(email, access).then((res) => {
-      showAlert({
-        type: 'success',
-        message: `Invitation sent to ${email}`
-      });
-      history.push('/account/users');
-    })
+
+    return inviteUser(email, access)
```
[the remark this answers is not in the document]
jasonrhodes: :sparkles: prettier :sparkles:

src/components/auth/ProtectedRoute.js
bizob2828: oh right, the pageNotFound isn't a route. sorry I'm slow

src/selectors/tests/navItems.test.js
```diff
@@ -1,22 +1,111 @@
-import selectNavItems from '../navItems';
+import { prepareNavItems } from '../navItems';
+
+jest.mock('src/config/routes', () => ([
```
bizob2828: we could add another route with no condition to hit that path

PR #254, opened 2018-01-25
FAD-6100: Sending domains status section and delete

ewandennis:
### TODO
 - [x] Edit page layout
 - [x] Delete
 - [x] Status
 - [x] Share with subdomains
 - [x] Delete subaccount domain
 - [ ] Reconcile with other domains tickets
 - [x] API error to alert action conversion helper


src/pages/sendingDomains/EditPage.js
```diff
@@ -30,6 +71,20 @@ export class EditPage extends Component {
     />;
   }
 
+  renderSections() {
+    return <Panel sectioned>
```
rajumsys: @jonambas suggested we shouldn't use panel for each section. @gwalls and I are sharing a structure not made with panels for top level sections.

rajumsys: https://github.com/SparkPost/2web2ui/pull/255/files#diff-99042ebfcae80c577d42f72b7ac5aab9

src/actions/sendingDomains.js
```diff
@@ -15,7 +15,29 @@ export function get(id) {
     type: 'GET_SENDING_DOMAIN',
     meta: {
       method: 'GET',
+      url: `/sending-domains/${id}`,
+      id
+    }
+  });
+}
+
+export function remove(id) {
```
bizob2828: you need to pass x-msys-subaccount header if sending domain is for a subaccount

src/pages/sendingDomains/EditPage.js
```diff
@@ -30,29 +84,53 @@ export class EditPage extends Component {
...
+      domain={domain}
+      onShareDomainChange={this.shareDomainChange} />;
+  }
+
   render() {
-    const { getLoading, getError, match: { params: { id }}} = this.props;
+    const { domain, getError, match: { params: { id }}} = this.props;
 
-    if (getLoading) {
+    if (domain.id !== id) {
```
bizob2828: why did this change? shouldn't the getLoading prop be enough

src/pages/sendingDomains/components/DomainStatus.js
```diff
@@ -0,0 +1,35 @@
...
+import { Grid, Panel } from '@sparkpost/matchbox';
+
+import { resolveStatus, resolveReadyFor } from 'src/helpers/domains';
+
+import VerificationHelp from './VerificationHelp';
+import StatusDescription from './StatusDescription';
+import ShareWithSubaccounts from './ShareWithSubaccounts';
+
+export const DomainStatus = ({ domain, onShareDomainChange }) => {
+  const status = resolveStatus(domain.status);
```
bizob2828: i think i left this on @gwalls but shouldn't this stuff go into the selector so the redux store always has this data?

src/pages/sendingDomains/components/DomainStatus.js
```diff
@@ -0,0 +1,35 @@
...
+  const readyFor = resolveReadyFor(domain.status);
+
+  return <Grid>
+    <Grid.Column xs={12} md={4}>
+      <h1>Status</h1>
+      <VerificationHelp status={status} />
+    </Grid.Column>
+    <Grid.Column xs={12} md={8}>
+      <Panel accent>
+        <StatusDescription
```
bizob2828: wouldn't it be better to just pass in domain as 1 prop then destructure in there?

src/pages/sendingDomains/components/ReadyFor.js
```diff
@@ -27,7 +27,7 @@ const ReadyFor = ({ bounce, dkim, sending, bounceDefault }) => {
 
   return (
     <span className={styles.ReadyFor}>
-      <small>Ready For:</small>
+      { showLabel && <small>Ready For:</small> }
```
bizob2828: I don't see anywhere that showLabel is being passed, do we need this?

src/pages/sendingDomains/components/ShareWithSubaccounts.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+
+import { Panel, Grid, Toggle } from '@sparkpost/matchbox';
+
+import styles from './DomainStatus.module.scss';
+
+const ShareWithSubaccounts = ({ domain, onChange }) => {
```
bizob2828: I think this is copy pasta.  templates edit already had this toggle `src/pages/templates/components/SubaccountSection.js` look for ToggleBlock

src/helpers/tests/apiMessages.test.js
```diff
@@ -0,0 +1,37 @@
...
+    response: {
+      data: {
+        errors: [
+          { description: 'it all went baaad' }
+        ]
+      }
+    }
+  };
+
+  it('produces an error alert with details', () => {
```
bizob2828: i don't see a test for the else path where errrs is falsey or errs isn't an array or empty array

src/pages/sendingDomains/tests/domain.js
```diff
@@ -0,0 +1,28 @@
+export const domain = {
```
bizob2828: the keys have strings, you can remove that. also this is in coverage. little weird

src/pages/sendingDomains/EditPage.js
```diff
@@ -1,26 +1,80 @@
...
+      showAlert
+    } = this.props;
+
+    return deleteDomain(id, subaccountId)
+      .then(() => {
+        showAlert({
+          type: 'success',
+          message: `Domain ${id} deleted.`
+        });
+        this.afterDelete();
```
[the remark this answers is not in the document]
ewandennis: yup :(

src/pages/sendingDomains/EditPage.js
```diff
@@ -1,26 +1,80 @@
...
+      }).catch((err) => {
+        showAlert(apiResponseToAlert(err, 'Could not delete domain'));
+        this.afterDelete();
```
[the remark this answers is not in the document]
ewandennis: Yeah, the most common reason for failing to delete is someone deleted it from under you.

src/pages/sendingDomains/EditPage.js
ewandennis: `getLoading` starts `false` which isn't great when you're trying to detect loading. It wasn't an issue previously because we didn't depend on the loaded result until now. Now that we do, we need to detect non-yet-loaded.

More generally, I'm not 100% comfortable with throwing loading vars into redux to detect the loading edge. It requires complex tests in the component and the state of the actual loaded object can be used in the same way like ☝️ .

src/pages/sendingDomains/components/DomainStatus.js
ewandennis: Yeah, could do. Sounds like it needs coordinated across tickets though.

src/pages/sendingDomains/components/DomainStatus.js
```diff
@@ -0,0 +1,35 @@
...
+import ShareWithSubaccounts from './ShareWithSubaccounts';
+
+export const DomainStatus = ({ domain, onShareDomainChange }) => {
+  const status = resolveStatus(domain.status);
+  const readyFor = resolveReadyFor(domain.status);
+
+  return <Grid>
+    <Grid.Column xs={12} md={4}>
+      <h1>Status</h1>
+      <VerificationHelp status={status} />
```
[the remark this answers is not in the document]
ewandennis: It is redundant at some level. However, I have a separate concern here - not showing a big green tick when a setting is "good" isn't great UX IMO. To improve this (later), we'll want to review it across list and edit pages everywhere.

For the moment, I'm following the convent from the list page by _either_ saying unverified/pending/blocked _or_ ready-for-whatever.

src/pages/sendingDomains/EditPage.js
```diff
@@ -1,26 +1,80 @@
...
+      });
+  };
+
+  shareDomainChange = () => {
```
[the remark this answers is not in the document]
ewandennis: I left EditPage as the controlling component for a few reasons:
 - It becomes the only redux-connected component which makes it easier to reason about
 - All user interaction and responses (alerts etc) stay 1 place
 - All the subcomponents become just encapsulated markup, easy to understand and test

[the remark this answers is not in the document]
ewandennis: I imagine all the other sections have more involved event handling logic. For that reason, this might be better in `DomainStatus`.

src/pages/sendingDomains/components/DomainStatus.js
```diff
@@ -0,0 +1,35 @@
...
+    </Grid.Column>
+    <Grid.Column xs={12} md={8}>
+      <Panel accent>
```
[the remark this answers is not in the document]
ewandennis: If you mean we'll have to use similar structure across all the sections, I agree. Hopefully, that'll be the main task for the Final Merger.

src/pages/sendingDomains/EditPage.js
jasonrhodes: Not necessary right this minute but we should chat about this via discussion issue or in a Slack channel or something. Redux loading state is sort of a fundamental part of the async flow and a big reason we use Redux -- could possibly add a `loaded` pattern that goes false -> true on first load. Anyway, something to talk about and settle on. :)

src/pages/sendingDomains/EditPage.js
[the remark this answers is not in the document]
bizob2828: or a 5xx

src/pages/sendingDomains/components/ReadyFor.js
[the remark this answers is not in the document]
ewandennis: Reinstated.

ewandennis: Leftover from an earlier iteration. Removed.

src/pages/sendingDomains/components/StatusDescription.js
```diff
@@ -0,0 +1,27 @@
...
+    <LabelledValue label='Status'>
+      { status !== 'verified' && <DomainStatusTag status={status} /> }
+      <ReadyFor {...readyFor} bounceDefault={bounceDefault} />
+    </LabelledValue>
+  </Panel.Section>
+  {
+    (subaccount || shared) &&
+      <Panel.Section>
+        <LabelledValue label='Subaccount'>
+          <SubaccountTag id={subaccount} all={shared} />
```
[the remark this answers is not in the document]
ewandennis: Yup, slavishly followed the existing app design. Removed after letting Jon know.

src/pages/sendingDomains/components/ShareWithSubaccounts.js
ewandennis: Good spot. Didn't know that component existed.

src/actions/sendingDomains.js
```diff
@@ -16,7 +16,18 @@ export function get(id) {
     type: 'GET_SENDING_DOMAIN',
     meta: {
       method: 'GET',
-      url: `/sending-domains/${id}`
+      url: `/sending-domains/${id}`,
+      id
+    }
+  });
+}
+export function update({ id, ...updateFields }) {
```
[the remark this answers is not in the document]
ewandennis: Picked up the impl from Raju's ticket to simplify the merge. (Not a fan of `data` as var name though)

bizob2828:
merged into FAD-6065

PR #255, opened 2018-01-25
FAD-6064: Sending Domain Edit, Bounce section

rajumsys:
Not ready, created PR so people can use shared comp

Notes: 
- Need to decide how to handle when update fails (set default bounce domain). Just show error and do nothing vs. show error and reload domain info so UI gets in sync. 
- This ticket will require A-D, A-E works too (to setup bounce domains info in production.conf)
 - - https://fisheye.int.messagesystems.com/cru/APP-3119

### DIFF
--- a/src/actions/sendingDomains.js
+++ b/src/actions/sendingDomains.js
@@ -35,7 +35,6 @@ export function create(data) {
   });
 }
 
-
 export function verify(id, type) {
   const data = {};
   data[`${type}_verify`] = true;
--- a/src/actions/sendingDomains.js
+++ b/src/actions/sendingDomains.js
@@ -34,3 +34,18 @@ export function create(data) {
     }
   });
 }
+
+
+export function verify(id, type) {
+  const data = {};
+  data[`${type}_verify`] = true;
+
+  return sparkpostApiRequest({
+    type: 'VERIFY_SENDING_DOMAIN',
+    meta: {
+      method: 'POST',
+      url: `/sending-domains/${id}/verify`,
+      data
+    }
+  });
+}
--- a/src/actions/sendingDomains.js
+++ b/src/actions/sendingDomains.js
@@ -49,3 +49,14 @@ export function verify(id, type) {
     }
   });
 }
+
+export function update(id, data) {
+  return sparkpostApiRequest({
+    type: 'UPDATE_SENDING_DOMAIN',
+    meta: {
+      method: 'PUT',
+      url: `/sending-domains/${id}`,
+      data
+    }
+  });
+}
--- a/src/actions/sendingDomains.js
+++ b/src/actions/sendingDomains.js
@@ -35,7 +35,9 @@ export function create(data) {
   });
 }
 
-export function verify(id, type) {
+export function verify(id, type, subaccount) {
+  const headers = setSubaccountHeader(subaccount);
+
   const data = {};
   data[`${type}_verify`] = true;
 
@@ -44,18 +46,22 @@ export function verify(id, type) {
     meta: {
       method: 'POST',
       url: `/sending-domains/${id}/verify`,
-      data
+      data,
+      headers
     }
   });
 }
 
-export function update(id, data) {
+export function update(id, data, subaccount) {
+  const headers = setSubaccountHeader(subaccount);
+
   return sparkpostApiRequest({
     type: 'UPDATE_SENDING_DOMAIN',
     meta: {
       method: 'PUT',
       url: `/sending-domains/${id}`,
-      data
+      data,
+      headers
     }
   });
 }
--- a/src/config/default.js
+++ b/src/config/default.js
@@ -81,6 +81,7 @@ const config = {
   dateFormat: 'YYYY/MM/DD',
   timeFormat: 'HH:mm',
   bounceDomains: {
+    allowDefault: true,
     cnameValue: 'sparkpostmail.com'
   }
 };
--- a/src/config/default.js
+++ b/src/config/default.js
@@ -79,7 +79,10 @@ const config = {
     enabled: false
   },
   dateFormat: 'YYYY/MM/DD',
-  timeFormat: 'HH:mm'
+  timeFormat: 'HH:mm',
+  bounceDomains: {
+    cnameValue: 'uat-public.mail.e.sparkpost.com'
+  }
 };
 
 export default config;
--- a/src/config/default.js
+++ b/src/config/default.js
@@ -81,7 +81,7 @@ const config = {
   dateFormat: 'YYYY/MM/DD',
   timeFormat: 'HH:mm',
   bounceDomains: {
-    cnameValue: 'uat-public.mail.e.sparkpost.com'
+    cnameValue: 'sparkpostmail.com'
   }
 };
 
--- a/src/pages/sendingDomains/EditPage.js
+++ b/src/pages/sendingDomains/EditPage.js
@@ -2,7 +2,6 @@ import React, { Component } from 'react';
 import { connect } from 'react-redux';
 import { Link } from 'react-router-dom';
 import { withRouter } from 'react-router-dom';
-import { reduxForm } from 'redux-form';
 
 import { get as getDomain } from 'src/actions/sendingDomains';
 import { Loading, ApiErrorBanner } from 'src/components';
@@ -14,8 +13,6 @@ const breadcrumbAction = {
   Component: Link,
   to: '/account/sending-domains'
 };
-
-const FORM_NAME = 'sendingDomainEdit';
 
 export class EditPage extends Component {
   componentDidMount() {
@@ -40,8 +37,9 @@ export class EditPage extends Component {
     if (!domain) {
       return null;
     }
+
     return (
-      <EditBounce form={FORM_NAME} id={this.props.match.params.id} domain={domain} />
+      <EditBounce id={this.props.match.params.id} domain={domain} />
     );
   }
 
@@ -74,9 +72,4 @@ const mapStateToProps = ({ sendingDomains: { domain, getError, getLoading }}) =>
 });
 
 
-const formOptions = {
-  form: FORM_NAME,
-  enableReinitialize: true // required to update initial values from redux state
-};
-
-export default withRouter(connect(mapStateToProps, { getDomain })(reduxForm(formOptions)(EditPage)));
+export default withRouter(connect(mapStateToProps, { getDomain })(EditPage));
--- a/src/pages/sendingDomains/EditPage.js
+++ b/src/pages/sendingDomains/EditPage.js
@@ -33,8 +33,13 @@ export class EditPage extends Component {
   }
 
   renderPage() {
+    const { domain } = this.props;
+
+    if (!domain) {
+      return null;
+    }
     return (
-      <EditBounce domain={this.props.match.params.id} />
+      <EditBounce name={this.props.match.params.id} domain={domain} />
     );
   }
 
@@ -50,7 +55,7 @@ export class EditPage extends Component {
         title={`Edit ${id}`}
         breadcrumbAction={breadcrumbAction}
       >
-        {getError ? this.renderError() : this.renderPage()}
+        {getError ? this.renderError() : this.renderPage(id)}
 
       </Page>
     );
--- a/src/pages/sendingDomains/EditPage.js
+++ b/src/pages/sendingDomains/EditPage.js
@@ -15,6 +15,8 @@ const breadcrumbAction = {
   to: '/account/sending-domains'
 };
 
+const FORM_NAME = 'sendingDomainEdit';
+
 export class EditPage extends Component {
   componentDidMount() {
     this.loadDomainProps();
@@ -39,7 +41,7 @@ export class EditPage extends Component {
       return null;
     }
     return (
-      <EditBounce name={this.props.match.params.id} domain={domain} />
+      <EditBounce form={FORM_NAME} id={this.props.match.params.id} domain={domain} />
     );
   }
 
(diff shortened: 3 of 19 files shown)


src/pages/sendingDomains/components/EditBounce.js
```diff
@@ -0,0 +1,133 @@
...
+    return (<Grid>
+      <Grid.Column xs={12} md={5}>
+        <div>
+          <p><strong>To use this domain for bounces</strong>, add this CNAME record to your DNS settings.</p>
+          <p><em>Note: Bounce domains must be verified via DNS.</em></p>
+        </div>
+      </Grid.Column>
+      <Grid.Column xs={12} md={7}>
+        <Panel sectioned>
+          <Grid>
```
rajumsys: These hurt my eyes! But not sure how can I make them not!

jonambas: Feel free to abstract the Grid out into something reusable to avoid this clutter.

Even something simple like `SectionRight` & `SectionLeft` that just wraps grid columns

jonambas: Also, you could maybe try to use Table from MB for the DNS info

src/pages/sendingDomains/components/EditBounce.js
```diff
@@ -0,0 +1,133 @@
...
+            <Grid.Column xs={12} md={4}>
+              <p>{ name }</p>
+            </Grid.Column>
+            <Grid.Column xs={12} md={6}>
+              <p>{ config.bounceDomains.cnameValue }</p>
+            </Grid.Column>
+          </Grid>
+
+          <Grid>
+            <Grid.Column xs={12} md={12} className={styles.Center}>
```
rajumsys: @jonambas i put this as a button than a link on the top-right cell. let me know if you like this or we should do as it's now

screen shot 2018-01-25 at 12 43 15 pm

src/pages/sendingDomains/components/EditBounce.js
```diff
@@ -0,0 +1,150 @@
...
+
+  toggleDefaultBounce = () => {
+    const { id, update, domain, change, form } = this.props;
+
+    return update(id, {
+      is_default_bounce_domain: !domain.is_default_bounce_domain
+    })
+      .catch((err) => {
+        //restore previous value
+        change(form, 'is_default_bounce_domain', domain.is_default_bounce_domain);
```
rajumsys: i'll probably get rid of this and will just alert (and possibly dispatch `get`)

src/pages/sendingDomains/EditPage.js
```diff
@@ -25,11 +29,22 @@ export class EditPage extends Component {
...
   }
 
+  renderPage() {
+    const { domain } = this.props;
+
+    if (!domain) {
+      return null;
+    }
+    return (
+      <EditBounce form={FORM_NAME} id={this.props.match.params.id} domain={domain} />
```
bizob2828: commented on this in greg's. why are we passing in the form? they should be distinct forms right?

rajumsys: they can be but i didn't expect want to be. no specific reason. that's on us. if we think it'll be better we can do that. i just thought it should be one single form unless there is other reason not to.

bizob2828: they are separate forms.  there's not one submit action. basically every section has separate updates/submits at least in current

src/pages/sendingDomains/components/EditBounce.js
```diff
@@ -0,0 +1,126 @@
...
+      <Grid.Column xs={12} md={6}>
+        <Panel sectioned>
+          <div>Your domain is ready to be used as a bounce domain.</div>
+        </Panel>
+        <Panel sectioned
+        >
+          <Field
+            name='is_default_bounce_domain'
+            component={ToggleBlock}
+            label='Use this domain as your default bounce domain?'
```
gwalls: Our current UI also has a tooltip with a description for setting the default bounce domain

src/pages/sendingDomains/components/EditBounce.js
```diff
@@ -0,0 +1,126 @@
...
+      </Grid>
+    );
+  }
+
+  renderNotReady() {
+    const { id } = this.props;
+    return (<Grid>
+      <Grid.Column xs={12} md={5}>
+        <div>
+          <p><strong>To use this domain for bounces</strong>, add this CNAME record to your DNS settings.</p>
```
gwalls: I went ahead and added an icon representing the current status (check mark or exclamation point), like in the current UI. I think the icon is a nice first-glance indicator.

src/pages/sendingDomains/components/EditBounce.js
```diff
@@ -0,0 +1,126 @@
+import React, { Component } from 'react';
+import { Field } from 'redux-form';
+import { Link } from 'react-router-dom';
+import { connect } from 'react-redux';
+
+import { verify, update } from 'src/actions/sendingDomains';
+
+import { Panel, Grid, Banner } from '@sparkpost/matchbox';
+import ToggleBlock from 'src/pages/templates/components/ToggleBlock';
```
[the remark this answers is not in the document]
jonambas: This is already pulled out in develop, moved it over in my templates ticket

src/pages/sendingDomains/components/EditBounce.js
```diff
@@ -0,0 +1,126 @@
...
+
+  verifyDomain = () => {
+    const { id, verify, showAlert } = this.props;
+    return verify(id, 'cname')
+      .then((results) => {
+        const readyFor = resolveReadyFor(results);
+        if (readyFor.bounce) {
+          showAlert({ type: 'success', message: `You have successfully verified CNAME record of ${id}` });
+        } else {
+          showAlert({ type: 'error', message: `Unable to verify CNAME record of ${id}. ${results.dns.cname_error}` });
```
gwalls: Needs test coverage

src/config/default.js
```diff
@@ -79,7 +79,10 @@ const config = {
     enabled: false
   },
   dateFormat: 'YYYY/MM/DD',
-  timeFormat: 'HH:mm'
+  timeFormat: 'HH:mm',
+  bounceDomains: {
+    cnameValue: 'uat-public.mail.e.sparkpost.com'
```
gwalls: It looks like the default is currently `sparkpostmail.com`

src/pages/sendingDomains/components/EditBounce.js
```diff
@@ -0,0 +1,126 @@
...
+        }
+      })
+      .catch((err) => {
+        showAlert({ type: 'error', message: `Unable to verify CNAME record of ${id}. ${err.message}` });
+      });
+  }
+
+  toggleDefaultBounce = () => {
```
gwalls: Add test coverage around this

src/pages/sendingDomains/components/EditBounce.js
```diff
@@ -0,0 +1,126 @@
...
+    const { id, update, domain, showAlert } = this.props;
+
+    return update(id, {
+      is_default_bounce_domain: !domain.is_default_bounce_domain
+    })
+      .catch((err) => {
+        showAlert({ type: 'error', message: `Failed to update default bounce option. ${err.message}` });
```
jonambas: Do you need to reset the form value here? maybe try `this.props.reset` from redux form?

rajumsys: I did that initially with `change`. didn't know about `reset`. 

@gwalls and I wanted to figure out best action at this point 
- show fail alert and do nothing
- show fail alert and re-fetch the domain. this way we won't to manually update UI

bizob2828: i think reset grabs whatever is in store so you wouldn't have to re-fetch domain

bizob2828: however, i was talking to @jonambas and i think we're going to create a form UX story to solve this holistically

src/pages/sendingDomains/EditPage.js
```diff
@@ -25,11 +29,22 @@ export class EditPage extends Component {
...
+      message='Sorry, we seem to have had some trouble loading your Sending Domain.'
       reload={this.loadDomainProps}
     />;
   }
 
+  renderPage() {
+    const { domain } = this.props;
+
+    if (!domain) {
+      return null;
```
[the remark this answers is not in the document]
rajumsys: yes. when component is mounted, it'll have `getLoading` and `getLoading` as `undefined` as the request isn't triggered yet. so at that time it'll throw error.

src/actions/sendingDomains.js
```diff
@@ -34,3 +34,28 @@ export function create(data) {
     }
   });
 }
+
+export function verify(id, type) {
```
gwalls: add the subaccount for verify and update. See @bizob2828 's PR.

src/pages/sendingDomains/EditPage.js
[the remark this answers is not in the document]
bizob2828: i solved this in mine by defaulting domain to empty object https://github.com/SparkPost/2web2ui/pull/261/files#diff-06e95fb1f0ceeae76f874abeda9a67e4

src/pages/sendingDomains/EditPage.js
```diff
@@ -42,7 +57,7 @@ export class EditPage extends Component {
         title={`Edit ${id}`}
         breadcrumbAction={breadcrumbAction}
       >
-        {getError ? this.renderError() : 'Coming soon'}
+        {getError ? this.renderError() : this.renderPage(id)}
```
bizob2828: you should be able to grab id from props in renderPage

src/pages/sendingDomains/EditPage.js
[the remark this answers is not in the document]
jasonrhodes: Reducer initial state is the way to go for these types of things, I think. Only time to be careful is if you need to know the difference between "haven't tried to load yet" vs "tried and found nothing" but in a single GET you won't have that problem (tried and found nothing would be a 404 error)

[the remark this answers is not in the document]
rajumsys: That's a good idea indeed

src/pages/sendingDomains/EditPage.js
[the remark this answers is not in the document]
jasonrhodes: FWIW not sure there's much to gain from this destructure

src/pages/sendingDomains/components/EditBounce.js
```diff
@@ -0,0 +1,126 @@
...
+      <Grid.Column xs={12} md={6}>
+        <div>This domain is all set up and ready to be used as a bounce domain. Such alignment, wow!</div>
+      </Grid.Column>
+      <Grid.Column xs={12} md={6}>
+        <Panel sectioned>
+          <div>Your domain is ready to be used as a bounce domain.</div>
+        </Panel>
+        <Panel sectioned
+        >
+          <Field
```
gwalls: This should only show based on config. See https://github.com/SparkPost/webui/blob/master/src/app/_pages/domains/edit.html#L193-L218

src/config/default.js
gwalls: Add `allowDefault` flag

src/pages/sendingDomains/components/EditBounce.js
```diff
@@ -0,0 +1,126 @@
...
+            name='is_default_bounce_domain'
+            component={ToggleBlock}
+            label='Use this domain as your default bounce domain?'
+            type='checkbox'
+            parse={(value) => !!value} // Prevents unchecked value from equaling ""
+            disabled={false}
```
gwalls: should this be `disabled={this.props.submitting}`?

src/pages/sendingDomains/EditPage.js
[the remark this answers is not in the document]
jasonrhodes: @rajumsys that's a good point, I was thinking about that after I commented last night. Probably better to "exit early" than to try to default out all of those deep props.

src/pages/sendingDomains/components/EditBounce.js
```diff
@@ -0,0 +1,126 @@
...
+          <div>Your domain is ready to be used as a bounce domain.</div>
+        </Panel>
+        <Panel sectioned
+        >
+          <Field
+            name='is_default_bounce_domain'
+            component={ToggleBlock}
+            label='Use this domain as your default bounce domain?'
+            type='checkbox'
+            parse={(value) => !!value} // Prevents unchecked value from equaling ""
```
bizob2828: how can this happen?

src/pages/sendingDomains/EditPage.js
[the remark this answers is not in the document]
gwalls: I got away with `{ domain: { status: {}, dkim: {}} }` for my initial state in the reducer. But I can see how that would get messy if there are more sub-keys to add...

src/pages/sendingDomains/components/EditBounce.js
rajumsys: good point. but i'll use `updateLoading`. I guess submitting works with a real form submission (need to review doc). but `updateLoading` totally works

src/pages/sendingDomains/EditPage.js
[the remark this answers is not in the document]
jasonrhodes: Yeah you could make a case to grab `match` there and then just use `match.params.id` here, but I could go either way on that. My main arguments for destructuring are 

a) when we re-use a value a few times in that block so it's easier to reference it directly, 
b) it allows us to use object shorthand later `const { a } = thing; return { a };` 
c) it doesn't go beyond 1-2 levels `level1: { level2 }` bc I think it gets kinda hairy to follow beyond that (subjective and unlinted opinion so take it with a grain of salt)

src/pages/sendingDomains/EditPage.js
gwalls: can remove during merge

src/actions/sendingDomains.js
gwalls: This will be fixed after the merge. :shipit:

bizob2828:
I merged this into FAD-6065.

PR #258, opened 2018-01-26
FAD-6123 Fix sandbox change in from email typeahead (FAD-5609 follow up)

jonambas:


src/pages/templates/components/FromEmail.js
```diff
@@ -22,6 +22,10 @@ export class FromEmail extends Component {
     this.setState({ value: this.props.value });
   }
 
+  componentWillReceiveProps(nextProps) {
+    this.setState({ value: nextProps.value });
```
jasonrhodes: Any reason to check `if (nextProps.value !== this.props.value)` in case this gets called in ways we don't realize?

jonambas: Yeah good idea

jasonrhodes:
Validated the subaccount/sending domain stuff, and the template ID validation change.

If you want, we can remove template ID help text when not in create mode, too.

```jsx
helpText={newTemplate ? 'A Unique ID for your template, we\'ll fill this in for you.' : null}
```

Up to you. Otherwise :shipit:

PR #259, opened 2018-01-27
FAD-5933 (b) ALL post-login redirects handled by new default redirect component

jasonrhodes:
Realized while finishing up FAD-5933 that I hadn't accounted for the Register or SSO flows which also needed complicated redirect logic. Rather than spreading all of that logic all over the place, I thought it was better to just always `history.push` or `<Redirect>` to a new page component route that will do all of that logic in one place, which will:
* Show a brief loading state in the app to show the user they successfully logged in a second or two sooner
* Wait for access control state to load (user, account, grants)
* Get the config splash page value and any other config values we need
* Make a decision on where the user should go
* Redirect the user to that route, with `{ replace: true }` so the intermediate route disappears from browser history

Also makes this logic easier to test.

**Note: I used a new `constants` file to load the route name so we can change it easily if we think of a better one. :)**


src/pages/users/tests/ListPage.test.js
```diff
@@ -41,7 +42,7 @@ describe('Page: Users List', () => {
       isCurrentUser: false,
       access: 'admin',
       email: 'testemail',
-      last_login: '2017-05-05T12:00:00'
+      last_login: utc({ year: 2017, month: 12, day: 15, hour: 12 })
```
bizob2828: this test is now broken

PR #260, opened 2018-01-29
FAD-6110 Audit and fix subaccount experience

jonambas:
### What I did

API Keys
- Removed history.push on update.
- Moved `history.push`s to the pages, to not redirect on error.

Webhooks
Details flow:
1) Details page getWebhook pulls subaccount id from qp.
2) Subaccount id stored under `webhooks.webhook.subaccount` after the GET.
3) Update, delete, test use this key.

- New `listAllWebhooks` action
- Removed id column
- Added new tag variation for "Master & all subaccounts"
- Used `showAlert` instead of banners
- Added `helpText` to most fields to match old ui
- Used history.push instead of Redirect

Selectors
- Tried to make these consistent and reusable
  - passing in both `(state, props)` to some selectors that dont need state, just so they can be shared with other `reselect` selectors

Subaccount Typeahead 
- added `helpText` support
- field value now shows id - `${item.name} (${item.id})`
- removes clear button if `disabled`

---
### Verify
- [x] Webhook create & redirect
- [x] Webhook update
- [x] Webhook delete
- [x] Webhook test
- [x] Webhook batch status
- [x] ApiKeys create & redirect
- [x] ApiKeys update
- [x] ApiKeys delete
- [ ] Templates still works


src/pages/webhooks/ListPage.js
```diff
@@ -3,36 +3,63 @@ import { connect } from 'react-redux';
...
+
+  getRowData = ({ id, name, target, subaccount_id }) => {
+    const { hasSubaccounts } = this.props;
+    const nameLink = <Link to={`/webhooks/details/${id}${setSubaccountQuery(subaccount_id)}`}>{name}</Link>;
+    const row = [nameLink, target];
+
+    if (hasSubaccounts) {
+      let subaccountTag = null;
+
+      if (subaccount_id) {
```
bizob2828: we need to pass master true to SubaccountTag if subaccount_id === 0

src/actions/webhooks.js
```diff
@@ -1,65 +1,96 @@
...
     type: 'TEST_WEBHOOK',
     meta: {
       method: 'POST',
       url: `/webhooks/${id}/validate`,
-      data: { message: payload }
+      data: { message: data },
+      headers
     }
   });
 }
```
bizob2828: missing subaccount header on getBatches action

bizob2828: i also noticed from that 404 that we aren't catching this call in the BatchTab.js so we get Uncaught errors, which wil leventually log in sentry, up to you if you want to fix now, at the very least put in the fix up webhooks ticket

bizob2828: I also noticed we aren't setting batches back to [] in `GET_WEBHOOK_BATCHES_FAL` so you could see other batch data on failure

bizob2828: all of this is my bad BTW so, i'd be more than happy to fix in this ticket

src/actions/webhooks.js
```diff
@@ -1,65 +1,96 @@
...
-      url: '/webhooks'
+      url: '/webhooks',
+      headers
     }
   });
 }
 
-export function getWebhook(id) {
+export function listAllWebhooks() {
+  return (dispatch) => allSettled([
```
bizob2828: so i don't think this is working as you may think. i hit a weird case cuz my vpn crapped out that masterAccountWebhooks never returns so when it got into the map in mergeWebhooks it failed.  so either we put a catch on this in here or at a higher level to avoid these Unhandled rejections

src/pages/api-keys/ListPage.js
```diff
@@ -24,6 +27,38 @@ export class ListPage extends Component {
...
+        ? <SubaccountTag id={subaccount_id} />
+        : null;
+
+      rowData.push(subaccountTag);
+    }
+
+    return rowData;
+  };
+
+  getColumns = () => {
```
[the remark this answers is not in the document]
bizob2828: i guess this is because rebase hasn't happened but that doesn't have subaccount column. seems valuable to sort on this

src/pages/suppressions/ListPage.js
```diff
@@ -42,8 +42,8 @@ export class ListPage extends Component {
   }
```
bizob2828: should we use subaccount tag on this list page?

src/pages/templates/containers/EditPage.container.js
```diff
@@ -5,7 +5,7 @@ import { reduxForm } from 'redux-form';
 import { getDraft, getPublished, update, deleteTemplate, publish, getTestData } from 'src/actions/templates';
 import { showAlert } from 'src/actions/globalAlert';
 import { selectTemplateById, selectTemplateTestData } from 'src/selectors/templates';
-import { getSubaccountIdFromQuery, hasSubaccounts } from 'src/selectors/subaccounts';
+import { selectSubaccountIdFromQuery, hasSubaccounts } from 'src/selectors/subaccounts';
```
bizob2828: the subaccount field doesn't show the name just the id

jonambas: Yeah I noticed, I need to pull in the actual subaccount object, easy fix

src/pages/webhooks/components/SubaccountSection.js
```diff
@@ -0,0 +1,91 @@
...
+  { label: 'Master account only', value: 'master' },
+  { label: 'Master and all subaccounts', value: 'all' },
+  { label: 'Single Subaccount', value: 'subaccount' }
+];
+
+/**
+ * Component produces the follow redux form fields
+ * If new
+ * - assignTo 'master' | 'all' | 'subaccount'
+ * - subaccount if assignTo === 'subaccount'
```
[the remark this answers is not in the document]
bizob2828: oh this is happening, mind showing me where and how

src/pages/webhooks/DetailsPage.js
```diff
@@ -3,33 +3,54 @@ import { connect } from 'react-redux';
...
   componentDidMount() {
-    this.props.getWebhook(this.props.match.params.id);
+    const { showAlert, getWebhook, hasSubaccounts, subaccounts, getSubaccountsList, match, history, subaccountId: subaccount } = this.props;
+
+    getWebhook({ id: match.params.id, subaccount }).catch((err) => {
+      history.push('/webhooks/');
+      showAlert({ type: 'error', message: 'Unable to find webhook', details: err.message });
+    });
+
+    // Get subaccounts list here, so page doesn't rely on SubaccountTypeahead to pull the list
```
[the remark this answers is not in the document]
jonambas: I think I can remove this, seems to work without it. I put this in cause the SubaccountSection component might not render the typeahead, it could load the radio group or a textfield for "Master only" "Master & all subaccounts".

src/pages/webhooks/components/SubaccountSection.js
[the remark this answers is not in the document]
jonambas: Yeah, its in `createWebhook` on the create page

src/actions/webhooks.js
```diff
@@ -1,65 +1,96 @@
...
+export function createWebhook({ webhook, subaccount }) {
+  const headers = setSubaccountHeader(subaccount);
   return sparkpostApiRequest({
     type: 'CREATE_WEBHOOK',
     meta: {
       method: 'POST',
       url: '/webhooks',
-      data: webhook
+      data: webhook,
+      subaccount, // Pass through subaccount to redirect
```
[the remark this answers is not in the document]
jonambas: I can expand on the comment
basically it redirects in `componentDidUpdate` with the ID that this call returns, and need to pass subaccount through as well to set the query param

src/pages/webhooks/components/SubaccountSection.js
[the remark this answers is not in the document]
jonambas: i could move it to the action for more visibility

src/pages/webhooks/ListPage.js
jonambas: Should we show this tag? On other lists we hide the tag if assigned to master

src/pages/api-keys/ListPage.js
```diff
@@ -24,6 +27,38 @@ export class ListPage extends Component {
...
 
+  getRowData = ({ id, label, short_key, subaccount_id }) => {
+    const { hasSubaccounts } = this.props;
+    const rowData = [
+      <Link to={`/account/api-keys/details/${id}${setSubaccountQuery(subaccount_id)}`}>{label}</Link>,
+      <code>{short_key}••••••••</code>
+    ];
+
+    if (hasSubaccounts) {
+      const subaccountTag = (hasSubaccounts && subaccount_id)
```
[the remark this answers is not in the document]
jonambas: Might be a good idea, can avoid this ternary

src/pages/webhooks/components/EditTab.js
```diff
@@ -40,19 +36,18 @@ class EditTab extends Component {
     to the updateWebhook action.
   */
   handleSubmit = (values, webhook, allEvents) => {
+    const { getWebhook, updateWebhook, showAlert } = this.props;
+
     const update = prepareWebhookUpdate(values, webhook, allEvents);
 
-    if (Object.keys(update).length !== 0) {
```
jonambas: Removed this check, submit is already disabled on pristine

src/pages/suppressions/ListPage.js
jonambas: Oh I missed this one, but I'll probably leave out for now cause this is getting massive

src/pages/webhooks/components/EditTab.js
bizob2828: nice, good catch

PR #261, opened 2018-01-30
FAD-6065 Sending Domains Edit Page

bizob2828:


src/reducers/messageEvents.js
```diff
@@ -28,6 +28,7 @@ export default (state = initialState, { type, payload }) => {
       return {
         ...state,
         historyLoading: false,
+        // TODO: fix this https://sentry.io/sparkpost/alpha-web-ui-production/issues/435911822/
```
bizob2828: i really want to fix this, i need to look at impacts though so if i run out of time ill defer but keep TODO

src/pages/sendingDomains/components/AssignTrackingDomain.js
```diff
@@ -0,0 +1,97 @@
...
+        message: 'Could not update tracking domain assignment.',
+        details: err.message
+      }));
+  };
+
+  renderSection() {
+    return (<Grid>
+      <Grid.Column xs={12} md={6}>
+        <div>Link to a tracking domain to track opens, clicks, and unsubscribes.</div>
+        <Link to={'/account/tracking-domains'}>Create a tracking domain.</Link>
```
gwalls: This has parity with the current UI, but shouldn't this link directly to the track-domains create page? (and not list)

src/pages/sendingDomains/components/AssignTrackingDomain.js
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  renderTrackingDomains() {
+    const { listLoading, trackingDomains, pristine, submitting } = this.props;
+
+    if (listLoading) {
+      return <PanelLoading />;
+    }
+
+    return (
```
gwalls: It appears that the tracking domains list will be populated with tracking domains "owned" by a subaccount, but I can't assign that tracking domain to a sending domain owned by a master account (I think) - it results in a 400
`"Tracking domain 'bob.spappteam.com' is not a registered tracking domain"`
This is regardless of the domain's `shared_with_subaccounts` flag. I'm not exactly sure what the behavior should be here. We might just have to pass the tracking domain's subaccount id to the sending domain request if `shared_with_subaccounts`.

The Angular UI doesn't populate the list with such tracking domains.

src/pages/sendingDomains/components/AssignTrackingDomain.js
```diff
@@ -0,0 +1,97 @@
...
+        <h5>Linked tracking domain</h5>
+        <form onSubmit={this.props.handleSubmit(this.updateTrackingDomain)}>
+          <Field
+            component={SelectWrapper}
+            name='trackingDomain'
+            options={trackingDomains}
+            disabled={submitting}
+            helpText='Domains must be verified to be linked to a sending domain.'
+          />
+          <Button submit primary disabled={submitting || pristine}>
```
gwalls: Once the value has been successfully updated, it should be "pristine". It currently detects it's still "dirty" so that I can press the button repeatedly.

src/pages/sendingDomains/components/AssignTrackingDomain.js
bizob2828: good point

src/pages/sendingDomains/components/AssignTrackingDomain.js
bizob2828: oh, i noticed this in sending domains. i think i can just get list of all and filter after response

src/pages/sendingDomains/components/AssignTrackingDomain.js
bizob2828: yea i noticed that too, i'll have to ask others because i couldn't figure this out

src/pages/auth/components/LoginForm.js
```diff
@@ -29,7 +29,7 @@ export class LoginForm extends Component {
           name='password'
           id='password'
           label='Password'
-          placeholder='Your Password'
+          placeholder='IheartLilSebastian'
```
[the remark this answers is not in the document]
bizob2828: trying to strengthen that Parks and Rec ❤️ around here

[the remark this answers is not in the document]
bizob2828: it was either that our M0useRatRulez

src/pages/sendingDomains/components/AssignTrackingDomain.js
bizob2828: derp, forgot a reducer for UPDATE_SENDING_DOMAIN

src/pages/sendingDomains/components/SetupSending.js
```diff
@@ -0,0 +1,125 @@
...
+
+  render() {
+    return (
+      <SendingDomainSection
+        title="Set Up For Sending"
+        sectioned
+      >
+        <Grid>
+          <Grid.Column xs={12} md={4}>
+            {this.renderBanner()}
```
jasonrhodes: We should move this banner up above the grid so it's full-width, like it is in the Bounce section

jasonrhodes:
Vertical alignment is off on the status section:
screen shot 2018-02-04 at 5 32 10 pm

jasonrhodes:
Status section header is top-aligned inside of the left side of the grid, but the other sections have the heading above. I like it inside but with the need for full-width alerts, probably should just have them all sit above for now. Either way, just needs to be consistent.

jasonrhodes:
Maybe for later, but could probably just have the change event on that tracking domains select just auto-save, so we don't need any update buttons on this page at all. (Like how the subaccount toggle works now.)

ewandennis:
- Status section / unverified domains
     - Domain setup guide link anchor is broken
     - Fix vertical alignment on status header
 - Setup for bounce section / bounce-ready domains
     - column widths should match other sections
     - note on left column duplicates message on right
 - Setup for bounce section / unverified domains
     - missing alert icon to left of cname settings
     - verify CNAME record button show disable while checking, like DKIM check

For later:
 - Possibly rework the status section header

PR #263, opened 2018-02-02
Feature/fad 5605

bizob2828:
TODO:

- [x] write tests
- [x] A-D updates to add new config
- [x] rip out the dynamic retrieval of API keys to optionally show password, just show a message about what the password is and link to api keys page


src/config/default.js
```diff
@@ -75,11 +75,16 @@ const config = {
...
+    host: 'no-default-set',
+    port: 587,
+    username: 'SMTP_Injection'
+  },
   sso: {
     enabled: false
   },
-  dateFormat: 'MMM D YYYY',
-  timeFormat: 'h:mma'
+  dateFormat: 'YYYY/MM/DD',
```
bizob2828: revert this

src/pages/smtp/SmtpPage.js
```diff
@@ -0,0 +1,112 @@
...
+        { this.renderTracking() }
+      </Page>
+    );
+
+  }
+}
+
+const mapStateToProps = ({ account, ...state }) => {
+  const apiKeys = selectApiKeysForSmtp(state);
+  const shortKey = _.get(apiKeys, '0.short_key', null);
```
bkemper: Since `apiKeys` isn't used again and to test this logic if the key is not present, this logic could be pulled into a selector, `selectShortApiKeysForSmtp`.

src/pages/smtp/SmtpPage.js
```diff
@@ -0,0 +1,112 @@
...
+        />
+      </Panel>
+    );
+  }
+
+  render() {
+    return (
+      <Page title='SMTP Relay'>
+        { this.renderSMTPDeets() }
+        { this.renderTracking() }
```
bkemper: You obviously don't need to take this advice, but if `renderSMTPDeets` and `renderTracking` were extracted into components, it would make writing tests easier and produce smaller easier to read snapshots.

src/pages/smtp/SmtpPage.js
```diff
@@ -0,0 +1,112 @@
...
+      ...account
+    }
+  };
+};
+
+const formOptions = {
+  form: 'smtpDefaultOptions'
+};
+
+export default connect(mapStateToProps, { updateAccount, listApiKeys })(reduxForm(formOptions)(SmtpPage));
```
bkemper: I don't know that redux form is needed for a single field form.

src/config/default.js
bkemper: Reverted... 4aba7ac7

bkemper:
Also, I noticed when a user visits the page directly it take a second to fetch the account which causes this toggle of the password.

password-toggle

src/pages/smtp/SmtpPage.js
bizob2828: im ripping this stuff out because the A/C says to just link to the api keys page

bizob2828:
A-D changes here: https://fisheye.int.messagesystems.com/cru/APP-3131

src/pages/smtp/tests/SmtpPage.test.js
```diff
@@ -0,0 +1,60 @@
...
+    instance = wrapper.instance();
+  });
+
+  it('should render page correctly with defaults', () => {
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  describe('toggleTracking', () => {
+    it('should show success alert after updating account', async() => {
+      await instance.toggleTracking();
```
bkemper: I don't know what the exact savings is or risk, but I have started to not rely on enzyme when testing instance functions.

```
const instance = new SmtpPage(props);
await instance.toggleTracking();
```

bkemper:
Just need to resolve the merge conflict.

PR #267, opened 2018-02-04
FAD-6157 updates tenant config location

jasonrhodes:
Associated deployment review: https://fisheye.int.messagesystems.com/cru/APP-3130

CONFIG_CACHE_ID is a timestamp that will fingerprint each deploy's config file for caching. NGINX will rewrite that url to ignore the cache ID and swap in the hostname:

```nginx
  location ~ ^/static/tenant-config/[0-9]*/(.*)$ {
    alias /opt/msys/app/phoenix/current/static/tenant-config/$host/$1;
  }
```


config/env.js
```diff
@@ -74,6 +75,9 @@ function getClientEnvironment(publicUrl) {
         // This should only be used as an escape hatch. Normally you would put
         // images into the `src` and `import` them in code to get their paths.
         PUBLIC_URL: publicUrl,
+        // Useful for a simple, cacheable fingerprint when the server rewrites requests
+        BUILD_TIME: now,
+        CONFIG_CACHE_ID: process.env.NODE_ENV !== 'development' ? now : ''
```
jasonrhodes: === "production"

rajumsys:
I guess we'll have to test it by deploying; but otherwise looks good to me.

PR #269, opened 2018-02-05
FAD-6166 Sending Domains Clean Up

jonambas:
- [x] UI Cleanup
- [x] Default Bounce Toggle should only show under right conditions
- [x] Verify TXT record action needs to disable and switch to "Verifying..." when action is pending
  - Verify action appends `type` to the action name. Reducer listens for `_CNAME` and `_DKIM` verify actions.
- [x] Status section - Domain setup guide link anchor is broken
  - This doesn't appear broken to me?
- [x] ~~_set this domain up for sending via email_ is not implemented~~
  - For another ticket
- [x] Align Type, Hostname content to top
  - No longer using SimpleTable
- [x] Column widths should match other sections
  - Created reusable Left and Right components under SendingDomainSection
- [x] Missing alert icon to left of CNAME settings


src/pages/sendingDomains/components/tests/SendingDomainSection.test.js
```diff
@@ -14,4 +14,20 @@ describe('Component: SendingDomainSection', () => {
...
+    };
+    const wrapper = shallow(<SendingDomainSection.Left {...props} />);
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('should render Right section', () => {
+    const props = {
+      children: ['Bill', 'Brasky']
+    };
+    const wrapper = shallow(<SendingDomainSection.Left {...props} />);
```
jonambas: Oops

src/actions/sendingDomains.js
```diff
@@ -67,8 +67,10 @@ export function verify({ id, subaccount, type }) {
   const data = {};
   data[`${type}_verify`] = true;
 
+  const actionType = type ? `_${type.toUpperCase()}` : null;
```
ewandennis: I think `something_verify` must be set in all calls to `/verify` so do we need to account for `!type`?
If we do, please add test coverage for that case.

src/pages/sendingDomains/components/EditBounce.js
```diff
@@ -57,69 +57,84 @@ export class EditBounce extends Component {
...
-      </Grid.Column>
-    </Grid>);
+    const { updateLoading, id, domain } = this.props;
+    const readyFor = resolveReadyFor(domain.status);
+
+    // Allow default bounce toggle if:
+    // Config flag is true
+    // Domain is verified
+    // Domain is not assigned to SubaccountTag
+    // Domain is ready for bounce
```
ewandennis: I ❤️  this literate, "spec in comments converted into named condition variable" style.

The "ready for bounce" condition isn't tested here because the caller handles it. Do we want to test `readyFor.bounce` again here or say `// Domain is ready for bounce - handled by caller` so future us don't introduce a subtle bug?

src/pages/sendingDomains/components/ReadyFor.module.scss
```diff
@@ -3,8 +3,8 @@
 .ReadyFor > * {
   margin-right: spacing(small);
   margin-bottom: spacing(smaller);
-  margin-top: spacing(smaller);
-  
+  // margin-top: spacing(smaller);
```
ewandennis: Remove?

src/reducers/sendingDomains.js
```diff
@@ -28,15 +28,25 @@ export default (state = initialState, { type, payload, meta }) => {
...
       // augment current domain's status property
-      return { ...state, verifyLoading: false, domain: { ...state.domain, status: payload }};
+      return { ...state, verifyCnameLoading: false, domain: { ...state.domain, status: payload }};
 
-    case 'VERIFY_SENDING_DOMAIN_FAIL':
-      return { ...state, verifyLoading: false, verifyError: payload };
+    case 'VERIFY_SENDING_DOMAIN_CNAME_FAIL':
+      return { ...state, verifyCnameLoading: false, verifyCnameError: payload };
+
+    case 'VERIFY_SENDING_DOMAIN_DKIM_PENDING':
```
ewandennis: We have test for this section so lets add base coverage or the `_DKIM` actions.

src/pages/sendingDomains/components/EditBounce.js
ewandennis: `tst` testing:
 - Tested `config.bounceDomains.allowDefault = false`
 - Tested unverified domain `5580master.com`
 - Tested ready for bounce domain `100yearsv2.com`
 - Tested *not* ready for bounce domain `100years.com`
 - Tested subaccount domain `uat-subaccount.splango.net` in unverified, pending and verified states

jonambas: I can add `readyFor.bounce` here again for clarity, may also help avoid a bug down the road if this ever gets restructured.

src/actions/sendingDomains.js
```diff
@@ -67,7 +67,7 @@ export function verify({ id, subaccount, type }) {
   const data = {};
   data[`${type}_verify`] = true;
 
-  const actionType = type ? `_${type.toUpperCase()}` : null;
+  const actionType = type ? `_${type.toUpperCase()}` : '';
```
rajumsys: looks like the way we've set `data[`${type}_verify`] = true;`, `type` is must provided. also if `type` is not passed, see the test it sets it like `"undefined_verify": true,`. i guess we should either default to something or throw if it's not passed.

src/actions/tests/sendingDomains.test.js
```diff
@@ -26,6 +26,10 @@ describe('Action Creator: Sending Domains', () => {
     it('it should verify domain owned by subaccount', () => {
```
rajumsys: too many `it` :)

src/actions/tests/sendingDomains.test.js
```diff
@@ -26,6 +26,10 @@ describe('Action Creator: Sending Domains', () => {
...
       expect(sendingDomains.verify({ id: 'sub.com', subaccount: 101, type: 'cname' })).toMatchSnapshot();
     });
+
+    it('it should not append action type if type is not provided', () => {
```
rajumsys: too many `it` :)

src/pages/sendingDomains/components/EditBounce.js
```diff
@@ -103,9 +103,9 @@ export class EditBounce extends Component {
     // Allow default bounce toggle if:
     // Config flag is true
     // Domain is verified
-    // Domain is not assigned to SubaccountTag
     // Domain is ready for bounce
-    const showToggle = config.bounceDomains.allowDefault && readyFor.sending && !domain.subaccount_id;
+    // Domain is not assigned to subaccount
+    const showToggle = config.bounceDomains.allowDefault && readyFor.sending && readyFor.bounce && !domain.subaccount_id;
```
rajumsys: a more explanatory name could be `showDefaultBounce`/`shouldShowDefaultBounce`/`showDefaultBounceToggle` or something like that.

src/actions/sendingDomains.js
jonambas: Yea the API returns with a 4xx error. Which is I think better than checking if `type` is provided and throwing an error manually. I can add a default though, but not sure what value to use

PR #270, opened 2018-02-06
FAD-6139 message events detail console error

rajumsys:


src/actions/messageEvents.js
```diff
@@ -52,16 +50,17 @@ export function getMessageEvents(options = {}) {
...
-export function getMessageHistory({ messageId, params = {}}) {
+export function getMessageHistory({ messageId, ...rest }) {
   return sparkpostApiRequest({
     type: 'GET_MESSAGE_HISTORY',
     meta: {
       method: 'GET',
       url: '/message-events',
       params: {
-        ...params,
+        to: moment().utc().format(apiDateFormat),
```
rajumsys: i put it above `...rest` so it can be overridden from calling function though we don't have an usage for now.

src/actions/messageEvents.js
```diff
@@ -52,16 +50,17 @@ export function getMessageEvents(options = {}) {
...
+        ...rest,
         message_ids: messageId,
-        from: format(subDays(Date.now(), 10), 'YYYY-MM-DDTHH:MM')
+        from: moment().subtract(10, 'days').utc().format(apiDateFormat)
```
rajumsys: got rid of moment as @jasonrhodes suggested. also fixed the formatter. `MM` means month. for minutes we need `mm`

src/actions/messageEvents.js
ewandennis: Perfect. Maybe a short comment saying that will help future readers.

src/actions/messageEvents.js
ewandennis: Although the implementation looks correct, I'm confused about the comment. Moment is intact in the code. Did you mean to remove moment? (I hope not - we don't yet have a clear way to test date code outside UTC-5 without moment).

ewandennis: Lets put the message events retention period in config. Something like `messageEvents: { retentionPeriodDays: 10 }}` ... ?

jasonrhodes: He means he got rid of datefns here and replaced it with moment :)

PR #271, opened 2018-02-07
2FA toggling, basic set up (FAD-5604)

jasonrhodes:
So far:
* Decided to put this content on the user profile where it feels at home
* Talked it over with Jon and we decided to use modals for most of the state transitions -- he had said maybe a new page for the enable flow but after playing with it, the complications of what to do if you go to that route directly etc just didn't feel worth it and the modal seems to work ok
* Set up a framework for the BackupCodesModal so that it can be worked on separately without needing to touch the same files, for the most part.

Still have work to do to get this working all the way. :)

### DIFF
--- a/src/pages/profile/components/EnableTfaModal.js
+++ b/src/pages/profile/components/EnableTfaModal.js
@@ -1,0 +1,90 @@
+import React, { Component } from 'react';
+import { QRCode } from 'react-qr-svg';
+import Modal from 'src/components/modals/Modal';
+import { Panel, Button, TextField, Grid } from '@sparkpost/matchbox';
+
+export default class EnableTfaModal extends Component {
+
+  state = {
+    code: '',
+    showErrors: false
+  };
+
+  componentDidUpdate(oldProps) {
+    if (this.props.enabled && this.props.open) {
+      this.props.onClose();
+    }
+    if (oldProps.open && !this.props.open) {
+      this.setState({
+        code: '',
+        showErrors: false
+      });
+    }
+    if (!oldProps.toggleError && this.props.toggleError) {
+      this.setState({
+        showErrors: true
+      });
+    }
+  }
+
+  handleInputChange = ({ target }) => {
+    this.setState({ code: target.value });
+  }
+
+  enable = () => {
+    this.props.enable(this.state.code);
+  }
+
+  render() {
+    const { open, onClose, secret, username, togglePending, toggleError } = this.props;
+
+    if (!secret) {
+      return null;
+    }
+
+    const qrData = `otpauth://totp/${username}?secret=${encodeURIComponent(secret)}&issuer=SparkPost`;
+
+    return (
+      <Modal open={open}>
+        <Panel title='Enable Two-Factor Authentication' accent actions={[{ content: 'Close', onClick: onClose }]}>
+          <Panel.Section>
+            To enable 2FA, you'll need to have a 2FA auth app installed on your phone or tablet (examples include Google Authenticator, Duo Mobile, and Authy).
+          </Panel.Section>
+          <Panel.Section>
+            <Grid>
+              <Grid.Column xs={12} md={6}>
+                <p><strong>Step 1: Configure your 2FA app</strong></p>
+                <p>Most apps will let you get set up by scanning our QR code from within the app. If you prefer, you can type in the key manually.</p>
+                <p><code>KEY: {secret}</code></p>
+              </Grid.Column>
+              <Grid.Column xs={12} md={6} style={{ textAlign: 'center' }}>
+                <QRCode
+                  bgColor="#FFFFFF"
+                  fgColor="#000000"
+                  level="Q"
+                  style={{ width: 256 }}
+                  value={qrData}
+                />
+              </Grid.Column>
+            </Grid>
+          </Panel.Section>
+          <Panel.Section>
+            <Grid>
+              <Grid.Column xs={12} md={6}>
+                <p><strong>Step 2: Enter a 2FA code</strong></p>
+                <p>Generate a code from your newly-activated 2FA app to confirm that you're all set up.</p>
+                <TextField error={(this.state.showErrors && toggleError) ? 'Problem verifying your code, please try again' : ''} placeholder='Enter your code' onChange={this.handleInputChange} value={this.state.code} />
+              </Grid.Column>
+            </Grid>
+          </Panel.Section>
+          <Panel.Section>
+            <Button primary onClick={this.enable}>
+              {togglePending ? 'Verifying Code...' : 'Enable 2FA'}
+            </Button>
+          </Panel.Section>
+        </Panel>
+      </Modal>
+    );
+  }
+
+}
--- a/src/pages/profile/components/EnableTfaModal.js
+++ b/src/pages/profile/components/EnableTfaModal.js
@@ -1,7 +1,8 @@
 import React, { Component } from 'react';
 import { QRCode } from 'react-qr-svg';
 import Modal from 'src/components/modals/Modal';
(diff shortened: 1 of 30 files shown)


rajumsys:
Was curious about new look. So checked it out. So much better (👍)
- Getting this on console
> Warning: `value` prop on `input` should not be null. Consider using an empty string to clear the component or `undefined` for uncontrolled components.
(let me know if you need full stack trace)
- what is this?
> code: input y u no controlled

seems that suffix the code I put in the text field
- I guess you are aware but just fyi, linked support article will need to be updated

jonambas:
> linked support article will need to be updated

There are going to be a TON of support articles we'll need to update, not just this one :D

src/pages/profile/components/EnableTfaModal.js
```diff
@@ -0,0 +1,89 @@
...
+        <Panel title='Enable Two-Factor Authentication' accent actions={[{ content: 'Close', onClick: onClose }]}>
+          <Panel.Section>
+            To enable 2FA, you'll need to have a 2FA auth app installed on your phone or tablet (examples include Google Authenticator, Duo Mobile, and Authy).
+          </Panel.Section>
+          <Panel.Section>
+            <Grid>
+              <Grid.Column xs={12} md={6}>
+                <p><strong>Step 1: Configure your 2FA app</strong></p>
+                <p>Most apps will let you get set up by scanning our QR code from within the app. If you prefer, you can type in the key manually.</p>
+                <p><code>KEY: {secret}</code></p>
```
jonambas: this looks good to me:

```html
<p>
  <div><small>2FA Key</small></div>
  <strong><code>{secret}</code></strong>
</p>
```

or this

```html
<p><strong><Icon name="Key"/> <code>{secret}</code></strong></p>
```

jasonrhodes:
@rajumsys those first 2 questions are related, I can't get that controlled input to work and I realize I don't know how they are supposed to work. Need to investigate more.

>seems that suffix the code I put in the text field

What's that mean?

src/pages/profile/components/EnableTfaModal.js
```diff
@@ -0,0 +1,89 @@
...
+  render() {
+    const { open, onClose, secret, togglePending, toggleError } = this.props;
+
+    if (!secret) {
+      return null;
+    }
+
+    return (
+      <Modal open={open}>
+        <Panel title='Enable Two-Factor Authentication' accent actions={[{ content: 'Close', onClick: onClose }]}>
```
jonambas: We don't typically use panel actions for closing modals, usually a button at the bottom of a panel floated to the right

src/pages/profile/components/EnableTfaModal.js
```diff
@@ -0,0 +1,89 @@
...
+                <p><strong>Step 1: Configure your 2FA app</strong></p>
+                <p>Most apps will let you get set up by scanning our QR code from within the app. If you prefer, you can type in the key manually.</p>
+                <p><code>KEY: {secret}</code></p>
+              </Grid.Column>
+              <Grid.Column xs={12} md={6} style={{ textAlign: 'center' }}>
+                <QRCode
+                  bgColor="#FFFFFF"
+                  fgColor="#000000"
+                  level="Q"
+                  style={{ width: 256 }}
```
jonambas: Maybe decrease this a little bit.
`style={{ width: 230 }}` looks okay with 7/5 grid columns instead of 6/6

rajumsys:
screen shot 2018-02-07 at 3 30 31 pm

src/pages/profile/components/EnableTfaModal.js
```diff
@@ -0,0 +1,89 @@
...
+                  value={secret}
+                />
+              </Grid.Column>
+            </Grid>
+          </Panel.Section>
+          <Panel.Section>
+            <Grid>
+              <Grid.Column xs={12} md={6}>
```
jonambas: `md={6}` can be removed. Since this is a modal, should try to decrease as much height as possible

jasonrhodes:
@rajumsys oh yeah, see the "WIP" label :)

src/pages/profile/components/TfaManager.js
```diff
@@ -0,0 +1,95 @@
...
+      content: 'Generate New Backup Codes',
+      onClick: () => this.setState({ openModal: 'backupCodes' })
+    }, {
+      content: 'Disable 2FA',
+      onClick: () => this.setState({ openModal: 'disable' })
+    }];
+
+    return (
+      <Panel sectioned title='2FA' actions={enabled ? enabledActions : disabledActions}>
+        <p>Two-factor authentication is currently <strong>{enabled ? 'enabled' : 'disabled'}</strong>.</p>
```
jonambas: Not really sure how to structure this, but this looks okay to me

```js
<Panel sectioned title='Two-factor Authentication' actions={enabled ? enabledActions : disabledActions}>
  <LabelledValue label='Status' value={enabled ? 'Enabled' : 'Disabled'}></LabelledValue>
```

src/pages/profile/components/EnableTfaModal.js
jonambas: actually Grid and Grid.Column can probably be removed

jonambas:
Oh also if we leave this out 
```
Store the above key in a secure place. You'll need it to activate a new device should your current device be lost or otherwise inaccessible.
```
We should maybe suggest they create backup codes somewhere in the enable modal

src/pages/profile/components/EnableTfaModal.js
jasonrhodes: `md={6}` is how we decrease height, what am I missing?

jonambas: `Generate a code from your newly-activated 2FA app to confirm that you're all set up.` wraps over to another line because of width.

I don't care too much, being nit-picky

jasonrhodes: Yeah I hear you, I was thinking this was the other group and not using that would force the QR code down to its own section. Is there a way to make a `TextField` not be 100% width besides a CSS style? That was my main gripe with going full width in this section.

src/pages/profile/components/TfaManager.js
jonambas: LabelledValue with a backup warning
```js
<LabelledValue label='Status'>
  <h6>Enabled</h6>
  <p><Icon name="ErrorOutline" size={15}/> We recommend creating backup codes should your current device be lost or otherwise inaccesible.</p>
</LabelledValue>
```

src/pages/profile/components/EnableTfaModal.js
jonambas: No not without a grid. Its fine as is, doesn't really make a big difference

src/pages/profile/components/EnableTfaModal.js
```diff
@@ -0,0 +1,91 @@
...
+                />
+              </Grid.Column>
+            </Grid>
+          </Panel.Section>
+          <Panel.Section>
+            <Grid>
+              <Grid.Column xs={12} md={7}>
+                <h6>Step 2: Enter a 2FA code</h6>
+                <p>Generate a code from your newly-activated 2FA app to confirm that you're all set up.</p>
+                <TextField error={(this.state.showErrors && toggleError) ? 'Problem verifying your code, please try again' : ''} placeholder='Enter your code' onChange={this.handleInputChange} value={this.state.code} />
```
jonambas: Needs a `label`

jasonrhodes: Happy to add one but the `<p>` above + the placeholder sort of already says it all, so let me know how you want to handle/re-arrange that?

jasonrhodes:
This got mergd into the other ticket so I'm closing.

PR #273, opened 2018-02-07
FAD-5861 Webhooks Fix Up

jonambas:


src/components/reduxFormWrappers.js
```diff
@@ -55,6 +55,7 @@ export const CheckboxWrapper = ({ input, meta, ...rest }) => {
     <Checkbox
       id={input.name}
       {...input}
+      checked={input.value}
```
jonambas: Not sure how any of our checkboxes worked before this change

jest.config.json
```diff
@@ -19,7 +19,7 @@
   ],
   "coverageThreshold": {
     "global": {
-      "statements": 53
+      "statements": 80
```
jonambas: increased test coverage threshold

rajumsys: is this the current coverage rate?

src/pages/webhooks/ListPage.js
```diff
@@ -24,26 +25,36 @@ export class WebhooksList extends Component {
 
   getColumns = () => {
     const { hasSubaccounts } = this.props;
-    const columns = [{ label: 'Name', sortKey: 'name' }, 'Target'];
+    const columns = [
+      { label: 'Name', sortKey: 'name' },
+      'Target',
+      { label: 'Last Success', sortKey: 'last_successful', width: '18%' },
```
rajumsys: i'm little unsure what should be ideal case of sorting here. i see some records don't have last success/failure. so when sorting descending, records w/o that comes first and then rest are sorted correctly. should we put the records without these records at the end (i mean always at the end; for both asc & desc)

jonambas: Yeah agreed, is there a way to do that through `sortKey`?

PR #275, opened 2018-02-08
Feature/fad 5603

bizob2828:


src/pages/profile/components/tests/2
```diff
@@ -0,0 +1,59 @@
...
+  });
+
+  it('should call generate when generateCodes is called', () => {
+    wrapper.setState({ password: 'Password1' });
+    instance.generateCodes();
+    expect(instance.props.generate).toHaveBeenCalledWith('Password1');
+  });
+
+  it('should handles changes to password', () => {
+    wrapper.find('TextField').simulate('change', { target: { value: 'pass' } });
```
bizob2828: missing assertion

src/pages/profile/components/tests/2
```diff
@@ -0,0 +1,59 @@
+import { shallow } from 'enzyme';
```
bizob2828: this can be deleted

src/pages/profile/components/BackupCodesModal.js
```diff
@@ -1,20 +1,82 @@
...
+      codes
+    } = this.props;
+    const hasCodes = codes.length > 0;
+
+    return (
+      <Modal open={open}>
+        <Panel title='Generate Two-factor Backup Codes' accent>
+          <form onSubmit={(e) => e.preventDefault()}>
+            <Panel.Section>
+              { !hasCodes && <Banner status="warning" >
```
jasonrhodes: should be `{ hasCodes` 

screen shot 2018-02-14 at 2 47 55 pm

src/pages/profile/components/BackupCodesModal.js
```diff
@@ -1,20 +1,82 @@
...
+    const { pending, onClose } = this.props;
+    const hasCodes = this.props.codes.length > 0;
+    return (
+      <div>
+        { !hasCodes && <Button type='submit' disabled={pending} primary onClick={this.generateCodes}>
+          { pending ? 'Generating...' : 'Generate' }
+        </Button> }
+        { hasCodes && <Button primary onClick={onClose}>Close</Button> }
+        { !hasCodes && <Button onClick={onClose} className={styles.Cancel}>Cancel</Button> }
+      </div>
```
jasonrhodes: Might be good to switch these for an if/else that returns the forks separately in this function, since we're out here in a render* function already? Might be clearer to see what's happening.

jasonrhodes:
Think we might want the backup reducer to mark `hasCodes` as `true` on generate success, so that if you click to generate again it knows you have codes the 2nd time.

Also: are we displaying the codes forever? We might need more logic to handle those cases, we should merge this into the other TFA ticket to find out.

src/pages/profile/components/BackupCodesModal.js
bizob2828: 🤔 dafuq. ok let me change

PR #276, opened 2018-02-09
FAD-6194 refactor report refresh cycle

jasonrhodes:
Re: some performance concerns with how we retrieve the Typeahead cache, this change stops us from refreshing on each report page change and only initializes the cache once.

This PR doesn't change "refreshing on date range change" behavior, but I think this is far less strain on our DB than the existing UI. I did a simple search for 3-4 filters in the existing UI and generated 116 requests, 20 of which were just for the `metrics/domains` endpoint. The new UI will only make one set of metrics requests for each new date range, then use that cache for every filter search.

screen shot 2018-02-08 at 10 30 29 pm

### DIFF
--- a/src/actions/reportFilters.js
+++ b/src/actions/reportFilters.js
@@ -32,14 +32,14 @@ export function initTypeaheadCache(params) {
     if (Array.isArray(metrics.domains)) {
       return;
     }
-    const requests = metricLists.map((list) => dispatch(list(params)));
+    const requests = []; //metricLists.map((list) => dispatch(list(params)));
     requests.push(dispatch(listTemplates()), dispatch(listSubaccounts()), dispatch(listSendingDomains()));
     return Promise.all(requests);
   };
 }
 
-// TODO: rename refreshTypeaheadCache
-export function refreshTypeaheadCache(params) {
+export function refreshTypeaheadCache(options) {
+  const params = getQueryFromOptions(options);
   return (dispatch) => {
     const requests = metricLists.map((list) => dispatch(list(params)));
     return Promise.all(requests);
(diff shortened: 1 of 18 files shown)


src/actions/reportFilters.js
```diff
@@ -19,15 +19,27 @@ const metricLists = [
...
+ * Returns a thunk that gets the metrics filter lists to populate cache,
+ * plus templates, subaccounts, and sending domains
+ *
+ * The thunk is a no-op if the cache has already been initialized, i.e.
+ * metrics.domains is an array and not undefined
+ */
+export function initTypeaheadCache(params) {
+  return (dispatch, getState) => {
+    const { metrics } = getState();
+    if (Array.isArray(metrics.domains)) {
```
bizob2828: can you explain why this is bailing.  I assume this is just an indicator it already has something from the metrics api but a comment here would be good for future peeps

src/actions/reportFilters.js
```diff
@@ -19,15 +19,27 @@ const metricLists = [
...
+      return;
+    }
     const requests = metricLists.map((list) => dispatch(list(params)));
     requests.push(dispatch(listTemplates()), dispatch(listSubaccounts()), dispatch(listSendingDomains()));
     return Promise.all(requests);
   };
 }
 
-export function refreshListsByTime(params) {
+// TODO: rename refreshTypeaheadCache
```
bizob2828: don't forget to rename

src/actions/reportFilters.js
```diff
@@ -19,15 +19,27 @@ const metricLists = [
...
-export function refreshTypeaheadCache(params) {
-  return (dispatch) => {
+/**
+ * Returns a thunk that gets the metrics filter lists to populate cache,
+ * plus templates, subaccounts, and sending domains
+ *
+ * The thunk is a no-op if the cache has already been initialized, i.e.
+ * metrics.domains is an array and not undefined
+ */
+export function initTypeaheadCache(params) {
```
bizob2828: the initial call isn't scoping this with from/to and getting all.  if we are refreshing when date ranges change would be good to figure out how to set this correclty

bizob2828: this also doesn't appear to be refreshing when date ranges change

jasonrhodes: This won't ever refresh, that's why I changed the name. We call this from `cDM` in each report page to make sure we have a cache -- it bails if we discover we've already loaded it (from another `cDM` ... the way I'm checking if we have the cache loaded is hacky for now but it'll do. Would be better to fire an action during init that sets a redux flag "typeaheadCacheInitialized" or something, then check for that here instead of assuming that `metrics.domains` existence means we loaded it already. Think it's worth doing that now?

src/actions/reportFilters.js
jasonrhodes: oops, already renamed, need to remove comment thanks

PR #277, opened 2018-02-09
Turn off billing for enterprise accounts (FAD-6186)

jasonrhodes:
Uses condition helpers to turn off billing routes if you are on plan 'ent1' or on service_level 'enterprise'

Also turned off IP Pools routes and Dashboard if these same conditions are met, if that makes sense?


jasonrhodes:
More I think about it, I don't think we want to turn off Dashboard or IP Pools if they are on service level enterprise, so I'll just revert those changes until we understand them more. I'll find out if we want to turn off IP Pools at least for ent1 plan holders.

src/config/routes.js
```diff
@@ -340,31 +339,31 @@ const routes = [
...
     component: billing.ChangePlanPage,
-    condition: hasGrants('account/manage'),
+    condition: composeConditions(hasGrants('account/manage'), notEnterprise()),
     layout: App
   },
   {
     path: '/account/ip-pools',
     component: ipPools.ListPage,
-    condition: hasGrants('ip_pools/manage'),
+    condition: composeConditions(hasGrants('ip_pools/manage'), notEnterprise()),
```
bizob2828: we don't want this.  i'd revert it.  some enterprise tenants already have ip pools because they have been migrated to shared MTA. in an effort to have one less feature flag that has to be flipped just remove this check

src/config/routes.js
```diff
@@ -195,7 +194,7 @@ const routes = [
   {
     path: '/templates',
     component: templates.ListPage,
-    condition: composeConditions(hasGrants('templates/modify'), ({ account }) => account.status === 'active'),
+    condition: hasGrants('templates/modify'),
```
bizob2828: this was done intentionally.  we don't want accounts modifying templates if they are suspended

jasonrhodes: I added this and only did it as a proof of concept for `composeConditions` when I first rolled it out.

src/config/routes.js
jasonrhodes: Yep, already have a few comments here saying I'm reverting everything but the billing conditions.

src/config/routes.js
bkemper: Wouldn't this condition be on most routes?

jasonrhodes: Many for sure, let's punt on that for now -- the API calls will fail if they try to update. Can solve holistically.

PR #283, opened 2018-02-10
FAD-6203 Refactor refresh cycle for all report pages

jasonrhodes:
We're doing some baaaaad things. Must stop.


jasonrhodes:
This is obviously a bit larger than we like to do at once but this flow was a big mess and made me nervous if we needed to tune calls to the metrics API etc., so I think this puts us in a much better place. Here are the main themes:

* **Dates, filters, current location updates, and share modal syncing is all done in `<Filters>` now.** I think we have more work to do to separate out the link update/share modal and have that done via Redux, but we're in a much better place to do that now that it's centralized. There is some hackishness to this to allow for the summary chart to work, but otherwise it's pretty clean.

* **Report range updates and typeahead cache refreshes are managed in one central location.** 

* **Report actions are consistent across the board now, as a single `refreshReport` function** (still finishing those renames up ATM) which makes multiple calls under the hood for you, instead of pushing that `Promise.all()` into the components.

* **Unnecessary secondary actions are removed.** The summary chart needs this for now because it gets data using a metrics action creator and then dispatches a secondary action with the results and a few other details as well (there are other ways we could do this too but that's why it happens). The other report action creators copied this pattern, but they don't have anything else to add to the secondary action, so I removed them and made the reducers listen to the metrics action instead.

* **Formatting data should happen in selectors if possible.** Moved some formatting we were doing in `.then()` out of actions and into selectors instead.

src/actions/reportFilters.js
```diff
@@ -32,30 +32,20 @@ export function initTypeaheadCache(params) {
     if (Array.isArray(metrics.domains)) {
       return;
     }
-    const requests = metricLists.map((list) => dispatch(list(params)));
+    const requests = []; //metricLists.map((list) => dispatch(list(params)));
```
jasonrhodes: clean this up, init only needs to do the 3 bottom calls and can probably check to see if they've been loaded or not before requesting each

src/config/default.js
```diff
@@ -41,7 +41,7 @@ const config = {
     { time: 60 * 2, value: '5min', format: 'ha' },
     { time: 60 * 4, value: '15min', format: 'ha' },
     { time: 60 * 24 * 2, value: 'hour', format: 'ha' },
-    { time: 60 * 24 * 7, value: '12hr', format: 'MMM Do' },
+    { time: 60 * 24 * 7, value: 'day', format: 'MMM Do' },
```
jasonrhodes: This fixes a bug Nancy mentioned in the #phoenix-feedback channel where the last 7 days view had a weird looking UI ... I checked the existing app and we use 'day' precision for this and that fixes it.

src/index.js
```diff
@@ -12,12 +12,19 @@ import './critical.scss';
 import './index.scss';
 import App from './App';
 
+const debug = () => (next) => (action) => {
```
jasonrhodes: remove (redux dev tools are crashing a lot less for me now that i got rid of some weird action duplication etc)

src/index.js
```diff
@@ -12,12 +12,19 @@ import './critical.scss';
...
+  next(action);
+};
+
 // necessary for redux devtools in development mode only
 const composeEnhancers = process.env.NODE_ENV !== 'production' && window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose; // eslint-disable-line no-mixed-operators
 const store = createStore(
   rootReducer,
   composeEnhancers(
     applyMiddleware(thunk),
+    applyMiddleware(debug),
```
jasonrhodes: remove

src/pages/reports/components/Filters.js
```diff
@@ -1,68 +1,138 @@
...
 import styles from './Filters.module.scss';
 
 export class Filters extends Component {
+  state = {
+    modal: false,
+    query: {}
+  }
+
+  componentDidMount() {
+    this._mounted = true;
```
jasonrhodes: This fixes the redirect bug

src/reducers/reportFilters.js
```diff
@@ -1,27 +1,25 @@
...
   switch (action.type) {
-    case 'SET_EXACT_TIME':
-      return { ...state, ...action.payload };
+    // case 'SET_EXACT_TIME':
+    //   return { ...state, ...action.payload };
 
-    case 'SET_RELATIVE_TIME':
-      return { ...state, ...action.payload };
+    // case 'SET_RELATIVE_TIME':
+    //   return { ...state, ...action.payload };
```
jasonrhodes: I think I can remove these

src/selectors/acceptedReport.js
```diff
@@ -0,0 +1,38 @@
...
+  [selectAggregates, selectAttempts],
+  (aggregates, attempts) => {
+    if (!aggregates.count_accepted) {
+      return [];
+    }
+    return reshapeAttempts(attempts);
+  }
+);
+
+// TODO: Share the "accepted metrics" as a constant somehow since it's not dynamic
```
jasonrhodes: done, remove comment

src/selectors/bounceReport.js
```diff
@@ -0,0 +1,37 @@
...
+// TODO: decide if these should live in selectors or in eventual containers
+export const mapStateToProps = createStructuredSelector({
+  chartLoading: selectChartLoading,
+  tableLoading: selectTableLoading,
+  reasons: selectReasons,
+  aggregates: selectFormattedAggregates,
+  categories: selectReshapedClassifications,
+  types: selectBandTypes,
+  reportFilters: selectReportFilters
+});
```
jasonrhodes: This is new, I think I like it but I'm not sure

bizob2828:
this is requesting lists on every page change.  It does however have the right from/to filters applied.  But I thought we were going to request once then only request new list when ranges change?

src/actions/metrics.js
```diff
@@ -37,9 +37,9 @@ export function fetchMetricsIpPools(params = {}) {
   return fetch({ type, path, params });
 }
 
-export function fetchDeliverability(params = {}) {
+export function fetchDeliverability({ params, type }) {
```
bizob2828: you can remove all the `params = {}` looks like all places calling them are using the same method of getQueryFromOptions which builds an object

src/actions/engagementReport.js
```diff
@@ -1,32 +1,27 @@
 import * as metrics from 'src/actions/metrics';
 import { getQueryFromOptions } from 'src/helpers/metrics';
 
-export function getAggregateMetrics({ getMetrics = metrics.fetch } = {}) {
-  return (dispatch, getState) => {
-    const params = getQueryFromOptions(getState().reportFilters);
+export function refreshEngagementReport(updates = {}, { getMetrics = metrics.fetch } = {}) {
```
bkemper: 👍 On passing in updates!  I wanted to be able to do that.

src/helpers/metrics.js
```diff
@@ -73,13 +73,20 @@ export function getPrecisionType(precision) {
   return (indexedPrecisions[precision].time <= (60 * 24 * 2)) ? 'hours' : 'days';
 }
 
-export function getMetricsFromKeys(keys = []) {
+export function _getMetricsFromKeys(keys = []) {
```
bizob2828: why are you exporting this if it's "private"

src/helpers/metrics.js
```diff
@@ -112,6 +119,7 @@ export function buildCommonOptions(reportFilters, updates = {}) {
   return {
     ...reportFilters,
     ...updates,
+    ...getRelativeDates(reportFilters.relativeRange),
```
bizob2828: what's this for?

src/pages/reports/components/DateFilter.js
```diff
@@ -1,15 +1,16 @@
...
 import { subMonths, format } from 'date-fns';
 import { getStartOfDay, getEndOfDay, relativeDateOptions } from 'src/helpers/date';
+import { refreshReportOptions } from 'src/actions/reportFilters';
 import { Button, Datepicker, TextField, Select, Popover } from '@sparkpost/matchbox';
 import DateForm from './DateForm';
 import styles from './DateFilter.module.scss';
 
 export class DateFilter extends Component {
-  format = 'MMM DD, YY h:mma';
+  format = 'MMM D YYYY h:mma';
```
bizob2828: shouldn't we put this in constants?

src/pages/reports/components/DateForm.js
```diff
@@ -6,7 +6,7 @@ import { Grid, TextField, Icon } from '@sparkpost/matchbox';
 import styles from './DateForm.module.scss';
 
 export default class DateForm extends Component {
-  dayFormat = 'YYYY-M-D';
+  dayFormat = 'YYYY-MM-DD';
```
bizob2828: same, put in constants?

bizob2828: 🙏 this bothered me before

bizob2828: why is this YYYY to start and other ends with it

src/pages/reports/components/Filters.js
bizob2828: can you put a comment explaining why this is here

src/actions/engagementReport.js
```diff
@@ -1,32 +1,27 @@
...
+      })),
+      dispatch(getMetrics({
+        params: {
+          ...params,
+          metrics: 'count_clicked,count_raw_clicked_approx'
+        },
+        path: 'deliverability/link-name',
+        type: 'GET_ENGAGEMENT_LINK_METRICS'
+      }))
+    ]);
```
bkemper: Should we catch and ignore since each promise will have a change to dispatch the error action?

src/pages/reports/components/DateFilter.js
bkemper: There is a date and time format in config/default.js.  Should it be moved to constants and used here?

src/helpers/metrics.js
jasonrhodes: so I can test it without the memoize

src/pages/reports/components/Filters.js
bkemper: I don't think this is needed anymore.  The redirect bug I was seeing before was caused by `history.push` being called in a promise.  You fixed that by handing the update here in `componentDidUpdate`.

src/pages/reports/engagement/EngagementPage.js
```diff
@@ -1,33 +1,26 @@
...
-  onLoad = () => Promise.all([
-    this.props.getAggregateMetrics().catch(this.onLoadFail('Unable to load engagement chart.')),
-    this.props.getLinkMetrics().catch(this.onLoadFail('Unable to load link metrics.'))
-  ])
 
-  onLoadFail = (message) => (error) => {
-    const details = _.get(error, 'response.data.errors[0].message', error.message);
-    this.props.showAlert({ details, message, type: 'error' });
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.reportFilters !== this.props.reportFilters) {
```
bkemper: Could this logic be moved to `Filters` and exposed with a `onLoad` (or some event handler) prop?

src/pages/reports/summary/components/Table.js
```diff
@@ -22,7 +22,6 @@ export class Table extends Component {
 
   handleRowClick = (item) => {
     this.props.addFilters([item]);
-    this.props.refresh();
   }
 
   getColumnHeaders() {
```
bkemper: I don't know how you will want to handle this error, so here are the steps to reproduce.

1. Load Summary Report with [a shared link](http://localhost:3100/reports/summary?from=2018-02-13T20:52:00Z&metrics=count_targeted&metrics=count_rendered&metrics=count_accepted&metrics=count_bounce&range=day&to=2018-02-14T20:52:00Z)
2. Scroll down to the data table and select any option (e.g. "Sending Domains") from the "Group By" dropdown.

screen shot 2018-02-14 at 3 53 24 pm

The request for new data will fail because the `metrics` param is empty.

I triaged the issue a little.  I found that the summary `Table` component calls `getTableData` without `params` which forces a manual construction of params from state.

```
// Gets filters and metrics for params
if (!params) {
  params = getQueryFromOptions({ ...state.summaryChart, ...state.reportFilters });
}
```

Here the correct list of metric objects gets spread in from `summaryChart` state.

```
[
  { key: "count_targeted", ... },
  { key: "count_rendered", ... }, 
  ...
]
```

Then immediately overrode by a list of metric strings from `reportFilters` state.

```
["count_targeted", "count_rendered", "count_accepted", "count_bounce"]
```

The reason the resulting `metrics` is set to an empty string is because `getQueryFromOptions` was expecting objects not strings.

src/actions/engagementReport.js
jasonrhodes: I think we should wait and solve this holistically for all API calls.

src/actions/rejectionReport.js
```diff
@@ -1,64 +1,21 @@
...
-    maybeRefreshTypeaheadCache(dispatch, options, updates);
-    dispatch(refreshReportRange(options));
-
-    return dispatch(fetchDeliverability(params))
-      .then((aggregates) => {
-        dispatch(refreshRejectionMetrics(aggregates));
-      });
+import { getQueryFromOptions, getMetricsFromKeys } from 'src/helpers/metrics';
+
+const REJECTION_METRICS = getMetricsFromKeys([
```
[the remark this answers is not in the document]
jasonrhodes: Idea is that if any of those keys need to be translated for computation, it'll just happen -- if we only hard-code the keys, if you added a client-side-computed key later it wouldn't work. But if we know the list is made up of only real metrics, it's probably fine?

src/actions/acceptedReport.js
```diff
@@ -1,68 +1,18 @@
...
+export function refreshAcceptedReport(updates = {}) {
+  return (dispatch) => {
+    // TODO: consider having action creators that already pre-load their hard-coded metrics or as a selector
+    // TODO: This can probably become a selector for each query
+    const params = getQueryFromOptions({ ...updates, metrics: ACCEPTED_METRICS });
+
+    return Promise.all([
+      // TODO: Should this just be an action like getAcceptedAggregates with hard-coded metrics and type?
+      dispatch(fetchDeliverability({ type: 'GET_ACCEPTED_AGGREGATES', params })),
+      dispatch(fetchDeliveriesByAttempt(params))
```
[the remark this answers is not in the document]
jasonrhodes: Doesn't look like you can specify metrics for this endpoint: 
https://developers.sparkpost.com/api/metrics.html#metrics-deliveries-by-attempt-get

src/actions/messageEvents.js
```diff
@@ -3,14 +3,14 @@ import config from 'src/config';
 import _ from 'lodash';
```
bizob2828: i don't see tests for this

src/actions/suppressions.js
```diff
@@ -3,7 +3,7 @@ import config from 'src/config';
 import _ from 'lodash';
 import sparkpostApiRequest from 'src/actions/helpers/sparkpostApiRequest';
 import setSubaccountHeader from 'src/actions/helpers/setSubaccountHeader';
-import { refreshReportOptions } from 'src/actions/reportFilters';
+import { refreshReportOptions } from 'src/actions/reportOptions';
```
bizob2828: i don't see tests for this action

src/helpers/tests/date.test.js
```diff
@@ -49,6 +50,14 @@ describe('Date helpers', () => {
     'for a quarter ago': { range: '90days', subtractArgs: [90, 'days']}
   });
 
+  it('should return for a custom relative date range', () => {
+    expect(getRelativeDates('custom')).toEqual({ relativeRange: 'custom' });
+  });
+
+  it('should return an empty object for an unknown range', () => {
+    expect(getRelativeDates('einstein')).toEqual({});
```
bkemper: 👍

src/pages/sendingDomains/components/tests/EditBounce.test.js
```diff
@@ -5,7 +5,7 @@ import config from 'src/config';
 
 jest.mock('src/config', () => ({
   zuora: {}, //axiosInstance throws without this
-  authentication: { cookie: {}}, //authCookie throws without this
+  authentication: { cookie: {}}, //authCookie throws without this,
```
bkemper: To avoid the stubbing, should we just spread in the actual values?

```
jest.mock('src/config', () => ({
  ...require.requireActual('src/config'),
}));
```

jasonrhodes: `require.requireActual.noSeriouslyPlz()` ugh

but yeah probably a good idea so long as that never breaks in weird magical ways

PR #286, opened 2018-02-12
FAD-6181 added sending domains tab, fixing bug in api keys tab, need …

bizob2828:
I had to abstract some sending domains components so they can be shared on list and then the sending domains tab in sending domains edit

I ended up fixing two bugs while testing things:
 * Editing an API key from subaccounts details tab
 * Creating a subaccount with ip pool(was passing nothing, edit was ok)
 * I tried to add tooltip to grants but CSS is f'ed. I asked jon to help and if it's simple we'll keep otherwise I'll restore
 * I also removed duplicate grantCheckboxes code and updated so both places use one component
 * there was some PermissionsCheckList that was also not being used anywhere, but it had tests!


src/pages/subaccounts/DetailsPage.js
```diff
@@ -41,8 +61,9 @@ export class DetailsPage extends Component {
   }
 
   render() {
-    const { loading, location, tabs } = this.props;
-    const selectedTab = location.pathname.endsWith('keys') ? 1 : 0;
+    const { loading, location, id } = this.props;
+    const tabs = buildTabs(id);
+    const selectedTab = tabs.findIndex(({ to }) => location.pathname === to);
```
bizob2828: write some tests for this

src/components/grantBoxes/GrantsCheckboxes.js
```diff
@@ -12,15 +12,16 @@ const cx = classnames.bind(styles);
...
-    <Field
-      key={grant.key}
-      name={`grants[${grant.key}]`}
-      label={grant.label}
-      component={CheckboxWrapper}
-      type="checkbox"
-      normalize={(value) => (value ? 'true' : '')} // TODO is this necessary? otherwise we get proptype warnings
-    />
+    // TODO: fix css
+    <Tooltip key={grant.key} dark content={grant.description}>
```
bizob2828: tried to do this but css is f'ed. I can revert unless someone has a clever fix _hint_ @jonambas _hint_

jonambas: I'd check MetricsModal in the summary report and maybe try the same css

src/components/reduxFormWrappers.js
```diff
@@ -55,7 +55,7 @@ export const CheckboxWrapper = ({ input, meta, ...rest }) => {
     <Checkbox
       id={input.name}
       {...input}
-      checked={input.value}
+      checked={input.value || false}
```
bizob2828: i can fix this by tweaking the initial values of the grantsRadio

src/selectors/subaccounts.js
```diff
@@ -14,6 +13,7 @@ const formatSubaccount = ({ compliance_status = 'active', status = 'active', ...
 };
 
 export const hasSubaccounts = (state) => state.currentUser.has_subaccounts;
+export const getSubaccount = (state) => state.subaccount;
```
bizob2828: we don't need that

src/pages/subaccounts/DetailsPage.js
```diff
@@ -5,12 +5,13 @@ import { Page, Tabs } from '@sparkpost/matchbox';
...
 import { getSubaccount } from 'src/actions/subaccounts';
 import { ApiKeySuccessBanner } from 'src/components';
-import { selectSubaccount, selectDetailTabs } from 'src/selectors/subaccounts';
+import { selectSubaccount } from 'src/selectors/subaccounts';
 import { listPools } from 'src/actions/ipPools';
 import { listApiKeys, hideNewApiKey } from 'src/actions/api-keys';
 
 import ApiKeysTab from './components/ApiKeysTab';
 import EditTab from './components/EditTab';
+import SendingDomainsTab from './components/SendingDomainsTab';
```
gwalls: As discussed irl, load sending domains on `componentDidMount` of the `DetailsPage`

PR #293, opened 2018-02-14
FAD-6206 Onboarding

jonambas:
Supporting Matchbox PR here for the `CodeBlock` component: sparkpost/matchbox#109

### What I Did:
- **SmtpDetails** - Moved to shared components folder, which is reused in onboarding
- **SmtpDetails** - shows a CopyField if provided an apiKey
- **AutoKeyGenerator** - a container component for the two onboarding pages that require an apiKey
- **CenteredLogo** - component for the Sparkpost logo. Should reuse in login, and join as well
- **Steps** - renders onboarding completion based on location pathname. Will need to tweak when final urls & all join pages are up.

### Routes:
- `/super-hidden-route/sending-domain`
- `/super-hidden-route/email`
- `/super-hidden-route/email/smtp`
- `/super-hidden-route/email/api`
- using `email` there since those 3 pages share the same step


src/config/routes.js
```diff
@@ -375,6 +377,22 @@ const routes = [
     component: ipPools.EditPage,
     condition: hasGrants('ip_pools/manage'),
     layout: App
+  },
+  {
+    path: '/super-hidden-route/sending-domain',
```
bizob2828: you were out yesterday, but i asked we gate these with a feature flag condition.  I think it's fine to keep the routes what they will be for real as long as we have a feature flag.  Previously this was environment === 'sp'. we are doing away with that. I'd sync up with @rajumsys but has_signup as the feature flag makes sense to me. then we can add in A-E for spc and spc-eu later. but not we can add for uat and staging to verify it

src/helpers/onboarding.js
```diff
@@ -0,0 +1,30 @@
...
+  return `curl -X POST
+  ${getTransmissionsUri()}
+  -H "Authorization: ${apiKey}"
+  -H "Content-Type: application/json"
+  -d '{
+    "options": {
+      "sandbox": true
+    },
+    "content": {
+      "from": "sandbox@sparkpostbox.com",
```
bizob2828: i'd make this a config just in case we change in the future, similar to the other emails we ref in our app

src/pages/onboarding/SendingDomainPage.js
```diff
@@ -0,0 +1,59 @@
...
+import SkipLink from './components/SkipLink';
+import { required, domain } from 'src/helpers/validation';
+import { LINKS } from 'src/constants';
+
+export class SendingDomainPage extends Component {
+  handleDomainCreate = (values) => {
+    const { createDomain, showAlert, history } = this.props;
+
+    return createDomain(values).then(() => {
+      history.push('/super-hidden-route/email');
```
bizob2828: update ref once you change the route

src/pages/onboarding/SendingDomainPage.js
```diff
@@ -0,0 +1,59 @@
...
+                label='Domain Name'
+                placeholder='email.example.com'
+                name='domain'
+                validate={[required, domain]}
+                disabled={submitting}
+              />
+            </Panel.Section>
+            <Panel.Section>
+              <Button primary submit disabled={submitting}>{submitting ? 'Adding Domain...' : 'Add Domain'}</Button>
+              <SkipLink to='/super-hidden-route/email'>Skip for now</SkipLink>
```
bizob2828: update ref after route is updated

src/pages/onboarding/SmtpOrApiPage.js
```diff
@@ -0,0 +1,43 @@
...
+const SmtpOrApiPage = () => (
+  <Fragment>
+    <CenteredLogo />
+    <Panel accent title='Sending Your First Email'>
+      <Panel.Section>
+        <p>How will you send email with us?</p>
+        <Grid>
+          <Grid.Column xs={12} md={6}>
+            <OptionTile
+              to='/super-hidden-route/email/smtp'
```
bizob2828: update ref after route is updated

src/pages/onboarding/SmtpOrApiPage.js
```diff
@@ -0,0 +1,43 @@
...
+              wrapper={Link}
+              label='SMTP'
+              content='Set up your own mail server to send through our SMTP relay service.'
+            />
+          </Grid.Column>
+          <Grid.Column xs={12} md={6}>
+            <OptionTile
+              to='/super-hidden-route/email/api'
```
bizob2828: update ref after route is updated

src/helpers/tests/onboarding.test.js
```diff
@@ -0,0 +1,11 @@
+import * as onboarding from '../onboarding';
+
+describe('Onboarding helpers', () => {
+  it('Should get correct transmission uri', () => {
```
bizob2828: missing a test that omits the protocol so it hits that reg ex if statement

bizob2828:
* There's a failing test
 * Also, if I'm on the onboarding and hit the back button there's this weird step where the sparkpost logo takes up the whole screen then renders the dashboard

src/components/smtpDetails/SmtpDetails.js
```diff
@@ -0,0 +1,51 @@
...
+    return (
+      <Fragment>
+        <p>{smtpDescContent}</p>
+        <LabelledValue label='Host' value={smtpAuth.host}/>
+        <LabelledValue label='Port' value={smtpAuth.port}/>
+        { config.smtpAuth.alternativePort && <LabelledValue label='Alternative Port' value={`${smtpAuth.alternativePort}`}/> }
+        <LabelledValue label='Authentication' value='AUTH LOGIN'/>
+        <LabelledValue label='Encryption' value='STARTTLS'/>
+        <LabelledValue label='Username' value={smtpAuth.username}/>
+        <LabelledValue label='Password'>{passwordContent}</LabelledValue>
```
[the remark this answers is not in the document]
jonambas: this is intended, to avoid wrapping. old UI does the same thing, but Copyfield also selects all on focus, and gives better feedback on copy click

src/pages/onboarding/SendingDomainPage.js
jasonrhodes: fwiw it's probably better to do these history.pushes in `cDU`, once the reducer has changed some state after a create -- otherwise if you navigate away while the create is happening you'll have the same bug we had in reports

might not be a big deal here bc it's harder to "navigate away" but something to keep in mind

jonambas:
@bizob2828 i'm not seeing this
> Also, if I'm on the onboarding and hit the back button there's this weird step where the sparkpost logo takes up the whole screen then renders the dashboard

react-router should be handling back/forward, usually what you saw means react router isn't picking up the event

src/config/default.js
```diff
@@ -35,6 +35,7 @@ const config = {
     allow_anyone_at_verification: false
   },
   gaTag: 'no-default-set',
+  hasSignup: true,
```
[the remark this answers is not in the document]
bizob2828: I think we're trying to keep stuff in the feature flags object too.  That way A-D already injects that object.  I'd also sync up with @rajumsys to make sure he's using the same flag

PR #294, opened 2018-02-15
FAD-5916 signup form

rajumsys:
To test join page, flip the value of `hasSignup` on `config/default.js` to `true`. 

Todos:
- [x] retrieve salesforce_data
- [x] retrieve sfdcid
- [ ] ~retrieve analytics session id~
- [x] handle and retrieve recaptcha
- [x] track signup after signup is completed
- [x] fix TOU checkbox alignment when validation error is shown (Jon allowed children for label in matchbox. so we don't need separate element for it)
- [x] review recaptcha lib loading  (@bkemper is working with similar, so let's sync. Update: we both agreed to use `react-load-script`) 
- [x] Enable eslint
- [x] expose this to `join` route but make it configurable and hidden for now
- [x] Tests 
- [x] Tests for `PublicRoute`
- [x] A-E/A-D changes


src/config/default.js
```diff
@@ -97,7 +97,22 @@ const config = {
...
+    cookieName: 'attribution',
+    cookieDuration: 60 * 24 * 30,
+    // cookieDomain: '.sparkpost.com' //TODO change it before merge
+    cookieDomain: 'localhost'
+  },
+  salesforceDataParams: ['sfdcid', 'src', 'utm_source', 'utm_medium', 'utm_campaign', 'utm_content', 'utm_term'],
+  links: {
+    submitTicket: 'https://support.sparkpost.com/customer/portal/emails/new'
+  },
+  hasSignup: false
```
bizob2828: commented on this in @jonambas . If you put within the feature flags object, it's more centralized and requires 0 A-D changes to add this to the template.  When we want to enable for a tenant, we just add to that featureFlags object within A-E for each tenant

rajumsys: fair. i'll move it to `featureFlags`.

src/pages/join/JoinPage.js
```diff
@@ -0,0 +1,153 @@
+/* eslint-disable */
```
jonambas: Specify max length (or remove)

src/config/default.js
```diff
@@ -97,6 +98,20 @@ const config = {
...
+  },
+  attribution: {
+    cookieName: 'attribution',
+    cookieDuration: 60 * 24 * 30,
+    // cookieDomain: '.sparkpost.com' //TODO change it before merge
+    cookieDomain: 'localhost'
+  },
+  salesforceDataParams: ['sfdcid', 'src', 'utm_source', 'utm_medium', 'utm_campaign', 'utm_content', 'utm_term'],
+  links: {
+    submitTicket: 'https://support.sparkpost.com/customer/portal/emails/new'
```
jonambas: Move this to `constants/index.js`

src/pages/join/JoinPage.js
```diff
@@ -0,0 +1,153 @@
...
+    const attributionData = this.getAndSetAttributionData();
+
+    const signupData = { ...values, salesforce_data: attributionData, sfdcid: attributionData.sfdcid };
+
+    return register(signupData)
+      .then((accountData) => {
+        this.trackSignup();
+        return authenticate(accountData.username, values.password);
+      })
+      .then(() => this.props.history.push(DEFAULT_REDIRECT_ROUTE));
```
jonambas: May need to move this history.push out of the promise and into `componentDidMount` to avoid redirecting unless this component is still mounted (like the reports redirect bug we had before).

I implemented this in the onboarding/sending-domain form if you want to take a look. You'll need to manually throw redux-form `SubmissionError` so that `submitFailed` and `submitSucceeded` are correct

src/pages/join/JoinPage.js
```diff
@@ -0,0 +1,153 @@
...
+    return (
+      <Error error={this.handleSignupFailure(errors)} />
+    );
+  };
+
+  render() {
+    const { createError } = this.props.account;
+    return (
+      <div>
+        <div className={styles.LogoWrapper}>
```
jonambas: can use `CenteredLogo` component

src/pages/join/JoinPage.js
```diff
@@ -0,0 +1,153 @@
...
+const { attribution, salesforceDataParams, links, gaTag } = config;
+
+
+export class JoinPage extends Component {
+
+  state = {
+    formData: {}
+  };
+
+  injectGA() {
```
jonambas: could maybe move this function and also `trackSignup` into helpers since they don't need rely on props or state

src/pages/join/components/JoinForm.js
```diff
@@ -0,0 +1,129 @@
...
+    const { formValues, onSubmit } = this.props;
+    this.state.reCaptchaInstance.reset();
+    onSubmit({ ...formValues, recaptcha_response: response, recaptcha_type: 'invisible' });
+  };
+
+  executeRecaptcha = () => {
+    this.state.reCaptchaInstance.execute();
+  };
+
+  linkRecaptcha = (instance) => {
```
jonambas: Can you comment in this file how the recaptcha flow works

src/pages/join/JoinPage.js
jonambas: Also this will need to go to `/join/plan` or `/onboarding/plan` for #307

src/pages/join/components/JoinForm.js
```diff
@@ -0,0 +1,129 @@
...
+          label='Email'
+          validate={[required, email]}
+          disabled={!reCaptchaReady || loading}
+          autoComplete='username email'
+        />
+        <Field
+          name='password'
+          component={TextFieldWrapper}
+          label='Password'
+          validate={[required, minLength(8)]}
```
jonambas: May want the `endsWithWhitespace` validator. 

If password ends with whitespace, account will be created but the auth will fail

src/pages/join/JoinPage.js
```diff
@@ -0,0 +1,153 @@
...
+            <SparkPost.Logo />
+          </UnstyledLink>
+        </div>
+
+        <Panel sectioned accent title="Log In">
+        { createError && this.renderError(createError)}
+
+          <JoinForm onSubmit={this.registerSubmit} />
+        </Panel>
+        Already have an account? <Link to={'/auth'}>Log In</Link>.
```
jonambas: Wrap with `<small>`. 

Also, in the `/auth` page, can we add "Don't have an account? Sign up.", will need to wrap this in `AccessControl`

jonambas:
Need to update package and package-lock to specify matchbox `2.2.1`

src/pages/join/JoinPage.js
rajumsys: it was supposed to gone!

src/pages/join/JoinPage.js
rajumsys: I'll redirect it to correct route now. let's defer the other issue until it's we see or believe it'll appear.

src/config/default.js
```diff
@@ -33,7 +33,7 @@ const config = {
   featureFlags: {
     allow_mailbox_verification: true,
     allow_anyone_at_verification: false,
-    has_signup: false
+    has_signup: true
```
jonambas: set this to `false`, this is set to `true` in production.js which overrides for local dev

src/config/routes.js
```diff
@@ -93,6 +94,12 @@ const routes = [
     public: true,
     component: RegisterPage
   },
+  {
+    path: '/join',
+    public: true,
+    component: JoinPage,
+    condition: configEquals('featureFlags.has_signup', true)
```
jonambas: I believe `configFlag('featureFlags.has_signup')` is alright here

src/helpers/loadScript.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import Script from 'react-load-script';
+
+function loadScript({ url, ...rest }) {
+  return <Script url={url} {...rest} />;
```
jonambas: Nice

src/pages/join/JoinPage.js
```diff
@@ -0,0 +1,93 @@
...
+      <div>
+        {loadScript({ url: LINKS.RECAPTCHA_LIB_URL })}
+        <CenteredLogo />
+
+        <Panel sectioned accent title="Log In">
+          { createError && <Error error={<JoinError errors={createError} data={formData} />} /> }
+
+          <JoinForm onSubmit={this.registerSubmit} />
+        </Panel>
+        <small>Already have an account? <Link to={'/auth'}>Log In</Link>.</small>
```
jonambas: `UnstyledLink`

src/pages/join/components/JoinForm.module.scss
```diff
@@ -0,0 +1,5 @@
+@import '~@sparkpost/matchbox/src/styles/config.scss';
+
+.spacer {
+  margin-bottom: 16px;
```
jonambas: use `margin-bottom: rem(16)`

src/pages/join/components/tests/JoinForm.test.js
```diff
@@ -0,0 +1,102 @@
+/* eslint max-lines: ["error", 200] */
```
jonambas: Don't think you need this, file is only 102 lines long

src/pages/join/components/JoinForm.js
```diff
@@ -0,0 +1,133 @@
...
+          disabled={!reCaptchaReady || loading}
+          validate={required}
+          label={<span>I agree to SparkPost's <UnstyledLink to={LINKS.TOU} external>Terms of Use</UnstyledLink></span>}
+        />
+
+        <Button primary
+          disabled={loading || pristine || invalid || !reCaptchaReady }
+          onClick={this.executeRecaptcha}
+        >
+          { loading ? 'Loading' : 'Create Account' }
```
jonambas: There's a delay between create account success -> login success, maybe we should also disable this if login is pending?

src/config/default.js
rajumsys: yeah, it will be false by default. mistakenly committed.

src/config/routes.js
rajumsys: `configFlag`, will disregard value. so a `false` will also enable it. no?

src/pages/join/components/tests/JoinForm.test.js
rajumsys: no. no need.

src/pages/join/components/JoinForm.js
rajumsys: loading is that. `  loading: state.account.createLoading`

jonambas: Yeah maybe createLoading + loginLoading (or whatever the key is)

src/config/routes.js
jasonrhodes: No, it will return the value at that path (defaulting to false if the path didn't exist) and treat it as boolean, ie a "flag"

```
export const configFlag = (path) => () => _.get(config, path, false);
```

src/pages/join/components/JoinForm.js
rajumsys: Added some comments. Let me know if it makes sense.

src/pages/auth/AuthPage.js
```diff
@@ -83,6 +83,7 @@ export class AuthPage extends Component {
 
   render() {
     const { errorDescription } = this.props.auth;
+    const hasSignup = (config.featureFlags || {}).has_signup;
```
bizob2828: i know you're merging this but lodash get is your friend here. `_.get(config, 'featureFlags.has_signup', false);`

PR #295, opened 2018-02-15
FAD-5899 Changed 5th graph line to contrast more; added initial rende…

gwalls:
…r metrics; fixed log(0) error; changed Sq Rt to Pow button; Added Aggregate Total Group By

### DIFF
--- a/src/actions/summaryChart.js
+++ b/src/actions/summaryChart.js
@@ -77,9 +77,10 @@ export function getTableData({ params, metrics, groupBy }) {
       params = getQueryFromOptions({ ...state.summaryChart, ...state.reportFilters });
     }
 
+    const path = activeGroup === 'aggregate' ? 'deliverability' : `deliverability/${activeGroup}`;
     const options = {
       type: 'FETCH_TABLE_DATA',
-      path: `deliverability/${activeGroup}`,
+      path,
       params
     };
 
--- a/src/pages/reports/summary/components/ChartHeader.js
+++ b/src/pages/reports/summary/components/ChartHeader.js
@@ -32,11 +32,11 @@ class ChartHeader extends Component {
           <div className={styles.Controls}>
             <Button size='small' onClick={onMetricsToggle}>Select Metrics</Button>
             {/* {this.renderTimeMode()} */}
-            <Button.Group className={styles.ButtonSpacer}>
-              {this.renderScaleButton('linear', 'Linear')}
-              {this.renderScaleButton('log', 'Log')}
-              {this.renderScaleButton('pow', 'Pow')}
-            </Button.Group>
+            {/*<Button.Group className={styles.ButtonSpacer}>*/}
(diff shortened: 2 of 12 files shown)


src/pages/reports/summary/components/tableConfig.js
```diff
@@ -1,4 +1,8 @@
 export const GROUP_CONFIG = {
+  'aggregate': {
+    label: 'Aggregate Total',
```
jonambas: Aren't these two words that mean the same thing? Not sure what the label should be, but just `Aggregates` or `Totals` would be fine

src/actions/summaryChart.js
```diff
@@ -77,9 +77,10 @@ export function getTableData({ params, metrics, groupBy }) {
       params = getQueryFromOptions({ ...state.summaryChart, ...state.reportFilters });
     }
 
+    const path = activeGroup === 'aggregate' ? 'deliverability' : `deliverability/${activeGroup}`;
```
jonambas: Looks like this won't be affected by #283 👍

src/pages/reports/summary/components/ChartHeader.js
```diff
@@ -35,7 +35,7 @@ class ChartHeader extends Component {
             <Button.Group className={styles.ButtonSpacer}>
               {this.renderScaleButton('linear', 'Linear')}
               {this.renderScaleButton('log', 'Log')}
-              {this.renderScaleButton('sqrt', '√ Sq Rt')}
+              {this.renderScaleButton('pow', 'Pow')}
```
jonambas: This wasn't included in AC

@jasonrhodes did you have any thoughts on this?

jasonrhodes: I asked him to do it. I only changed to sqrt because pow was blowing up for me but it doesn't seem to be now. We'll need to keep an eye on it.

jonambas: 👍

src/pages/reports/summary/components/tableConfig.js
jasonrhodes: I think they're always "totals" -- these are "ungrouped totals", or "combined totals", so "aggregates" would be better than "totals" alone IMO, although "aggregate totals" works for me too to distinguish from the group by content. Could also solve with a better section header/description, but not sure that's necessary unless we get confused feedback.

gwalls: The API docs use the term "Summary" for that endpoint. So perhaps "Summary Totals" or "Aggregate Metrics"? Idk, nothing sounds correct anymore 😄

src/pages/reports/summary/components/ChartHeader.js
```diff
@@ -35,7 +35,7 @@ class ChartHeader extends Component {
             <Button.Group className={styles.ButtonSpacer}>
```
jonambas: Comment out this button group

PR #297, opened 2018-02-16
FAD-6109 Support necessary Sift Science events

bkemper:
_Refer to [FAD-6109](https://jira.int.messagesystems.com/browse/FAD-6109) for AC_

The intention of this PR is to integrate with Sift Science (SS) similar to the webui setup.

**About SS JavaScript Snippet**

Here are the [installation instructions](https://siftscience.com/developers/docs/javascript/javascript-api).  Communication is done by pushing one of the following events on a global variable, `window._sift`, and letting the snippet process them as needed.

* _setAccount - set with [Javascript Snippet Key](https://console.siftscience.com/developer/api-keys)
* _setUserId - set with user's customer ID (account prefixed if available)
* _setSessionId - set with a UUID that is shared our other sites (e.g. [sparkpost.com](https://github.com/SparkPost/sparkpost.com/blob/8bf34fd430e4ed55843040c07a0078288e321dd6/wordpress/wp-content/themes/jolteon/js/siftscience.js))
* _trackPageview - this event is magical, it doesn't accept any arguments/values, it grabs all it needs from window

NOTE: We currently do not set _setCookieDomain.  It is my understanding that this could be an alternative for _setSessionId to identify unique visitors without needing to create a UUID and store it in a cookie to share with our other sites.  Instead, we would set the cookie domain to 'sparkpost.com' and let SS do the rest.

**Overview of Webui Setup**

1. [index.html](https://github.com/SparkPost/webui/blob/master/src/index.html#L40) loads the inline-scripts.js script
1. [common/inline-scripts.js](https://github.com/SparkPost/webui/blob/master/src/common/inline-scripts.js#L20-L35) loads the SS snippet [onload](https://developer.mozilla.org/en-US/docs/Web/Events/load) after all other resources have finished loading and defines `window._sift`
1. The Angular app during configuration [calls Analytics.setup()](https://github.com/SparkPost/webui/blob/master/src/app/app.js#L153)
1. [setup()](https://github.com/SparkPost/webui/blob/master/src/common/services/analytics.js#L58-L76) registers a function to call Sift Science events.
1. On `$stateChangeSuccess` (a page change), [Analytics.setCustomData(userData)](https://github.com/SparkPost/webui/blob/master/src/app/app.js#L227) and [Analytics.trackPageView($location.url(), $rootScope.stateData.title)](https://github.com/SparkPost/webui/blob/master/src/app/app.js#L230) are called.
1. `setCustomData` calls [getCustomDataArgs](https://github.com/SparkPost/webui/blob/master/src/common/services/analytics.js#L86-L88) which executes the [case 'user_data'](https://github.com/SparkPost/webui/blob/master/src/common/services/analytics.js#L70-L72) to set _setUserId.
1. `trackPageView` calls [getPageViewArgs](https://github.com/SparkPost/webui/blob/master/src/common/services/analytics.js#L77-L85) which executes the [case 'page_view'](https://github.com/SparkPost/webui/blob/master/src/common/services/analytics.js#L67-L69) to call _trackPageview.

NOTE: There are no additional events captured ([`getEventArgs`](https://github.com/SparkPost/webui/blob/master/src/common/services/analytics.js#L89-L91) is a noop).  The API events come from accusers-api.

**How to Test**

For local development, testing, and staging, we **_MUST_** use the SS sandbox account.  For production, create a throwaway SP account with the [fake credit card](https://confluence.int.messagesystems.com/pages/viewpage.action?pageId=43157618).

- [ ] Happy path - all is setup and pushing events to SS
- [ ] Without SS snippet key - shouldn't load SS snippet or push any events to `window._sift`
- [ ] Without customer ID (unauthorized user) - should [set _setUserId to empty string](https://support.siftscience.com/hc/en-us/articles/208370598), leave _setSessionId set to current value and call _trackPageview to continue tracking page views
- [ ] With account prefix (specifically, in EU) - should set _setUserId with a prefix (e.g. 'eu-')
- [ ] With a chocolate chip cookie - should expire after a minute, should persist like a normal cookie for our application (e.g. a cross tabs, etc.), and should be accessible by sparkpost.com

**Todo**
- [x] ~Add configuration variables to production.js.js2 template~ [APP-3162](https://fisheye.int.messagesystems.com/cru/APP-3162)
- [x] Setup [JavaScript snippet](https://siftscience.com/developers/docs/javascript/javascript-api)
- [x] Why does `getPageViewArgs` have the [`resp.meta = { auth: false };`](https://github.com/SparkPost/webui/blob/master/src/common/services/analytics.js#L80-L82) condition.
- [ ] File ticket to see if _setCookieDomain is a better solution than _setSessionId

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,8 +1,10 @@
 import React from 'react';
 import { PublicRoute, ProtectedRoute, AuthenticationGate } from 'src/components/auth';
 import { Support, GlobalAlertWrapper, RouteWatch } from 'src/components';
+import SiftScience from 'src/components/siftScience/SiftScience';
 import Layout from 'src/components/layout/Layout';
 import routes from 'src/config/routes';
+import config from 'src/config';
 
 import {
   BrowserRouter as Router,
@@ -13,6 +15,7 @@ import {
 const App = () => (
   <Router>
     <div>
+      {config.siftScience && <SiftScience config={config.siftScience} />}
       <RouteWatch />
       <AuthenticationGate />
       <Layout>
--- a/package-lock.json
+++ b/package-lock.json
@@ -4973,6 +4973,11 @@
         "strip-eof": "1.0.0"
       }
     },
+    "exenv": {
+      "version": "1.2.2",
+      "resolved": "https://registry.npmjs.org/exenv/-/exenv-1.2.2.tgz",
+      "integrity": "sha1-KueOhdmJQVhnCwPUe+wfA72Ru50="
+    },
     "exit": {
       "version": "0.1.2",
       "resolved": "https://registry.npmjs.org/exit/-/exit-0.1.2.tgz",
@@ -10428,7 +10433,7 @@
             "stringstream": "0.0.5",
             "tough-cookie": "2.3.3",
             "tunnel-agent": "0.4.3",
-            "uuid": "3.1.0"
+            "uuid": "3.2.1"
           }
         },
         "sntp": {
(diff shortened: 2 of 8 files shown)


src/App.js
```diff
@@ -13,6 +15,7 @@ import {
 const App = () => (
   <Router>
     <div>
+      {config.siftScience && <SiftScience config={config.siftScience} />}
```
jasonrhodes: My first reaction is that we don't handle this kind of forking in this file usually, and instead would do that in the SiftScience component. But your explanation when walking through this code made sense, bc we'd have to check for the config existence in a few places in that component. If you're good with it, then it works for me.

PR #299, opened 2018-02-16
Public access control

jasonrhodes:
This is a tiny PR into @rajumsys 's existing sign up PR so we can wrap public routes in config flag conditions. @rajumsys if you can look through and when you understand what it's doing, merge this into your PR and let whoever reviews that review this as well. It should probably have some tests added too. Thanks!


rajumsys:
Tested and working fine. Let's add the test in the main ticket. I'll also read a bit of `react-router-dom` doc.

PR #304, opened 2018-02-20
FAD-6195 Fixed NaN and infinite rate calculations for chart data

gwalls:


src/config/metrics.js
```diff
@@ -1,10 +1,11 @@
 /* eslint-disable max-lines */
 function average(item, keys = []) {
-  return item[keys[0]] / item[keys[1]];
+  const avg = item[keys[0]] / item[keys[1]] || 0;
+  return isFinite(avg) ? avg : 0;
```
gwalls: should infinite numbers be treated as zeroes?

jonambas: yep, this is correct

PR #305, opened 2018-02-20
FAD-6241: Missing billing summary and dedicated IPs purchase forms

ewandennis:
For non-free, self-billed accounts, `/account/billing` is meant to show a billing address and CC summary and a form for buying dedicated IPs.

Neither currently show because they use a condition variable based on `/account?include=billing`. Auth stopped loading billing details some time ago and so both parts of the billing page stayed hidden.

The fix below is to have the billing page preload account billing details on mount.

## Testing
 - Create a self-billed non-free account
    [This page](https://confluence.int.messagesystems.com/pages/viewpage.action?pageId=43157618) may be useful in creating that account ;)
 - Visit /account/billing
 - Verify that billing address and CC details are visible
 - Verify that you can buy dedicated IPs

### DIFF
--- a/src/pages/billing/tests/SummaryPage.test.js
+++ b/src/pages/billing/tests/SummaryPage.test.js
@@ -11,6 +11,7 @@ describe('Page: SummaryPage', () => {
     },
     billing: {},
     dedicatedIpPrice: jest.fn(() => ''),
+    fetchAccount: jest.fn(),
     getPlans: jest.fn(),
     getBillingCountries: jest.fn(),
     getSendingIps: jest.fn(),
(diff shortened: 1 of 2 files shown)


src/pages/billing/tests/SummaryPage.test.js
```diff
@@ -27,6 +28,13 @@ describe('Page: SummaryPage', () => {
     wrapper = shallow(<SummaryPage {...props} />);
   });
 
+  it('gets account billing details on mount', () => {
+    wrapper.instance().componentDidMount();
+    expect(props.fetchAccount).toBeCalledWith(expect.objectContaining({
```
rajumsys: this is fine, but, just for fyi, i guess it should work simply with
```toBeCalledWith({include: billing})```

rajumsys:
Verified 3 cases:
free: doesn't show 
paid: shows
manual: doesn't show

so LGTM

PR #306, opened 2018-02-20
FAD-6234 Hide view details button when message_id does property does not exist

rajumsys:
See ticket for the detail context

Didn't find a easy way to manually test it. It would be good if I could modify redux store to modify the data but didn't find a way even though feel like there is a way. 

However, this is pretty trivial and easy enough to verify from snapshot. 

*Update*
Testing:
You can test it using redux dev tool. Click 'Dispatcher'. 

put something like:

```
{
type: 'GET_MESSAGE_EVENTS_SUCCESS',
payload: [{
        formattedDate: 'formatted',
        type: 'injection',
        friendly_from: 'mean@friendly',
        rcpt_to: 'tom.haverford@pawnee.state.in.us',
        message_id: '123abc',
        event_id: '456xyz'
      }, {
        formattedDate: 'formatted',
        type: 'policy_rejection',
        friendly_from: 'mean@friendly',
        rcpt_to: 'tom.haverford@pawnee.state.in.us',
        event_id: '456xyz'
      }]
}
```
And then click tiny little almost microscopic `Dispatch` button on the right. UI should be updated.

### DIFF
--- a/src/pages/reports/messageEvents/MessageEventsPage.js
+++ b/src/pages/reports/messageEvents/MessageEventsPage.js
@@ -50,9 +50,11 @@ export class MessageEventsPage extends Component {
       snakeToFriendly(type),
       rcpt_to,
       friendly_from,
-      <div style={{ textAlign: 'right' }}>
-        <Button onClick={() => this.handleDetailClick({ message_id, event_id })} size='small'>View Details</Button>
-      </div>
+      message_id
+        ? <div style={{ textAlign: 'right' }}>
+          <Button onClick={() => this.handleDetailClick({ message_id, event_id })} size='small'>View Details</Button>
+        </div>
+        : null
     ];
   }
 
--- a/src/pages/reports/messageEvents/MessageEventsPage.js
+++ b/src/pages/reports/messageEvents/MessageEventsPage.js
@@ -2,10 +2,12 @@ import React, { Component } from 'react';
 import { connect } from 'react-redux';
 
 import { snakeToFriendly } from 'src/helpers/string';
-import { Page, Banner, Button, Panel } from '@sparkpost/matchbox';
+import { Page, Banner, Panel } from '@sparkpost/matchbox';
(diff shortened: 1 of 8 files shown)


jasonrhodes:
FWIW @rajumsys you can modify redux store using dev tools, you just fire the action you want with the data you want.

rajumsys:
> FWIW @rajumsys you can modify redux store using dev tools, you just fire the action you want with the data you want.

Thanks. I figured out the 'Dispatcher' button after your comment.

src/pages/reports/messageEvents/MessageEventsPage.js
```diff
@@ -50,9 +50,11 @@ export class MessageEventsPage extends Component {
       snakeToFriendly(type),
       rcpt_to,
       friendly_from,
-      <div style={{ textAlign: 'right' }}>
-        <Button onClick={() => this.handleDetailClick({ message_id, event_id })} size='small'>View Details</Button>
-      </div>
+      message_id
```
jasonrhodes: What do you think about moving this into a small component?
```jsx
<ViewDetails messageEvent={rowData}>
```
In there you could wrap it with `withRouter`, so the onClick inline, and exit early if there is no message_id ... would be a bit easier to see what's going on, a bit easier to test clearly, and can get rid of that inline CSS.

rajumsys: it's fine. just test boilerplate :). 

will do

src/pages/reports/messageEvents/components/ViewDetailsButton.js
```diff
@@ -0,0 +1,30 @@
...
+  };
+
+  render() {
+    const { message_id, event_id } = this.props;
+
+    if (!message_id) {
+      return null;
+    }
+
+    return (<div style={{ textAlign: 'right' }}>
```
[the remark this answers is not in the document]
rajumsys: moved it as it is. new file? too much work :D

PR #307, opened 2018-02-20
FAD-6207 Added selecting plan to onboarding

bizob2828:
I removed the language around SMS, push in this PR too.


bizob2828:
There is a console log when starting to type the month of expiration date, but only on the onboarding. Not sure what's up here

src/pages/billing/forms/ChangePlan.js
```diff
@@ -99,7 +77,7 @@ export class ChangePlan extends Component {
   }
 
   render() {
-    const { account, submitting, pristine, currentPlan, selectedPlan, plans } = this.props;
+    const { account, submitting, pristine, currentPlan, selectedPlan, plans, onboarding = false } = this.props;
```
jasonrhodes: let's call this prop something like `showSidebar` or `hideSidebar`

src/pages/billing/forms/ChangePlan.js
```diff
@@ -36,30 +34,10 @@ export class ChangePlan extends Component {
...
-    // Creates Zuora account
-    return billingCreate(values)
-      .then(() => history.push('/account/billing'))
-      .then(() => showAlert({ type: 'success', message: 'Subscription Upgraded' }))
-      .catch(() => showAlert({ type: 'error', message: 'Plan Upgrade Failed' }));
+    // capturing the useSavedCC state and passing to parent submit function
+    return handleSubmit(values, this.state.useSavedCC);
   }
 
   renderCCSection = () => {
```
jonambas: Can we hide the "Add a Credit Card" panel title only for onboarding

src/pages/billing/forms/ChangePlan.js
```diff
@@ -109,8 +87,10 @@ export class ChangePlan extends Component {
...
       : plans;
 
+    const buttonText = selectedPlan && selectedPlan.isFree ? 'Get Started' : 'Confirm Plan';
+
     return (
-      <form onSubmit={this.props.handleSubmit(this.onSubmit)}>
+      <form onSubmit={this.onSubmit}>
         <Grid>
           <Grid.Column>
             <Panel title='Select A Plan'>
```
jonambas: So we're loading two panels inside another panel (in the parent component), looks ok though 👍

src/config/routes.js
```diff
@@ -375,6 +375,11 @@ const routes = [
     condition: hasGrants('ip_pools/manage'),
     layout: App
   },
+  {
+    path: '/onboarding/plan',
```
rajumsys: just fyi, this is the route i've forwarded after signup. so in case it changes, you'll need to update that in JoinPage module.

src/pages/billing/components/tests/__snapshots__/Banners.test.js.snap
```diff
@@ -107,7 +107,7 @@ exports[`Billing Banners:  renders Enterprise banner 1`] = `
   title="Enterprise"
 >
   <p>
-    Enterprise-grade guarantees and SLAs across email, push, and SMS - manage, analyzed, and reported from a single console.
+    Enterprise-grade guarantees and SLAs across email - manage, analyzed, and reported from a single console.
```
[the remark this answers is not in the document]
jasonrhodes: Yeah I think it needs more work than even that 🤔

src/config/routes.js
```diff
@@ -375,6 +375,11 @@ const routes = [
...
+    component: onboarding.ChoosePlan,
+    condition: configFlag('featureFlags.has_signup')
```
jonambas: Jason mentioned that we only really need this flag on the sign up page, and not the onboarding pages. But I'll leave that up to you

src/pages/onboarding/ChoosePlan.js
```diff
@@ -0,0 +1,115 @@
...
+import { onboardingInitialValues } from 'src/selectors/accountBillingForms';
+import PaymentForm from 'src/pages/billing/forms/fields/PaymentForm';
+import BillingAddressForm from 'src/pages/billing/forms/fields/BillingAddressForm';
+
+const FORM_NAME = 'selectInitialPlan';
+
+export class OnboardingPlanPage extends Component {
+  componentDidMount() {
+    this.props.getPlans();
+    this.props.getBillingCountries();
```
jonambas: Do we need to `.catch` these?

src/pages/onboarding/ChoosePlan.js
```diff
@@ -0,0 +1,115 @@
...
+    return (
+      <form onSubmit={this.props.handleSubmit(this.onSubmit)}>
+        <CenteredLogo />
+        <Grid>
+          <Grid.Column>
+            <Panel title='Select A Plan'>
+              <PlanPicker disabled={this.props.submitting} plans={plans} />
+              {this.renderCCSection()}
+              <Panel.Section>
+                <Button primary={true} type='submit' size='large' fullWidth={true}>Get Started</Button>
```
jonambas: Disable when `submitting`

src/pages/onboarding/components/Steps.js
```diff
@@ -7,8 +7,6 @@ import styles from './Steps.module.scss';
 // TODO Finalize when whole register flow is complete
```
jonambas: Can remove comment

src/pages/onboarding/ChoosePlan.js
jonambas: And perhaps even redirect to the dashboard or the next step if these fail? The form doesn't load properly. I'm considering ApiErrorBanner, but redirecting might be the better experience.

src/config/routes.js
jasonrhodes: Oh yeah, we should remove the condition for everything but `/join`

src/pages/onboarding/ChoosePlan.js
bizob2828: I thought we were holistically going to solve the uncaught requests from cDM

src/config/routes.js
bizob2828: i did this because current UI doesn't register onboarding for non self serve customers

src/pages/onboarding/ChoosePlan.js
jonambas: Is that the plan? I'm out of the loop with whats planned there

jasonrhodes: These are two different issues:
1) Catching/swallowing API errors so they don't cause uncaught promise rejections when we've already handled the error via _FAIL actions (should solve holistically)
2) Dealing with individual error cases as needed

We should probably be dealing with (2) as we encounter it, so we don't add to the problem while we are waiting to holistically solve (1) -- which means _not_ using `.catch()` here but handling plan/country errors one way or another with a redirect like @jonambas suggests or something similar.

src/config/routes.js
bizob2828: discussed in slack. misunderstanding....leaving as is with this feature flag check

src/pages/onboarding/ChoosePlan.js
bizob2828: handling in cDU and redirect to next step

jonambas:
After you merge in develop / rebase, think you could sneak this into `JoinForm`?:

Need to add checkbox group around the two checkboxes:
```js
<Checkbox.Group>
          <Field />
          <Field />          
</Checkbox.Group>
```

src/pages/onboarding/ChoosePlan.js
```diff
@@ -0,0 +1,126 @@
...
+    return (
+      <form onSubmit={this.props.handleSubmit(this.onSubmit)}>
+        <CenteredLogo />
+        <Grid>
+          <Grid.Column>
+            <Panel title='Select A Plan'>
+              <PlanPicker disabled={this.props.submitting} plans={plans} />
+              {this.renderCCSection()}
+              <Panel.Section>
+                <Button disabled={this.props.submitting} primary={true} type='submit' size='large' fullWidth={true}>Get Started</Button>
```
jonambas: This always takes a few seconds for me, maybe flip `Get Started` to `Updating Subscription...` when loading

(not sure on language, whatever makes sense here)

src/pages/billing/forms/fields/PaymentForm.js
```diff
@@ -91,7 +95,7 @@ export class PaymentForm extends Component {
               onChange={this.handleExpiry}
               placeholder='MM/YYYY'
               component={TextFieldWrapper}
-              validate={required}
+              validate={[required, minLength(9)]}
```
jonambas: the validator error message shows `Must be at least 9 characters`. Can we write a wrapper around `minLength` in this component with a better error message?

PR #308, opened 2018-02-21
Fix dependency ranges and update packages with small updates

jasonrhodes:
A bunch of greenkeeper PRs were sitting here because our dependencies were version-locked in package.json. We use package-lock.json now for locking deps, and we use TravisCI and greenkeeper to help us understand when small updates are safe.

With that said, this PR bumps semver ranges to use `^` and updates a few of the small patch/minor libs that were PRd so we can push them all together. This should make things slightly less noisy going forward with greenkeeper.

Packages updated here:

Package | Old version | New version
:--- | :---: | :---:
webpack | 3.10.0 | 3.11.0
chalk | 2.3.0 | 2.3.1
style-loader | 0.20.1 | 0.20.2
eslint | 4.16.0 | 4.18.0
file-loader | 1.1.6 | 1.1.7


package.json
```diff
@@ -73,32 +73,32 @@
     "react-timeago": "^3.4.3",
     "recharts": "^1.0.0-alpha.2",
```
jonambas: we're on an alpha release? 😱

PR #311, opened 2018-02-22
FAD-6203 refresh typeahead cache when all caches are empty (first load)

jasonrhodes:


bizob2828:
another issue from the original PR is if i pick a custom range and navigate to other reports, it isn't requesting the api calls for that page.

PR #312, opened 2018-02-22
Wired up the email verification. route has changed will handle in cutover

bizob2828:
TODO:

- [x] verify UX works
- [x] write unit tests

To verify the success path:
 * Create new account via /join
 * Go to dashboard and click re-send email
 * connect to redis  `redis-cli -h etl-spc-tst.xzz0pa.ng.0001.usw2.cache.amazonaws.com`
 * get token `get accusers-email-verification-tokens:uat:<username>`
 * go to `http://localhost:3100/account/email-verification/<token>`

Fail case:
 * go to `http://localhost:3100/account/email-verification/this-is-fake`


src/components/emailVerification/EmailVerification.js
```diff
@@ -0,0 +1,54 @@
+import React, { Component } from 'react';
+import { connect } from 'react-redux';
+import { verifyEmailToken } from 'src/actions/currentUser';
+import { Loading } from 'src/components';
+import { showAlert } from 'src/actions/globalAlert';
+
+export class EmailVerification extends Component {
```
jonambas: What i like about this is that we could put this on any route or even inside another page and it will work, as long as the token is provided

src/components/emailVerification/EmailVerification.js
```diff
@@ -0,0 +1,54 @@
...
+    if (!prevProps.error && error) {
+      showAlert({
+        type: 'error',
+        message: 'The verification token is either invalid or expired.',
+        details: error.message
+      });
+      history.push('/dashboard');
+    }
+
+    if (!error && verifyingEmail === false) {
```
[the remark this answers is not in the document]
jonambas: or `prevProps.error`

src/components/emailVerification/EmailVerification.js
bizob2828: yea, that was the idea

PR #313, opened 2018-02-23
FAD-6086: Add Marketing email opt-in to signup form

ewandennis:
### Accusers Dependency
This new field is accepted by `/account` as a salesforce_data field. It is not stored in the app, just passed through directly to SalesForce as `HasOptedOutOfEmail`. That work was deployed under FAD-6150 on 2018-02-22.

Note: The form has an `email_opt_in` field but `POST /account` expects `salesforce_data.email_opt_out`. I added an explicit test for that negation.

### Testing
 - Create accounts through `/join` with the new field checked and unchecked.
 - Verify the resulting call to `POST /account` in the browser.
 - The contact for the main user on each account you create will appear in test.salesforce.com (creds in LastPass)
 - That contact will include the `HasOptedOutOfEmail` field with the value you set at signup.

### DIFF
--- a/src/pages/join/components/JoinForm.js
+++ b/src/pages/join/components/JoinForm.js
@@ -109,6 +109,15 @@ export class JoinForm extends Component {
         />
 
         <Field
+          name='email_opt_in'
+          id='email_opt_in'
+          component={CheckboxWrapper}
+          disabled={!reCaptchaReady || loading}
+          type='checkbox'
+          label={<span>I'm happy to receive marketing email from SparkPost</span>}
+        />
+
+        <Field
           name="tou_accepted"
           id='tou_accepted'
           component={CheckboxWrapper}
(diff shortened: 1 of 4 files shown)


src/pages/join/components/JoinForm.js
```diff
@@ -117,6 +118,15 @@ export class JoinForm extends Component {
           label={<span>I agree to SparkPost's <UnstyledLink to={LINKS.TOU} external>Terms of Use</UnstyledLink></span>}
         />
 
+        <Field
+          name='email_opt_in'
```
bizob2828: I'm sure marketing is not happy about this.  Are we able to change language to Opt out and make them click it to do so?

src/pages/join/components/JoinForm.js
```diff
@@ -117,6 +118,15 @@ export class JoinForm extends Component {
           label={<span>I agree to SparkPost's <UnstyledLink to={LINKS.TOU} external>Terms of Use</UnstyledLink></span>}
```
bizob2828: feels like we should have TOU right above button

src/pages/join/components/JoinForm.js
ewandennis: GDPR consent requires `a positive opt in`. [Here are the UK Information Commissioner's Office notes on consent](https://ico.org.uk/for-organisations/guide-to-the-general-data-protection-regulation-gdpr/lawful-basis-for-processing/consent/).

PR #314, opened 2018-02-23
FAD-6207 ChoosePlan page using deeper billing components

jasonrhodes:


jasonrhodes:
@bizob2828 I think this is more in line with what @jonambas had in mind with those tiny billing form components. It took me a while for it to click but this worked out when we had to also add different handling for what to do when the free plan is selected, and which plan is the default to start with. 

If we go this way you can revert the changes in the Change Plan page and form, then just need some test updates.

PR #315, opened 2018-02-24
FAD-6195 changed all rate calculation to be done by common function

gwalls:


src/helpers/metrics.js
```diff
@@ -132,3 +132,6 @@ export function rate(item, keys) {
   return average(item, keys) * 100;
 }
 
+export function getRate(numerator, denominator) {
```
jasonrhodes: If it's not too much trouble, could we switch this around? I'd like to see `safeDivide` and `safeRate` functions in some kind of number helpers file that just accept `(numerator, denominator)` like you've got for `getRate` here, and then let the metrics helper versions use those methods but accept their weirder, more specific APIs i.e. `(item, keys)` 

Does that make sense?

gwalls: Absolutely, I wasn't set on the naming / placement at all.
I think I see what you're after. Let me whip something up and you can lmk if it's similar to what you had in mind.

src/pages/reports/accepted/components/AcceptedChart.js
```diff
@@ -61,11 +62,9 @@ export default class AcceptedChart extends Component {
     const { aggregates } = this.props;
     const { hoveredItem } = this.state;
 
-    const getRate = (n) => `${(100 * n).toFixed(2)}%`;
-
     return hoveredItem
-      ? { name: hoveredItem.name, value: getRate(hoveredItem.count / aggregates.count_accepted) }
-      : { name: 'Accepted Rate', value: getRate(aggregates.count_accepted / aggregates.count_targeted) };
+      ? { name: hoveredItem.name, value: `${getRate(hoveredItem.count, aggregates.count_accepted).toFixed(2)}%` }
```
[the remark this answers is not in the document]
gwalls: formatPercent takes care of the fixed decimal places as well

src/pages/reports/bounce/components/BounceChart.js
```diff
@@ -62,11 +63,9 @@ export default class BounceChart extends Component {
     const { aggregates } = this.props;
     const { hoveredItem } = this.state;
 
-    const getRate = (n) => `${(100 * n).toFixed(2)}%`;
-
     return hoveredItem
-      ? { name: hoveredItem.name, value: getRate(hoveredItem.count / aggregates.countBounce) }
-      : { name: 'Bounce Rate', value: getRate(aggregates.countBounce / aggregates.countTargeted) };
+      ? { name: hoveredItem.name, value: `${safeRate(hoveredItem.count, aggregates.countBounce).toFixed(2)}%` }
+      : { name: 'Bounce Rate', value: `${safeRate(aggregates.countBounce, aggregates.countTargeted).toFixed(2)}%` };
```
jonambas: Can we add `formatPercent` around these

src/pages/reports/accepted/components/TopLevelMetrics.js
```diff
@@ -41,7 +42,7 @@ const TopLevelMetrics = ({ filters, metrics, aggregates }) => {
   return (
     <div>
       <MetricsSummary
-        rateValue={(count_accepted / count_targeted) * 100}
+        rateValue={getRate(count_accepted, count_targeted)}
```
bizob2828: we could avoid passing in the rate by just taking a numerator, dominator then taking care of calculating the rate within the MetricsSummary component

src/pages/reports/accepted/components/TopLevelMetrics.js
```diff
@@ -41,7 +42,7 @@ const TopLevelMetrics = ({ filters, metrics, aggregates }) => {
   return (
     <div>
       <MetricsSummary
-        rateValue={(count_accepted / count_targeted) * 100}
+        rateValue={safeRate(count_accepted, count_targeted)}
```
bizob2828: i think we should just pass in numerator, denominator and do the safeRate in the MetricsSummary

src/pages/reports/accepted/components/tests/__snapshots__/AcceptedChart.test.js.snap
```diff
@@ -42,14 +42,14 @@ Object {
 exports[`Accepted ChartGroup:  should handle mouse in and out 2`] = `
 <PieChart.ActiveLabel
   name="first attempt"
-  value="10.00%"
```
gwalls: Are we okay with this change?

jonambas: We should add `toFixed` to the percent component if we can, but don't mind if you want to defer.

src/helpers/math.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * Safely divides two numbers, returning 0 for NaN or infinite values
+ * @param numerator
+ * @param denominator
+ * @return {number}
+ */
+export function safeDivide(numerator, denominator) {
+  const value = numerator / denominator || 0;
```
bizob2828: you don't need the or here. `isFinite(Nan)` is false

src/helpers/units.js
```diff
@@ -74,6 +74,10 @@ export const formatNumber = (value) => {
     return value;
   }
 
+  if (!isFinite(value)) {
```
bizob2828: not seeing tests for this

src/helpers/units.js
```diff
@@ -108,6 +112,10 @@ export const formatPercent = (value) => {
...
```
bizob2828: not seeing tests for this

src/pages/reports/engagement/components/tests/__snapshots__/EngagementSummary.test.js.snap
```diff
@@ -15,7 +15,9 @@ exports[`EngagementSummary renders metric summary 1`] = `
       xs={12}
     >
       <h1>
-        0.42%
+        <Percent
```
bizob2828: these should be fixed to 2 places

src/pages/reports/accepted/components/tests/__snapshots__/AcceptedChart.test.js.snap
gwalls: This particular instance doesn't use the Percent component

jonambas: It runs through `formatPercent` though, which is the same helper. But looks okay visually to me, without the forced decimals

src/pages/reports/engagement/components/tests/__snapshots__/EngagementSummary.test.js.snap
gwalls: this is just the value passed to the component (which takes care of the fixed decimal places)

PR #321, opened 2018-02-26
FAD-6273: Reports new user UX issues

ewandennis:
- Rates >100% now renders as `> 100%`
 - Chart legend items with 0 values now style their values correctly
 - Bounce chart empty state now works again
 - Cleaned up some uses of `formatPercent` in favour of `<Percent />`

### Testing
 - Use a new account with no metrics data


jonambas:
I did a first pass, but suspect much will need to change once #315 is merged. Will take a look again once that happens.

ewandennis:
@jonambas Minimal changes required to work in with #315 luckily. I backed away from using `<Percent />` everywhere though. We'll need a formatting helper / component policy but it's not in the way today.

src/components/pieChart/LegendItem.js
```diff
@@ -37,7 +37,7 @@ const LegendItem = ({
       { fill && <span className={styles.Color} style={{ backgroundColor: fill }}/> }
       { breadcrumbMarkup }
       <span className={styles.Name}>{name}{expandMarkup}</span>
-      { count && <span className={styles.Count}>{count.toLocaleString()}</span> }
+      { count !== undefined && <span className={styles.Count}>{count.toLocaleString()}</span> }
```
jonambas: Good catch 👍

PR #324, opened 2018-02-26
FAD-6245 Overlay Portals

jonambas:
Updates for SparkPost/matchbox#111

Will update package version once  the MB pr has been merged

### DIFF
--- a/src/pages/reports/components/DateFilter.js
+++ b/src/pages/reports/components/DateFilter.js
@@ -136,7 +136,6 @@ export class DateFilter extends Component {
       <Popover
         wrapper='div'
         className={styles.Popover}
-        manualTrigger
         trigger={dateField}
         onOutsideClick={this.cancelDatePicker}
         open={this.state.showDatePicker} >
--- a/src/pages/reports/components/DateFilter.js
+++ b/src/pages/reports/components/DateFilter.js
@@ -1,7 +1,6 @@
 /* eslint max-lines: ["error", 200] */
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
-import ReactDOM from 'react-dom';
 import { subMonths, format } from 'date-fns';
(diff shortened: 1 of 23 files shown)


src/pages/reports/components/DateFilter.js
```diff
@@ -149,6 +138,7 @@ export class DateFilter extends Component {
         className={styles.Popover}
         manualTrigger
         trigger={dateField}
+        onOutsideClick={this.cancelDatePicker}
```
bizob2828: 👍

bizob2828:
from the MB pr "the copy to clipboard messages on the tfa backup codes, api key in onboarding(smtp, api) show the alert then in top left. also that delay of 300 is odd for this feels like copy didn't work. it also stays there on hovers after you click" . this seems better on the onboarding but still is wonky.  Also the tfa backup codes scenario does not appear to work unless you click and hover

jonambas:
> the copy to clipboard messages on the tfa backup codes, api key in onboarding(smtp, api) show the alert then in top left

are you still seeing tooltips in the top left?

> also that delay of 300 is odd for this feels like copy didn't work. it also stays there on hovers after you click" . this seems better on the onboarding but still is wonky. Also the tfa backup codes scenario does not appear to work unless you click and hover

Yeah tooltips have always been hover only, if we want to always show for a set time after hitting copy, we'll need to add a `visible` prop or something to tooltips

There shouldn't be anymore debouncing unless you are scrolling or resizing. Did some digging around the delay though. Previously, we were using native browser `:hover` to handle the css, now mb uses `setState` on `mouseOver` to add a class. CopyField also doesn't pre-mount the tooltip so initial render, setState, and mouseOver all try to happen at the same time.

```js
connectRight = copied
        ? <Tooltip dark content='Copied to clipboard!'>{ button }</Tooltip>
        : button;
```

That's enough to delay render by ~50ms in a non-production build, crazy that its enough to notice.

PR #329, opened 2018-03-01
FAD-6287: Can sign up without accepting TOU

ewandennis:
Fix: Wait for logout _before_ initialising form to avoid obliterating redux-form state.

Also fixed:
 - Sign-up and Login links on `/join` and `/auth` no longer reload the whole app.
 - TFA login page no longer shows a Sign-up link.

### Testing
1. Log in
1. Visit /join
1. Verify that form validation is working correctly

Click back and forth between /join and /auth using the sign-up and sign-in links at the bottom of the forms.

Log in with TFA and verify missing Sign-up link,


src/pages/join/JoinPage.js
```diff
@@ -24,7 +24,7 @@ export class JoinPage extends Component {
     formData: {}
   };
 
-  componentDidMount() {
+  componentWillMount() {
```
rajumsys: :facepalm:

jonambas: DM'd in slack  - 

I don't think `willMount` is the right solution here. Here's what worked for me if I understand this issue correctly:

```js
componentDidMount() {
  if (this.props.loggedIn) {
    this.props.logout();
  }
}

// in render
if (this.props.loggedIn) {
  return <Loading/>;
}

// in mstp
loggedIn: state.auth.loggedIn,
```

rajumsys:
There are two others small issues related to join page
- Click Log In and Sign Up reloads the page. Seems we just need to pass `component`
- screen shot 2018-03-01 at 10 34 17 am So on TFA page, probably we can hide the "Don't have an account" thing.
_posting here as they're small; we can create new ticket if you don't wanna do it here_.

src/pages/auth/AuthPage.js
```diff
@@ -101,7 +102,7 @@ export class AuthPage extends Component {
           { !tfa.enabled && <LoginForm onSubmit={this.loginSubmit} ssoEnabled={this.state.ssoEnabled}/> }
         </Panel>
 
-        {hasSignup && <small>Don't have an account? <UnstyledLink to={'/join'}>Sign up</UnstyledLink>.</small> }
+        {hasSignup && !tfa.enabled && <small>Don't have an account? <UnstyledLink Component={Link} to={'/join'}>Sign up</UnstyledLink>.</small> }
```
jonambas: Missing test for this branch

src/pages/join/JoinPage.js
```diff
@@ -69,8 +69,13 @@ export class JoinPage extends Component {
   };
 
   render() {
-    const { createError } = this.props.account;
+    const { loggedIn, account: { createError }} = this.props;
     const { formData } = this.state;
+
+    if (loggedIn) {
+      return <Loading />;
```
jonambas: Missing test for this

ewandennis:
Closed in favour of #333

PR #331, opened 2018-03-01
FAD-6224 removed functions in favor of built in rechart props

gwalls:


jasonrhodes:
I reviewed this and from what I can tell it LGTM ... @jonambas do you have any concerns with this?

PR #333, opened 2018-03-02
FAD-6281 Password Reset

jonambas:
Updated the cutover ticket FAD-6270 to include updating password reset template


src/pages/password/ResetPasswordPage.js
```diff
@@ -0,0 +1,88 @@
...
+      ? 'Creating Password..'
+      : 'Create New Password';
+
+    return (
+      <Fragment>
+        <CenteredLogo />
+        <Panel accent sectioned title='Create a New Password'>
+          <form onSubmit={handleSubmit(this.handleResetPassword)}>
+            <Field
+              name='newPassword'
```
jonambas: naming this `password` causes chrome to autofill

src/pages/auth/tests/__snapshots__/AuthPage.test.js.snap
```diff
@@ -68,5 +73,18 @@ exports[`should display tfa form when TFA is enabled 1`] = `
       onSubmit={[Function]}
     />
   </Panel>
+  <Panel.Footer
```
jonambas: Hm don't think this is supposed to show with TFA enabled

src/actions/password.js
```diff
@@ -0,0 +1,68 @@
+import { sparkpost as sparkpostRequest } from 'src/helpers/axiosInstances';
```
jasonrhodes: can we call this action and reducer "reset-password" to help when looking through quickly?

src/actions/password.js
```diff
@@ -0,0 +1,68 @@
...
+      method: 'POST',
+      url: '/users/password/forgot',
+      data
+    })
+      .then(() => dispatch(emailSuccess()))
+      .catch((err) => dispatch(emailError(err)));
+  };
+}
+
+function emailPending() {
```
jasonrhodes: I think it'll be fine to move these six functions directly inline to the `dispatch()` calls

i.e.
```js
.then(() => dispatch({ type: 'SEND_PASSWORD_EMAIL_SUCCESS' }))
```
etc, rather than have those six separate functions -- we may do that other places but probably only necessary if they have logic etc

src/pages/password/ForgotPasswordPage.js
```diff
@@ -0,0 +1,67 @@
...
+  message: 'If you have an account with us, please check your email for your password reset instructions.'
+};
+
+const errorAlert = {
+  type: 'error',
+  message: 'Unable to send your password reset email.'
+};
+
+export class ForgotPasswordPage extends Component {
+  componentDidUpdate(prevProps) {
```
jasonrhodes: should we log out on cWM or redirect to /landing if the user is logged in? Or maybe redirect to the user profile where a logged in user would change their password? Maybe it doesn't matter here but it feels a little weird...

src/pages/password/ResetPasswordPage.js
jonambas: raju suggested `autoComplete='off'`

src/pages/auth/AuthPage.js
```diff
@@ -88,20 +88,17 @@ export class AuthPage extends Component {
...
         <Panel sectioned accent title="Log In">
           { errorDescription && this.renderLoginError(errorDescription)}
 
           { tfa.enabled && <TfaForm onSubmit={this.tfaSubmit} /> }
           { !tfa.enabled && <LoginForm onSubmit={this.loginSubmit} ssoEnabled={this.state.ssoEnabled}/> }
         </Panel>
-
-        {hasSignup && <small>Don't have an account? <UnstyledLink to={'/join'}>Sign up</UnstyledLink>.</small> }
+        <Panel.Footer
+          left={hasSignup && !tfa.enabled && <small>Don't have an account? <UnstyledLink Component={Link} to='/join'>Sign up</UnstyledLink>.</small>}
```
jasonrhodes: So in this case, `tfa.enabled` really means that we're in the 2nd "TFA" step of login where they are entering their code, is that right? I was confused at first thinking we were blocking the sign up link if a user had TFA, but we wouldn't even know that until they logged in.

Should we just not render the Panel.Footer at all if in this state? I kind of want to fork earlier and have a whole separate Panel for each fork but maybe at least something like this for now:
```jsx
<Panel sectioned accent title="Log In">
  {error stuff}
  {tfa.enabled ? this.renderTfaStep() : this.renderLogin()}
```

jasonrhodes: If you think it makes more sense to have those be separate functional components like Brian has been advocating for, I think that's fine too, but render* functions are still fine with me until they totally blow up the file.

PR #342, opened 2018-03-07
FAD-6308 De-couple date management, reduxify suppressions and message events pages

jasonrhodes:
The DateFilter and DateForm components that used to be inside src/pages/reports/components are now in src/components/datePicker to be shared app-wide, and are called DatePicker and ManualEntryForm, respectively.

To Do:
* [x] Write new tests for uncovered/new code paths
* [x] Add prop types to DatePicker component for documentation

To test:
* Visit the Message Events page and make sure all combinations of searches work as expected
* Visit the Suppressions Page and make sure all combinations of searches work as expected
* Run through all of the other reports pages and make sure their date management is still working as expected
* Verify that dates picked in the regular reports are now de-coupled from both message events and suppressions (and that message events and suppressions are not coupled to each other, either).


src/__testHelpers__/renderHelpers.js
```diff
@@ -1,47 +0,0 @@
-import React from 'react';
```
jasonrhodes: We stopped using this file a while ago when @jonambas and I discovered that the latest versions of Enzyme and React handle this for us out of the box.

src/components/ipPoolTypeahead/PoolTypeahead.js
```diff
@@ -1,81 +0,0 @@
-import classnames from 'classnames/bind';
```
jasonrhodes: This wasn't used anywhere at all -- we can bring it back from history if we need it.

src/components/reduxFormWrappers.js
```diff
@@ -1,95 +0,0 @@
-import React from 'react';
```
jasonrhodes: This file should have been removed when we moved all these components into the reduxFormWrappers directory. Since it was still here, it was being used instead of the components in the folder, so I removed.

src/pages/reports/messageEvents/components/MessageEventsSearch.js
```diff
@@ -80,10 +64,28 @@ export class BasicFilter extends Component {
...
+              'hour',
+              'day',
+              '7days',
+              '10days',
+              'custom'
+            ]}
+            disabled={loading}
+            onChange={refreshMessageEventsDateRange}
+            datePickerProps={{
+              disabledDays: {
```
jonambas: Add `canChangeMonth: false,`

src/pages/reports/messageEvents/components/MessageEventsSearch.js
```diff
@@ -80,10 +64,28 @@ export class BasicFilter extends Component {
...
+                after: now,
+                before: moment(now).subtract(10, 'days').toDate()
```
jonambas: Datepicker will also need to validate/reset if entered date is outside this range

src/pages/reports/messageEvents/components/MessageEventsSearch.js
```diff
@@ -1,39 +1,31 @@
...
-
-  // TODO: move this to redux
-  state = {
-    recipients: ''
-  };
+export class MessageEventsSearch extends Component {
 
   componentDidMount() {
-    this.props.refreshReportOptions();
+    const relativeRange = this.props.search.dateOptions.relativeRange || 'hour';
```
jonambas: wonder if the `hour` default should be more obvious or maybe add a comment about it

src/pages/reports/messageEvents/components/MessageEventsSearch.js
```diff
@@ -80,10 +64,28 @@ export class BasicFilter extends Component {
...
 
   render() {
+    const { search, refreshMessageEventsDateRange, loading, now = new Date() } = this.props;
     return (
       <Grid>
         <Grid.Column xs={12} md={6}>
-          <DateFilter />
+          <DatePicker
+            {...search.dateOptions}
+            relativeDateOptions={[
```
jonambas: maybe move to constants file or somewhere else outside render like the other pages

src/selectors/suppressions.js
```diff
@@ -0,0 +1,26 @@
+import convertListToBoolHash from 'src/helpers/convertListToBoolHash';
```
jonambas: Need tests for this file

src/pages/reports/messageEvents/components/MessageEventsSearch.js
jasonrhodes: Per our talk IRL, we're punting on this for now

src/components/datePicker/DatePicker.js
```diff
@@ -153,15 +154,24 @@ export class DateFilter extends Component {
...
   }
 }
 
-const mapStateToProps = ({ reportOptions }) => ({ filter: reportOptions });
-export default connect(mapStateToProps, { refreshReportOptions })(DateFilter);
+AppDatePicker.propTypes = {
+  now: PropTypes.instanceOf(Date),
+  from: PropTypes.instanceOf(Date).isRequired,
+  to: PropTypes.instanceOf(Date).isRequired,
+  relativeRange: PropTypes.string,
```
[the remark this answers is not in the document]
jasonrhodes: This ended up being too hard with mocks and tests, etc :/

PR #345, opened 2018-03-09
FAD-6121 Add redirects for email links

jonambas:


src/config/emailRoutes.js
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Handles redirect for existing profile route
+ * and the templates 'email-verification', 'welcome-user'
+ * /account/profile?verify={{token}}
+ * to /account/email-verification/{{token}}
+ */
+export function emailVerificationRedirect(props) {
+  const search = qs.parse(props.location.search);
+  return search.verify ? <Redirect to={`/account/email-verification/${search.verify}`}/> : <ProfilePage {...props} />;
```
bkemper: How does this work if the user is not logged in?

src/config/emailRoutes.js
```diff
@@ -0,0 +1,44 @@
...
+}
+
+/**
+ * Handles redirect for template 'forgotten-password'
+ * /reset-password?token={{token}}
+ * to /reset-password/{{token}}
+ */
+function passwordReset(props) {
+  const search = qs.parse(props.location.search);
+  return <Redirect to={search.token ? `/reset-password/${search.token}` : '/'}/>;
```
bkemper: It's a shame this doesn't work out of the box.

src/config/emailRoutes.js
```diff
@@ -0,0 +1,44 @@
...
+  {
+    path: '/account/webhooks',
+    redirect: '/webhooks'
+  },
+  {
+    path: '/reset-password',
+    public: true,
+    forceLogout: true,
+    component: passwordReset
+  }
```
bkemper: Do we need this extra file or should we just include this in the `config/routes.js`.

src/config/emailRoutes.js
jonambas: I think ProtectedRoute should handle the redirect to the correct route

src/config/emailRoutes.js
jonambas: Wanted to keep these separate to easily remove later

PR #348, opened 2018-03-12
FAD-6288 Suspended User Experience

jasonrhodes:
The work for a _suspended_ user is simple:

* A yellow banner shows up when you first log in and it does not auto-dismiss. If you try to do any non-read action while that banner is up, nothing else happens.
* Once you dismiss the initial banner, if you try to do any non-read action, the yellow banner re-appears (and auto-dismisses with the default timeout).
* While suspended, a 403 response never logs you out

For the "suspended for billing reasons" experience, I had to make more changes:

* The banner already existed, but the condition to make it appear was inaccurate. Instead of pushing a computed boolean like `isSuspendedForBilling` onto the account object via selector, I opted to lean into our access control condition logic so that these decisions all flow through one place. This involved:
  * Making condition helpers _positive_ in direction (isEnterprise instead of notEnterprise)
  * Introducing a `not()` helper that reverses a condition
  * Re-naming `composeConditions` to `all()` to represent the case where you want to grant access when _all_ of the given conditions are true
  * Introducing `any()` as a counterpart to `all()`, for the case where you want to grant access when _any one_ of the given conditions is true
  * Introducing a `<ConditionSwitch>` component, modeled after the react-router `<Switch>`, that forces only one of the given component/condition pairs to render (the first where the condition returns true), to guard against increasingly complex if/then logic in a page like Billing
* The update payment form now appears beneath the banner when you are suspended for billing

### How to test

To test these cases, you can either use the internal REST API to set your account to status = "suspended" for the regular case, and then add "status_reason" = "Suspended following payment failure in Zuora" for the "suspended for billing case".

Alternatively, use Redux Dev Tools to re-emit the "FETCH_ACCOUNT_SUCCESS" action but modify status to be "suspended" for the regular case, and add "status_reason_category" to be "100.01" for the "suspended for billing" case.

### DIFF
--- a/src/actions/globalAlert.js
+++ b/src/actions/globalAlert.js
@@ -1,8 +1,19 @@
 export function showAlert(alert) {
   return {
     type: 'SHOW_GLOBAL_ALERT',
-    payload: { ...alert }
+    payload: alert
   };
+}
+
+export function showSuspensionAlert(overrides = {}) {
+  return showAlert({
+    type: 'warning',
+    message: 'This account is currently suspended',
+    details: 'For more information about account suspension, please email compliance@sparkpost.com',
+    maxWidth: 800,
+    dedupeId: 'SUSPENSION_NOTICE',
+    ...overrides
+  });
 }
 
 export function clear(id) {
--- a/src/actions/helpers/sparkpostApiRequest.js
+++ b/src/actions/helpers/sparkpostApiRequest.js
@@ -1,6 +1,6 @@
 import requestHelperFactory from 'src/actions/helpers/requestHelperFactory';
 import { refresh, logout } from 'src/actions/auth';
-import { showAlert } from 'src/actions/globalAlert';
+import { showAlert, showSuspensionAlert } from 'src/actions/globalAlert';
 import { useRefreshToken } from 'src/helpers/http';
 import { resolveOnCondition } from 'src/helpers/promise';
 import _ from 'lodash';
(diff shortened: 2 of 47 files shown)


src/actions/globalAlert.js
```diff
@@ -1,10 +1,21 @@
...
-    payload: { ...alert }
+    payload: alert
   };
 }
 
+export function showSuspensionAlert(overrides = {}) {
+  return showAlert({
+    type: 'warning',
+    message: 'This account is currently suspended',
+    details: 'For more information about account suspension, please email compliance@sparkpost.com',
```
gwalls: Is there an easy way to make this a href with mailto: ?

src/actions/helpers/sparkpostApiRequest.js
```diff
@@ -91,10 +94,17 @@ const sparkpostRequest = requestHelperFactory({
       meta
     });
 
+    // if the account is suspended, a 403 should not log the user out
+    if (response.status === 403 && account.status === 'suspended') {
+      dispatch(showSuspensionAlert());
+      // TODO: dispatch(getAccount());
```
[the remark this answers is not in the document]
jasonrhodes: Nah I'll do it here, thanks.

src/actions/globalAlert.js
jasonrhodes: Let me give it a try.

src/pages/reports/engagement/components/EngagementTable.js
```diff
@@ -71,7 +71,7 @@ export default function EngagementTable({ data, loading }) {
   // Must include percentage in data for sorting
   const dataWithPercentage = data.map((row) => ({
     ...row,
-    percentage_clicked: formatPercent(safeRate(row.count_clicked, totalClicks))
+    percentage_clicked: safeRate(row.count_clicked, totalClicks)
```
gwalls: 🙈

PR #353, opened 2018-03-15
FAD-6340 wired up save as csv, need to fix css and handle objects wit…

bizob2828:
- [ ] fix message events where certain keys have objects within
- [x] fix css and figure out positioning

### DIFF
--- a/src/components/collection/Pagination.js
+++ b/src/components/collection/Pagination.js
@@ -3,22 +3,24 @@ import { Pagination, Button } from '@sparkpost/matchbox';
 import classnames from 'classnames';
 import styles from './Pagination.module.scss';
 import Papa from 'papaparse';
+import _ from 'lodash';
 
 export const defaultPerPageButtons = [10, 25, 50, 100];
 
 class CollectionPagination extends Component {
   formatToCsv = () => {
     const { data: rows } = this.props;
-    const data = Papa.unparse(rows);
+    // we are doing this because certain keys are objects/array which papa parse doesn't stringify
+    const mappedRows = _.map(rows, (row) => _.mapValues(row, (value) => _.isObject(value) || _.isArray(value) ? JSON.stringify(value) : value));
+    const data = Papa.unparse(mappedRows);
     return `data:text/csv;charset=utf-8,${encodeURI(data)}`;
   }
 
   renderCSVSave() {
-    const { saveCsv, data, perPage } = this.props;
+    const { saveCsv, data, perPageButtons } = this.props;
 
-    // mimick behavior with "old" webui where we do not show
-    // save as csv button when less than a page
-    if (!saveCsv || data.length < perPage) { return null; }
+    // do not show save as csv if only 1 page of smallest per page increment
+    if (!saveCsv || data.length <= Math.min(...perPageButtons)) { return null; }
 
     const now = Math.floor(Date.now() / 1000);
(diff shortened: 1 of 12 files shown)


src/components/collection/Pagination.js
```diff
@@ -60,6 +76,7 @@ class CollectionPagination extends Component {
         </div>
         <div className={styles.PerPageButtons}>
           {this.renderPerPageButtons()}
+          {this.renderCSVSave()}
```
jasonrhodes: I'm not reviewing this but can we check what happens in the existing app re: when this button shows up? If our current UI doesn't show it with < x results, we should duplicate.

src/components/collection/Pagination.js
```diff
@@ -2,10 +2,28 @@ import React, { Component } from 'react';
...
+  formatToCsv = () => {
+    const { data: rows } = this.props;
+    const data = Papa.unparse(rows);
+    return `data:text/csv;charset=utf-8,${encodeURI(data)}`;
+  }
+
+  renderCSVSave() {
+    const { saveCsv, data, perPage } = this.props;
+
+    // mimick behavior with "old" webui where we do not show
```
bizob2828: bob, learn how to spell

src/components/collection/Pagination.js
```diff
@@ -2,10 +2,28 @@ import React, { Component } from 'react';
...
+    // save as csv button when less than a page
+    if (!saveCsv || data.length < perPage) { return null; }
+
+    const now = Math.floor(Date.now() / 1000);
+    return <Button download={`sparkpost-csv-${now}.csv`} to={this.formatToCsv()}>Save As CSV</Button>;
```
rajumsys: this is fine but in case you want to match current ui, it says `export_data_<uniq>`

src/components/collection/Pagination.js
```diff
@@ -14,9 +14,11 @@ class CollectionPagination extends Component {
   }
 
   renderCSVSave() {
-    const { saveCsv } = this.props;
+    const { saveCsv, data, perPage } = this.props;
 
-    if (!saveCsv) { return null; }
+    // mimick behavior with "old" webui where we do not show
+    // save as csv button when less than a page
+    if (!saveCsv || data.length < perPage) { return null; }
```
rajumsys: why is that? now if I choose 100 per page and data length is 99, it won't show the button.

rajumsys: screen shot 2018-03-15 at 4 41 14 pm
old webui shows the button even with 4 records.

jasonrhodes: Not sure what the logic is in that page you're looking at @rajumsys because Bob and I verified pagination doesn't show up for small data sets in the existing UI ... 

At any rate, at the very least we should mimic the logic in `renderPerPageButtons` and not `renderPageButtons`, i.e.:
```javascript
    if (data.length <= Math.min(...perPageButtons)) {
      return null;
    }
```

rajumsys: what's the downside of showing it when `data.length > 0`?

jasonrhodes: Because a) a download CSV link for 1-5(ish) rows seems pretty silly and b) the existing UI didn't seem to expose this so we didn't want to add _more_ silly CSV functionality than we currently had in the app

src/components/collection/Pagination.js
bizob2828: yea i didn't

PR #356, opened 2018-03-16
FAD-6237 Send more CTA!

rajumsys:
Todos:
- [x] Tests for new action creator
- [x] Tests for support ticket form
- [x] IP address pricing bug fix for AWS customer

Notes:
- Unlike the old UI, I've put the support ticket note in the form. Can move if not liked! 

**Testing**
- Login as 'free' account (cta to upgrade), account which email isn't verified (cta to verify email) and non-free account (cta to create support request) to test three cases of the CTA. 
- Login as a free aws account not-to see the upgrade CTA but login as a non-free account to see it. Reach out if you need help with setting a aws mktplace account.


ewandennis:
Initial UX notes:
 - The form should probably be a modal.
 - Can we include a cancel button?
 - The daily limit field should use positive integer validation. There are `integer()` and `minNumber()` validators in `src/helpers/validation` that should help.
 - I wonder if we can position the CTA better than the last UI.
    - @jonambas: Would it work as a panel action if we move the verify email and upgrade sub-CTAs into the modal?

Now, to the code.

src/components/usageReport/RequestForm.js
```diff
@@ -0,0 +1,59 @@
+import React, { Component } from 'react';
+import { Field, reduxForm } from 'redux-form';
+import { Button, Panel, Banner } from '@sparkpost/matchbox';
+import { TextFieldWrapper } from 'src/components';
+import { required, minLength, hasNumber } from 'src/helpers/validation';
+
+const formName = 'requestForm';
```
ewandennis: Given that redux-form needs an app-unique identifier for each form, maybe we should say something like:

`const formName = 'dailyLimitIncreaseRequestForm';`

src/components/usageReport/RequestForm.js
```diff
@@ -0,0 +1,59 @@
...
+        </h6>
+        <Banner status='info'>
+          <p>Please note this process may take up to 2 business days.</p>
+        </Banner>
+
+      </Panel.Section>
+      <form onSubmit={handleSubmit(onSubmit)}>
+        <Panel.Section>
+          <Field
+            name='daily_limit'
```
ewandennis: Small point but maybe use camel case?

src/components/usageReport/RequestForm.js
```diff
@@ -0,0 +1,59 @@
...
+            label='Daily Limit'
+            inlineErrors={true}
+            disabled={submitting}
+            validate={[required, hasNumber]}
```
ewandennis: `validate={[required, integer, minNumber(...)]}`

I suppose we could use a minimum of `1` but since the parent has the current daily limit, we _could_ use it here.

Also, if you include `required={true}` on `Field`, [the label gets a little asterisk[(https://sparkpost.github.io/matchbox/?selectedKind=TextField&selectedStory=required&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel).

src/components/usageReport/RequestForm.js
```diff
@@ -0,0 +1,59 @@
...
+            component={TextFieldWrapper}
+          />
+
+          <Field
```
ewandennis: `required={true}`

src/components/usageReport/RequestForm.js
```diff
@@ -0,0 +1,59 @@
...
+            name='reason'
+            label='Reason for Increase?'
+            inlineErrors={true}
+            disabled={submitting}
+            validate={[required, minLength(5)]}
+            component={TextFieldWrapper}
+          />
+        </Panel.Section>
+        <Panel.Section>
+          <Button submit primary disabled={pristine || invalid || submitting}>
```
ewandennis: Cancel button please :)

src/components/usageReport/tests/SendMoreCTA.test.js
```diff
@@ -0,0 +1,86 @@
...
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('renders correctly for aws free plan (cta to upgrade)', () => {
+    wrapper.setProps({ currentPlan: { code: 'aws_free' }});
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('renders correctly for customers who can request limit bump', () => {
+    wrapper.setProps({ currentPlan: { code: 'pro1' }});
```
ewandennis: Small point but this is a deprecated plan. Maybe use `250K-0817` or similar?

src/pages/billing/components/BillingSummary.js
```diff
@@ -56,6 +56,7 @@ export default class BillingSummary extends Component {
       count={this.props.sendingIps.length}
       plan={this.props.currentPlan}
       onClick={this.handleIpModal}
+      isAWSAccount={this.props.isAWSAccount}
```
ewandennis: I could use some notes on how to test the AWS cases. I think we have these:

 - Free AWS accounts should not be able to request limit increases
 - AWS accounts should see hourly rates for dedicated IPs

src/components/usageReport/RequestForm.js
rajumsys: Cancel button can be added but unlike old UI, i made the `Submit a request` to toggle the form.

src/pages/billing/components/BillingSummary.js
rajumsys: ah, sure. i've added some notes a few mins ago (see the PR description).

src/components/usageReport/SendMoreCTA.js
```diff
@@ -0,0 +1,110 @@
...
+        return showAlert({
+          type: 'success',
+          message: 'We got your request and you should hear from us soon.'
+        });
+      });
+  }
+
+  render() {
+    const { currentUser: { email_verified: emailVerified }, currentPlan } = this.props;
+    const allowSendingLimitRequest = currentPlan.status !== 'deprecated' && !/free/.test(currentPlan.code);
```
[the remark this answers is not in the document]
rajumsys: I'm fine either way. Let me know if you prefer a selector.  Using `isFree` sounds good idea though.

src/components/usageReport/RequestForm.js
rajumsys: Good idea (both of them)

rajumsys:
> The form should probably be a modal.

The ticket didn't say it to be a modal. So I didn't. If we want to change it, let me know but may be it's a significant refactoring. 

> Can we include a cancel button?

We can if you think that'll be helpful. I just made the CTA link toggle-able. 

> The daily limit field should use positive integer validation. There are integer() and minNumber() validators in src/helpers/validation that should help.

(Y)

> I wonder if we can position the CTA better than the last UI.
@jonambas: Would it work as a panel action if we move the verify email and upgrade sub-CTAs into the modal?

Yeah, I initially thinking as panel option but later decided to stay on course of parity ;).

ewandennis:
Looking good. Coupla quick points:

 - The `dailyLimit` field should probably validate on `minNumber(currentLimit+1)`
 - The appteam account in `tst` has no daily limit which breaks `dailyLimit` validation. `currentLimitSelector()` should probably default to `1` [here](https://github.com/SparkPost/2web2ui/blob/e05dd46d5423feb40889b2a4c09569c09deec5db/src/selectors/support.js#L18) and then we can remove the default value in `RequestForm` [here](https://github.com/SparkPost/2web2ui/blob/e05dd46d5423feb40889b2a4c09569c09deec5db/src/components/usageReport/RequestForm.js#L12).
 - [Since we added new selectors and tests to a suite with only 1](https://github.com/SparkPost/2web2ui/blob/e05dd46d5423feb40889b2a4c09569c09deec5db/src/selectors/tests/support.test.js#L10), could we also wrap the existing tests in a `describe()`?
 - Can we explicitly set a `plan.status` for our `allowSendingLimitRequestSelector` tests? We're currently testing if it succeeds in the absence of a field it depends on which might trip us up later.

PR #359, opened 2018-03-19
FAD-6372 and friends

ewandennis:
### FAD-6372: For ... of
Issue: for...of isn't well supported on DOM collections
Fix: Use `[...collection].forEach()`

Testing: Use Edge (!) or an old Safari to preview a template containing a link. Or just eyeball the patch in this PR.

###  FAD-6385: AWS users should see hourly dedicates IP pricing
Issue: we show AWS users incorrect monthly IP pricing in the billing summary
Fix: Pull the AWS account state from subscription details and the correct pricing from config

Testing: Use React dev tools to set isAWSAccount on the BillingSummary under `/account/billing` and voila.

Also fixed a type on `/account/smtp`.

### FAD-6377: Webhooks "active" checkbox state doesn't match underlying `active` field
Issue: Active webhooks appear inactive in our edit form and confuse the rest of the form's `dirty` state.
Fix: Tie our `<Checkbox />` `checked` state to `input.value` from redux-form.

Testing: Visit an active webhook and verify:
 - the active checkbox shows the correct initial state
 - the submit button goes active when unchecking active
 - the submit button goes inactive when re-checked active

### DIFF
--- a/src/pages/webhooks/components/Fields.js
+++ b/src/pages/webhooks/components/Fields.js
@@ -65,14 +65,16 @@ const AuthDropDown = () => (
   />
 );
 
-const ActiveField = () => (
-  <Field
+const ActiveField = () => {
+  // Artificially set a checked prop so the underlying input displays properly.
+  const CheckableCheckbox = ({ input, ...rest }) => <CheckboxWrapper input={input} {...rest} checked={Boolean(input.value)} />;
+  return <Field
     name='active'
     label='Active'
-    component={CheckboxWrapper}
+    component={CheckableCheckbox}
     helpText='An inactive webhook will not transmit any data.'
-  />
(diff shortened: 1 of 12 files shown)


src/pages/webhooks/components/Fields.js
```diff
@@ -65,14 +65,16 @@ const AuthDropDown = () => (
   />
 );
 
-const ActiveField = () => (
-  <Field
+const ActiveField = () => {
+  // Artificially set a checked prop so the underlying input displays properly.
```
jasonrhodes: Mind just adding a TODO in here to fix this up later?

ewandennis: FAD-6388

PR #365, opened 2018-03-20
FAD-6126: Template list/preview access for reporting users

bkemper:
_Refer to [FAD-6126](https://jira.int.messagesystems.com/browse/FAD-6126)_

**Todo**
- [x] Visible template navigation link
- [x] Load template list page
- [x] Hide create template button on list page
- [x] Hide delete, duplicate, publish, and save buttons on edit page
- [x] Disabled form and editor for edit page (only needed for drafts, published are already disabled)
- [x] Hide send button and to field on preview page 
- [x] Save substitution data when click preview and send button

### DIFF
--- a/src/config/routes.js
+++ b/src/config/routes.js
@@ -236,7 +236,7 @@ const routes = [
   {
     path: '/templates',
     component: templates.ListPage,
-    condition: all(hasGrants('templates/modify'), ({ account }) => account.status === 'active'),
+    condition: hasGrants('templates/view'),
     layout: App
   },
   {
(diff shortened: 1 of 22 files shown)


ewandennis:
There's a stealth ListPage re-org in the AC of this ticket which isn't addressed in the PR:
>Templates list page unlinks name and uses "Draft" "Published" "Preview" buttons in far right column (may remove other columns to make room) for all users

I tried this out and it's ok if we use active voice on the buttons and find a simple way to lay them out cleanly. It raises wider questions though, like "why no delete action?" and "should we put actions on all list pages?".

So I wonder if we want to tackle a list page cleanup separately and leave this PR as is. If so, let's cut that improvement.

bkemper:
Ewan and I paired on this.  He gave his approval to deploy.  I'll file a follow-up ticket to address his concern about list pages.

PR #375, opened 2018-03-26
FAD-5564 Restricts editing API keys to owners only

jasonrhodes:
screen shot 2018-03-26 at 2 25 22 pm

When a key belongs to another admin user on your account, the owner's name is shown and the label is no longer a link to the edit page.

screen shot 2018-03-26 at 2 25 28 pm

On hover, a tooltip describes why the label is not linked.

screen shot 2018-03-26 at 3 51 36 pm

Attempting to visit the edit page for a key you don't own redirects you back to the list page with the red error alert.

---

*Added in here to improve the filterbox experience*

screen shot 2018-03-26 at 3 48 48 pm

The existing FilterBox shows a placeholder _and_ help text, which is redundant. It also uses fake data like "some-id" and "some-key" instead of showing an example of real data. 

screen shot 2018-03-26 at 3 42 31 pm

This ticket removes the help text and improves the placeholder, adding the first value for each supplied key that it can find in the data set.

### DIFF
--- a/src/pages/api-keys/DetailsPage.js
+++ b/src/pages/api-keys/DetailsPage.js
@@ -1,6 +1,6 @@
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
-import { Link, withRouter } from 'react-router-dom';
+import { Link, Redirect } from 'react-router-dom';
 import { Page, Panel } from '@sparkpost/matchbox';
 
 import { deleteApiKey, getApiKey, updateApiKey, listGrants, listSubaccountGrants } from 'src/actions/api-keys';
@@ -61,10 +61,14 @@ export class ApiKeysDetailsPage extends Component {
   };
 
   render() {
-    const { apiKey, loading } = this.props;
+    const { apiKey, error, loading } = this.props;
 
     if (loading) {
       return <Loading />;
+    }
+
+    if (error) {
+      return <Redirect to='/account/api-keys' />;
     }
 
     return (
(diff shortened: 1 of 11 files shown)


src/pages/api-keys/DetailsPage.js
```diff
@@ -101,6 +105,4 @@ const mapStateToProps = (state, props) => {
   };
 };
 
-export default withRouter(
-  connect(mapStateToProps, { getApiKey, updateApiKey, listGrants, listSubaccountGrants, deleteApiKey, showAlert })(ApiKeysDetailsPage)
-);
+export default connect(mapStateToProps, { getApiKey, updateApiKey, listGrants, listSubaccountGrants, deleteApiKey, showAlert })(ApiKeysDetailsPage);
```
bkemper: Looks like `handleDelete` still need `withRouter`.

src/selectors/tests/api-keys.test.js
```diff
@@ -112,4 +112,36 @@ describe('ApiKey Selectors', () => {
...
+          {
+            // no username and no subaccount_id (should never happen but should produce false if so)
+            lol: 'wut'
+          }
+        ]
+      }
+    };
+
+    const list = apiKeys.selectKeysForAccount(store);
+    expect(list.map((key) => key.isOwnedByCurrentUser)).toEqual([true, false, true, false]);
```
bkemper: Do we need this test and the snapshot?

src/selectors/tests/__snapshots__/api-keys.test.js.snap
```diff
@@ -70,3 +70,24 @@ Object {
...
+    "isOwnedByCurrentUser": false,
+    "username": "other",
+  },
+  Object {
+    "isOwnedByCurrentUser": true,
+    "subaccount_id": 123,
+  },
+  Object {
+    "isOwnedByCurrentUser": false,
+    "lol": "wut",
```
bkemper: 😄

src/pages/api-keys/DetailsPage.js
jasonrhodes: Page components should never need `withRouter`

PR #376, opened 2018-03-26
FAD-6423 Collect payments after updating payment info

rajumsys:
How to test:
- Go to /billing; open your console
- Click `Update Payment Information`
- Put the new credit card info
- Submit the form. You should see a request to `/collect` endpoint

Notes:
- Contrary to what ticket said, seems api doesn't respond with 400 when nothing to be collected. Seems API now responds a 200 with a description. [See](https://github.com/SparkPost/accusers-api/blob/master/api/resources/account-controller.js#L878-L880)
- [Test credit cards](https://developer.merchante-solutions.com/#/payment-gateway-testing#validation)

### DIFF
--- a/src/actions/billing.js
+++ b/src/actions/billing.js
@@ -149,7 +149,6 @@ export function billingCreate(values) {
 
 /**
  * attempts to collect payments (like when payment method is updated) to make sure pending payments are charged
- * should suppress 400 from api as they are good here (means nothing to collect)
  */
 
 export function collectPayments() {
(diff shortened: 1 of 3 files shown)


src/actions/billing.js
```diff
@@ -147,6 +147,21 @@ export function billingCreate(values) {
   };
 }
 
+/**
+ * attempts to collect payments (like when payment method is updated) to make sure pending payments are charged
+ * should suppress 400 from api as they are good here (means nothing to collect)
```
jasonrhodes: remove this part of the comment now, yay

jasonrhodes:
One small change to remove unnecessary comment, otherwise LGTM!

src/actions/billing.js
rajumsys: what? i removed that! how come it's still here!

rajumsys: ok, i haven't (force) pushed earlier

PR #380, opened 2018-03-28
FAD-6367 Disable editor validation unless json

jonambas:


src/pages/templates/components/async/AceWrapper.js
```diff
@@ -33,7 +34,7 @@ const AceWrapper = ({
       highlightActiveLine
       showPrintMargin={false}
       setOptions={{
-        // useWorker: false, - disables error checking
+        useWorker: syntaxValidation,
```
jonambas: Thought about just doing `mode === 'json'` here, but thought itd be better to be explicit about it

src/pages/templates/components/async/tests/AceWrapper.test.js
```diff
@@ -0,0 +1,60 @@
...
+    expect(props.input.onChange).toHaveBeenCalled();
+
+    wrapper.find('ReactAce').simulate('blur');
+    expect(props.input.onBlur).toHaveBeenCalledWith('test');
+
+    wrapper.find('ReactAce').simulate('focus');
+    expect(props.input.onFocus).toHaveBeenCalledWith('test');
+  });
+
+  it('should show errors correctly', () => {
```
ewandennis: Great to have more coverage and most of these grouped tests are fine. In the `should show errors correctly` case though, could we split the test cases up and give the meaningful names? For example, it's useful for newcomers to see at a glance that when the editor is active / has focus, we don't show errors.

PR #381, opened 2018-03-29
FAD-6286: fixes

ewandennis:
- Show the banner before `state.accessControlReady` since we want to show it on `/auth` and `/join`
 - Reinstate `state.cookieConsent.cookieSet` from the cookie on `LOGOUT` to avoid showing the banner to a consenting user after logout
- [x] tests

### DIFF
--- a/src/components/cookieConsent/CookieConsent.js
+++ b/src/components/cookieConsent/CookieConsent.js
@@ -40,11 +40,7 @@ export class CookieConsent extends React.Component {
   }
 
   render() {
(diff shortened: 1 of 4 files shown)


rajumsys:
LGTM

The only problem is that in a smaller (height) screen, this will block the part of the form. if @jonambas is ok with this, i'm good. 
screen shot 2018-03-29 at 1 22 24 pm

jonambas:
As long as that's dismissible, shouldn't be a problem

PR #383, opened 2018-03-29
FAD-6168: Link to SPC-EU or SPC on sign-up page (phoenix ui)

bkemper:


src/pages/join/components/JoinLink.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { UnstyledLink } from '@sparkpost/matchbox';
+
+import config from 'src/config';
+import { LINKS, SPC_EU_TENANT } from 'src/constants';
+
+export default function JoinLink({ location, tenant = config.tenant }) {
```
rajumsys: does it really need to accept `tenant` arg? haven't seen a use.

src/pages/join/components/JoinLink.js
```diff
@@ -0,0 +1,19 @@
...
+  const brand = tenant === SPC_EU_TENANT
+    ? { label: 'US', url: LINKS.SPC_US_URL }
+    : { label: 'EU', url: LINKS.SPC_EU_URL };
+
+  return (
+    <small>
+      <UnstyledLink to={`${brand.url}${location.pathname}${location.search}`}>
```
rajumsys: this will reload the app than internal routing. i believe that makes sense and intentional. just wanted to call out as we did this mistakenly in Join page :) 

minor: probably would be easier to read by prepping the url in a const first and then just using it here.

src/pages/join/components/JoinLink.js
ewandennis: `tenant` is used in tests.

src/pages/join/components/JoinLink.js
ewandennis: Yes, deliberate. Loading a different tenant means pulling a different tenant config file.

ewandennis:
Made a few tweaks:
- Show the cross links _only_ for SPC and SPCEU tenants
 - Customize the signup page title to make *SparkPost EU* more prominent at Jon's suggestion

(scuse the rude rebase)

src/config/tenant.js
```diff
@@ -0,0 +1,5 @@
+import config from '.';
+import { SPC_TENANT, SPC_EU_TENANT } from 'src/constants';
+
+export const inSPC = () => config.tenant === SPC_TENANT;
+export const inSPCEU = () => config.tenant === SPC_EU_TENANT;
```
bkemper: Do these need to be functions?

src/constants/index.js
```diff
@@ -38,3 +40,11 @@ export const ENTERPRISE_PLAN_CODES = ['ent1'];
 export const AWS_COOKIE_NAME = 'aws-mkt';
 
 export const DAILY_LIMIT_REQUEST_TEMPLATE = 'daily-limit-increase';
+
+export const SPC_EU_TENANT = 'spceu';
+export const SPC_TENANT = 'spc';
```
bkemper: Isn't this only true for production?  Should we handle for uat and staging?

src/pages/join/JoinPage.js
```diff
@@ -65,12 +68,14 @@ export class JoinPage extends Component {
   render() {
     const { createError } = this.props.account;
     const { formData } = this.state;
+    const title = inSPCEU() ? 'Sign Up For SparkPost EU' : 'Sign Up';
```
bkemper: 👍

src/pages/join/components/JoinLink.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import { UnstyledLink } from '@sparkpost/matchbox';
+
+import config from 'src/config';
+import { CROSS_LINK_MAP } from 'src/constants';
+
+export default function JoinLink({ location, crossLinkTenant = config.crossLinkTenant }) {
+  const brand = CROSS_LINK_MAP[crossLinkTenant];
+  if (!brand) {
```
bkemper: Smart!  I assumed only self service users would only ever see the join page.

src/constants/index.js
bkemper: I see now.  This is only intended to be used with cross linking which is only set up for cross linking production environments.  👍

PR #384, opened 2018-03-30
FAD-6309 implemented action onSuccess callback for billing update; ch…

gwalls:
…anged SparkpostApiError to be class


gwalls:
Note: we won't be able to _not_ throw errors from request wrappers until we can ensure there are no longer any catch's dependent on them.

src/actions/account.js
```diff
@@ -1,11 +1,12 @@
 import sparkpostApiRequest from 'src/actions/helpers/sparkpostApiRequest';
 
-export function fetch(params = {}) {
+export function fetch(params = {}, onSuccess) {
```
bkemper: Can we include `onSuccess` as an argument of `params`?

src/actions/helpers/sparkpostApiRequest.js
```diff
@@ -35,7 +35,7 @@ const sparkpostRequest = requestHelperFactory({
       meta
     });
 
-    return results;
+    return meta.onSuccess ? dispatch(meta.onSuccess({ results })) : results;
```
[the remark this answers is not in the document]
jasonrhodes: I think if we do what you suggest, you wouldn't end up with the last results of the chain, you'd end up with the first ... like `chain(a, b, c, d)()` would return a promise fulfilled with the results of a ... now in theory we'll _never_ need those results because we're never chaining manually but it's still a bit weird I think?

Unless I'm misunderstanding which is very possible with this mind bender :)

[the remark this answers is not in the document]
bkemper: Nope, I think you nailed it.  That makes sense.  Let's leave it.

src/actions/helpers/chainActions.js
```diff
@@ -0,0 +1,18 @@
...
+          actions[4].action({ ...results, meta: { onSuccess: undefined })
+        }})
+      }})
+    }})
+  }});
+ **/
+export default function chainActions(...actions) {
+  return actions.reduceRight((last, current) => {
+    const meta = { onSuccess: last };
+    return (results = {}) => current({ ...results, meta });
```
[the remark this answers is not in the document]
gwalls: I could do that, but I think it was @jasonrhodes that thought spreading `meta` over the action's request properties would allow for more flexibility in what gets passed.

src/actions/helpers/requestHelperFactory.js
```diff
@@ -10,7 +10,7 @@ function defaultOnSuccess({ types, response, dispatch, meta, action }) {
   return response;
 }
 
-function defaultOnFail({ types, err, dispatch, meta, action }) {
```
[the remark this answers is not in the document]
gwalls: will it hurt anything? I probably just removed it because Webstorm complained about it now being used. But I suppose it makes sense to keep it in the defaultOnFail so that others know what that method accepts, if they overwrite it.

src/actions/helpers/sparkpostApiError.js
```diff
@@ -1,17 +1,18 @@
...
+    this.stack = error.stack;
 
-  // Intentionally assigning additional properties
-  Object.assign(this, error);
+    // Intentionally assigning additional properties
+    Object.assign(this, error);
+  }
 }
 
-SparkpostApiError.prototype = Object.create(Error.prototype);
```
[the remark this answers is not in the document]
gwalls: If that's important, then I might have to revert this whole whole error class. I think it was babel that didn't let me assign to the read-only protoype property of the SparkpostApiError class (but apparently doesn't care if SparkpostApiError is a function instead).

src/actions/tests/billingCreate.test.js
```diff
@@ -0,0 +1,73 @@
...
+    billing.createZuoraAccount = jest.fn(({ meta }) => ({ meta }));
+    billing.syncSubscription = jest.fn(({ meta }) => ({ meta }));
+    // fetch = jest.fn(({ meta }) => ({ meta }));
+    // chainActionsMock = jest.genMockFromModule('src/actions/helpers/chainActions').default;
+    // chainActionsMock = jest.spyOn(actionHelper, 'default');
+  });
+
+  it('should dispatch a chained billing create action', async() => {
+    await billingCreate({ some: 'test-values' })(dispatchMock, getStateMock);
+    // expect(billing.cors).toHaveBeenCalledWith({ meta: { onSuccess: () => {} }, context: 'update-billing', data: { some: 'test-cors-data' }});
```
gwalls: I just could not get anything working here :(

src/actions/helpers/chainActions.js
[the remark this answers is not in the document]
jasonrhodes: Yeah I did ... but Brian talked me out of it by bullying me in a Slack DM.

But seriously, we should make it _as hard as possible_ to extend these horrors in the future, so cleaning that up and slimming it down a touch is fine with me if you're cool with it @gwalls

src/pages/onboarding/ChoosePlan.js
```diff
@@ -46,7 +47,11 @@ export class OnboardingPlanPage extends Component {
     // Note: billingCreate will update the subscription if the account is AWS
     return billingCreate(values)
       .then(() => history.push(NEXT_STEP))
-      .then(() => showAlert({ type: 'success', message: 'Added your plan' }));
+      .then((res) => {
+        if (!(res instanceof Error)) {
```
gwalls: hmm, note to self: this might have only worked because I was reproducing with Zuora `Error`'s and not `SparkpostApiError`'s

src/actions/helpers/chainActions.js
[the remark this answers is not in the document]
gwalls: 👍 works for me!

src/actions/billing.js
```diff
@@ -52,30 +52,32 @@ export function updateBillingContact(data) {
   });
 
   return (dispatch) => dispatch(action)
-    .then(() => dispatch(fetchAccount({ include: 'usage,billing' })));
+    .then(() => dispatch(fetchAccount({ params: { include: 'usage,billing' }})));
```
bkemper: This doesn't need to change.

src/actions/helpers/requestHelperFactory.js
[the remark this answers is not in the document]
bkemper: Nope.  You can leave it removed.

src/components/usageReport/UsageReport.js
```diff
@@ -36,7 +36,7 @@ const ProgressLabel = ({ title, secondaryTitle }) => (
 
 export class UsageReport extends Component {
   componentDidMount() {
-    this.props.getAccount({ include: 'usage' });
+    this.props.getAccount({ params: { include: 'usage' }});
```
bkemper: This shouldn't need to change.

src/pages/billing/ChangePlanPage.js
```diff
@@ -11,7 +11,7 @@ import { canChangePlanSelector } from 'src/selectors/accountBillingInfo';
 
 export class ChangePlanPage extends Component {
   componentWillMount() {
-    this.props.fetchAccount({ include: 'billing' });
+    this.props.fetchAccount({ params: { include: 'billing' }});
```
bkemper: This shouldn't need to be changed.

src/pages/billing/SummaryPage.js
```diff
@@ -15,7 +15,7 @@ import SuspendedForBilling from './components/SuspendedForBilling';
 export class BillingSummaryPage extends Component {
 
   componentDidMount() {
-    this.props.fetchAccount({ include: 'billing' });
+    this.props.fetchAccount({ params: { include: 'billing' }});
```
bkemper: This shouldn't need to be changed.

src/pages/billing/forms/ChangePlan.js
```diff
@@ -54,7 +56,11 @@ export class ChangePlan extends Component {
 
     return action
       .then(() => history.push('/account/billing'))
-      .then(() => showAlert({ type: 'success', message: 'Subscription Updated' }));
+      .then((res) => {
+        if (!(res instanceof Error)) {
```
bkemper: Is this condition needed?

src/pages/billing/forms/UpdatePayment.js
```diff
@@ -21,9 +22,11 @@ export class UpdatePayment extends Component {
 
   onSubmit = (values) => {
     const { billingUpdate, onCancel, showAlert } = this.props;
-    return billingUpdate(values).then(() => {
-      showAlert({ type: 'success', message: 'Payment Information Updated' });
-      typeof onCancel === 'function' && onCancel();
+    return billingUpdate(values).then((res) => {
+      if (!(res instanceof Error)) {
```
bkemper: Is this condition needed?

src/pages/onboarding/ChoosePlan.js
bkemper: You are correct.  Do we need this condition?

src/actions/helpers/chainActions.js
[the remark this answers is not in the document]
gwalls: So I have no problem doing this, but I might suggest we defer to a later time since this ticket is dragging on 😢 
I'll leave that up to you guys.

src/pages/billing/forms/ChangePlan.js
gwalls: Nope. I'm still throwing error at this point, so I can revert this.

src/actions/billing.js
```diff
@@ -1,40 +1,40 @@
...
+    sparkpostApiRequest({
       type: 'UPDATE_SUBSCRIPTION',
       meta: {
         method: 'PUT',
-        url: url,
-        data: { code }
+        url: isAws(getState()) ? '/account/aws-marketplace/subscription' : '/account/subscription',
+        data: { code },
+        ...meta,
+        onSuccess: meta.onSuccess ? meta.onSuccess : () => fetchAccount({ include: 'usage,billing' })
```
bkemper: Should this be a ternary or should we keep the current behavior and pass along `meta`?

`onSuccess: () => fetchAccount({ include: 'usage,billing', meta })`

gwalls: I think the ternary is fine to keep from fetching the account twice when updateSubscription happens to already be part of a chain that fetches that account at the end.

PR #385, opened 2018-03-30
FAD-6429 better examples for filter box

jasonrhodes:


src/components/collection/helpers/exampleSearch.js
```diff
@@ -0,0 +1,68 @@
...
+    // Return the key:value pair in string format
+    return `${modifier}:${value}`;
+  });
+
+  // Filter out falsey values
+  const results = converted.filter((pair) => pair);
+
+  // Don't guess at keys that are searchable or return an empty list, use a generic example instead
+  if (results.length === 0) {
+    return ['myKey:123dF9', 'myKey:"Some Value"', 'myKey:true'];
```
jonambas: When would these ever show?

It might be better not to show an example at all if the keys aren't real

jasonrhodes: If you didn't provide options for `exampleModifiers` or if the ones you picked were all invalid values. I prefer to show _something_ in those cases rather than nothing, and this at least will show you how to use the syntax in a generic way although providing real modifiers is always better.

jonambas:
Do you think you could fix up the bounce page filterbox?

```js
filterBox={{
        show: true,
        keyMap: {
          category: 'bounce_category_name',
          classification: 'bounce_class_name',
          reason: 'bounce_class_description' // this line can be removed
        },
        // replace bounce_class_description below with reason i think
        itemToStringKeys: ['bounce_category_name', 'bounce_class_name', 'domain', 'bounce_class_description'],
        exampleModifiers: ['domain', 'category']
}}
```

PR #388, opened 2018-03-30
FAD-6462: Suppress Sentry errors from Chrome and Firefox extensions

bkemper:
I used [Ferdinand](https://gist.github.com/bkemper/4c1e9a87f4b0b921711ea6791c265cc0) to test.

### DIFF
--- a/src/helpers/errorTracker.js
+++ b/src/helpers/errorTracker.js
@@ -74,6 +74,10 @@ class ErrorTracker {
     const options = {
       breadcrumbCallback,
       dataCallback: getEnricherOrDieTryin(store),
+      ignoreUrls: [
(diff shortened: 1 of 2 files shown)


jasonrhodes:
Any reason we should just block everything except our own hostname for errors? Is that too aggressive?

bkemper:
I can't confidently say that we should.  I think we should wait and address it on a case by case basis.  Ignoring extension errors is a no brainer because it is something that the user has installed at their own risk.  However, if one of the many things we load from tag manager causes a problem with our application, I think we should report those errors.

jasonrhodes:
That's fair. LGTM then.

PR #391, opened 2018-04-02
FAD-6438 Update hosted font files

jonambas:
There's not much to verify here until this is in production, aside from checking this hostname and path-to-fonts are correct

screen shot 2018-04-02 at 9 45 48 am

Until ^ that is verified, anything loading 2.4.3 matchbox will run from a "development" font project (bandwidth based instead of pageviews).

I duplicated the existing font projected instead of replacing its hostname - so prd is not affected.


jasonrhodes:
What changed in matchbox that's related to these font files?

jasonrhodes:
Oh I always forget we include fonts from inside matchbox ... we should probably change that.

jonambas:
@jasonrhodes 
this: 
```
@import url('//cloud.typography.com/6240112/7125792/css/fonts.css');
```

I can move this into this repo if you'd like. Could try moving it into a link tag at the bottom of body, but haven't experimented there with load times, etc

jasonrhodes:
>I can move this into this repo if you'd like. Could try moving it into a link tag at the bottom of body, but haven't experimented there with load times, etc

I like that idea but doesn't have to be conflated with this change. :+1:

PR #396, opened 2018-04-03
Feature/fad 6418

nancyhabecker:


src/components/reduxFormWrappers/SendingDomainTypeaheadWrapper.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import SendingDomainTypeahead from '../subaccountTypeahead/SendingDomainTypeahead';
+
+/**
+ * @example
+ * // This plugs into a redux-form Field like so:
+ * <Field
+ *   name="sendingDomain"
+ *   component={SendingDomainTypeahead}
+ *   results={sendingDomainsFromReduxStore}
```
nancyhabecker: can i do this? would this use some cached results from the store?

src/components/subaccountTypeahead/Typeahead.js
```diff
@@ -0,0 +1,92 @@
...
+    getItemProps,
+    getLabelProps,
+    highlightedIndex,
+    inputValue,
+    openMenu,
+    selectedItem,
+    clearSelection,
+    isOpen
+  }) => {
+    const { name, results, disabled, label = 'Subaccount', placeholder = (isOpen ? 'Type to search' : 'None'), error, helpText } = this.props;
```
nancyhabecker: "Subaccount" oops

src/pages/ipPools/EditPage.js
```diff
@@ -40,7 +40,7 @@ export class EditPage extends Component {
      * have been reassigned by the user.
      */
     const changedIpKeys = Object.keys(values).filter((key) =>
-      key !== 'name' && values[key] !== id);
+      key !== 'name' && key !== 'signing_domain' && values[key] !== id);
```
nancyhabecker: i need to do something about `signing_domain`, since it's now an object.

src/components/subaccountTypeahead/SendingDomainTypeahead.js
```diff
@@ -0,0 +1,29 @@
...
+    name: 'sendingDomain'
+  };
+
+  componentDidMount() {
+    this.props.listSendingDomains();
+  }
+}
+
+const mapStateToProps = (state) => {
+  const verifiedDomains = _.map(selectVerifiedDomains(state), (domain) => ({
```
jasonrhodes: This should move into a selector and use reselect

src/pages/ipPools/EditPage.js
jasonrhodes: Probably worth re-thinking how we're doing this

src/components/subaccountTypeahead/SendingDomainTypeahead.js
```diff
@@ -0,0 +1,29 @@
...
+export class SendingDomainTypeahead extends Typeahead {
+  static defaultProps = {
+    name: 'sendingDomain'
+  };
+
+  componentDidMount() {
+    this.props.listSendingDomains();
+  }
+}
+
```
jasonrhodes: I think you need a `render` method here to render the `<Typeahead>` component, right?

src/components/reduxFormWrappers/SendingDomainTypeaheadWrapper.js
jasonrhodes: You would connect to the store value but probably also retrieve them through a connected action in `componentDidMount` so that if they haven't been retrieved yet, they will be. The retrieval can be conditional on whether there are results in the store if you're ok with the slight chance of stale data there, but usually we just re-fetch in the background.

src/components/subaccountTypeahead/SubaccountTypeahead.js
```diff
@@ -1,112 +1,43 @@
...
 
     return (
-      <div className={cx('Typeahead')}>
-        <ActionList className={listClasses} actions={mappedItems} />
-        <TextField {...textFieldProps} onFocus={openMenu} />
-      </div>
+      <Typeahead
+        renderItem={(item) => <Item name={item.name} id={item.id} /> }
+        itemToString={(item) => (item ? `${item.name} (${item.id})` : '')}
+        {...this.props}
```
jasonrhodes: need to pass a `label` prop here of "Subaccount"

src/components/subaccountTypeahead/SendingDomainTypeahead.js
```diff
@@ -0,0 +1,39 @@
...
+  render() {
+    const { hasVerifiedDomains } = this.props;
+
+    if (!hasVerifiedDomains) {
+      return null;
+    }
+
+    return (
+      <Typeahead {...this.props} />
```
jasonrhodes: Should probably pass a `label` prop here, before the `{...this.props}` spread so it can be overwritten

jasonrhodes:
Let's rename the `subaccountTypeahead` directory to just `typeahead`

PR #397, opened 2018-04-03
Feature/fad 6629

bizob2828:


jasonrhodes:
I was going to suggest clearing the codes out on disable but I think it would be better to call the clear backup codes action on unmount of the modal component (and then we can remove it from being called when the modal is closed), so they never stick around past when they are supposed to.

As it is, I can get in a weird state by hitting my back button after I generate codes (I didn't close the modal so they aren't cleared, so if I try to generate new ones they are still in the store).

Should be a quick add if you can sneak it in here? Otherwise this is looking good.

jasonrhodes:
Hm I actually can't confirm this is working because the API is still returning codes after I disable/re-enable.

jasonrhodes:
Also @bizob2828 your commit message references FAD-6629 but the ticket is FAD-6229 :)

PR #398, opened 2018-04-03
FAD-6279 Gracefully handle and log app load errors

jonambas:
Sentry issue: https://sentry.io/sparkpost/phoenix-ui-local/issues/512705488/?query=is:unresolved

Error screen:
screen shot 2018-04-03 at 5 38 54 pm

Easiest way to test is to block the url `localhost:3100/static/js/bundle.js` in chrome dev tools

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -44,7 +44,7 @@
       }
 
       window.addEventListener('load', function() {
-        document.getElementById('critical').className += ' ready'; // Kills loading screen
+        document.getElementById('loading').className += ' ready'; // Kills loading screen
 
         if (!window.SPARKPOST_LOADED) {
           document.getElementById('error').className += ' show';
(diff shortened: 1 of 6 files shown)


public/index.html
```diff
@@ -54,10 +77,18 @@
...
     <div id="critical">
       <svg class='critical__icon' xmlns="http://www.w3.org/2000/svg" viewBox="0 0 127.5 260">
         <path class='critical__flame' d="M106.5 101.3c-13.4 10.1-16 28-16.4 40.1C68.8 116.8 129.3 44.6 60.6 0 103.1 55.1 0 109.3 0 190.4c0 31.8 19.9 59.9 63.5 69.6 42.8-9.1 64-37.8 64-69.6 0-47.4-29.5-63.3-21-89.1zM63.6 234c-23.3 0-42.2-18.9-42.2-42.2 0-23.3 18.9-42.2 42.2-42.2 23.3 0 42.2 18.9 42.2 42.2.1 23.3-18.8 42.2-42.2 42.2z"/>
       </svg>
     </div>
+
+    <!-- Error screen -->
+    <div id="error">
+      <h3>Oh no, something went wrong.</h3>
+      <p>Please try again soon.</p>
```
jasonrhodes: Maybe link to status.sparkpost.com, the Slack channel, our Twitter (?), and any other communication channels someone might be able to use? Not everyone can submit a support ticket I think, but we should check with CST about how to handle this situation...

public/index.html
```diff
@@ -54,10 +77,18 @@
...
+    <!-- Loading screen -->
     <div id="critical">
       <svg class='critical__icon' xmlns="http://www.w3.org/2000/svg" viewBox="0 0 127.5 260">
         <path class='critical__flame' d="M106.5 101.3c-13.4 10.1-16 28-16.4 40.1C68.8 116.8 129.3 44.6 60.6 0 103.1 55.1 0 109.3 0 190.4c0 31.8 19.9 59.9 63.5 69.6 42.8-9.1 64-37.8 64-69.6 0-47.4-29.5-63.3-21-89.1zM63.6 234c-23.3 0-42.2-18.9-42.2-42.2 0-23.3 18.9-42.2 42.2-42.2 23.3 0 42.2 18.9 42.2 42.2.1 23.3-18.8 42.2-42.2 42.2z"/>
       </svg>
     </div>
+
+    <!-- Error screen -->
+    <div id="error">
+      <h3>Oh no, something went wrong.</h3>
```
jasonrhodes: Oh no, SparkPost is having trouble loading.

(Can we put the logo or the flame in the background too to help them know they are in the right place, maybe?)

public/index.html
ewandennis: Can we also include a call to action?

> [Try again](https://app.sparkpost.com/) or check our [status page](https://status.sparkpost.com/)

public/static/tenant-config/production.js
```diff
@@ -15,6 +15,10 @@ window.SP.productionConfig = {
   },
   gaTag: 'UA-111136819-2',
   gtmId: 'GTM-P87NNJ4',
+  sentry: {
+    projectId: 232588,
+    publicKey: '63149b1565df4b7199db939a4410b47a'
```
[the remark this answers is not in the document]
jonambas: They are, should I remove?

public/index.html
jonambas: Don't think twitter is appropriate here, and not sure about slack 🤔 .

public/static/tenant-config/production.js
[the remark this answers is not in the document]
jasonrhodes: Yeah remove from default.js please

public/index.html
jasonrhodes: The problem is this experience will be frustrating for a user and it'd be nice to give them options for reaching out somehow. If they don't have support access (we don't know if they do), they can visit the status page but if we don't know about the problem, it'd be good to suggest a way for them to get in touch with someone somehow... open to other ideas

jasonrhodes: But at the very least the status page will be good :)

jonambas:
screen shot 2018-04-04 at 12 50 50 pm

rajumsys:
Another way to simulate error loading after waiting sometime is to:
- block the bundle URL as @jonambas mentioned
- Include this in index.html header section. 

``` 
<script src="https://httpstat.us/500?sleep=10000" type="text/javascript"></script>
```
So, it'll show the error after about 10 seconds.

PR #399, opened 2018-04-04
FAD-6466: Reinstate cookie consent timestamp storage

ewandennis:
In the related accusers PR, we're giving Heroku and Azure folks the `users/self-manage` grant. The knock-on effect is that the UI will now show those accounts `/accounts/profile`. That is useful for support since ProfilePage shows the account ID. However, the page needs some love to avoid offering capabilities these SSO accounts shouldn't have.

- Make ProfilePage hide some stuff from those accounts using AccessControl
- Ensure we attempt to set the cookie consent flag only once to avoid a loop on error

### Testing
 - Run a local accusers from the hotfix/FAD-6466 branch
 - Fake up a heroku user login (? TBD)
 - Verify that profile page is visible in diminished form

### DIFF
--- a/src/pages/profile/ProfilePage.js
+++ b/src/pages/profile/ProfilePage.js
@@ -7,6 +7,7 @@ import { updateUser } from 'src/actions/users';
 import { get as getCurrentUser } from 'src/actions/currentUser';
 import { confirmPassword } from 'src/actions/auth';
 import { showAlert } from 'src/actions/globalAlert';
+import { isSSOAccountSelector } from 'src/selectors/accountBillingInfo';
 
 import NameForm from './components/NameForm';
 import PasswordForm from './components/PasswordForm';
@@ -40,7 +41,8 @@ export class ProfilePage extends Component {
     const {
       username,
       email,
-      customer
+      customer,
+      isSSOAccount
     } = this.props.currentUser;
 
     return (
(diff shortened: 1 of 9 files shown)


src/pages/profile/ProfilePage.js
```diff
@@ -51,23 +53,28 @@ export class ProfilePage extends Component {
           <LabelledValue label='Email Address' value={email}/>
         </Panel>
 
-        <TfaManager />
+        { !isSSOAccount && <TfaManager /> }
```
jasonrhodes: (Note: I can probably take this review but will let everyone know later this morning if support work gets in the way.)

I think we should use `<AccessControl>` + condition helpers for these.
```js
import { all } from 'src/helpers/conditions/compose';
import not from 'src/helpers/conditions/not';
import { isHeroku, isAzure } from 'src/helpers/conditions/user';

// ...

<AccessControl condition={all(not(isHeroku), not(isAzure))}>
  // all the components that should be hidden, or each one wrapped individually, either way
</AccessControl>
```
I know that stuff isn't super well-defined or understood yet but anytime we reach for account or user conditions, those work well and create a centralized place for that logic that can be shared between AccessControl, ConditionSwitch, and routes.

ewandennis: Actually pretty well-defined in docs/. I just wasn't aware of it. Updated.

Not sure how we test components using `AccessControl`. For instance, [these 2 tests](https://github.com/SparkPost/2web2ui/blob/master/src/components/navigation/tests/Footer.test.js#L19) both produce the same snapshot.

The thing we want to test is presumably just the condition so we could break that out. Will give that a go.

src/pages/profile/tests/ProfilePage.test.js
```diff
@@ -32,6 +35,15 @@ describe('ProfilePage', () => {
...
 
+  cases('handles the various account types', ({ access_level, result }) => {
+    const condition = wrapper.find(AccessControl).first().prop('condition');
+    expect(condition({ currentUser: { access_level }, ready: true })).toEqual(result);
+  }, {
+    'user': { access_level: 'user', result: true },
+    'heroku': { access_level: 'heroku', result: false },
+    'azure': { access_level: 'azure', result: false }
+  });
+
```
jasonrhodes: oh this is nice :+1:

PR #401, opened 2018-04-04
FAD-6475 Message events reports filter does not handle + in email addresses properly

jonambas:
To verify:
- Go to message events list
- Add an email to recipients filter with a `+`, like `1+2@test.com`
- Refresh
- Check that the active filter and request does not replace the + with a space between 1 and 2


rajumsys:
It's working correctly but in the Filters list, it's showing without `+` sign like `Recipient: me me@gmail.com` instead of `Recipient: me+me@gmail.com`

PR #402, opened 2018-04-04
FAD-6455: Improve UX for message events details page when message_id is undefined

bkemper:
Refer to [FAD-6455](https://jira.int.messagesystems.com/browse/FAD-6455)

### DIFF
--- a/src/actions/messageEvents.js
+++ b/src/actions/messageEvents.js
@@ -3,7 +3,7 @@ import config from 'src/config';
 import _ from 'lodash';
 import sparkpostApiRequest from 'src/actions/helpers/sparkpostApiRequest';
 
-const { apiDateFormat, messageEvents } = config;
+const { apiDateFormat } = config;
 
 export function getMessageEvents(options = {}) {
   const { dateOptions, ...rest } = options;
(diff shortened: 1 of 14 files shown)


bkemper:
Lets hold off, see how the new component functions in production and I'll file a follow-up tech debt ticket.

src/actions/messageEvents.js
```diff
@@ -82,17 +82,19 @@ export function removeFilter(filter) {
...
       params: {
-        to: moment().utc().format(apiDateFormat), // it's above ...rest so it can be overriden when needed
-        ...rest,
         message_ids: messageId,
-        from: moment().subtract(messageEvents.retentionPeriodDays, 'days').utc().format(apiDateFormat)
+        // Must pass a time range because the defaults are too narrow (now to 24 hours ago) and
+        // must cast a wide time range (even wider than the standard 10 day retention) to avoid
+        // missing message events
+        to: moment.utc().add(1, 'minute').format(apiDateFormat),
+        from: moment.utc().subtract(1, 'month').format(apiDateFormat)
```
gwalls: Can we keep the retentionPeriodDays and just add a buffer to that config value. So (retentionPeriodDays + 10) or whatever. My only fear is that that time increases for one customer or another and no one thinks to looks here to manually increase the 'from' time, but updates the tenant's value in ansible.

PR #404, opened 2018-04-05
FAD-6451 Update bounce chart to reflect using 'sent' as denominator

jonambas:
Turns out the percentages in the table are: item bounce / total bounce, so no change was needed there.


src/actions/bounceReport.js
```diff
@@ -11,7 +11,7 @@ export function refreshBounceReport(updates = {}) {
         type: 'GET_BOUNCE_REPORT_AGGREGATES',
         params: {
           ...params,
-          metrics: 'count_targeted,count_bounce,count_inband_bounce,count_outofband_bounce'
+          metrics: 'count_sent,count_bounce,count_inband_bounce,count_outofband_bounce'
```
rajumsys: we'll still need `count_targeted` for metrics like admin bounce

jonambas: We don't show rate though for admin bounce, just the count

src/pages/reports/bounce/BouncePage.js
```diff
@@ -87,8 +88,13 @@ export class BouncePage extends Component {
...
 
   renderTopLevelMetrics() {
-    const { chartLoading, aggregates } = this.props;
-    const { countBounce, countTargeted } = aggregates;
+    const { chartLoading, aggregates, adminBounces } = this.props;
+    const { countBounce, countSent } = aggregates;
+
+    // TODO Add support doc link - <UnstyledLink to={LINKS.ADMIN_BOUNCE} external>Learn more</UnstyledLink>.
+    const adminBounceText = adminBounces
+      ? <Fragment>{maybePluralize(adminBounces, 'message')} were categorized as Admin Bounces.</Fragment>
```
jonambas: err this wont work `1 message were categorized...`

src/pages/reports/components/MetricsSummary.js
```diff
@@ -50,7 +50,7 @@ MetricsSummary.propTypes = {
   relativeRange: PropTypes.string,
   rateValue: PropTypes.number,
   rateTitle: PropTypes.string,
-  secondaryMessage: PropTypes.string
+  secondaryMessage: PropTypes.node
```
bkemper: Should we allow either string or node?

src/selectors/bounceReport.js
```diff
@@ -22,10 +24,23 @@ const selectFormattedAggregates = createSelector(
...
 );
 
-const selectReshapedClassifications = ({ bounceReport }) => {
-  const { classifications = []} = bounceReport;
-  return reshapeCategories(classifications);
-};
+const selectReshapedClassifications = createSelector(
+  [selectClassifications],
+  (classifications) => reshapeCategories(classifications)
+);
```
bkemper: 👍

src/selectors/bounceReport.js
```diff
@@ -22,10 +24,23 @@ const selectFormattedAggregates = createSelector(
...
+
+const selectCategories = createSelector(
+  [selectReshapedClassifications],
+  (classifications) => _.filter(classifications, ({ name }) => name !== 'Admin')
+);
+
+const selectAdminBounces = createSelector(
+  [selectReshapedClassifications],
+  (classifications) => {
+    const adminBounces = _.filter(classifications, ({ name }) => name === 'Admin')[0];
```
bkemper: Should we add this category to `/constants`?

bkemper: Also, there might be multiple classifications with 'Admin' category.  Should filter then sum?

bkemper: Or, should we get the value directly from the api with count_admin_bounce metric?

src/pages/reports/components/MetricsSummary.js
jonambas: I believe `node` allows for strings, numbers, and react components - will double check

bkemper: You are right.

```
// Anything that can be rendered: numbers, strings, elements or an array
// (or fragment) containing these types.
optionalNode: PropTypes.node,
```
_https://reactjs.org/docs/typechecking-with-proptypes.html#proptypes_

src/selectors/bounceReport.js
jonambas: we can get it from aggregates, but the sum should already added up at this point from `reshapeCategories`

bkemper: I didn't give you the proper context for why I suggested to use the count_admin_bounce metric. 
 You are right it should.  I was getting ahead of myself with some of the possible changes to the metrics-api.

Let me know what you think about my other two questions.

jmartin4563:
In discussions with Daeyon, count_bounce is going to be including admin bounces. This means that the webui cannot use this metrics when displaying "total bounces" in the webui. Instead, the bounce report will need to use count_inband_bounce + count_out_of_band_bounce so we do not include admin bounces in our rate calculations. We will be making an additional api change to change count_inband_bounce to NOT include admin bounces, thus making count_inband_bounce + count_out_of_band_bounce the actual number of ISP related bounces, which when divided by and updated sent metric, will truly represent your "actual" bounce rate wrt ISP interactions

PR #407, opened 2018-04-09
FAD-6481 Fixed test webhook batch

gwalls:


src/pages/webhooks/components/TestTab.js
```diff
@@ -13,12 +11,24 @@ import RequestBlock from './RequestBlock';
 export class TestTab extends Component {
   state = {
     testSent: false,
-    buildRequest: _.once(this.buildTestRequest)
+    sampleGenerated: null
```
rajumsys: as discussed may be we should rename it to `generatedSample` or something like that as this sounds like a boolean.

PR #413, opened 2018-04-11
FAD-6414: Sanitize HTTP error messages

bkemper:
_Refer to [FAD-6414](https://jira.int.messagesystems.com/browse/FAD-6414)_

**Testing**

1. Navigate to change plan page
1. Select any paid plan
1. Fill-in all billing info fields as normal except for credit card enter an invalid card number, "123"
1. Click "Upgrade Plan" button

You should get an error message.

Was:
> Transaction declined.<br>CREDIT_CARD_NUMBER_H - Credit card type is not accepted by this merchant account.\nCredit card number must be 12-19 digits.

Now: 
> Transaction declined. CREDIT_CARD_NUMBER_H - Credit card type is not accepted by this merchant account. Credit card number must be 12-19 digits.

**Note:**  I looked at a number of solutions and packages.  The `createElement` solution was cool, but I feel it might leave us exposed to edge cases.  The node-html-to-text, striptags, and string-strip-html packages would work, but for different reasons are too much for what we need.


jasonrhodes:
>The createElement solution was cool, but I feel it might leave us exposed to edge cases.

Which edge cases are you worried about?

jasonrhodes:
Yeah, my understanding is that using regex to parse HTML is the ̷r̵o̵o̸t̸ ̶ŏ̸̜f̵͕̓ ̵͓͝a̶̡̨̝̪̣̤͋́̉͒̔́͂͑̈́̅̈͝͝l̷̢̨͎̼̥̫̲̭̗̪̙͓͆́̕̚̚ͅͅļ̵̢̡̼͙̘̣͍̞̝̯̮͈͎̫͎͍̦͇̀̀̔̏͝ ̵̙͉̻̞̗̰̭̪̮̠̙̇͑̒͌̑͘͠z̶̡̩̜̩̟̗̝̜̫̲͔͎͇̙̫̝̮̱̰̪̲̎̒̎́̈́̏͛̐̃̈́̕ͅa̴̡̙̺̱͇̎̾̉͊̂̓̇͜͝͠l̷̻̃͐̽̍͒͒͂̉́͊̋͌́̑͘͝g̵̨̢̡̬̤̖͔̯̗͚͈̝̮̲̜̼͂̄̀̏͛̀̐̍̈́̈́͆͂̂̉̂̆͋̂͒͜͜͝͠͝ȏ̵̧̖̮̖̭̬͙̙̘̼̐̽̾͜͜

jasonrhodes:
I like the other stuff you added in there though for new lines and spaces...

bkemper:
@jasonrhodes Fixed.

jasonrhodes:
Something else I just realized: in grooming we talked about doing this transformation for _all_ error messages and not just Zuora ones. How difficult would it be to move this helper into an interceptor that is applied at the axios instance level for each? Then we can build on that for other message transformations we want to do later.

jasonrhodes:
Alternatively we can log a ticket (if it doesn't already exist) to implement error message transformations in the interceptor level and note that we want to move this check there when we do that work.

bkemper:
Interceptors would not be hard to implement -- I played around with it before when creating `SparkpostApiError`.  I'm just not sold that it is something we need.  This is a very specific case.  I hope we don't experience it again.  If and when we do, then I think we should reconsider interceptors.

jasonrhodes:
So a few separate things here:
1. IMO we will definitely need interceptors so we can transform error messages on a wider scale, particularly ones from our own APIs that return status codes that could be used to create better messages. We can also eventually improve the Zuora error messages this way.
1. Once those exist, we should move all error message transformation into those to keep it centralized and make it invisible to the API call process.
1. We probably won't really know if this comes up again in another API unless a user happens to report/screenshot it. I agree it's probably a Zuora-only thing, and that's definitely the case we're trying to solve right now, but it probably doesn't hurt to just strip HTML and normalize new lines/spaces across the board.

Let's do the 2nd option and include this all in an "error interceptor" ticket. I'll file and reference this one.

PR #419, opened 2018-04-13
FAD-6504 Filtered rows should now respect sort column

gwalls:


src/components/collection/Collection.js
```diff
@@ -57,14 +57,18 @@ export class Collection extends Component {
...
-      update.filteredRows = objectSortMatch({
+      const filteredRows = objectSortMatch({
         items: rows,
         pattern,
         getter: objectValuesToString(itemToStringKeys),
         keyMap
       });
+
+      // ultimately respect the sort column, if present
+      update.filteredRows = sortColumn ? _.orderBy(filteredRows, sortColumn, sortDirection || 'asc') : filteredRows ;
```
gwalls: Alternatively, I could do this logic somewhere in `sortMatch.js` (possibly `filterAndSortByScore`?), but it would require passing sortColumn and sortDirection to those functions.

gwalls:
The issue here was in our order of operations. We were sorting the rows according to the column selection, filtering those rows, and then sorting according to relevance to the search term (https://github.com/SparkPost/2web2ui/blob/4f9c9feff3ed2d41d8adc2d5b5b08277325ceada/src/helpers/sortMatchScorers.js).

With this fix, if a sort column is provided as well as a search filter, then rows will be sorted primarily according to the column and secondarily according to search relevance.

jasonrhodes:
So ... because the filter box is trying to sort based on the best match, and including things that are very loosely related to the search string, re-sorting after the results are returned is going to produce some really weird results where a very loosely related match will be at the top based on the dumb column sort.

Suggestion:
* When calling `objectSortMatch`, pass a `lowestMatchScore` or `matchThreshold` or whatever you want to call it, and use that here instead of hard-coded 0 https://github.com/SparkPost/2web2ui/blob/master/src/helpers/sortMatch.js#L9
* Can also pass in a "sorter" method to `objectSortMatch` that defaults to sorting by score like it already does. Then you could pass a function that has access to the sorting things you need without having to pass those down for cases where you want to do a custom sort.

Not sure what the match threshold score should be tbh, but probably 5 or above is a good place to start? https://github.com/SparkPost/2web2ui/blob/master/src/helpers/sortMatchScorers.js#L52

ewandennis:
> Can also pass in a "sorter" method to objectSortMatch that defaults to sorting by score like it already does. Then you could pass a function that has access to the sorting things you need without having to pass those down for cases where you want to do a custom sort.

`sortMatch.js` manages to keep it's scoring, filtering and sorting concerns mostly to itself. Adding into that the ability to do some arbitrary post-hoc sort adds complexity and leaks client concerns into `sortMatch.js`. I think we should leave `sortMatch` alone and let client code do whatever use case-centric sorting it needs on its own.

In this case, that means `Collection` is responsible for doing final column sorting on filtered results. Will push a commit to that effect and we can haggle if required.

jasonrhodes:
I think eventually we will want/need the ability to override the default sorting that sortMatch is doing. Whether that means we allow the sort-by-score default to be overridden by passing in a custom sorter that is given access to each item's `score` value, or we stop sorting altogether and return a list of items that have each been given a `._score` property or similar, the next evolution of `FilterBox` is going to require being able to sort on multiple dimensions that will include score.

src/components/collection/Collection.js
```diff
@@ -48,21 +48,29 @@ export class Collection extends Component {
...
 
     if (pattern) {
-      update.filteredRows = objectSortMatch({
+      const filteredRows = objectSortMatch({
         items: rows,
         pattern,
         getter: objectValuesToString(itemToStringKeys),
-        keyMap
+        keyMap,
+        matchThreshold
```
jasonrhodes: My thought about this was that we would dynamically make a distinction between "the default sort we dumbly present data in" vs. "the user specifically requested a sort dimension", and only in the latter case we would need to up the threshold significantly. That way, when a user hasn't specifically requested any sort, they still get the full benefit of scored matches.

PR #422, opened 2018-04-16
FE-61: Accept 24-hour time in date-time inputs

bkemper:
_Refer to [FE-61](https://jira.int.messagesystems.com/browse/FE-61)_


config/jest/matchers/index.js
```diff
@@ -0,0 +1,5 @@
+import toBeValid from './toBeValid';
+
+export default {
+  toBeValid
```
jonambas: Does this work here?
```js
export { default as toBeValid } from './toBeValid';
```

src/constants/index.js
```diff
@@ -5,11 +5,25 @@ export const COOKIE_DOMAIN = '.sparkpost.com';
 
 export const DATE_FORMATS = {
   READABLE_DATE_YEAR_TIME: 'MMM Do YYYY h:mma',
-  READABLE_DATE_TIME: 'MMM Do h:mma',
-  INPUT_DATE: 'YYYY-MM-DD',
-  INPUT_TIME: 'h:mma'
+  READABLE_DATE_TIME: 'MMM Do h:mma'
 };
 
+export const DEFAULT_DATE_FORMAT = 'YYYY-MM-DD';
```
jonambas: Wondering if these should be moved to a new file `constants/datepicker.js`, and maybe prefix with something.
We have `dateFormat` `timeFormat` in config/default, just worried it will be a little confusing.

src/helpers/date.js
```diff
@@ -117,3 +125,10 @@ export function formatTime(time) {
 export function formatDateTime(datetime) {
   return `${formatDate(datetime)}, ${formatTime(datetime)}`;
 }
+
+export const formatFormDate = (date) => moment(date).format(DEFAULT_DATE_FORMAT);
+export const formatFormTime = (time) => moment(time).format(DEFAULT_TIME_FORMAT);
+export const formatFormDateTime = (...args) => moment(args.join(' ')).format(DEFAULT_DATE_TIME_FORMAT);
+export const parseDate = (str) => moment(str, ACCEPTED_DATE_FORMATS, true);
+export const parseTime = (str) => moment(str, ACCEPTED_TIME_FORMATS, true);
+export const parseDateTime = (...args) => moment(args.join(' '), ACCEPTED_DATE_TIME_FORMATS, true);
```
jonambas: Same thought here, maybe `helpers/datepicker.js`?

src/components/datePicker/ManualEntryForm.js
```diff
@@ -23,15 +21,15 @@ export default class ManualEntryForm extends Component {
...
+      toTime: formatFormTime(to),
+      fromDate: formatFormDate(from),
+      fromTime: formatFormTime(from)
     });
   }
 
-  handleFieldChange = (e, key) => {
-    this.setState({ [key]: e.target.value });
+  handleFieldChange = (e) => {
+    this.setState({ [e.target.id]: e.target.value });
```
jonambas: 👍

src/components/datePicker/ManualEntryForm.js
```diff
@@ -76,17 +73,17 @@ export default class ManualEntryForm extends Component {
...
               value={fromDate} />
           </Grid.Column>
           <Grid.Column >
             <TextField
-              id="fieldFromTime"
-              label='From Time' labelHidden placeholder='12:00am'
-              onChange={(e) => this.handleFieldChange(e, 'fromTime')}
+              id="fromTime"
+              label='From Time' labelHidden placeholder='HH:MMam'
+              onChange={this.handleFieldChange}
```
jonambas: We discussed leaving out `onChange` on the time fields - you think this works well?

config/jest/matchers/index.js
bkemper: That would work.  It would name export the custom matches and I would need to change how I'm importing in the `jest/setup.js`.

src/constants/index.js
bkemper: I think you are right.  I'm going to namespace the format variables with `FORMATS`.  As for config defaults, we have a follow-up audit, [FE-65](https://jira.int.messagesystems.com/browse/FE-65), that I expect with address the consolidation.

src/components/datePicker/ManualEntryForm.js
bkemper: I completely forgot that the onChange needs to be in place to setState for the input value.

PR #424, opened 2018-04-17
FE-135: Reinstate gtm.js tag manager event

ewandennis:
This is a minimal patch to our Google Tag Manager integration to emit a `gtm.js` event. This event produces the so-called "Page View" event in Tag Manager and our marketing folks have various tracking actions triggered by that it.

[Some background material](https://www.simoahava.com/analytics/trigger-guide-google-tag-manager/)

### Testing

All GTM events are pushed onto a global named `dataLayer` which you can review in the console.
1. Log in to the app
1. Verify that the `gtm.js` event is listed on the `dataLayer` variable.

You can also do this in production to verify that the event is missing.


src/helpers/analytics.js
```diff
@@ -0,0 +1,48 @@
+// Google Tag Manager tracking helpers
```
bkemper: Is this file needed?  I don't see it being used.

src/components/googleTagManager/GoogleTagManager.js
```diff
@@ -11,7 +11,10 @@ export class GoogleTagManager extends Component {
   }
 
   componentDidMount() {
-    window.dataLayer = [];
+    window.dataLayer = [{
+      'gtm.start': new Date().getTime(),
+      event: 'gtm.js'
+    }];
```
bkemper: 👍

src/components/googleTagManager/tests/GoogleTagManager.test.js
```diff
@@ -49,16 +59,14 @@ describe('Component: GoogleTagManager', () => {
...
-      }
-    ]);
+    expect(window.dataLayer).toHaveLength(2);
+    expect(window.dataLayer.pop()).toEqual({
+      event: 'content-view',
+      'content-name': '/some/path?fun=times',
+      'content-title': 'Test Public Title',
+      username: undefined
+    });
     wrapper.setProps({ username: 'loaded' });
```
bkemper: Is this still needed now that the test below was removed?

PR #426, opened 2018-04-18
FE-291 API Key details view read only

rajumsys:
This renders api key details page in read only view. 

Flow:
- (Unlike current behavior) Allows clicks to all links (even api key not owned by current user) 
- The detail page picks up the api key id (and subaccount) from path/query string 
- if the currentUser doesn't match apikey owner, it renders the page in ready only mode
 - all fields read only
 - hides delete button/modal
 - hides submit button

Test:
- create an api key as "normal user"
- create a superuser and login as superuser and go to the api key created in prev step. you should see the form in readonly mode
- rest of the features should work as is. 

Todos

- [x] Tests

Note: It seemed lot of work for a simple problem (to see grants). We could show it in list like legacy UI. To de-clutter ui, we could show it inline or in modal on a button click.


src/components/grantBoxes/GrantsCheckboxes.js
```diff
@@ -18,6 +18,7 @@ const GrantsCheckboxes = ({ grants, show }) => {
           label={grant.label}
           component={CheckboxWrapper}
           type="checkbox"
+          disabled={disabled}
```
rajumsys: this could be done by adding the prop in each item of `grants` but we don't have a use case where we would want to disable certain grant. so i added it globally.

src/selectors/api-keys.js
```diff
@@ -1,13 +1,13 @@
 import _ from 'lodash';
```
rajumsys: tests missing for a couple of paths. will fix that along with other review feedbacks

ewandennis:
Your point about doing something lighter weight on the list page makes some sense. However, we know that grants lists are large so inline display would still bloat the list quite a bit and we already have a link to details per key. A second button per item would likely be confusing.

We _could_ try putting a tooltip in place on each item with grants listed under it. It's your decision whether you mock up an inline implementation for the team to see, or to do with this impl.

On the level of effort, from our discussion I actually thought we were going to create a new read-only page which used the `ApiKeyForm`. That would simplify the redux state extraction, rendering and tests as well as not making `DetailsPage` do 2 jobs.

This implementation looks good though. A few visual notes on the read-only view:
 - The banner is large: lets put the text all on 1 line: `This API key is only editable by the owner: {apiKey.username}`
 - We don't need to see the all/select radio buttons, just the checkboxes
 - The allowed IP placeholder text looks like real content. Maybe best to remove it.

On to the code (even if you choose not to use it)!

src/pages/api-keys/DetailsPage.js
```diff
@@ -1,13 +1,13 @@
...
-import { Page, Panel } from '@sparkpost/matchbox';
+import { Page, Panel, Banner } from '@sparkpost/matchbox';
 
-import { deleteApiKey, getApiKey, updateApiKey, listGrants, listSubaccountGrants } from 'src/actions/api-keys';
+import { listApiKeys, deleteApiKey, getApiKey, updateApiKey, listGrants, listSubaccountGrants } from 'src/actions/api-keys';
 import { showAlert } from 'src/actions/globalAlert';
 
 import { hasSubaccounts } from 'src/selectors/subaccounts';
-import { getFormLoading, selectApiKeyId } from 'src/selectors/api-keys';
+import { getFormLoading, selectApiKeyId, getCurrentAPIKey, selectKeysForAccount, isFormReadyOnly } from 'src/selectors/api-keys';
```
ewandennis: `s/isFormReadyOnly/isFormReadOnly/g`

src/pages/api-keys/DetailsPage.js
```diff
@@ -60,8 +62,19 @@ export class ApiKeysDetailsPage extends Component {
     }));
   };
 
+  renderReadOnlyAlert() {
+    const { apiKey } = this.props;
+
+    return (<Banner
+      status='info'
+    >
+      <p>API keys are only editable by their owner.</p>
```
ewandennis: Lets merge these 2 paras into something smaller:

> This API key is only editable by its owner: {apiKey.username}

src/pages/api-keys/DetailsPage.js
```diff
@@ -71,21 +84,25 @@ export class ApiKeysDetailsPage extends Component {
       return <Redirect to='/account/api-keys' />;
     }
 
+    const secondarActions = isReadOnly ? [] : [{ content: 'Delete', onClick: this.onToggleDelete }];
```
ewandennis: `s/secondarActions/secondaryActions/g`

src/pages/api-keys/DetailsPage.js
```diff
@@ -71,21 +84,25 @@ export class ApiKeysDetailsPage extends Component {
...
         breadcrumbAction={breadcrumbAction}
-        secondaryActions={[{ content: 'Delete', onClick: this.onToggleDelete }]}>
+        secondaryActions={secondarActions}>
+        {isReadOnly && this.renderReadOnlyAlert()}
         <Panel>
-          <ApiKeyForm apiKey={apiKey} onSubmit={this.onSubmit} />
+          <ApiKeyForm apiKey={apiKey} onSubmit={this.onSubmit} isReadOnly={isReadOnly} />
         </Panel>
-        <DeleteModal
+        {!isReadOnly && <DeleteModal
```
ewandennis: Do we need this condition on the `DeleteModal`? It's not visible unless you click `Delete` which is also invisible. 

One less condition makes the next reader's eyes happier :)

src/selectors/api-keys.js
```diff
@@ -1,13 +1,13 @@
...
-import { getSubaccountIdFromProps } from './subaccounts';
+import { getSubaccountIdFromProps, selectSubaccountIdFromQuery } from './subaccounts';
 /*
  * generic apiKeys selectors
  */
 const getApiKeys = (state) => state.apiKeys.keys;
 const getGrantsArray = (state) => state.apiKeys.grants;
 const getSubaccountGrantsArray = (state) => state.apiKeys.subaccountGrants;
-export const selectApiKeyId = (props) => props.match.params.id;
+export const selectApiKeyId = (state, props) => props.match.params.id;
```
ewandennis: Not using `state` param

src/selectors/api-keys.js
```diff
@@ -27,24 +27,23 @@ export const getFormLoading = createSelector(
     grantsLoading || subaccountGrantsLoading
 );
 
-/*
- * ApiKeyForm selectors
- */
-const getFormApiKey = (state, props) => props.apiKey || {};
+export const getCurrentAPIKey = createSelector(
```
ewandennis: Since we replaced `getFormApiKey` with `getCurrentAPIKey`, lets include the default `|| {}` case and remove it from the 2 uses below.

Let's also follow the naming convention: `s/getCurrentAPIKey/getCurrentApiKey/g`

src/selectors/api-keys.js
```diff
@@ -27,24 +27,23 @@ export const getFormLoading = createSelector(
...
-    isNew || _.size(grants) <= _.size(apiKey.grants) ? 'all' : 'select'
+  [getGrants, getCurrentAPIKey, getIsNew],
+  (grants, apiKey, isNew) => isNew || _.size(grants) <= _.size(apiKey.grants) ? 'all' : 'select'
 );
 
 export const getInitialValues = createSelector(
-  [getGrants, getFormApiKey],
-  (grantsList, apiKey) => {
+  [getGrants, getCurrentAPIKey],
+  (grantsList, apiKey = {}) => {
```
ewandennis: When `getCurrentAPIKey` returns a default empty object, we can remove this default value.

src/selectors/api-keys.js
```diff
@@ -87,7 +86,27 @@ export const selectApiKeysForSmtp = createSelector(
   (apiKeys) => apiKeys.filter((key) => key.grants.includes('smtp/inject'))
 );
 
+export const isKeyOwnedByCurrentUser = (key, currentUsername) => Boolean((key.username === currentUsername) || (!key.username && key.subaccount_id));
```
ewandennis: Nice little refactor. Do we know what the 2nd clause is for? `!key.username && key.subaccount_id` doesn't sound like "owned by me".

src/selectors/api-keys.js
```diff
@@ -87,7 +86,27 @@ export const selectApiKeysForSmtp = createSelector(
...
+
+export const isFormReadyOnly = createSelector(
+  [getCurrentUsername, getCurrentAPIKey, getIsNew],
+  (currentUsername, currentKey, isNew) => {
+
+    if (isNew) {
+      return false;
+    }
+
+    if (isKeyOwnedByCurrentUser(currentKey || {}, currentUsername)) {
```
ewandennis: We can remove this default value when `getCurrentAPIKey` returns a default empty object.

src/selectors/api-keys.js
```diff
@@ -87,7 +86,27 @@ export const selectApiKeysForSmtp = createSelector(
...
 export const selectKeysForAccount = createSelector(
   [getApiKeys, getCurrentUsername],
-  (keys, currentUsername) => keys.map((key) => ({ ...key, isOwnedByCurrentUser: Boolean((key.username === currentUsername) || (!key.username && key.subaccount_id)) }))
+  (keys, currentUsername) => keys.map((key) => ({
+    ...key,
+    isOwnedByCurrentUser: isKeyOwnedByCurrentUser(key, currentUsername)
+  }))
 );
+
+export const isFormReadyOnly = createSelector(
```
ewandennis: `s/isFormReadyOnly/isFormReadOnly/g`

src/pages/api-keys/components/ApiKeyForm.js
```diff
@@ -46,8 +42,17 @@ export class ApiKeyForm extends Component {
     });
   }
 
+  getGrandOptions() {
```
ewandennis: `s/getGrandOptions/getGrantOptions/g`

src/pages/api-keys/components/ApiKeyForm.js
```diff
@@ -71,23 +77,26 @@ export class ApiKeyForm extends Component {
...
+            options={this.getGrandOptions()}
           />
-          <GrantsCheckboxes grants={this.availableGrants} show={showGrants} />
+          <GrantsCheckboxes grants={this.availableGrants} show={showGrants} disabled={isReadOnly}/>
           <Field
             name='validIps'
             component={TextFieldWrapper}
             label='Allowed IPs'
             helpText='Leaving the field blank will allow access by valid API keys from any IP address.'
             placeholder='10.20.30.40, 10.20.30.0/24'
```
ewandennis: Lets hide the placeholder text in read-only mode

src/selectors/api-keys.js
ewandennis: 🎉

src/components/grantBoxes/GrantsCheckboxes.js
ewandennis: Agreed

src/selectors/tests/api-keys.test.js
```diff
@@ -144,4 +184,78 @@ describe('ApiKey Selectors', () => {
...
+          }
+        },
+        location: {
+          search: '?subaccount=123'
+        },
+        apiKey: { id: 'key1' }
+      };
+    });
+
+    it('returns true when form is not new and api kew not owned by current user ', () => {
```
ewandennis: `s/kew/key/`

src/pages/api-keys/components/ApiKeyForm.js
```diff
@@ -58,6 +63,7 @@ export class ApiKeyForm extends Component {
             component={TextFieldWrapper}
             validate={required}
             label='API Key Name'
+            disabled={isReadOnly}
```
ewandennis: Lets also fix the helpText typo on line 70: `s/assigment/assignment/`

src/pages/api-keys/DetailsPage.js
rajumsys: it seems harmless when look at the code but seems bothering me when looked at snapshot.

src/selectors/api-keys.js
rajumsys: yes, it doesn't but changed the signature to be able to use it in other selector. is there an way to do without putting `state` as first arg?

src/pages/api-keys/DetailsPage.js
rajumsys: :facepalm:

src/pages/api-keys/components/ApiKeyForm.js
rajumsys: good point!

src/selectors/api-keys.js
rajumsys: yeah probably that's better as can easily add tests for that and not in every other references

src/selectors/api-keys.js
rajumsys: yeah, it confused me too. so basically, it's like `can I edit this api key`? but i named after the property it sets. i'll see if i can easily change the name.

src/selectors/api-keys.js
rajumsys: :headdesk:

ewandennis:
When editing a subaccount key, the form shows in create mode. Looks like the `getIsNew` selector relies on `getCurrentApiKey` which has a small bug. `key.subaccount_id` is a Number but the `subaccountId` URL param is a string so `===` will fail.

When navigating direct to an edit page, the page should list API keys on `cDM` since it uses `getCurrentApiKey` which uses the list.

Lets hide `grantsRadio` in read-only mode.

I found this read-only alert looked more in-keeping with our other banners:
```
<Banner
      status='info'
      title='This key is read-only'
    >
      This API key is only editable by the owner: {apiKey.username}.
    </Banner>
```

On the `ApiKeyForm`, I also found hiding the `helpText` on `subaccount`and `validIps` made more sense since the help text is really meant for creating and editing.

rajumsys:
> When editing a subaccount key, the form shows in create mode. Looks like the getIsNew selector relies on getCurrentApiKey which has a small bug. key.subaccount_id is a Number but the subaccountId URL param is a string so === will fail.

Nice catch! for fun: we've an api endpoint which returns this as string :D. i casted to int. 

> When navigating direct to an edit page, the page should list API keys on cDM since it uses getCurrentApiKey which uses the list.

ah! i wanted to minimize the changes on edit page. but all those selector changes caused this regression. i restored `getFormApiKey` selector and it's usage which should fix it in edit page. so it's just details page that will use `getCurrentApiKey` selector. 

> Lets hide grantsRadio in read-only mode.

Didn't get this. Didn't we want to see the grants? 

> I found this read-only alert looked more in-keeping with our other banners:

I know and deliberately skipped title as it makes it bigger. But i added as you said. 

> On the ApiKeyForm, I also found hiding the helpText on subaccountand validIps made more sense since the help text is really meant for creating and editing.

I thought about it but then thought it's a same for but in read only. also not for end-users. but anyway, hidden in read only view now

PR #427, opened 2018-04-19
FE-295: Fix Suppression delete operation

ewandennis:
## Issue 1: Deleting a master account suppression throws
Cause: Delete modal expects the suppression item to exist after deletion
Solution: Make delete modal resilient to missing state

## Issue 2: Deleting a subaccount suppression causes 401 which logs the user out
No solution yet


bkemper:
I confirmed that Issue 2 will need to be addressed on the suppressions-api.

PR #428, opened 2018-04-19
FE-300 updated react and react-dom to 16.3 latest, fixes a prop type bug

jasonrhodes:
All tests pass and some manual QA looks good. There were two small PropTypes bugs on the billing page that I fixed while I was QA-ing, now that plans don't need to have the overage key present.


rajumsys:
oh double review. same report

PR #430, opened 2018-04-19
FE-292 Allow deep linking to in-app support

jonambas:
Jira ticket: https://jira.int.messagesystems.com/browse/FE-292

---

Support ticket UI can be triggered through query params or through redux actions

```js
// Support component opens form and hydrates on mount
app.sparkpost.com?supportTicket=true&supportMessage=message&supportSubject=subject
```

```js
// Toggles the support panel UI
toggleSupportPanel()

// Toggles support ticket form (algolia search shown when false)
toggleTicketForm()     

// Fills support ticket form
hydrateTicketForm({ message, subject })
```

### DIFF
--- a/src/components/support/Support.js
+++ b/src/components/support/Support.js
@@ -10,12 +10,6 @@ import { SearchPanel } from './components/SearchPanel';
 import styles from './Support.module.scss';
 
 export class Support extends Component {
-
-  state = {
-    showPanel: false,
-    showForm: false
-  };
-
   componentDidMount() {
     const { location, toggleSupportPanel, toggleTicketForm, hydrateTicketForm } = this.props;
     const { supportTicket, supportMessage: message, supportSubject: subject } = qs.parse(location.search);
--- a/src/components/support/Support.js
+++ b/src/components/support/Support.js
@@ -4,13 +4,24 @@ import { withRouter } from 'react-router';
 import qs from 'query-string';
 import { Portal, Icon, Popover } from '@sparkpost/matchbox';
 import { entitledToSupport } from 'src/selectors/support';
-import { createTicket, toggleSupportPanel, toggleTicketForm, hydrateTicketForm } from 'src/actions/support';
+import * as supportActions from 'src/actions/support';
 import SupportForm from './components/SupportForm';
 import { SearchPanel } from './components/SearchPanel';
 import styles from './Support.module.scss';
 
 export class Support extends Component {
   componentDidMount() {
+    this.maybeOpenTicket();
+  }
+
+  componentDidUpdate(prevProps) {
+    if (this.props.location.search !== prevProps.location.search) {
+      this.maybeOpenTicket();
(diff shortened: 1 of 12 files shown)


src/components/support/Support.js
```diff
@@ -23,29 +36,23 @@ export class Support extends Component {
...
-
-    this.setState(stateUpdates);
+    toggleSupportPanel();
   }
 
-  resetPanel = () => {
-    this.setState({ showPanel: false }, () => {
-      this.props.clearSupportForm();
-    });
-  };
```
jonambas: Removed this, didn't see `resetPanel` used anywhere

src/components/support/Support.js
```diff
@@ -1,18 +1,25 @@
...
-
-  state = {
-    showPanel: false,
-    showForm: false
-  };
+  componentDidMount() {
+    const { location, toggleSupportPanel, toggleTicketForm, hydrateTicketForm } = this.props;
+    const { supportTicket, supportMessage: message, supportSubject: subject } = qs.parse(location.search);
+
+    if (supportTicket) {
```
jonambas: hmm, do i need to gate this with `!loggedIn || !entitledToSupport`

src/components/support/Support.js
```diff
@@ -1,18 +1,25 @@
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
+import { withRouter } from 'react-router';
+import qs from 'query-string';
 import { Portal, Icon, Popover } from '@sparkpost/matchbox';
 import { entitledToSupport } from 'src/selectors/support';
-import { createTicket, clearSupportForm } from 'src/actions/support';
+import { createTicket, toggleSupportPanel, toggleTicketForm, hydrateTicketForm } from 'src/actions/support';
```
jasonrhodes: At this point I think it's ok to do something like:
```js
import * as supportActions from 'src/actions/support';

// component

export default withRouter(connect(mapStateToProps, supportActions)(Support));
```
It's less typing which is nice, but it also prevents those weird bugs where we accidentally use the imported action instead of the dispatch-bound action from `this.props`.

src/components/support/Support.js
```diff
@@ -1,18 +1,25 @@
...
 import { SearchPanel } from './components/SearchPanel';
 import styles from './Support.module.scss';
 
 export class Support extends Component {
-
-  state = {
-    showPanel: false,
-    showForm: false
-  };
+  componentDidMount() {
```
jasonrhodes: One down side to doing this in `cDM` is you can't link to a page with the support panel open from inside the app, something like:
```js
<Link to='/account/billing?supportTicket=true' />
```
We could probably move this logic into a function and then call it from `cDM` and `cDU`?

jasonrhodes: Maybe `if (location.search && location.search !== prevProps.location.search)` but yeah

PR #431, opened 2018-04-23
FE-98 Add phone support text in submit a ticket pop up

jonambas:
To verify, make sure you are on a plan with phone support. Use appteam account to verify the negative case where online support is still available but phone is not

### DIFF
--- a/src/components/support/components/SupportCta.js
+++ b/src/components/support/components/SupportCta.js
@@ -4,7 +4,7 @@ import { entitledToPhoneSupport } from 'src/selectors/support';
 import { Button } from '@sparkpost/matchbox';
 import styles from './SupportCta.module.scss';
 
-const SupportCta = ({ entitledToPhoneSupport, toggleForm }) => {
+export const SupportCta = ({ entitledToPhoneSupport, toggleForm }) => {
   let leftContent = <span>Do you need more assistance?</span>;
 
   if (entitledToPhoneSupport) {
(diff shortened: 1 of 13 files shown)


src/components/support/components/SupportCta.js
```diff
@@ -0,0 +1,35 @@
...
+      <div className={styles.Phone}>
+        <p>Phone Support: <strong>415-751-0928</strong></p>
+        <p>Available: <strong>9am - 7pm EST, Mon-Fri</strong></p>
+      </div>
+    );
+  }
+
+  return (
+    <Fragment>
+      {leftContent}
```
rajumsys: looks like `do you ned more assistance` part will still make sense with phone support. adding one line on top phone and ticket cta probably won't look too bad.

rajumsys:
- Overall looks good. Just one comment I left in the code about showing `do you need more assistance` text. 
- also (exists in master) the panel is showing if i logout while it's open and i log back in regardless of the account i'm logging in.

src/components/support/components/SupportCta.js
```diff
@@ -0,0 +1,35 @@
...
+import styles from './SupportCta.module.scss';
+
+export const SupportCta = ({ entitledToPhoneSupport, toggleForm }) => {
+  let leftContent = <span>Do you need more assistance?</span>;
+
+  if (entitledToPhoneSupport) {
+    leftContent = (
+      <div className={styles.Phone}>
+        <p>Do you need more assistance?</p>
+        <p><strong>+1 415-751-0928</strong> <span className={styles.Spacer}>—</span> <strong>9am – 7pm EST, Mon – Fri</strong></p>
```
rajumsys: any reason for not adding `-` after `+1`?

jonambas: 🤔

jonambas:
Awaiting confirmation on phone number & hours

PR #432, opened 2018-04-23
FE-306: Fix hiding on hover on ff

rajumsys:
(Temporary; less annoying) Fixes hiding of pie chart on hover with one caveat (in ff); see photos below:

Before:
screen shot 2018-04-23 at 4 48 21 pm

After:
screen shot 2018-04-23 at 4 50 04 pm
(does not set opacity for other slices)

Other things tried:
- Upgrading `recharts` to latest version `1.0.0-beta.10`. Reverted as didn't solve it
- Tried `-moz-opacity: 0.4`. Didn't work. Nor does `-moz-opacity: 1`.

### DIFF
--- a/src/components/pieChart/async/Chart.scss
+++ b/src/components/pieChart/async/Chart.scss
@@ -6,7 +6,8 @@
 
 .Chart--hover {
   .recharts-sector {
-    opacity: 0.4;
+    opacity: 1;
+    -webkit-opacity: 0.4;
   }
 }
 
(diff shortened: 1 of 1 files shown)


jonambas:
Using `fill-opacity` instead of `opacity` seems to work for me:

```css
.recharts-sector {
  fill-opacity: 1;
  transition: 0.3s;
}

.Chart--hover {
  .recharts-sector {
    fill-opacity: 0.4;
  }
}

.recharts-sector.Chart--forceOpacity {
  fill-opacity: 1 !important;
}
```

rajumsys:
Thanks it works on FF. Now Chrome looks funny :). Probably we can use this for ff but use `opacity` for chrome. 
screen shot 2018-04-23 at 6 24 32 pm 1
screen shot 2018-04-23 at 6 25 15 pm

PR #433, opened 2018-04-23
FE-35: Error not surfaced to user when SSO sign-in fails

bkemper:
_Refer to [FE-35](https://jira.int.messagesystems.com/browse/FE-35)_

**How to test locally**
- Load auth page with error param set to any base64 encoded string (e.g. http://app.sparkpost.test/auth?error=VW5hYmxlIHRvIGNyZWF0ZSB0b2tlbiBmb3IgU0FNTCB1c2VyLg%3D%3D)

**How to test in UAT**
1. Navigate to https://app-speuat.tst.sparkpost.com _[should load SparkPost authentication form and only ask for email or username]_
1. Enter "appteam+sso" _[should redirect to OneLogin authentication form]_
1. Enter "appteam" credentials from LastPass _[should redirect back to SparkPost authentication page with an error message]_

### DIFF
--- a/src/helpers/string.js
+++ b/src/helpers/string.js
@@ -70,3 +70,11 @@ export function stripTags(html) {
     .replace(spacesRegex, space) // avoid multiple spaces between words
     .trim();
 }
+
+export function decodeBase64(str) {
+  try {
+    return atob(str);
+  } catch (e) {
+    return undefined; // ignore parsing error
+  }
+}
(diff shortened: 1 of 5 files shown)


rajumsys:
Did you figure out the other error (possible race condition or whatever) that we during screenshare? asking as haven't seen any note and/or obvious code change.

bkemper:
@rajumsys I rewrote that code to not have a race condition and I was still unable to get subsequent requests to authenticate.  I couldn't reproduce it in any other environment, so I'm going to chalk it up to a development environment issue.  Also, I setup that speuat environment which is a much better place to test this.

PR #434, opened 2018-04-24
FE-303 Fix support form attachments

jasonrhodes:
In this ticket I:

* [x] Used the native FileReader API to convert the incoming File object to a Base64-encoded string, first by reading it as a "data URL" and then by stripping the data URL `data:*/*;base64,` prefix off of that string. This seemed like the easiest way since there was no "read as base64 string directly" method.
* [x] Abstracted the FileReader logic into a helper
* [x] Replaced the bespoke File upload field used in the Support panel with the general FileFieldWrapper used everywhere else in the app
* [x] Tweaked the FileFieldWrapper to have more drag/drop affordance
* [x] Updated snapshot tests for those changes
* [x] Wrote new tests for the helper and a few other logic updates added here
* [x] Turned off linting during webpack development so that we can incrementally add lint rules without blowing up the code base, using lint-staged and husky pre-commit hooks
* [x] Added (2) fixable lint rules
  * Always have a space between `function` and `name`, `()`, etc, including always enforcing a space between `async` and an arrow function
  * Enforce (no) spacing in JSX curlies

---
* [x] Fixed a small bug with support ticket deep linking -- we were doing a blind toggle in cDM and cDU that meant that certain cases opened then closed the panel. I created an OPEN action that is explicit so that multiple calls won't close it when we know we want to be opening it.
* [x] Fixed a one character typo on the Sending Domains page


bkemper:
May need to shorten the textbox to make room for this validation message.

screen shot 2018-04-24 at 4 03 32 pm

src/helpers/file.js
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Returns only the base 64 encoded string version of a
+ * JS File object, using the data URL API and stripping the prefix
+ *
+ * @param {File} file object
+ * @return {Promise -> String} base 64 encoded contents of the file
+ */
+export function getBase64Contents(file) {
+  return getDataUrl(file).then((dataUrl) => dataUrl.replace(/^data:.*\/.*;base64,/, ''));
```
bkemper: To avoid this replace, do you think it makes sense to use `reader.readAsText` and then `btoa` the result?

jonambas:
That popover has a fixed height, might need to make that Error appear inline next to the Label

.eslintrc
```diff
@@ -11,6 +11,7 @@
     ],
     "complexity": "off",
     "react/jsx-no-bind": "off",
+    "space-before-function-paren": ["error", "always"],
```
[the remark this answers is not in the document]
jasonrhodes: I didn't see a reason to break out the rule for separate cases?

src/components/support/components/SupportForm.js
```diff
@@ -76,22 +68,22 @@ export class SupportForm extends Component {
           <Field
             type='file'
             name='attachment'
-            label='Attachment'
-            component={AttachmentField}
+            label='Attach a file (optional)'
```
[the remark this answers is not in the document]
jasonrhodes: We don't need it. It just felt nice to me because we don't use the required asterisk (which can be a lot of noise if all fields are required but one). But I don't feel strongly :)

src/helpers/file.js
jasonrhodes: It doesn't work because `readAsText` gives you back the text representation of the image file, i.e.
```
�PNG


IHDR�Z>l�gAMA���a@IDATx�]|Et�@I�:���X@@A�.�� ��A)*ME��
```
etc.

If you try to `btoa` that you get `Uncaught DOMException: Failed to execute 'btoa' on 'Window': The string to be encoded contains characters outside of the Latin1 range.`

.eslintrc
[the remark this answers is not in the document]
jasonrhodes: I will let everyone know about the change.

src/helpers/tests/file.test.js
```diff
@@ -0,0 +1,44 @@
...
+      this.readAsDataURL = jest.fn(() => {
+        this.result = 'data:image/png;base64,<BASE64 ENCODED CONTENTS>';
+        this.cb();
+      });
+    };
+    file = new File(['some contents'], 'test.png');
+  });
+
+  afterEach(() => {
+    global.FileReader = FR;
```
bkemper: I wish there was an easier way to do this.  Dependency injection is a less brittle way, but it junks up your source. 😞

jasonrhodes: Yeah, in a case like this I don't mind mocking the global.

src/helpers/file.js
```diff
@@ -0,0 +1,29 @@
+/**
+ * Returns the data:image/png;base64,<encoded> data URL
+ * version of a JS File object
+ *
+ * @param {File} file object
+ * @return {Promise -> String} data url of the file
+ */
+export function getDataUrl (file) {
+  if (!(file instanceof File)) {
+    return Promise.reject('Only native File objects can be converted to data URL');
```
bkemper: 👍

bkemper:
@jasonrhodes There were two merge conflict that I had to fix.

PR #439, opened 2018-04-26
FE-304: Reinstate analytics events

ewandennis:
- Added some generic analytics helpers under `src/helpers/analytics`
 - Added conversion-specific helpers under `src/helpers/conversionTracking`
 - Moved existing analytics calls over to new helpers
 - Reinstated the events that the legacy app used to emit
 - Much testing

### Testing
The `dataLayer` global variable is an array that contains an entry for each event the app emits. You can verify an event is emitted by taking some triggering action in the app, then checking the contents of `dataLayer`.

[The set of events emitted is listed here, along with their structure](https://confluence.int.messagesystems.com/display/ENG/The+UI+And+Google+Tag+Manager).


src/helpers/analytics.js
```diff
@@ -0,0 +1,56 @@
+// Google Tag Manager tracking helpers
+// Adapted from webui/src/common/services/analytics.js
+
+// Call once at app load time to initialise the GTM dataLayer variable
+// and trigger the GA 'Page View' event.
+function setup() {
+  if (!window.dataLayer) {
+    window.dataLayer = [{
```
jasonrhodes: Should we just create the data layer if it doesn't exist, and then outside of the `if` do a `dataLayer.push()` with the `gtm.start` event? Just in case it gets created as an empty array early somehow? :shrug:

src/helpers/analytics.js
```diff
@@ -0,0 +1,56 @@
...
+  }
+}
+
+function pushEvent(evt) {
+  window.dataLayer.push(evt);
+}
+
+// Call on route change.
+// Used in Marketing GTM account.
+function trackPageview(path, title, username) {
```
jasonrhodes: Can we pass a destructured object here? Will make it easer to add info later if necessary, etc.

src/helpers/analytics.js
```diff
@@ -0,0 +1,56 @@
...
+  pushEvent({
+    event: 'content-view',
+    'content-name': path,
+    'content-title': title,
+    username
+  });
+}
+
+// Call to push a generic event to GTM.
+function trackEvent(category, action, data) {
```
jasonrhodes: Do you think we should do a destructured object here, too? I think 3 args is the line and you could go either way there, but I get nervous about having to add or reorder args later for some reason :D

src/helpers/analytics.js
```diff
@@ -0,0 +1,56 @@
...
+  pushEvent({ [name]: value });
+}
+
+export {
+  setup,
+  trackPageview,
+  trackEvent,
+  trackFormSuccess,
+  setVariable
+};
```
jasonrhodes: Can we export each function inline when it's defined, `export function` etc? The places where we do `export { }` are, I think, from before we realized you could export inline.

src/helpers/conversionTracking.js
```diff
@@ -0,0 +1,32 @@
...
+    form_type: action,
+    addon_key: addonType
+  });
+}
+
+function trackAddonRequest(addonType) {
+  analytics.trackEvent('Clicked button', addonType, { action: addonType });
+}
+
+export {
```
jasonrhodes: Same inline export change here.

src/helpers/conversionTracking.js
```diff
@@ -0,0 +1,32 @@
+import * as analytics from './analytics';
+
+// These events are consumed through GTM by various Marketing-owned services to trigger
+// follow-on actions.
+
+function trackPlanChange(allPlans, oldCode, newCode) {
```
jasonrhodes: I'm being annoying now but maybe another destructured object here? This one is interesting bc if we ever figure out a way to not have to pass allPlans to this function, would be nice to just be able to drop that without affecting the arg order...

src/components/googleTagManager/GoogleTagManager.js
```diff
@@ -11,10 +15,10 @@ export class GoogleTagManager extends Component {
   }
 
   componentDidMount() {
-    window.dataLayer = [{
-      'gtm.start': new Date().getTime(),
-      event: 'gtm.js'
-    }];
+    if (!this.state.dataLayerLoaded) {
```
jasonrhodes: This can't be true here if we only set it true at the bottom of this function, right? I think we only need to check this state variable in render to prevent loading the script before we've loaded our dataLayer, but I may misunderstand the flow there.

src/components/googleTagManager/GoogleTagManager.js
```diff
@@ -42,13 +46,11 @@ export class GoogleTagManager extends Component {
...
     const route = findRouteByPath(location.pathname);
 
-    window.dataLayer.push({
-      event: 'content-view',
-      'content-name': location.pathname + location.search, // duplicates angular 1.x ui-router "$location.url()" which is /path?plus=search
-      'content-title': route.title || location.pathname, // duplicate angular 1.x $rootScope.stateData.title
+    analytics.trackPageview(
+      location.pathname + location.search, // duplicates angular 1.x ui-router "$location.url()" which is /path?plus=search
+      route.title || location.pathname, // duplicate angular 1.x $rootScope.stateData.title
       username
```
jasonrhodes: I think the old angular app did a separate GTM push for username that we combined here because Daniel thought it might be the easiest fix for whatever problem we were seeing. If we've reverted everything else to the "old way", we might be able to do that too and remove some of this hairy waitForUser logic... but also this is me assuming you understand this flow much better than me now that you've looked into it for a few weeks :)

src/helpers/analytics.js
ewandennis: Agree in principle. As it is though, the guard in `setup()` ensures we don't emit >1 `gtm.js` events by calling twice. Since that's the `Page View` event which triggers a 💩-load of tags, I like this implementation for safety.

Granted, the only case this could happen today is if we unmount and remount `GoogleTagManager`.

src/components/googleTagManager/GoogleTagManager.js
ewandennis: Yup. Since `setup()` includes its own guard, its protected from abuse of `GoogleTagManager` to a degree.

PR #441, opened 2018-04-26
FE-296 Catch component errors with React Error Boundary

rajumsys:
How to test:
- Full page: throw an exception from public or protected router.
- In Page: throw an exception from any pages in `src/pages` or any component referenced in there. 
- In development, you'll still see the error overlay. Seems there isn't a way to hide it (it was making me crazy thinking my error boundary wasn't working). Just close it to see the page rendered by boundary component. For same reason, I believe, the errors will be logged to sentry even if we remove `ErrorTracker.report({ info }, error);` from `ErrorBoundary.js`.

Notes:
- Two error boundaries. One for whole app, another for just pages. So in that case the nav will be still rendered. If the error is specific to a page, users can still visit other pages. App boundary will, however, show full page message
- Full page one probably needs some styling. Let me know about the copy of exact message. 
- For full page error, i've put a button to reload. May be here (and in page error as well), we can [ask for user feedback](https://docs.sentry.io/learn/user-feedback/) :mindblown:
- Doesn't look like `report` in errorTracker working as expected (or i need to understand it). Seems the extra contexts aren't visible in sentry. may be we should expose apis to set context. that way we can make it explicit that error is coming from boundaries. 

Other note:
Seems App.js, index.js isn't under test coverage. Any reason not to?


jasonrhodes:
Not taking this review unless it's still available after I finish the GTM one, but just wanted to comment one thing for now: 
> Seems App.js, index.js isn't under test coverage. Any reason not to?

I vote for not doing that, I can't see a good reason to do it. If we want to test, for example, the logic of the route selection, I think we should move it out to another file and test that file directly instead of trying to instantiate the full `<App>` and test logic there. Etc.

src/components/errorBoundaries/ErrorBoundary.js
```diff
@@ -0,0 +1,30 @@
+import React, { Component } from 'react';
+import Error from './component/Error';
+import ErrorTracker from 'src/helpers/errorTracker';
+
+const primaryAction = {
+  content: 'Reload Page',
+  onClick: () => {
+    window.location.reload();
```
jonambas: Maybe `reload(true)` for a hard refresh?

src/components/errorBoundaries/ErrorBoundary.js
```diff
@@ -0,0 +1,30 @@
...
+
+  componentDidCatch(error, info) {
+    this.setState({ hasError: true });
+    ErrorTracker.report({ info }, error);
+  }
+
+  render() {
+    const { showAction } = this.props;
+    if (this.state.hasError) {
+      return <Error primaryAction={showAction ? primaryAction : null } />;
```
jonambas: Do you think we can call this component something else? Matchbox already has `Error` as a component, maybe `ErrorPage` `ErrorCatcher`?

jonambas: Also, I don't see why you couldn't move `EmptyState` to this component instead of splitting it up  into two

src/components/errorBoundaries/component/Error.js
```diff
@@ -0,0 +1,15 @@
...
+export default (props) => {
+  const { message, primaryAction } = props;
+
+  return <EmptyState
+    title='Sorry, something went wrong'
+    image='Generic'
+    primaryAction={primaryAction}
+    {...props}
+  >
+    <p>{message || 'Our engineers are notified and working on fixing this issue.'}</p>
```
jonambas: Some ideas: 

> We're having some technical issues. Our engineers have been notified and are working on getting this fixed.

> We're having some technical issues. Please try to refresh the page, or try again later.

src/components/errorBoundaries/component/Error.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { EmptyState } from '@sparkpost/matchbox';
+
+export default (props) => {
+  const { message, primaryAction } = props;
+
+  return <EmptyState
```
jonambas: Easy way to center this
```js
<div style={{ margin: '0 auto', maxWidth: 1080 }}>
```

src/components/errorBoundaries/ErrorBoundary.js
```diff
@@ -0,0 +1,30 @@
...
+};
+
+export default class ErrorBoundary extends Component {
+  state = {
+    hasError: false
+  }
+
+  componentDidCatch(error, info) {
+    this.setState({ hasError: true });
+    ErrorTracker.report({ info }, error);
```
jonambas: Yeah seems like you can remove this, i see double errors

src/components/errorBoundaries/component/tests/Error.test.js
```diff
@@ -0,0 +1,33 @@
...
+  });
+
+  it('renders correctly with custom message', () => {
+    wrapper.setProps({ message: 'fooooo' });
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('renders correctly with actions', () => {
+    wrapper.setProps({ primaryAction: { content: 'Reload' }});
+    expect(wrapper).toMatchSnapshot();
```
jonambas: To avoid redundant snapshots, could we change these assertions to 
`expect(wrapper.find('EmptyState').props().children).toMatchSnapshot();`
`expect(wrapper.find('EmptyState').props().primaryAction).toEqual({ content: 'Reload' });`

src/components/errorBoundaries/tests/__snapshots__/ErrorBoundary.test.js.snap
```diff
@@ -0,0 +1,24 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Component: ErrorBoundary renders correctly with error 1`] = `
+<Component
```
jonambas: You many want to set a `displayName` on this component so it shows up correctly in snapshots

PR #444, opened 2018-04-30
FE-321 Update react-ace, react-timeago

jonambas:
- Closes #420 
- Closes #410


jonambas:
I was on npm 6 which caused a huge package-lock diff, reverted back to 5.5 for now

PR #455, opened 2018-05-08
FE-317 User can make in-app unsuspension requests

jonambas:
Jira ticket: https://jira.int.messagesystems.com/browse/FE-317

---

New component: `<SuspensionAlerts />`  rendered under `<App />`

- Moved all code around these alerts from AuthGate and globalAlert actions into this component.
- It felt weird to render jsx inside an action
- All logic is now in one place

screen shot 2018-05-08 at 2 54 46 pm


src/components/auth/SuspensionAlerts.js
```diff
@@ -0,0 +1,61 @@
...
+    return this.props.isSuspendedForBilling
+      ? (
+        <Fragment>
+          <div>Your account is currently suspended due to a billing problem.</div>
+          <div>To make a payment and reactivate your account, <UnstyledLink component={Link} to='/account/billing'>visit the billing page</UnstyledLink>.</div>
+        </Fragment>
+      ) : (
+        <Fragment>
+          <div>Your account is currently suspended.</div>
+          <div>For any questions or to request reactivation, please <UnstyledLink onClick={this.openTicket}>submit a ticket</UnstyledLink>.</div>
```
jonambas: This currently poses a small problem. Free accounts will still see this alert, but clicking this link won't render the support UI.

src/components/auth/SuspensionAlerts.js
```diff
@@ -0,0 +1,61 @@
...
+    }
+  }
+
+  render () {
+    return null;
+  }
+}
+
+const mapStateToProps = (state) => ({
+  status: state.account.status,
```
bkemper: Should we used the [`hasStatus()`](https://github.com/SparkPost/2web2ui/blob/master/src/helpers/conditions/account.js#L10) helper?

src/components/auth/SuspensionAlerts.js
```diff
@@ -0,0 +1,61 @@
...
+    this.props.openSupportPanel({ view: 'ticket' });
+    this.props.hydrateTicketForm({ issueId: 'account_suspension' });
+  }
+
+  getMessage () {
+    return this.props.isSuspendedForBilling
+      ? (
+        <Fragment>
+          <div>Your account is currently suspended due to a billing problem.</div>
+          <div>To make a payment and reactivate your account, <UnstyledLink component={Link} to='/account/billing'>visit the billing page</UnstyledLink>.</div>
```
bkemper: `<PageLink />` could be used here.

src/components/auth/tests/SuspensionAlerts.test.js
```diff
@@ -0,0 +1,56 @@
...
+      isSuspendedForBilling: false,
+      openSupportPanel: jest.fn(),
+      hydrateTicketForm: jest.fn()
+    };
+
+    wrapper = shallow(<SuspensionAlerts {...props} />);
+  });
+
+  afterEach(() => {
+    jest.resetAllMocks();
```
bkemper: This can be removed.

PR #461, opened 2018-05-11
AC-78: Allow users to login via SSO without flag

bizob2828:
TODO:
- [x] write tests
- [x] update Ansible to no longer require sso enabled flag

Ansible review here: https://fisheye.int.messagesystems.com/cru/APP-3235#CFR-131427


src/pages/auth/AuthPage.js
```diff
@@ -1,119 +1,35 @@
...
-        <CenteredLogo />
-        <Panel sectioned accent title="Log In">
-          {loginError && <Error error={loginError} />}
-
-          {tfa.enabled && <TfaForm onSubmit={this.tfaSubmit} />}
-          {!tfa.enabled && <LoginForm onSubmit={this.loginSubmit} ssoEnabled={this.state.ssoEnabled}/>}
-        </Panel>
-        {footerMarkup}
-      </div>
+      <LoginPanel title={'Log In'} ssoEnabled={false} loginError={auth.errorDescription} handleSubmit={this.loginSubmit}>
```
bizob2828: this is techinally a breaking change. Because I have to update accusers to update the error route on consumption(https://github.com/SparkPost/accusers-api/pull/381) so I can put back parsing of the url to get an error for sso here temporarily

bizob2828: basically restore some of this https://github.com/SparkPost/2web2ui/pull/461/files#diff-aa34e41e34701685b3fa9d42f8083adaR39

src/pages/auth/components/TfaForm.js
```diff
@@ -29,16 +42,17 @@ export class TfaForm extends Component {
...
     );
   }
 }
 
-function mapStateToProps({ tfa }) {
+function mapStateToProps ({ tfa }) {
   return {
+    tfa,
     tfaPending: tfa.tfaPending,
     initialValues: {
```
bkemper: I don't think this is necessary.

bizob2828: it is and i don't recall why, but I remember when I was working on this with @jasonrhodes we had issues without it. i'm keeping as is for now

PR #463, opened 2018-05-11
ESCSP-2204

rajumsys:
log zuora payload (partially) to track card type issue

### DIFF
--- a/src/helpers/errorTracker.js
+++ b/src/helpers/errorTracker.js
@@ -33,7 +33,7 @@ export const BROWSER_EXTENSION_REGEX = new RegExp('^(chrome|chrome-extension|res
  * @param {number} crumb.timestamp - created at
  * @see https://docs.sentry.io/clients/javascript/config
  */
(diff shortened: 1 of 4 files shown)


src/helpers/errorTracker.js
```diff
@@ -104,10 +104,15 @@ class ErrorTracker {
...
    */
-  report(logger, error) {
+  report (logger, error, extra = {}) {
     // Silently ignore if Sentry is not setup
     if (!Raven.isSetup()) { return; }
-    Raven.captureException(error, { logger });
+    Raven.captureException(error, { logger, extra });
+  }
+
+  addBreadcrumb (data) {
```
bkemper: I think this is fine for now, but with raven-for-redux middleware in place, we should start using [breadcrumbDataFromAction](https://github.com/captbaritone/raven-for-redux#breadcrumbdatafromaction-function) to enrich all our breadcrumbs with non-sensitive data.

PR #465, opened 2018-05-14
disallow chrome to autofill cc hidden fields, reverts error tracking code

rajumsys:
There are multiple ways to solve this but IMO this is simplest. So far it seems to be working fine but real test would be on tst/stg. 

other approaches
- rename fields
- set state locally 
- calculate these upon submit

all those require more work than this is. renaming field may be trivial but still can't be tested locally. 

https://cs.chromium.org/chromium/src/components/autofill/core/common/autofill_regex_constants.cc?l=135-194 is the regex chrome use for field matching. as i couldn't enable cc auto fill locally, it's tough to test. also seemed this approach isn't very uncommon. 

notes: 
- also reverting the code that we used to log zuora payload
- someone may not ask "why not just autocomplete='off'". most browsers/extensions ignores that now. there is a SO thread about it but i lost it. I can find again if you (reviewer) is interested.

### DIFF
--- a/src/pages/billing/forms/fields/PaymentForm.js
+++ b/src/pages/billing/forms/fields/PaymentForm.js
@@ -114,7 +114,12 @@ export class PaymentForm extends Component {
           </Grid.Column>
         </Grid>
 
-        {/* Hidden redux-form connected fields */}
+        {
+        /*
+        Hidden redux-form connected fields.
+        autoComplete attr is to prevent chrome automatically filling credit card number.
+        */
+        }
         <div className={styles.hidden} >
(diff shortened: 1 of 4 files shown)


src/pages/billing/forms/fields/PaymentForm.js
```diff
@@ -116,9 +116,9 @@ export class PaymentForm extends Component {
 
         {/* Hidden redux-form connected fields */}
```
jonambas: Maybe add a comment here about `autoComplete` if you want, otherwise everything else looks good

PR #466, opened 2018-05-15
FE-325 Remove redundant webhooks call

rajumsys:


src/actions/webhooks.js
```diff
@@ -1,11 +1,10 @@
 import sparkpostApiRequest from 'src/actions/helpers/sparkpostApiRequest';
 import setSubaccountHeader from './helpers/setSubaccountHeader';
-import { mergeWebhooks } from 'src/helpers/webhooks';
 
-export function listWebhooks({ subaccount = null, type = 'LIST_WEBHOOKS' } = {}) {
+export function listWebhooks (subaccount = null) {
```
ewandennis: Do we use the `subaccount` parameter to `listWebhooks()`? If not, lets remove it. (I couldn't find an instance).

rajumsys: Right, deleted that.

rajumsys:
@ewandennis addressed the feedback. can ship it today once approved.

PR #469, opened 2018-05-16
AC-106 fixed issue with resetting error on pending password reset

bizob2828:


src/reducers/passwordReset.js
```diff
@@ -10,7 +10,7 @@ const initialState = {
 export default (state = initialState, { type, payload }) => {
   switch (type) {
     case 'SEND_PASSWORD_EMAIL_PENDING':
-      return { ...state, emailPending: true, emailSuccess: false };
+      return { ...state, emailPending: true, emailSuccess: false, emailError: null };
 
     case 'SEND_PASSWORD_EMAIL_SUCCESS':
       return { ...state, error: null, emailPending: false, emailSuccess: true };
```
bkemper: It is probably safe to remove `error`, as well?

bizob2828: this key isn't being used. im removing

PR #471, opened 2018-05-17
FE-334 Top Navigation

jonambas:
- Adds a top navigation bar
- Adds a new context provider - `WindowSizeContext` that determines if a mobile theme should be used.

```js
<WindowSizeContext.Consumer>
  {({ mobile }) => (mobile ? <MobileThing/> : <DesktopThing /> )}
</WindowSizeContext.Consumer>;
```

---
Matchbox PR: https://github.com/SparkPost/matchbox/pull/128

- Animation now uses RTG, contents are not rendered unless open
- `onClose` prop modified to handle both escape and outside click events - `handleOutsideClick` removed


src/components/navigation/components/Item.module.scss
```diff
@@ -0,0 +1,107 @@
+@import '~@sparkpost/matchbox/src/styles/config.scss';
```
jasonrhodes: The side nav items feel a tad tight together to me now, what do you think?

PR #477, opened 2018-05-21
FE-298 make browsers support official 

rajumsys:
- link http://browserl.ist/  in app not loading page. let me know objections
- Added browserslist dep as, otherwise, `npx browserslist` doesn't follow the rule in `package.json`
- explicitly disallowed all IE

we've about 83% coverage now

Notes: 
- Adding `not dead` isn't working during build (version issue, i guess). but > 1% and last 2 versions combination would probably give us something similar. 

- Browserlist repo suggested not to disallow unknown browsers but in our case our query disallowed browsers like baido, opera mobile, qq browsers. 
- There are some notes in ticket.

### DIFF
--- a/package.json
+++ b/package.json
@@ -161,14 +161,10 @@
   "browserslist": {
     "production": [
       "> 1%",
+      "last 2 versions",
       "not ie <= 11",
-      "Firefox ESR",
-      "last 2 opera versions",
-      "last 2 edge versions",
-      "last 2 safari versions",
-      "last 2 chrome versions",
-      "last 2 firefox versions",
-      "last 2 FirefoxAndroid versions"
+      "not ie_mob <=11",
+      "Firefox ESR"
     ]
   }
 }
--- a/package.json
+++ b/package.json
@@ -160,10 +160,15 @@
   },
   "browserslist": {
     "production": [
-      "last 4 versions",
-      "not < 1%",
+      "> 1%",
+      "not ie <= 11",
       "Firefox ESR",
-      "not ie <= 11"
+      "last 2 opera versions",
+      "last 2 edge versions",
+      "last 2 safari versions",
+      "last 2 chrome versions",
+      "last 2 firefox versions",
+      "last 2 FirefoxAndroid versions"
     ]
   }
 }
--- a/package.json
+++ b/package.json
@@ -104,6 +104,7 @@
     "babel-preset-stage-2": "^6.24.1",
     "babel-runtime": "^6.26.0",
     "brace": "^0.11.0",
+    "browserslist": "^3.2.8",
     "case-sensitive-paths-webpack-plugin": "^2.1.1",
     "chalk": "^2.3.1",
     "coveralls": "^3.0.0",
@@ -159,10 +160,11 @@
   },
   "browserslist": {
     "production": [
-      ">1%",
       "last 4 versions",
+      "not < 1%",
       "Firefox ESR",
-      "not ie < 11"
+      "not ie < 11",
+      "not ie 11"
     ]
   }
 }
--- a/package.json
+++ b/package.json
@@ -163,8 +163,7 @@
       "last 4 versions",
       "not < 1%",
       "Firefox ESR",
-      "not ie < 11",
-      "not ie 11"
+      "not ie <= 11"
     ]
   }
 }
--- a/public/index.html
+++ b/public/index.html
@@ -84,7 +84,7 @@
           <path class='logo__flame' d="M106.5 101.3c-13.4 10.1-16 28-16.4 40.1C68.8 116.8 129.3 44.6 60.6 0 103.1 55.1 0 109.3 0 190.4c0 31.8 19.9 59.9 63.5 69.6 42.8-9.1 64-37.8 64-69.6 0-47.4-29.5-63.3-21-89.1zM63.6 234c-23.3 0-42.2-18.9-42.2-42.2 0-23.3 18.9-42.2 42.2-42.2 23.3 0 42.2 18.9 42.2 42.2.1 23.3-18.8 42.2-42.2 42.2z"/>
         </svg>
         <h3>Oh no, SparkPost is having trouble loading.</h3>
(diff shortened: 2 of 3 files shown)


rajumsys:
Seems build is failing due to query ` not < 1%`. Could be a version issue. Need to check why/when that gets executed during build process.

public/index.html
```diff
@@ -84,7 +84,7 @@
           <path class='logo__flame' d="M106.5 101.3c-13.4 10.1-16 28-16.4 40.1C68.8 116.8 129.3 44.6 60.6 0 103.1 55.1 0 109.3 0 190.4c0 31.8 19.9 59.9 63.5 69.6 42.8-9.1 64-37.8 64-69.6 0-47.4-29.5-63.3-21-89.1zM63.6 234c-23.3 0-42.2-18.9-42.2-42.2 0-23.3 18.9-42.2 42.2-42.2 23.3 0 42.2 18.9 42.2 42.2.1 23.3-18.8 42.2-42.2 42.2z"/>
         </svg>
         <h3>Oh no, SparkPost is having trouble loading.</h3>
-        <p>Please <a id='reloadLink'>try again</a> soon.</p>
+        <p>Please <a id='reloadLink'>try again</a> soon. Try using a <a target="_blank" href="http://browserl.ist/?q=%3E+1%25%2C+not+ie+%3C%3D+11%2C+Firefox+ESR%2C+last+2+opera+versions%2C+last+2+edge+versions%2C+last+2+safari+versions%2C+last+2+chrome+versions%2C+last+2+opera+versions%2C+last+2+firefox+versions%2C+last+2+FirefoxAndroid+versions">supported browser</a>.</p>
```
jasonrhodes: Neat idea but I think we'd be better off with something more vague -- maybe something like:
```html
<p>
  We're very sorry for the inconvenience. Please try again soon.
  <br /><small>(Note: Some users visiting from a legacy or uncommon browser may want to consider trying another browser or upgrading to the latest version.)</small>
</p>
```

package.json
```diff
@@ -159,10 +160,15 @@
   },
   "browserslist": {
     "production": [
-      ">1%",
-      "last 4 versions",
+      "> 1%",
+      "not ie <= 11",
```
jasonrhodes: maybe just "not ie" ?

jasonrhodes:
What happened to this AC re: Sentry (I saw your comment in the Jira ticket but still not clear why we can't block the IE11 UA string, at least):
"Probably best to blacklist the specific UA string for IE11 to start"

public/index.html
rajumsys: Yeah, I later felt my copy `Try using...`, assumes they're using old browser which they may not. 

i can update with your copy. but do you want to say 'some users...'. i think it's fine about to talk this particular case like 'if you're using ....'

package.json
```diff
@@ -159,10 +160,15 @@
...
       "Firefox ESR",
-      "not ie < 11"
+      "last 2 opera versions",
+      "last 2 edge versions",
+      "last 2 safari versions",
+      "last 2 chrome versions",
+      "last 2 firefox versions",
+      "last 2 FirefoxAndroid versions"
```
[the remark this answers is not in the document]
rajumsys: yes, it was bringing some browsers which i have no way to test (it's true i haven't tested all anyway though) like baidu, blackberry, Android Browser4.4.3-4.4.4 etc.

package.json
rajumsys: no, it wouldn't work. it requires a version :(

rajumsys:
> "Probably best to blacklist the specific UA string for IE11 to start"

Yeah, the comment is the reason I didn't do it. we can block IE11. But why specifically IE11 only?

public/index.html
jasonrhodes: I like it being as vague as possible, but feel free to tweak what I suggested :)

public/index.html
```diff
@@ -28,21 +28,65 @@
...
+        var isIE = /*@cc_on!@*/false || !!document.documentMode;
+
+        switch(true){
+          // Opera 8.0+
+          case (!!window.opr && !!opr.addons) || !!window.opera || navigator.userAgent.indexOf(' OPR/') >= 0:
+            return 'Opera';
+
+          // Firefox 1.0+
+          case typeof InstallTrigger !== 'undefined':
+          return 'Firefox';
```
jasonrhodes: indent?

public/index.html
```diff
@@ -28,21 +28,65 @@
...
+          case !!window.chrome && !!window.chrome.webstore:
+            return 'Chrome';
+
+          case /constructor/i.test(window.HTMLElement) || (function (p) { return p.toString() === "[object SafariRemoteNotification]"; })(!window['safari'] || safari.pushNotification):
+            return 'Safari';
+
+          case isIE:
+            return 'IE';
+
+          case !isIE && !!window.StyleMedia:
```
jasonrhodes: Don't need the `!isIE` here since the case above handles that, right?

public/index.html
```diff
@@ -28,21 +28,65 @@
 
      <!-- Pushes app load failure to Sentry -->
     <script>
+      /**
+       * Returns browser's name and version
+       * @credit https://jsfiddle.net/311aLtkz/
+       * @returns {*}
```
jasonrhodes: returns what now? haha

public/index.html
```diff
@@ -28,21 +28,65 @@
...
+       */
+      function getAgent(){
+        var isIE = /*@cc_on!@*/false || !!document.documentMode;
+
+        switch(true){
```
jasonrhodes: whoa this is weird ...

public/index.html
rajumsys: oh yeah. that looks redundant!

public/index.html
jasonrhodes: Sorry that wasn't a super helpful comment. :) `switch (true)` feels like a bit of a hard-to-understand hack to me but for this inlined JS code, I don't care too much, so let's just go with it.

rajumsys: yeah, it's a hack. alternate is if...else if.....several.

PR #480, opened 2018-05-23
FE-341 fix string formatted template not showing in UI

rajumsys:
- The API allows creating template where `content.from` can be string. UI shows this blank as it expects from field to be `{name: <string>, email: <string>}`. 

I'm converting the string formatted email to this format. Converting it when
- storing to store
- getting preview object from content preview endpoint

notes:
- this will reformat the stored from address when template is saved. 
- looks like API also allows setting email address like `My Name <email@domain.com>`. But this isn't mentioned in the [API Spec](https://developers.sparkpost.com/api/templates.html#header-content-attributes). So didn't handle it.

### DIFF
--- a/src/pages/templates/containers/EditPage.container.js
+++ b/src/pages/templates/containers/EditPage.container.js
@@ -5,7 +5,7 @@ import { reduxForm } from 'redux-form';
 import { getDraft, getPublished, update, deleteTemplate, publish, getTestData, setTestData } from 'src/actions/templates';
 import { showAlert } from 'src/actions/globalAlert';
 import { hasGrants } from 'src/helpers/conditions';
-import { selectTemplateById, selectTemplateTestData, normalizeFromAddress } from 'src/selectors/templates';
+import { selectTemplateById, selectTemplateTestData } from 'src/selectors/templates';
 import { selectSubaccountIdFromQuery, hasSubaccounts, selectSubaccountFromQuery } from 'src/selectors/subaccounts';
 
 import EditPage from '../EditPage';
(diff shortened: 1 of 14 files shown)


src/pages/templates/containers/EditPage.container.js
```diff
@@ -14,7 +14,7 @@ const FORM_NAME = 'templateEdit';
 
 const mapStateToProps = (state, props) => {
   const template = selectTemplateById(state, props);
-  const values = template.draft || template.published; // For templates with published but no draft, pull in published values
+  const values = normalizeFromAddress(template.draft || template.published); // For templates with published but no draft, pull in published values
```
jonambas: Will this need to be added to the other containers? Published, and Preview (Draft/Published)?

Instead of adding the transformer here, it might be easier to wrap `selectDraftTemplate` and `selectPublishedTemplate` in selectors instead

src/selectors/templates.js
```diff
@@ -40,3 +40,34 @@ export const selectDomainsBySubaccount = createSelector(
...
+ * Converts string formatted from address to json structure.
+ * Example:
+ * before: content.from = 'me@domain.com'
+ * after: content.from = { name: null, email: 'me@domain.com'}
+ * @param template
+ * @returns {*} null or template with reformatted from field
+ */
+export const normalizeFromAddress = (template) => {
+  if (_.isEmpty(template)) {
+    return template; //just not to change whatever value it is
```
jonambas: What is this case for? Isn't this a required field? Don't think this will ever be an empty object

jonambas: Oh nvm, this is checking if template itself is empty

PR #488, opened 2018-05-31
FE-338 Add Notification Center to top nav bar

jonambas:
Matchbox PR to fix onClose behaviour: https://github.com/SparkPost/matchbox/pull/131

---

screen shot 2018-06-04 at 4 07 45 pm


src/actions/tests/notifications.test.js
```diff
@@ -3,6 +3,7 @@ import * as actions from '../notifications';
 import * as userActions from 'src/actions/currentUser';
 
 jest.mock('src/actions/currentUser');
+global.Date = jest.fn();
```
jasonrhodes: Let's do this so we still see the date appear in the snapshot:

```js
const mockNow = new Date('2018-05-15T12:00:00.000Z');
global.Date = jest.fn().mockImplementation(function () { return mockNow; });
```

jonambas:
If I don't get to deploy this today, make sure to bump matchbox out of a prerelease before merging :D

PR #489, opened 2018-06-01
FE-374 hide free plan if someone is on free1 plan

rajumsys:
just fyi, this selector is also used in onboarding flow

- not sure why that big diff for package-lock.json! doing `npm install` on `master` also update package-lock.json. could be npm version? `node 9.0.0, npm 5.5.1`.  update: updated npm to 6.1.0 and that diff gone.

### DIFF
--- a/src/selectors/accountBillingInfo.js
+++ b/src/selectors/accountBillingInfo.js
@@ -7,6 +7,7 @@ const pendingSubscriptionSelector = (state) => state.account.pending_subscriptio
 const plansSelector = (state) => state.billing.plans || [];
 const accountBillingSelector = (state) => state.account.billing;
 const isFreeLegacyPlanSelector = (state) => onPlan('ccfree1')(state);
+const isFreeLegacyFree1PlanSelector = (state) => onPlan('free1')(state);
 
 export const currentSubscriptionSelector = (state) => state.account.subscription;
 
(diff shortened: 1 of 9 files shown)


src/selectors/accountBillingInfo.js
```diff
@@ -7,6 +7,7 @@ const pendingSubscriptionSelector = (state) => state.account.pending_subscriptio
 const plansSelector = (state) => state.billing.plans || [];
 const accountBillingSelector = (state) => state.account.billing;
 const isFreeLegacyPlanSelector = (state) => onPlan('ccfree1')(state);
+const isFreeLegacyFree1PlanSelector = (state) => onPlan('free1')(state);
```
jasonrhodes: This is so confusing, but not your fault. Can we remove both of these since each is only used once in this file and instead use `onPlan` directly?

src/selectors/accountBillingInfo.js
```diff
@@ -76,8 +77,11 @@ export const selectAvailablePlans = createSelector(
 );
 
 export const selectVisiblePlans = createSelector(
-  [selectAvailablePlans],
-  (plans) => plans.filter(({ status }) => status === 'public')
+  [selectAvailablePlans, isFreeLegacyFree1PlanSelector],
```
jasonrhodes: So with my above comment, this would become:
```js
  [selectAvailablePlans, onPlan('free1')]
  (plans, isLegacyFree1) => ...etc
```

src/selectors/accountBillingInfo.js
jasonrhodes: So these lines: https://github.com/SparkPost/2web2ui/blob/6bfa3c11197efe3e37c929aa91c5c3cc8d3c55e5/src/selectors/accountBillingInfo.js#L49-L50

would become:
```js
  [currentPlanSelector, accountBillingSelector, onPlan('ccfree1')],
  (currentPlan, accountBilling, isCcFree) => (
```

PR #490, opened 2018-06-01
FE-372 Move support into account nav bar

bkemper:


ewandennis:
UX feedback:
 - Our support feature is no longer visible at a glance. That _looks_ to me like we're trying to discourage giving support. Since we are also removing the old web form and published email addresses, I might also empathise with customers who feel that way. Either we have to draw attention to the new location or we need to make it more visible, perhaps with a separate help icon on the top nav.
 - The support modal has no cancel button. Its clear to me how to dismiss it, but I'm not so sure about world+dog. Can we validate that approach before we deploy?
 - Do we really need a whole tab just for the phone number?
 - The phone number should still include the timezone. I would go longer form than `ET` though maybe: `Eastern` or `Eastern time`.

bkemper:
- I agree that it is a little hidden.  I don't think that was the intention.  I like the idea of a dedicated icon until users are use to it being a dropdown item.  Also, it is important to point out that the icon was accessible during on-boarding and will not be now.
- Okay.  I'll get more feedback.
- It doesn't have to be a whole tab, but I don't think it fits well on the other two tabs.  Let me think on it.
- 👍

jonambas:
I think its reasonable to have the support hidden inside the support dropdown - many apps do this.

We don't have a way to call out new features popover style (kinda like slack), but I think that would be the best way to introduce the change. It would be as simple as adding another popover around the dropdown that is defaulted to open with a message like -

> Looking to submit a support ticket? We've moved the form into this dropdown

.. but would have to do something similar to the notification center so that it doesn't show up again if already dismissed.

src/components/support/Support.js
```diff
@@ -33,63 +36,45 @@ export class Support extends Component {
...
+    this.props.openSupportPanel({ view });
   }
 
   render() {
-    const { loggedIn, showPanel, showTicketForm } = this.props;
+    const { authorizedToSubmitSupportTickets, closeSupportPanel, currentSupportView, loggedIn, showSupportPanel } = this.props;
+    const tabs = [
+      { content: 'Search Help', onClick: this.openSupportPanel('docs'), view: 'docs' },
+      { content: 'Submit A Ticket', onClick: this.openSupportPanel('ticket'), view: 'ticket' },
+      { content: 'Contact Us', onClick: this.openSupportPanel('contact'), view: 'contact' }
```
ewandennis: This should not show when the `entitledToPhoneSupport()` selector returns falsey. We can also probably simplify the `SupportContact` component since it doesn't need to handle that same case any more.

src/components/support/Support.js
```diff
@@ -33,63 +36,45 @@ export class Support extends Component {
...
-              {showPanel && (showTicketForm
-                ? <SupportForm onCancel={this.toggleForm} onContinue={this.toggleForm} />
-                : <SearchPanel toggleForm={this.toggleForm} />)}
-            </Popover>
-          </AccessControl>
-        </div>
-      </Portal>
+      <AccessControl condition={() => true}>
+        <Modal open={showSupportPanel} onClose={closeSupportPanel}>
+          {authorizedToSubmitSupportTickets && (
```
ewandennis: As discussed, there's a little work needed here to handle the access control to the `SupportContact` tab.

src/components/support/components/NoIssues.js
```diff
@@ -1,13 +1,11 @@
...
 
 const NoIssues = ({ onCancel }) => (
-  <div className={styles.NoIssues}>
-    <div className={styles.Message}>
-      <h6>Sorry, you are not authorized to submit a support ticket.</h6>
-      <Button flat color='orange' onClick={onCancel}>Search support articles</Button>
-    </div>
+  <div className={styles.SupportContainer}>
+    <h6>Sorry, you are not authorized to submit a support ticket.</h6>
+    <Button flat color='orange' onClick={onCancel}>Search support articles</Button>
```
ewandennis: Can we repeat the upgrade CTA from the `SupportForm` here?

src/components/support/components/SupportContact.js
```diff
@@ -0,0 +1,24 @@
...
+import styles from '../Support.module.scss';
+
+export const SupportContact = ({ entitledToPhoneSupport }) => {
+  if (!entitledToPhoneSupport) {
+    return null;
+  }
+
+  return (
+    <div className={styles.SupportContainer}>
+      <h6>We are available Monday through Friday, 9am to 8pm.</h6>
```
ewandennis: + `Eastern time` :)

src/components/support/components/SupportContact.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import { UnstyledLink } from '@sparkpost/matchbox';
+import { entitledToPhoneSupport } from 'src/selectors/support';
+import styles from '../Support.module.scss';
+
+export const SupportContact = ({ entitledToPhoneSupport }) => {
+  if (!entitledToPhoneSupport) {
```
ewandennis: We likely don't need this logic or state any more since the enclosing component handles it.

src/actions/tests/support.test.js
```diff
@@ -11,26 +11,12 @@ describe('Action Creator: Support', () => {
...
-    it('should clear ticket form', () => {
-      expect(support.clearTicketForm()).toMatchSnapshot();
-    });
-  });
-
-  describe('toggleTicketForm', () => {
-    it('should toggle ticket form', () => {
-      expect(support.toggleTicketForm()).toMatchSnapshot();
-    });
+  it('should close support panel', () => {
```
ewandennis: Add explicit coverage for `openSupportPanel()` ?

src/components/support/Support.js
```diff
@@ -33,63 +36,45 @@ export class Support extends Component {
...
-    // handling reseting doc search when closed
-    if (!showPanel && showTicketForm) {
-      toggleTicketForm();
-    }
-    toggleSupportPanel();
-  }
-
-  toggleForm = () => {
-    this.props.toggleTicketForm();
+  openSupportPanel = (view) => () => {
```
ewandennis: Explicit coverage for this would be great. No need to test the `Tabs.onClick` capability, just that this fires the correct action.

src/components/support/components/SupportContact.js
bkemper: Removed.

src/components/support/Support.js
bkemper: Agreed.  Fixed.

src/components/support/components/SupportContact.js
bkemper: Fixed.

src/components/support/Support.js
bkemper: Done.

src/actions/tests/support.test.js
bkemper: Done.

src/components/support/components/NoIssues.js
bkemper: I believe only reporting users will see this message.  Does it make sense to show an upgrade message to them?

bkemper:
@ewandennis This is ready for final review.

ewandennis:
lgtm. Some foibles:

 - Close button is good. However, the click-anywhere-to-close behaviour is a bit risky now. I can half-fill the form, accidentally click or hit escape and my form contents are lost.
 - The mobile view hides both the notification and help icons. We should keep help visible at all times.
 - The new `visible` prop on `Tab` is being passed down the tree all the way to the `<a>` under `UnstyledLink`. React is throwing an exception explaining that. `{...rest}` appears to be tripping us up there.

Depending on level of effort, we might want to file improvements for these points.

PR #492, opened 2018-06-04
FE-376 Fix from email address input validator in templates

jonambas:


src/pages/templates/components/validation.js
```diff
@@ -3,6 +3,10 @@ import _ from 'lodash';
 
 const ID_ALLOWED_CHARS = 'a-z0-9_-';
 
+function looseSubstitution(value) {
+  return value.includes('{{') && value.includes('}}') ? undefined : 'Substitution syntax error';
+}
+
 function substitution(value) {
```
rajumsys: we can delete this right?

jonambas: Yep, think so

PR #502, opened 2018-06-14
FE-43 Template stüf

jonambas:
Matchbox Pr: https://github.com/SparkPost/matchbox/pull/135

---


jasonrhodes:
After discussing in Slack, I recommend we do the following changes for now:
* [ ] Remove the "Edit" button from the right side and go with just the icon button
* [ ] The action list for the icon button should be based on template status: Published and Draft templates get "Edit" and "Preview", Published with Unpublished Changes get "Edit Saved Draft", "Preview Saved Draft", "Preview Published"
* [ ] The status tag will add an icon or asterisk for templates in "Published with Unpublished Changes" state

Things we are punting on that should be pulled into a new Templates UI epic:
* Add actions to the list for templates in "Published with Unpublished Changes" state:
  * "Discard Unpublished Changes" (we can just load up the published values and save them over the existing draft, then re-publish it)
  * "View Diff" (load published and draft side by side with diff view, including the data and html/text versions)
* For draft templates, consider a "Publish" action
* For published tempaltes, consider an "Unpublish" action (would require API changes, see TR-297)
* Move all substitution test data management to the preview screen
* Once we have a diff view, the "view published" view and route can be completely removed (unnecessary)
* Align all language and options present in the action lists in the list view with the secondary actions in the edit and preview views

jasonrhodes:
Two last things:

* For published templates without unpublished changes, the main template title link should go to the published URL (it's confusing to me when you click that link and end up on a page that says "(DRAFT)" at the top, with a "PUBLISH TEMPLATE" button in the top right.
* When on the "view published" page for now, instead of "View Draft" can we change that wording to just say "Edit"?

PR #513, opened 2018-06-25
FE-375 Update Jest deps

jonambas:
Jest changes
- `<React.Fragment>` now rendered in snapshots
- `<Context>` now rendered in snapshots
- `mockFn.mock.results` returned in snapshots for some reason

### DIFF
--- a/src/actions/tests/__snapshots__/support.test.js.snap
+++ b/src/actions/tests/__snapshots__/support.test.js.snap
@@ -27,6 +27,7 @@ exports[`Action Creator: Support openSupportTicketForm should open support panel
       },
     ],
   ],
+  "results": undefined,
 }
 `;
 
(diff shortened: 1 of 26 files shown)


src/actions/tests/__snapshots__/support.test.js.snap
```diff
@@ -27,6 +27,7 @@ exports[`Action Creator: Support openSupportTicketForm should open support panel
       },
     ],
   ],
+  "results": undefined,
```
rajumsys: this is supposed to be `undefined`?

rajumsys: ok, i see it. there are two tests of same name (but this one is under different suite).

jonambas: Yea, not sure theres anything we can do about this. This is just how jest.fn is being snapshotted

PR #516, opened 2018-06-26
FE-395:  Handle GET failures correctly in Subaccounts and Sending Domains

bkemper:
**How to test?**

- For subaccount page, open a link to a subaccount that does not exist (e.g. http://app.sparkpost.test/account/subaccounts/0).  The expected result is to be redirected to the subaccount list page with a warning.
- For sending domain page, open a link to a sending domain that does not exist (e.g. http://app.sparkpost.test/account/sending-domains/edit/0).  The expected result is to be redirected to the sending domains list page with a warning.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,12 @@
 language: node_js
 node_js:
   - "8"
-install: npm install -dd # change to npm ci when travis runs npm 5.7+ (currently 5.6)
-script: npm run test-ci && npm run build && npm run coveralls
+cache:
+  directories:
+    - build # shows bundle size diff per file/chunk
+    - "$HOME/.npm" # to avoid fetching modules from npm every time
 before_install: npm install -g greenkeeper-lockfile@1
+install: npm install --loglevel http # change to npm ci when travis runs npm 5.7+ (currently 5.6)
 before_script: greenkeeper-lockfile-update
+script: npm run test-ci && npm run build && npm run coveralls
 after_script: greenkeeper-lockfile-upload
--- a/src/pages/subaccounts/tests/__snapshots__/DetailsPage.test.js.snap
+++ b/src/pages/subaccounts/tests/__snapshots__/DetailsPage.test.js.snap
@@ -1,102 +1,18 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`edit path base props 1`] = `
-<Page
-  breadcrumbAction={
+exports[`DetailsPage should redirect and alert on error 1`] = `
+<Connect(RedirectAndAlert)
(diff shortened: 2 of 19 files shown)


src/pages/subaccounts/tests/__snapshots__/DetailsPage.test.js.snap
```diff
@@ -135,64 +107,26 @@ exports[`keys path base props 1`] = `
   <Switch>
     <Route
       exact={true}
-      path="/account/subaccounts/123"
+      path="/account/subaccounts/:id"
```
jonambas: Hm what happened here

.travis.yml
```diff
@@ -1,8 +1,12 @@
 language: node_js
 node_js:
   - "8"
-install: npm install -dd # change to npm ci when travis runs npm 5.7+ (currently 5.6)
-script: npm run test-ci && npm run build && npm run coveralls
+cache:
+  directories:
+    - build # shows bundle size diff per file/chunk
+    - "$HOME/.npm" # to avoid fetching modules from npm every time
```
jonambas: When we update deps, will we need to clear the cache on those branches?

src/pages/subaccounts/tests/__snapshots__/DetailsPage.test.js.snap
```diff
@@ -39,64 +51,24 @@ exports[`edit path base props 1`] = `
...
       render={[Function]}
```
jonambas: Do you think we need to add tests to cover these render functions?

.travis.yml
bkemper: Nope, it should install the latest dependency based on package-lock

src/pages/subaccounts/tests/__snapshots__/DetailsPage.test.js.snap
```diff
@@ -234,22 +165,24 @@ exports[`on create scenario should show api key banner on new key 1`] = `
   <Switch>
     <Route
       exact={true}
-      path="/account/subaccounts/123"
+      path="/account/subaccounts/:id"
```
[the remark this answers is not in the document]
bkemper: As discussed, I switch the Route path from a dynamic value set with the current location to a static value.

src/pages/subaccounts/tests/__snapshots__/DetailsPage.test.js.snap
bkemper: After a couple of attempts, I couldn't find an easy way to unit test this render prop.  I think this would be a good candidate for some integration test coverage.

PR #518, opened 2018-06-27
FE-382 - add new support category, remove technical support category

rajumsys:
Notes:
- We can't ship this until SC cutover
- As per ticket, we had to remove Technical Support. That will require updating, at least, https://www.sparkpost.com/docs/getting-started/how-to-get-help/. Support docs PR https://github.com/SparkPost/support-docs/pull/261
- I think it's worth renaming this part to something appropriate as that category is gone
  > Additional technical support is available on paid plans. Upgrade now.
 
  to like,
  > Additional support categories are available on paid plans. Upgrade now.

### DIFF
--- a/src/config/supportIssues.js
+++ b/src/config/supportIssues.js
@@ -127,7 +127,7 @@ const supportIssues = [
 const augmentIssuesWithConditions = function () {
   return _.map(supportIssues, (supportIssue) => ({
     ...supportIssue,
-    condition: idConditionsMap[supportIssue.id] || defaultCondition
+    condition: supportIssue.condition || idConditionsMap[supportIssue.id] || defaultCondition
   }));
 };
 
--- a/src/config/supportIssues.js
+++ b/src/config/supportIssues.js
@@ -1,3 +1,5 @@
+import _ from 'lodash';
+
 import { hasOnlineSupport, hasStatus, hasStatusReasonCategory, isSuspendedForBilling } from 'src/helpers/conditions/account';
 import { all, not, any } from 'src/helpers/conditions';
 import { isAdmin } from 'src/helpers/conditions/user';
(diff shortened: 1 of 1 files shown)


src/config/supportIssues.js
```diff
@@ -30,12 +30,68 @@ const SUPPORT = 'Support';
  */
 const supportIssues = [
   {
-    id: 'technical_errors',
-    label: 'Technical errors',
+    id: 'ui_errors',
+    label: 'UI errors',
     messageLabel: 'Tell us more about your issue',
     type: ERRORS,
     condition: all(hasOnlineSupport, hasStatus('active'))
```
jasonrhodes: We should move all of these conditions into a hash (can just be at the top of this file) and name each "type" of condition, then apply it to each issue below. That makes it a bit more intuitive for how to add new conditions later, and we can eventually look at the issue config being self-serve for support somehow.

rajumsys: ok, that sounds good and trivial than what I thought what i heard during standup. i'll address it.

src/config/supportIssues.js
```diff
@@ -19,58 +30,97 @@ const SUPPORT = 'Support';
...
+    label: 'Reporting & event issues',
+    type: SUPPORT
+  },
+  {
+    id: 'blacklisting',
+    label: 'IP blacklisting',
+    type: COMPLIANCE
+  },
+  {
+    id: 'product/support_feedback',
```
jonambas: Are we sure we want to restrict this to online support only?

rajumsys: No, i'm not sure. I've went by same permission for Type. The ticket doesn't say anything about restriction. If you you another group should get it, let's check with Bryan and i can update it.

PR #519, opened 2018-06-27
FE-411 Fixes support ticket visibility issues

jasonrhodes:
The main fix for this problem is:
* Use our existing support issue type selector in [the support action](https://github.com/SparkPost/2web2ui/compare/master...hotfix/FE-411#diff-b4c7380ea449443f42eadb932607da71) -- the selector decides which issues a customer has access to see, so that the query string value can only be applied if the issue type is in that visible list.
* This means the Support component must wait until the access control state is loaded before it can initialize
  * The Support component's `render` method already wrapped all of its contents with the `<AccessControl>` component, which forces those children to wait for the access control state to load before rendering
  * Unfortunately, the support ticket action is called from `componentDidUpdate` which means using `<AccessControl>` in the render method is too late and has no effect
  * For this reason, I changed the default export for the `Support` component so that the whole thing is wrapped in `<AccessControl>` before it ever mounts, which guarantees that all of its contents have full access to access control state before mounting and rendering
  * Unfortunately, the AccessControl component had a small bug where it didn't wait for access control state to be ready unless we were going to redirect, so I fixed that bug
  * Fixing that bug required a small change to the integration tests I recently wrote, to make sure that we set "access control ready" so that those access-controlled fields show up (before they showed up because we weren't waiting for "ready" to be true)


src/components/support/Support.js
```diff
@@ -67,37 +67,40 @@ export class Support extends Component {
...
     return (
-      <AccessControl condition={() => true}>
-        <Modal open={showSupportPanel} onClose={closeSupportPanel} showCloseButton={true}>
-          {visibleTabs.length > 1 && (
-            <Tabs
-              connectBelow={true}
-              selected={visibleTabs.findIndex((tab) => tab.view === currentSupportView)}
-              tabs={visibleTabs.map(({ content, onClick }) => ({ content, onClick }))}
-            />
+      <Modal open={showSupportPanel} onClose={closeSupportPanel} showCloseButton={true}>
```
jasonrhodes: Only change here is that I no longer wrap this stuff in `<AccessControl>` because the whole exported component is now wrapped so that we wait for that state before this component is even mounted.

src/components/support/Support.js
```diff
@@ -67,37 +67,40 @@ export class Support extends Component {
...
-        </Modal>
-      </AccessControl>
+        </Panel>
+      </Modal>
     );
   }
 }
 
 const mapStateToProps = (state) => ({
+  ready: state.accessControlReady,
```
jasonrhodes: Need to remove this, was debugging ^

src/components/support/Support.js
```diff
@@ -100,4 +98,8 @@ const mapStateToProps = (state) => ({
   showSupportPanel: state.support.showPanel
 });
 
-export default withRouter(connect(mapStateToProps, supportActions)(Support));
+export const ConnectedSupport = withRouter(connect(mapStateToProps, supportActions)(Support));
+
+export default function SupportWithAccessControlLoaded(props) {
+  return <AccessControl><ConnectedSupport {...props} /></AccessControl>;
```
jasonrhodes: This guarantees that this component won't mount until access control state is loaded in the app. (This is done for most of our other components by nature of them being loaded via the `ProtectedRoute` component, so this only appears as an issue for components that load outside of that flow)

src/actions/support.js
```diff
@@ -57,9 +57,13 @@ export function createTicket({ issueType, ...data }) {
...
 export function openSupportTicketForm({ issueId, message } = {}) {
-  const issue = _.find(supportIssues, { id: issueId });
 
-  return (dispatch) => {
+  return (dispatch, getState) => {
+    // this action assumes access control is ready when called, because
+    // selectSupportIssues uses AC state to decide on what issues are visible
+    const visibleIssues = selectSupportIssues(getState());
+    const issue = _.find(visibleIssues, { id: issueId });
+
```
jasonrhodes: This is the KEY change that solved the bug for support issues. You can test it by logging in and visiting a url with `supportTicket=true&supportIssue=<any_id_you_dont_have_access_to>`, which works for suspended users, free users, reporting role users, etc. each with their own list of issue types they aren't allowed to see. 

Before that would result in the first item in the list looking like it was selected but blowing up on submit if you never touched the select input. Now, it should work exactly like using a totally garbage issue type in the query string, which defaults to "Please select an issue"

rajumsys: This seems fine. But I feel like (may be in another ticket) ideal UX will be showing an error that customer don't have access to the selected issue type.

src/actions/tests/support.test.js
```diff
@@ -28,20 +32,26 @@ describe('Action Creator: Support', () => {
...
     it('should open support panel and set support ticket message', () => {
-      support.openSupportTicketForm({ message: 'test' })(dispatchMock);
+      support.openSupportTicketForm({ message: 'test' })(dispatchMock, getStateMock);
       expect(formActions.change).toHaveBeenCalledWith('supportForm', 'message', 'test');
       expect(formActions.change).toHaveBeenCalledTimes(1);
     });
 
-    it('should open support panel ane set support ticket issue', () => {
-      support.openSupportTicketForm({ issueId: 'technical_errors' })(dispatchMock);
+    it('should open support panel and set support ticket issue', () => {
```
rajumsys: hehe, i fixed that too.

PR #523, opened 2018-07-03
FE-429 Remove ip pool empty state & small ui improvements

jonambas:
- Removed the empty state from the ip pools list page
- Added the `shouldShowIpPurchaseCTA` selector to the form and edit pages, wrapped around the purchase links/copy
- Some clean up

### DIFF
--- a/src/pages/ipPools/components/PoolForm.js
+++ b/src/pages/ipPools/components/PoolForm.js
@@ -56,7 +56,7 @@ export class PoolForm extends Component {
       : null;
 
     return (
-      <React.Fragment>
+      <Fragment>
         <Panel.Section>
           <p>
             Add dedicated IPs to this pool by moving them from their current pool{purchaseCTA}.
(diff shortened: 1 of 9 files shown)


src/pages/ipPools/components/PoolForm.js
```diff
@@ -51,18 +51,16 @@ export class PoolForm extends Component {
...
 
-    // Empty pool
-    if (ips.length === 0) {
-      return <Panel.Section>
-        <p>Add a Dedicated IP to the pool by purchasing from the <UnstyledLink to="/account/billing" component={Link}>billing</UnstyledLink> page.</p>
-      </Panel.Section>;
-    }
-
+    const purchaseCTA = showPurchaseCTA
+      ? <React.Fragment>, or by purchasing new IPs from the <UnstyledLink to="/account/billing" component={Link}>billing page</UnstyledLink></React.Fragment>
```
rajumsys: the link might be on `purchase new IPs`.

jonambas:
@rajumsys I think the problem was that a TAM thought that you had to create the default pool even though it already exists by default. The idea was to always show the collection so they could see that default is already there

what you suggested I think is what was in place before already
```js
const isEmptyState = ipPools.length === 1 && ipPools[0].ips.length === 0;
```

PR #524, opened 2018-07-03
FE-428 Replace support ticket form with message for heroku users

jonambas:
Two things:
- Not sure how to test a heroku user locally
- Is this the right url?

screen shot 2018-07-03 at 5 15 02 pm


jonambas:
@bkemper I thought about that at first, but the tab would be hidden

PR #526, opened 2018-07-04
FE-430: be more careful with event handler arguments in Suppression Detail component

ewandennis:
- Add thunks from event handlers to suppression search `Results.toggleDeleteModal`
 - Also fixed some smart quotes in critical.scss

### DIFF
--- a/src/pages/suppressions/components/Results.js
+++ b/src/pages/suppressions/components/Results.js
@@ -93,7 +93,12 @@ export class Results extends Component {
     const { subaccounts, hasSubaccounts } = this.props;
 
     return (
-      <Detail suppression={data} open={open} onCancel={this.toggleDetailModal} subaccounts={subaccounts} hasSubaccounts={hasSubaccounts} />
+      <Detail
+        suppression={data}
+        open={open}
+        onCancel={() => this.toggleDetailModal(data)}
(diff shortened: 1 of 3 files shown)


src/pages/suppressions/components/Results.js
```diff
@@ -93,7 +93,12 @@ export class Results extends Component {
     const { subaccounts, hasSubaccounts } = this.props;
 
     return (
-      <Detail suppression={data} open={open} onCancel={this.toggleDetailModal} subaccounts={subaccounts} hasSubaccounts={hasSubaccounts} />
+      <Detail
+        suppression={data}
+        open={open}
+        onCancel={() => this.toggleDetailModal(data)}
```
jonambas: I don't think you need to pass `data` here

src/pages/suppressions/components/Results.js
```diff
@@ -128,7 +133,7 @@ export class Results extends Component {
       content={<p>This can not be undone!</p>}
       isPending={deleting}
       onConfirm={this.deleteSuppression}
-      onCancel={this.toggleDeleteModal}
+      onCancel={() => this.toggleDeleteModal()}
```
jonambas: 👍

PR #527, opened 2018-07-05
Added a non-engineer how-to on adding notifications

ewandennis:


ewandennis:
Merging into another hotfix.

PR #532, opened 2018-07-11
FE-12 Refactor and relayout webhook events

jasonrhodes:
Main refactor of webhooks is done. Now, when choosing the "Select individual events" radio button for a webhook, the events are laid out in ABC order, ungrouped and in a grid, with documentation-driven descriptions beneath each. See:

screenshot 2018-07-10 22 52 24

Code has been immensely simplified to show events, create webhooks, display existing webhooks, and edit existing webhooks. Integration tests were added for the creation flow, as well, which can be run with `npm test __integration__/webhooks`. To see how they were running before the refactor, checkout the branch called `webhook-integration-tests-prerefactor` and run the same tests to compare the output. The events may be in slightly different order but otherwise the snapshot output should be the same.


src/pages/webhooks/components/WebhookForm.js
```diff
@@ -1,55 +1,94 @@
...
+    return <OAuth2Fields />;
+  }
+  return null;
+}
+
+
+
+export class WebhookForm extends Component {
+
+  // componentDidUpdate(prevProps) {
```
jasonrhodes: Remove commented code

src/pages/webhooks/components/WebhookForm.js
```diff
@@ -1,55 +1,94 @@
...
+        <Panel.Section>
+          <Button submit primary disabled={disabled}>{submitText}</Button>
+        </Panel.Section>
+      </form>
+    );
+  }
+}
 
 
 const mapStateToProps = (state, props) => {
```
jasonrhodes: This function needs to be tested somehow.

src/pages/webhooks/CreatePage.js
```diff
@@ -118,17 +91,17 @@ export class WebhooksCreate extends Component {
     return (
       <Page title='Create Webhook' breadcrumbAction={{ content: 'Webhooks', Component: Link, to: '/webhooks' }} >
         <Panel>
-          <WebhookForm eventsTree={eventsTree} newWebhook={true} onSubmit={(values) => this.createWebhook(values, eventsTree)}/>
+          <WebhookForm newWebhook={true} onSubmit={(values) => this.createWebhook(values)}/>
```
bkemper: The `onSubmit` arrow function can be removed.

`onSubmit={(values) => this.createWebhook(values)}` ~> `onSubmit={this.createWebhook}`

src/pages/webhooks/components/EditTab.js
```diff
@@ -1,32 +1,18 @@
...
-    this.buildEventsTree = _.memoize(buildEventsTree);
-    this.getAllEvents = _.memoize(this.getAllEvents);
-  }
 
-  /*
-    Dispatches eventDocs if it isn't set on state already
-  */
   componentDidMount() {
-    if (!this.props.eventDocs) {
+    if (this.props.eventListing.length === 0) {
```
bkemper: Similar logic is in `CreatePage`.  Should we move this logic to the action?

src/pages/webhooks/components/EditTab.js
```diff
@@ -35,40 +21,28 @@ export class EditTab extends Component {
...
-      });
-  }
-
-  /*
-    Makes an array with all possible events from the eventsTree.
-    Memoized in constructor to limit unnecessary work
-  */
-  getAllEvents(eventsTree) {
-    return _.flatten(_.map(eventsTree, ({ events }) => _.map(events, ({ key }) => (key))));
+  update = async (values, webhook, allEvents) => {
```
bkemper: `allEvents` can be removed.

src/selectors/webhooks.js
```diff
@@ -5,6 +5,7 @@ import _ from 'lodash';
 
 const selectBatches = (state) => state.webhooks.batches;
 const formatStatus = (code) => _.inRange(code, 200, 300) ? 'Success' : 'Fail';
+const getCurrentWebhook = (state) => state.webhooks.webhook || {};
```
bkemper: Is it better to default in the reducer or here?

src/selectors/webhooks.js
```diff
@@ -37,3 +38,8 @@ export const selectInitialSubaccountValue = createSelector(
     return _.find(subaccounts, { id: Number(id) });
   }
 );
+
+export const getSelectedEvents = createSelector(
+  [getCurrentWebhook],
+  ({ events = []}) => events.reduce((hash, key) => ({ ...hash, [key]: true }), {})
```
bkemper: Tests?

src/selectors/eventListing.js
```diff
@@ -0,0 +1,45 @@
...
+      key,
+      ...group.events[key]
+    }));
+    return [...accumulated, ...listForGroup];
+  }, []);
+
+  list.sort(({ display_name: a = '' }, { display_name: b = '' }) => (a.toLowerCase() < b.toLowerCase()) ? -1 : 1);
+
+  return list;
+}
```
bkemper: This function needs some tests.

src/selectors/eventListing.js
```diff
@@ -0,0 +1,45 @@
...
+ * and returns a flattened list of event objects, e.g.:
+  [
+    { key: "event1", a: 1, b: 2 },
+    { key: "event2", a: 11, b: 22 },
+    { key: "event3": a: 111, b: 222 },
+    { key: "event4", a: 1111, b: 2222 }
+  ]
+ *
+ */
+export function selectEventListing(state) {
```
bkemper: Should this be a selector to limit the number of times it is run?  Something like...

```
import { createSelector } from 'reselect';

const getDocs = (state) => state.webhooks.docs || {};

export const selectEventListing = createSelector(
  [getDocs],
  (docs) => {
    // ...
  }
});
```

bkemper:
🥇 to @jasonrhodes for integration test work!

src/pages/webhooks/CreatePage.js
jasonrhodes: Almost forgot how we do this but also changed `this.createWebhook` to be an instance method via arrow function and then this worked.

src/pages/webhooks/components/EditTab.js
jasonrhodes: Just gonna remove the gate altogether, it's not like people are creating/updating hundreds of webhooks and slamming that endpoint, I don't think.

src/pages/webhooks/components/EditTab.js
jasonrhodes: I wonder why linting didn't catch this 🤔

src/pages/webhooks/components/EditTab.js
```diff
@@ -35,40 +21,28 @@ export class EditTab extends Component {
...
+      subaccount,
+      ...resolveAuthUpdates(values),
+      name,
+      target,
+      active,
+      events: (eventsRadio === 'all') ? eventKeys : eventKeys.filter((e) => events[e]) // if not "all", choose only keys whose value is "true"
+    });
+
+    showAlert({ type: 'success', message: 'Update Successful' });
+    getWebhook({ id, subaccount });
```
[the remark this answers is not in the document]
jasonrhodes: I don't think so. You may not always want to refetch, and I don't think we want to wait for that fetch to return before showing the "Update Successful" alert, necessarily. Seems ok here.

src/selectors/webhooks.js
jasonrhodes: For this I prefer to default here in case an action sets webhook to `null` or something.

PR #533, opened 2018-07-12
FE-442: Make Chrome translate users less unhappy

ewandennis:
This PR starts to address the [interaction between React and Chrome's page translation feature](https://github.com/facebook/react/issues/11538).

In essence, Chrome replaces text nodes in the DOM with translated equivalents. React expects exclusive access to the DOM so it throws when it next tries to do updates around translated content.

For instance, in the following:

 ```jsx
<h1>{ showIcon && 'This is bare text' }</h1>
```

If we toggle `showIcon` after Chrome translates the text, React tries to remove a now-nonexistent node.

### Issue Detection and Linting
I added an eslint rule to warn about the most common instance of the issue. It detects a bare string inside JSX with at least 1 sibling. For example:

```jsx
const MyThing = () => <div>This is bare text with this sibling: <Sibling /></div>;
```

This doesn't catch variables or expressions that evaluate to bare strings unfortunately.

### Minimal Fixes

The absolute simplest thing is to wrap bare text in `<span>` to insulate React from DOM edits underneath. It's ugly in complex cases though and there are some edges where it doesn't work.

### Outstanding Issues
 - The simple fix is ugly and, if applied as an autofix, would cruft up a lot of code unnecessarily
 - Complex cases are not auto-detected (think `<p>{ thisMightBeAString }</p` or `<p>{ funcWhichCouldReturnWhateverOrNothing() }</p>`)
 - Current lint rule warns unnecessarily on cases where the content won't change such as config and props: `<p>{ fixedValueProp && 'Some text' }</p>`
 - The PlanPicker renders incorrectly after translation due to some unknown translation-related DOM edit.
 - ChangePlan's component tree is so nested and spread over source files, it's difficult to comprehend.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -56,6 +56,13 @@
         "when": "never",
         "children": true
       }
+    ],
+    "no-restricted-syntax": [
+      "warn",
+      {
+        "message": "☀️ Plz avoid bare strings inside JSX ☀️",
+        "selector": "JSXElement[children.length > 1] > Literal[value=/\\w+/]"
+      }
     ]
   },
   "extends": [
--- a/src/components/billing/PlanPrice.js
+++ b/src/components/billing/PlanPrice.js
@@ -19,16 +19,16 @@ const PlanPrice = ({ plan, showOverage = false, showIp = false, ...rest }) => {
     : null;
 
   return (
-    <span>
+    <span className='notranslate'>
       <span className={styles.MainLabel} {...rest}>
-        <strong>{plan.volume.toLocaleString()}</strong> <span>emails/month</span>
+        <strong>{plan.volume.toLocaleString()}</strong> emails/month
         {priceInfo.price > 0 && <span> at <strong>${priceInfo.price.toLocaleString()}</strong>/{priceInfo.intervalShort}</span>}
-        {plan.isFree && <span>&nbsp;for Free</span>}
+        {plan.isFree && ' for Free'}
       </span>
       <span className={styles.SupportLabel}>
-        {showOverage && <span>{overage}</span>}
+        {showOverage && overage}
 
-        {showIp && <span>{ip}</span>}
+        {showIp && ip}
       </span>
     </span>
   );
--- a/src/components/billing/PlanPrice.js
+++ b/src/components/billing/PlanPrice.js
@@ -21,14 +21,14 @@ const PlanPrice = ({ plan, showOverage = false, showIp = false, ...rest }) => {
   return (
     <span>
       <span className={styles.MainLabel} {...rest}>
-        <strong>{ plan.volume.toLocaleString() }</strong> emails/month
+        <strong>{plan.volume.toLocaleString()}</strong> <span>emails/month</span>
         {priceInfo.price > 0 && <span> at <strong>${priceInfo.price.toLocaleString()}</strong>/{priceInfo.intervalShort}</span>}
-        {plan.isFree && ' for Free'}
+        {plan.isFree && <span>&nbsp;for Free</span>}
       </span>
       <span className={styles.SupportLabel}>
-        {showOverage && overage }
+        {showOverage && <span>{overage}</span>}
 
-        { showIp && ip }
+        {showIp && <span>{ip}</span>}
       </span>
     </span>
   );
--- a/src/pages/users/ListPage.js
+++ b/src/pages/users/ListPage.js
@@ -98,7 +98,7 @@ export class ListPage extends Component {
         onDelete={this.handleDelete}
         onCancel={this.handleCancel}
         open={isOpen}
-        content={<p>User "{name}" will no longer be able to log in or access this SparkPost account and all API keys associated with this user will be immediately deleted.</p>}
+        content={<p><span>User "</span><span>{name}</span><span>" will no longer be able to log in or access this SparkPost account and all API keys associated with this user will be immediately deleted.</span></p>}
         title="Are you sure you want to delete this user?"
       />
     );
@@ -120,7 +120,7 @@ export class ListPage extends Component {
           }}
           defaultSortColumn='name'
         />
-        { this.renderDeleteModal() }
+        {this.renderDeleteModal()}
       </div>
     );
   }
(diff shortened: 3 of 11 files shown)


src/components/billing/PlanPrice.js
```diff
@@ -21,14 +21,14 @@ const PlanPrice = ({ plan, showOverage = false, showIp = false, ...rest }) => {
   return (
     <span>
       <span className={styles.MainLabel} {...rest}>
-        <strong>{ plan.volume.toLocaleString() }</strong> emails/month
+        <strong>{plan.volume.toLocaleString()}</strong> <span>emails/month</span>
         {priceInfo.price > 0 && <span> at <strong>${priceInfo.price.toLocaleString()}</strong>/{priceInfo.intervalShort}</span>}
-        {plan.isFree && ' for Free'}
+        {plan.isFree && <span>&nbsp;for Free</span>}
```
rajumsys: not really reviewing, but just a quick comment: i think we can add a simple `Text` component. This will allow to update this in future. also we can optionally pass the condition to that component. 

`{plan.isFree && <Text>&nbsp;for Free</Text>}` 
or, 
`<Text condition={plan.isFree}>&nbsp;for Free</Text>`

src/pages/users/ListPage.js
```diff
@@ -98,7 +98,7 @@ export class ListPage extends Component {
         onDelete={this.handleDelete}
         onCancel={this.handleCancel}
         open={isOpen}
-        content={<p>User "{name}" will no longer be able to log in or access this SparkPost account and all API keys associated with this user will be immediately deleted.</p>}
+        content={<p><span>User "</span><span>{name}</span><span>" will no longer be able to log in or access this SparkPost account and all API keys associated with this user will be immediately deleted.</span></p>}
```
jasonrhodes: Yikes are these necessary? This all still renders to a single bare string inside a `<p>` tag, right?

jasonrhodes: Looked in codesandbox and I think this would be fine the way it was, just have to avoid a bare text node as a sibling of a non text node, after the JSX has been rendered...

.eslintrc
```diff
@@ -56,6 +56,13 @@
         "when": "never",
         "children": true
       }
+    ],
+    "no-restricted-syntax": [
+      "warn",
```
jasonrhodes: In our flow, is there any value in the "warn" level?

.eslintrc
```diff
@@ -56,6 +56,13 @@
...
+      {
+        "message": "☀️ Plz avoid bare strings inside JSX ☀️",
+        "selector": "JSXElement[children.length > 1] > Literal[value=/\\w+/]"
```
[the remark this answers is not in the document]
ewandennis: Yup, the code snippet in my PR description is wrong. However, the selector this comment is on does detect 1 common true case: a bare string literal in JSX with at least 1 other sibling.

src/pages/users/ListPage.js
ewandennis: Yup, it's ugly. [I was chasing this instance of the issue](https://sentry.io/sparkpost/phoenix-ui-prd/issues/535437352/events/24024618789/). I also repro'd it but no longer can. Clearly this butchery is not necessary but since it's the only patch to that page, I remain suspicious.

Incidentally, thats the sort of markup an autofix might produce :(

ewandennis: Yup, it's ugly. [I was chasing this instance of the issue](https://sentry.io/sparkpost/phoenix-ui-prd/issues/535437352/events/24024618789/). I also repro'd it but no longer can. Clearly this butchery is not necessary but since it's the only patch to that page, I remain suspicious.

Incidentally, thats the sort of markup an autofix might produce :(

.eslintrc
ewandennis: Prevention is tricky.

Throwing an error for every instance of this gives volumes of errors in places that won't exhibit the issue. Those are mostly cases where a condition is set early in component lifecycle and then it doesn't change.

So we could throw or autofix in all those cases just to be safe and lower cognitive load a little at the cost of dirtying up a lot of markup (~120 instances).

Alternately, we could give folks a tool to see the issue if they choose to. My next step was to find out who uses eslint in their editor, where the warning has value.

Given that info, we can decide what makes most sense.

In either case, we still don't have a full solution. I excluded the plan picker from translation entirely because the interaction between our code, Downshift and Chrome translate produced some really unintuitive results. If we can crack that, maybe we can work out a tighter lint rule and autofix.

src/pages/users/ListPage.js
ewandennis: Update: The repro steps include 2 DOM edits:
 - Show the delete modal (and trigger translation)
 - Hit cancel or confirm buttons on the modal, which hides the modal
 - 💥

ewandennis: Update 2: this issue in this case is that `handleCancel` updates `name` _and_ hides the modal. React then seems to update the content prop which contains the offending bare text, then hides the modal.

You can achieve a similar "double DOM edit" effect in your codesandbox by checking then unchecking the checkbox.

ewandennis: How about forcing the text to render into a single DOM node instead?

```jsx
<p>{`User "${name}" will no longer be able to log in or access this SparkPost account and all API keys associated with this user will be immediately deleted.`}</p>
```

jasonrhodes: That last example is different ... because it uses a template string instead of bare JSX?

I had some trouble reliably reproducing this in codesandbox, in a way that makes sense to me compared to what Google Translate is doing, but I see it happening on the user delete modal and I see it's fixed in your branch, so I think you have a handle on what's going on well enough. Whatever you think makes sense here is good with me.

src/components/billing/PlanPrice.js
```diff
@@ -19,16 +19,16 @@ const PlanPrice = ({ plan, showOverage = false, showIp = false, ...rest }) => {
     : null;
 
   return (
-    <span>
+    <span className='notranslate'>
```
jonambas: https://cloud.google.com/translate/faq#technical_questions
`<span translate="no"> </span>` seems to be another option, although I have no preference.

PR #534, opened 2018-07-12
FE-369 Render message events without message_id

rajumsys:
Notes:
- Change the routes to `/messageId/eventId`. It will allow bookmarking exact event.
- Redirects old routes to new routes
- Events without `message_id` will show `<empty>` in `messageId`. Not 100% sold for this; so happy to change to something more appropriate. 
- Show full width for view with MessageHistory.
- exported `mstp` with inspiration from `src/selectors/onboarding.js`.

Todo:

- [x]  Fix test coverage

Update: 07/16 11:30AM
Breaking? 
- It just happened that I clicked View Details for an event (with message_id) but it redirected me back to list page saying event wasn't found. i'll check whether something I broke! But it should not block reviewing.


src/pages/reports/messageEvents/EventPage.js
```diff
@@ -39,29 +33,26 @@ export class EventPage extends Component {
...
   }
 
   render() {
-    const { isMessageHistoryEmpty, loading, messageId, messageHistory, documentation } = this.props;
-    const { selectedEventId } = this.state;
-    const selectedEvent = _.find(messageHistory, (event) => event.event_id === selectedEventId);
+    const { isMessageHistoryEmpty, isOrphanEvent, loading, messageId, messageHistory, documentation, selectedEventId, selectedEvent } = this.props;
 
-    if (isMessageHistoryEmpty) {
+    if ((!isOrphanEvent && isMessageHistoryEmpty) || (isOrphanEvent && !selectedEvent)) {
```
jonambas: `(!isOrphanEvent && isMessageHistoryEmpty) ` will always return true on first load if history hasn't been loaded.

I think you need to move this to cDU, and check `prevprops.loading && !props.loading` before doing this redirect

src/pages/reports/messageEvents/EventPage.js
```diff
@@ -39,29 +33,26 @@ export class EventPage extends Component {
   }
```
jonambas: in the function above, we should check `isOrphanEvent`, no need to call `getMessageHistory` if it is loading a single event. Or, maybe check in componentDidMount

src/pages/reports/messageEvents/EventPage.js
```diff
@@ -39,29 +33,26 @@ export class EventPage extends Component {
...
   componentWillReceiveProps({ selectedEventId }) {
-    if (!this.state.selectedEventId && selectedEventId) {
-      // Saves selected event from location state, defaults to first event in message history
-      this.setState({ selectedEventId });
-
-      // Resets location state
-      this.props.history.replace({ ...this.props.location, state: {}});
+    const { messageId, match } = this.props;
+    if (selectedEventId && messageId && !match.params.eventId) {
+      this.props.history.replace(getDetailsPath(messageId, selectedEventId));
```
jonambas: i dont think `getDetailsPath` is even needed here? 

This is for updating the url when they click the history table - which will never happen if the page is loading a single event?

src/pages/reports/messageEvents/EventPage.js
```diff
@@ -39,29 +33,26 @@ export class EventPage extends Component {
   }
 
   componentWillReceiveProps({ selectedEventId }) {
```
jonambas: cWRP will be deprecated soon and should move this to cDU, if you wanted to make that update in the PR

src/pages/reports/messageEvents/EventPage.js
rajumsys: this is for backward compatibility. if someone has a old url (with messageId only), this will update path to new path pattern (messageId/eventId)

jonambas: ah gotcha, makes sense

src/selectors/messageEvents.js
```diff
@@ -47,5 +47,36 @@ export const selectMessageEventsSearchOptions = createSelector(
...
+  [selectMessageHistory, getEventIdParam],
+  (messageHistory, eventId) => _.find(messageHistory, (event) => event.event_id === eventId)
+);
+
+export const getSelectedEventFromEventsList = createSelector(
+  [getMessageEvents, getEventIdParam],
+  (messageEvents, eventId) => _.find(messageEvents, (event) => event.event_id === eventId)
+);
+
+//whether the event is without a message_id (defaulted to <empty>)
```
jonambas: Update comment here to `_noid_`

PR #510, opened 2018-06-20
FE-368 Track form events

rajumsys:
This WIP, tracks redux form events to Google Analytics via GTM. Currently, GTM and GA tst is configured to accept events. 

It tracks redux form is initialization, form validation error/success, form submission. Initially I've tracked with redux-beacon extension for GTM but seems that's unnecessary (replaced that with `pushEvents`). 

Todos:

- [x] Remove beacon logger
- [x] Track non-redux form
  - I guess we can just call `pushEvents` or make a wrapper like `trackEvent`.
- [x] Validate if these tracking events are measurable 
- [x] Write tests

Any early feedbacks are welcome.

Update:
1) Somewhat based on Bryan's suggestion, I've setting validation `label` like `<formName> JSON.stringify(<validation_errors>)`. We can use `regex` to filter by form name and also then what error actually occurred. 

2) could not test `eventsMap` the way i wanted. I wanted to mock `getFormDefinition` and assert it has been called with correct arg (like 'Initialize' for first item). 

Deployment tasks:
**TST:** 
- [x] Setup GTM (setup variables, triggers and tags in an appropriate folder)
- [ ] Ship Code
- [ ] Live Preview 
- [ ] Publish GTM Changes

**STG:** 
- [ ] Setup GTM (setup variables, triggers and tags in an appropriate folder)
- [ ] Ship Code
- [ ] Live Preview 
- [ ] Publish GTM Changes

**PRD** 
- [ ] Setup GTM (setup variables, triggers and tags in an appropriate folder)
- [ ] Ship Code
- [ ] Live Preview 
- [ ] Publish GTM Changes

**Testing**
- Start Preview Mode in GTM
- Start using any (redux) form normally. The events should be sent to GA via GTM. Check GTM preview/data layer for event details

PR #536, opened 2018-07-16
Fix error: val.replace is not a function

rajumsys:
(json serialize when val is object)

created this PR as hotfix as it's broken on `master` (though it's not deployed to prod yet)

### DIFF
--- a/src/helpers/analytics/utils.js
+++ b/src/helpers/analytics/utils.js
@@ -7,7 +7,7 @@ export function isWhitelistedForm(event) {
 }
 
 export function determineFormValidationState(action) {
-  const fieldErrors = _.map(action.payload.syncErrors, (val, key) => `${key}=${val.replace('=','%3D')}`); //in case any error message contains equal sign (=)
+  const fieldErrors = _.map(action.payload.syncErrors, (val, key) => `${key}=${_.isString(val) ? val.replace('=','%3D') : JSON.stringify(val)}`);
 
   return {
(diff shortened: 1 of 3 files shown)


src/helpers/analytics/utils.js
```diff
@@ -7,7 +7,7 @@ export function isWhitelistedForm(event) {
 }
 
 export function determineFormValidationState(action) {
-  const fieldErrors = _.map(action.payload.syncErrors, (val, key) => `${key}=${val.replace('=','%3D')}`); //in case any error message contains equal sign (=)
+  const fieldErrors = _.map(action.payload.syncErrors, (val, key) => `${key}=${_.isString(val) ? val.replace('=','%3D') : JSON.stringify(val)}`);
```
jasonrhodes: I think, for now, it might be safer just to JSON.stringify the whole syncErrors object, and use `action.payload.syncErrors.length` as the check below between Error vs. Success.

jasonrhodes: If we don't want to switch to JSON stringifying for now, we can silence the logging entirely for now until we can find a safer way to do this kind of transformation on the data. I would recommend looking for a library that is tested and does this kind of object stringification for us, but not in this bugfix ticket.

rajumsys: IMO, one option isn't safer than other (but suggested one definitely simpler than what this PR has). also there isn't a quick away to silent this. so i'm going to revert PR #510 from master until @ewandennis and I can decide a format for this.

PR #539, opened 2018-07-17
FE-407 Deprecated plans cannot submit sending limit increase requests

jasonrhodes:
* updated access control to include plans
* moved around some selectors so that they still work
* added a README to explain a few things about condition helpers
* updated SendMoreCTA so that deprecated plans prompt for upgrade instead of allowing sending limits support request
* updated test

TODO:
* [x] block accounts on deprecated plans from submitting sending request support ticket in support panel

### DIFF
--- a/docs/access-control.md
+++ b/docs/access-control.md
@@ -65,6 +65,33 @@ _Note: Routes default to exact: true because [it makes more sense](https://githu
 
 Condition helpers are functions that return a condition function. Any time you need to use an access control condition, create a helper for that condition scenario. Whenever that returned function is called, it will be passed an object containing whatever [the `accessConditionState` selector](#access-condition-selector) returns, and it should return a boolean that reflects the access decision.
 
+#### Using condition helpers as selectors
+
+IMPORTANT NOTE: These functions are _primarily_ meant to be used with the Access Control components in this repo. They _cannot_ be used directly as selectors without first wrapping them with the `selectCondition` method exported from the accessConditionState selector.
+
+```js
+const isCool = ({ account }) => account.isCool
+
+// Best way, use access control components
+<AccessControl condition={isCool}>
+  {...}
+</AccessControl>
+
+// Good, if you want to use the boolean value outside of access control components
+import { selectCondition } from 'src/selectors/accessConditionState'
+
+const mapStateToProps = (state) => ({
+  isCool: selectCondition(isCool)(state) // will break someday
+})
+
+// Bad, do not do
+const mapStateToProps = (state) => ({
+  isCool: isCool(state) // will break someday
+})
+```
+
+#### Condition helpers should return functions
+
 To avoid repetitiveness and noise, don't make helpers that work like this:
 ```js
 // helpers/conditions/hasGrants.js
--- a/docs/access-control.md
+++ b/docs/access-control.md
@@ -36,6 +36,28 @@ render() {
 
   _Note: The redirect will not happen before the `accessControlReady` flag is true in the global redux store. See the [Access Control Ready](#access-control-ready) section below for more information._
 
+## The `ConditionSwitch` and `Case` components
+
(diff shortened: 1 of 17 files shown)


src/helpers/conditions/README.md
```diff
@@ -0,0 +1,5 @@
+# Access Condition Helpers
+
+IMPORTANT NOTE: These functions are _only_ meant to be used with the Access Control components in this repo. They _cannot_ be used directly as selectors without first providing them with the "accessConditionState" which is provided by src/selectors/accessConditionState (there are a few account-related selectors in that file that make some of the account condition helpers available as selectors, as an example of how to do that).
```
jonambas: Hm have I been using these wrong? What does `without first providing them with the "accessConditionState"` mean?

```js
// Does that mean using this directly in a component wont work?
export const onPlan = (planCode) => ({ accountPlan }) => accountPlan.code === planCode;

// And we should instead wrap it first?
export const onPlan = (planCode) => createSelector(
  [selectAccessConditionState],
  (state) => accountConditions.onPlan(planCode)(state)
);
```

src/actions/accessControl.js
```diff
@@ -1,16 +1,12 @@
...
+import { fetch as fetchAccount, getPlans } from './account';
 import { get as getCurrentUser, getGrants } from './currentUser';
 
 // initialize some state used for access control
 export function initializeAccessControl() {
   return (dispatch) => Promise.all([
     dispatch(fetchAccount()),
-    dispatch(getCurrentUser())
-      .then(({ access_level }) => dispatch(getGrants({ role: access_level })))
+    dispatch(getPlans()),
```
jonambas: Just wondering, does this mean we can get rid of `getPlans` calls in  other pages?

src/helpers/conditions/README.md
jasonrhodes: Depends on what you mean by "directly in a component"—if you use them the way they are designed to be used, e.g. as the `condition` prop for the access control components or in the route config, they work great as-is. 

If we want to use their decision logic in other places, in an imperative way (say, in a selector or other helper), then they need to be provided with the access condition state using that AccessConditionState selector. I've transformed some account condition helpers into selectors inside of that accessConditionState selector file, as an example. How you all want to organize or re-organize that is up to you! :)

Let me know how I can help clear this up, I know it's a little bit messy.

src/actions/accessControl.js
jasonrhodes: Yeah, that should be fine since plans never get updated mid-session.

src/selectors/accessConditionState.js
```diff
@@ -1,11 +1,54 @@
...
+ *
+ * Without this, using conditions as selectors would fail
+ * because things like 'accountPlan' would not be available.
+ *
+ * There may be a better way to organize this. :)
+ */
+
+export const selectCondition = (condition) => createSelector([selectAccessConditionState], condition);
+
+// export const onPlan = (planCode) => createSelector(
```
jonambas: Can remove now, and maybe update the comment above.

docs/access-control.md
```diff
@@ -65,6 +87,33 @@ _Note: Routes default to exact: true because [it makes more sense](https://githu
...
+// Best way, use access control components
+<AccessControl condition={isCool}>
+  {...}
+</AccessControl>
+
+// Good, if you want to use the boolean value outside of access control components
+import { selectCondition } from 'src/selectors/accessConditionState'
+
+const mapStateToProps = (state) => ({
+  isCool: selectCondition(isCool)(state) // will break someday
```
jonambas: Yeah this makes sense to me now - select access condition state & give it a condition helper

src/actions/accessControl.js
```diff
@@ -1,16 +1,12 @@
-import { fetch as fetchAccount } from './account';
+import { fetch as fetchAccount, getPlans } from './account';
```
jonambas: we don't have tests for this file if you wanted to add some

PR #540, opened 2018-07-18
FE-368 InApp Event Tracking

rajumsys:
Based on #510 
+ Check form whitelist before evaluating action data

IMO, there isn't much benefit of removing redux-beacon and using our own middleware. Even if we create our own middleware, we'll receive an action and we've to decide based on that which is what we're doing now base. Moreover, we're getting a few other things free like the events mapping which we'll have to write ourselves and also write tests. Redux-beacon is providing that free. It provides a simple and understandable functionalities and may be helpful if our analytics logics needs extension in future. Similarly, we're using `flat` library but we could write our own flatten logic. So having decoupled library is a good reason itself :). Let me know your thoughts. 

Todos
- [x] Fix rebase issue (certain changes are lost)
- [x] Flatten `syncErrors`


src/helpers/analytics/utils.js
```diff
@@ -9,11 +10,25 @@ export function isValidEvent(event) {
...
+    return null;
+  }
+
+  const flattendSyncErrors = flatten(syncErrors);
+
+  return _.map(flattendSyncErrors, (val, key) => `${key}=${val}`).join(',');
+}
+
 export function determineFormValidationState(action) {
-  const fieldErrors = _.map(action.payload.syncErrors, (val, key) => `${key}=${val.replace('=','%3D')}`); //in case any error message contains equal sign (=)
```
rajumsys: removed replacing `=` as it seems creating more confusion. we'll solve it if that turns out to be a problem.

src/helpers/analytics/index.js
```diff
@@ -27,21 +30,25 @@ export function trackPageview ({ path, title }) {
...
-export function trackEvent ({ category, action, data }) {
+export function trackEvent({ category, action, data }) {
   pushEvent({
     event: category,
     'event-action': action,
     'event-data': data || {}
   });
 }
 
+export function pushEvents(events) {
```
ewandennis: As elementary as this is, we should add test coverage.

src/helpers/analytics/eventsDefinitions.js
```diff
@@ -0,0 +1,25 @@
+import _ from 'lodash';
+import * as utils from './utils';
+
+
+export function getFormDefinition(actionName, value) {
```
ewandennis: Lets be explicit about the purpose of this function. I think it's actually translating a redux action creator into an analytics form interaction event. How about one of these:

 - `actionToFormInteraction`
 - `actionToFormEvent`
 - `translateActionToEvent`

The filename also suggests that we should see event definitions in here. In fact we're seeing a translation function. Maybe we do something similar with the filename:

 - `actionTranslators.js`
 - `actionsToEvents.js`
 - `formInteractions.js`

?

src/helpers/analytics/eventsDefinitions.js
```diff
@@ -0,0 +1,25 @@
+import _ from 'lodash';
+import * as utils from './utils';
```
ewandennis: Can I please import just what we need from `utils`? When we use tree shaking, it works best if each module is explicit about which names it needs from each dependency. I'm not sure if that has an effect here but it's good practice.

Note to self: look into eslint autofixing this type of thing.

src/helpers/analytics/setup.js
```diff
@@ -0,0 +1,8 @@
+import { createMiddleware } from 'redux-beacon';
+import { pushEvents } from './index';
```
ewandennis: I believe you can just say`import { pushEvents } from '.';`

src/helpers/analytics/eventsMap.js
```diff
@@ -0,0 +1,15 @@
...
+
+const eventsMap = {
+  '@@redux-form/INITIALIZE': getFormDefinition('Initialize'),
+  '@@redux-form/START_SUBMIT': getFormDefinition('Submit'),
+  '@@redux-form/SET_SUBMIT_SUCCEEDED': getFormDefinition('Submit Success'),
+  '@@redux-form/SET_SUBMIT_FAILED': getFormDefinition('Submit Failure'),
+  '@@redux-form/UPDATE_SYNC_ERRORS': getFormDefinition(utils.determineFormValidationState)
+};
+
+export default _.mapValues(eventsMap, (val) => ensure(utils.isValidEvent, val));
```
ewandennis: It took me a while to decode this. Can we describe how this composition works in a comment maybe?

src/helpers/analytics/eventsDefinitions.js
```diff
@@ -0,0 +1,25 @@
...
+
+
+export function getFormDefinition(actionName, value) {
+  const labelPath = 'meta.form';
+
+  return (action) => {
+    const formName = _.get(action, labelPath);
+
+    if (!utils.isWhitelistedForm(formName)) {
```
ewandennis: I might say something like this here: 
> We return null to signal that we don't want to emit an event in this case

src/pages/join/components/JoinForm.js
```diff
@@ -32,22 +32,22 @@ const { recaptcha } = config;
...
     this.setState({ reCaptchaReady: true });
   };
 
+  onSubmit = (formValues) => this.props.onSubmit({ ...formValues, recaptcha_response: this.state.recaptcha_response, recaptcha_type: 'invisible' })
+
   recaptchaVerifyCallback = (response) => {
-    const { formValues, onSubmit } = this.props;
     this.state.reCaptchaInstance.reset();
-    onSubmit({ ...formValues, recaptcha_response: response, recaptcha_type: 'invisible' });
+    this.setState({ recaptcha_response: response }, this.props.handleSubmit(this.onSubmit));
```
ewandennis: 👍

src/pages/join/components/JoinForm.js
```diff
@@ -59,11 +59,14 @@ export class JoinForm extends Component {
...
       loading,
       pristine,
-      invalid
+      invalid,
+      submitting
     } = this.props;
 
     const { reCaptchaReady } = this.state;
 
+    const pending = loading || submitting || !reCaptchaReady;
```
ewandennis: 👍

src/helpers/analytics/tests/eventsMap.test.js
```diff
@@ -0,0 +1,17 @@
...
+
+const TEST_CASES = {
+  '@@redux-form/INITIALIZE': { arg: 'Initialize' },
+  '@@redux-form/START_SUBMIT': { arg: 'Submit' },
+  '@@redux-form/SET_SUBMIT_SUCCEEDED': { arg: 'Submit Success' },
+  '@@redux-form/SET_SUBMIT_FAILED': { arg: 'Submit Failure' },
+  '@@redux-form/UPDATE_SYNC_ERRORS': { arg: expect.any(Function) }
+};
+
+cases('Events Map', ({ name, arg }) => {
```
ewandennis: It's not very clear what is being tested here. Can we name the `cases()` clause with a little more detail? Maybe
> Analytics: action to event translation
or something similar.

Do we need to do any negative testing here? We're only hitting action names we know we handle.

src/helpers/analytics/tests/eventsMap.test.js
```diff
@@ -0,0 +1,17 @@
+import cases from 'jest-in-case';
+import eventsMap from '../eventsMap';
+
+jest.mock('../eventsDefinitions');
```
ewandennis: It might actually be clearer what is happening in these tests we didn't mock `eventsDefinitions` and the test cases included full action-creator-to-event test data. It would also act as documentation of the translation process.

src/helpers/analytics/tests/eventsDefinitions.test.js
```diff
@@ -0,0 +1,48 @@
...
+
+  const TEST_CASES = {
+    'Initialize': null,
+    'Submit': { value: 1 },
+    'Submit Failure': {},
+    'Submit Success': {},
+    'Validation Success': {}
+  };
+
+  cases('getFormDefinition', ({ name, value, ...rest }) => {
```
ewandennis: Maybe add some detail to the name of this `cases()` clause:
> getFormDefinition action transformations
?

src/helpers/analytics/tests/utils.test.js
```diff
@@ -0,0 +1,60 @@
...
+      formName: 'loginForm',
+      expectation: true
+    },
+    'false when form name is not in whitelist': {
+      formName: 'fooForm',
+      expectation: false
+    }
+  };
+
+  cases('isWhitelistedForm', ({ name, formName, expectation }) => {
```
ewandennis: There are only 2 cases for this clear true/false function. Do we need to use `cases()`? It can hurt readability so I would only use it where we have lots of very similar tests.

src/helpers/analytics/tests/utils.test.js
```diff
@@ -0,0 +1,60 @@
...
+      event: { event: 'Interactions', category: 'Form', action: 'Initialize', label: 'someForm' },
+      expectation: true
+    },
+    'false when event is falsey': {
+      event: null,
+      expectation: false
+    }
+  };
+
+  cases('isValidEvent', ({ name, event, expectation }) => {
```
ewandennis: Same comment about `cases()` here.

src/helpers/analytics/tests/utils.test.js
```diff
@@ -0,0 +1,60 @@
...
+    expect(utils.isValidEvent(event)).toBe(expectation);
+  }, isValidEventTestCases);
+
+  const determineFormValidationStateTestCases = {
```
ewandennis: This is a better use of `cases()` IMO.

src/helpers/analytics/utils.js
```diff
@@ -0,0 +1,34 @@
...
+  return !_.isEmpty(event);
+}
+
+function formatFieldErrors(action) {
+  const syncErrors = _.get(action, 'payload.syncErrors', null);
+  if (_.isEmpty(syncErrors)) {
+    return null;
+  }
+
+  const flattendSyncErrors = flatten(syncErrors);
```
ewandennis: `s/flattend/flattened` :)

src/helpers/analytics/utils.js
```diff
@@ -0,0 +1,34 @@
...
+
+  return _.map(flattendSyncErrors, (val, key) => `${key}=${val}`).join(',');
```
ewandennis: 👍 Later on, if we're still concerned about `val` containing `=`, we can always put it in quotes. I wouldn't do that until we need to though.

src/helpers/analytics/eventsDefinitions.js
rajumsys: This is actually defining an event that needs to go to gtm.  Action is just it's input. i'm, however, fine renaming to whatever makes sense.

src/helpers/analytics/eventsDefinitions.js
rajumsys: ok

PR #547, opened 2018-07-19
FE-399 - Remove datepicker from Matchbox

jamaral2:
Moved DatePicker (Matchbox) within 2web2ui, named DateSelector


src/components/dateSelector/DateSelector.module.scss
```diff
@@ -0,0 +1,157 @@
+@import '../../styles/config';
```
jonambas: So we want to point to the bundled version from node_modules, since matchbox is included as a dependancy
```
```

No need to actually copy all the source CSS from matchbox

jamaral2: Done and done, I'll push this out with the version # update

package.json
```diff
@@ -47,7 +47,7 @@
   },
   "dependencies": {
```
jonambas: whoa, looks like `react-day-picker` is already installed in our app for some reason.

matchbox used `v7.1.8`, would you be able to update the UI's version to at least that?
Maybe that will fix the styling issues

screen shot 2018-07-19 at 4 34 39 pm

PR #549, opened 2018-07-20
FE-377 - Remove "Leslie Knope" of parks and rec from sign-up page :-(

jamaral2:


src/pages/join/components/JoinForm.js
```diff
@@ -97,7 +97,7 @@ export class JoinForm extends Component {
           validate={[required, email]}
           disabled={!reCaptchaReady || loading}
           autoComplete='username email'
-          placeholder='leslie.knope@pawnee.indiana.state.us.gov'
+          placeholder='username@email.com'
```
jonambas: Can we revert join back? Let's only do auth for now and see if anyone complains

jamaral2:
Ran into a few issues when rebasing this with most recent version of develop. Cutting a new branch/PR to avoid conflicts with commits made previously.

PR #550, opened 2018-07-23
Feature/eo 190

ischer:


src/config/routes.js
```diff
@@ -526,6 +527,14 @@ const routes = [
     title: 'Edit IP Pool',
     supportDocSearch: 'ip pool'
   },
+  {
+    path: '/ab-testing',
+    component: abTesting.ListPage,
+    condition: hasUiOption('abtests'),
```
jasonrhodes: Just FWIW so that everyone knows, if at any point you want this to be a "soft block" and just hide the item from appearing in the nav, this same exact condition can be applied in the `navItems` file. Or keep it here to block it from _both_ the nav as well as direct URL access.

ischer: good to know - thanks!

ischer:
if you get 401s on /api/v1/ab-test, you will need to update your nginx config. i added the following...
```# ab test api
upstream sp-ab-testing-api {
  server app.tst.sparkpost:8888;
#  server localhost:3000;
}``` and this to both `location /api/v1` blocks ```    location /api/v1/ab-test {
      proxy_pass http://sp-ab-testing-api;
      include cors;
    }```

File: src/pages/abTesting/ListPage.js [hunk 2]
```diff
@@ -0,0 +1,151 @@
...
+  renderCollection() {
+    const { abTests } = this.props;
+    return (
+      <TableCollection
+        columns={this.getColumns()}
+        rows={abTests}
+        getRowData={this.getRowData}
+        pagination={true}
+        filterBox={filterBoxConfig}
+        defaultSortColumn='updated_at'
```
jonambas: think we want to add `defaultSortDirection='desc'` here

src/pages/abTesting/ListPage.js
```diff
@@ -0,0 +1,151 @@
...
+    return [
+      <React.Fragment>
+        <p className={styles.Name}>
+          <strong><UnstyledLink to={`/ab-testing/${id}`} component={Link}>{name}</UnstyledLink></strong>
+        </p>
+        <p className={styles.Id}>ID: {id}</p>
+      </React.Fragment>,
+      <StatusTag status={status}/>,
+      <p className={styles.Template}>{template}</p>,
+      <p className={styles.LastUpdated}>{formatDate(updated_at)}</p>,
```
jonambas: maybe `formatDateTime` instead of just `formatDate`

src/pages/abTesting/ListPage.js
```diff
@@ -0,0 +1,151 @@
...
+
+function mapStateToProps({ abTesting, ...state }) {
+  return {
+    abTests: abTesting.list,
+    loading: abTesting.listLoading,
+    error: abTesting.listError
+  };
+}
+
+export default withRouter(connect(mapStateToProps, { listAbTests })(ListPage));
```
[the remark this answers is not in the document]
ischer: 👍 i was wondering specifically what this did - thanks!

PR #556, opened 2018-07-25
FE-451 AB test details page 

jonambas:


src/config/routes.js
```diff
@@ -535,6 +535,14 @@ const routes = [
     title: 'A/B Testing',
     supportDocsSearch: 'A/B test'
   },
+  {
+    path: '/ab-testing/:id/:version?',
```
jonambas: remove `?` option

src/pages/abTesting/ViewMode.js
```diff
@@ -0,0 +1,88 @@
...
+    ];
+  }
+
+  render() {
+    const { breadcrumbAction, test, subaccountId } = this.props;
+    const { id } = this.props.test;
+
+    return (
+      <Page
+        title={id}
```
jonambas: Maybe use `name` instead of ID for page title

src/reducers/abTesting.js
```diff
@@ -40,6 +40,20 @@ export default (state = initialState, action) => {
         }
       };
 
+    /* Get Latest */
+    case 'GET_LATEST_AB_TEST_SUCCESS':
+      return {
+        ...state,
+        detailsLoading: false,
```
jonambas: Remove this

src/actions/tests/abTesting.test.js
```diff
@@ -14,4 +14,19 @@ describe('Action Creator: A/B Testing', () => {
     mockStore.dispatch(abTesting.listAbTests());
     expect(mockStore.getActions()).toMatchSnapshot();
   });
+
+  it('should dispatch a get test action', () => {
```
rajumsys: a test w/o subaccount?

src/config/routes.js
```diff
@@ -535,6 +535,14 @@ const routes = [
     title: 'A/B Testing',
     supportDocsSearch: 'A/B test'
   },
+  {
+    path: '/ab-testing/:id/:version',
```
rajumsys: Is version required parameter?

src/pages/abTesting/DetailsPage.js
```diff
@@ -0,0 +1,101 @@
...
+    if (!prevError && error && _.isEmpty(test)) {
+      this.setState({ shouldRedirect: true });
+    }
+  }
+
+  toggleDelete = (modal) => {
+    this.setState({ showDelete: !this.state.showDelete });
+  }
+
+  handleDelete = () => {
```
rajumsys: to help with reduced coverage, i think you can just delete it. IMO, no value having this placeholder. or leave it inside a comment :)

src/pages/abTesting/DetailsPage.js
```diff
@@ -0,0 +1,101 @@
...
+        <DetailPage {...this.getSharedProps()} />
+        {/* TODO - Add delete modal here */}
+      </Fragment>
+    );
+  }
+}
+
+function mapStateToProps(state, props) {
+  return {
+    test: selectAbTest(state, props),
```
rajumsys: will it be little more explanatory if we say `selectCurrentAbTest`?

src/pages/abTesting/DetailsPage.js
```diff
@@ -0,0 +1,101 @@
...
+    if (this.state.shouldRedirect) {
+      return (
+        <RedirectAndAlert
+          to='/ab-testing'
+          alert={{ type: 'warning', message: 'Unable to load A/B test', details: _.get(error, 'message') }}
+        />
+      );
+    }
+
+    const DetailPage = (status === 'draft' || status === 'scheduled') ? EditMode : ViewMode;
```
rajumsys: Just a thought, will it be easier to navigate to different route based on status? if the test is published we take to different state than if the test is in `draft`. that way, components will be simpler. 

we can still share logics using helpers, selectors etc.

src/pages/abTesting/ViewMode.js
```diff
@@ -0,0 +1,88 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { Page } from '@sparkpost/matchbox';
+import Section from './components/Section';
+import StatusPanel from './components/StatusPanel';
+import { StatusContent, SettingsContent, VariantsContent } from './components/content';
+
+export class ViewMode extends Component {
+
+  getPrimaryAction = () => {
```
rajumsys: this method needs test coverage

src/pages/abTesting/DetailsPage.js
jonambas: That's what I thought initially, but then everything in this component would be copy-pasted - handling error, loading, GET action, etc

rajumsys: there might be some repetition we can't avoid but i think having separate route will be easier and simpler to manage. this view still gonna grow a lot anyway.

jonambas: But we can avoid it, that's what this container does. Everything in this component is shared.

If anything this makes it easier to manage because you won't have separate code in two pages.  THIS component wont grow anymore (just missing the delete action) - everything in the follow up tickets goes into their individual pages. 

Here is the complete list of actions for both modes:
- Edit
  - save as draft
  - schedule
  - update scheduled
  - delete (shared)
- View
  - create new draft from completed
  - override
  - delete (shared)

The pages will end up past the 200 line limit if we don't reuse code

rajumsys: ok then. let's see how it goes. we can revisit if needed.

src/pages/abTesting/tests/DetailsPage.test.js
```diff
@@ -0,0 +1,65 @@
...
+    wrapper.setProps({ test: { ...props.test, status: 'running' }});
+    expect(wrapper.find(ViewMode)).toExist();
+    wrapper.setProps({ test: { ...props.test, status: 'cancelled' }});
+    expect(wrapper.find(ViewMode)).toExist();
+    wrapper.setProps({ test: { ...props.test, status: 'completed' }});
+    expect(wrapper.find(ViewMode)).toExist();
+  });
+
+  it('should get A/B test on mount', () => {
+    expect(props.getAbTest).toHaveBeenCalledWith({ id: props.id, version: props.version, subaccountId: props.subaccountId });
```
rajumsys: `props` isn't in `beforeEach`, so can't totally be sure whether it was ran during mount (after `shallow`) or by any other tests before this test.

src/pages/abTesting/components/fields/StatusFields.js
```diff
@@ -0,0 +1,32 @@
...
+import { TextFieldWrapper } from 'src/components/reduxFormWrappers';
+
+const StatusFields = ({ disabled }) => {
+  return (
+    <Fragment>
+      <Panel sectioned>
+        <Field
+          name='name'
+          component={TextFieldWrapper}
+          helpText='Internal use only'
```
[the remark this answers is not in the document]
jonambas: Yeah, I'll just remove it

src/config/routes.js
rajumsys: sharing url is a very good point.

PR #561, opened 2018-07-26
EO-191 Create A/B test with default screen

ischer:


src/reducers/abTesting.js
```diff
@@ -18,6 +19,28 @@ export default (state = initialState, action) => {
...
+    /* CREATE DRAFT */
+
+    case 'CREATE_AB_TEST_DRAFT_PENDING':
+      return { ...state, abTest: {}, createError: null };
+
+    case 'CREATE_AB_TEST_DRAFT_SUCCESS':
+      return { ...state, abTest: { id: action.payload.id }};
+
+    case 'CREATE_AB_TEST_DRAFT_FAIL':
+      return { ...state, abTest: {}, createError: action.payload };
```
ischer: is it ok that both create & create draft use `createError`? should they be different keys? not sure if i need `abTest`, either. i don't use the ID in the success message (maybe i should?).

also, nothing is currently using the `CREATE` stuff (just draft)... i created it and then realized i needed draft. should i nuke `CREATE` or leave it since it'll eventually be used?

jonambas: I would remove the action/reducer if its not used - I dont think we have plans to include that in the UI?

Also no need to pass through the `abTest` on create since the edit page will GET on mount.

src/pages/abTesting/CreatePage.js
```diff
@@ -0,0 +1,49 @@
...
+  create = (values) => {
+    const { createAbTestDraft, showAlert, history } = this.props;
+    const { id, name, subaccount = {}, default_variant = {}} = values;
+    const default_template = { template_id: default_variant.id };
+    const abTest = { id, name, default_template };
+
+    return createAbTestDraft({ abTest, subaccount: subaccount.id || 0 })
+      .then(() => {
+        showAlert({ type: 'success', message: 'A/B test created' });
+        history.push('/ab-testing');
```
jonambas: You should redirect to `/ab-testing/${id}/${version}${setSubaccountQuery(subaccount_id)}`

(the url i'm using in my ticket)

`version` can probably just be hardcoded to `1`

src/components/typeahead/TemplateTypeahead.js
```diff
@@ -0,0 +1,41 @@
...
+      />
+    );
+  }
+}
+
+function mapStateToProps(state) {
+  const templates = selectTemplates(state);
+
+  return {
+    results: templates,
```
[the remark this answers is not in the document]
jonambas: the template list should also be able to refresh whenever the subaccout typeahead is updated

src/pages/abTesting/components/AbTestCreateForm.js
```diff
@@ -0,0 +1,101 @@
...
+            />
+          </Panel.Section>
+          : null}
+        <Panel.Section>
+          <Field
+            name='default_variant'
+            component={TemplateTypeaheadWrapper}
+            label={'Select this test\'s default template'}
+            placeholder='Type to search'
+            helpText={<span>We will send this template by default when the test is not running. If you need to create a new template, <UnstyledLink to='#'>head over to the templates page</UnstyledLink>.</span>}
```
jonambas: This will need `component={Link}` and the correctly URL

src/pages/abTesting/CreatePage.js
```diff
@@ -0,0 +1,49 @@
...
+
+export class CreatePage extends Component {
+
+  create = (values) => {
+    const { createAbTestDraft, showAlert, history } = this.props;
+    const { id, name, subaccount = {}, default_variant = {}} = values;
+    const default_template = { template_id: default_variant.id };
+    const abTest = { id, name, default_template };
+
+    return createAbTestDraft({ abTest, subaccount: subaccount.id || 0 })
```
jonambas: If you select a subaccount, then clear it, the form fails to submit - `Cannot read property 'id' of null`

src/components/typeahead/TemplateTypeaheadItem.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import classnames from 'classnames/bind';
+import styles from './Typeahead.module.scss';
+classnames.bind(styles);
```
jonambas: No need to `.bind` or even import `classnames` here

src/pages/abTesting/components/tests/AbTestCreateForm.test.js
```diff
@@ -0,0 +1,53 @@
...
+  });
+
+  it('should render subaccount section if subaccounts exist', () => {
+    wrapper.setProps({ hasSubaccounts: true });
+    expect(wrapper.find({ name: 'subaccount' })).toMatchSnapshot();
+  });
+
+  it('should not render subaccount section if subaccounts do not exist', () => {
+    wrapper.setProps({ hasSubaccounts: false });
+    expect(wrapper.find({ name: 'subaccount' })).toMatchSnapshot();
```
jonambas: This can maybe be 

`expect(wrapper.find({ name: 'subaccount' }).html()).toBe(null);`
or `expect(wrapper.find({ name: 'subaccount' })).toHaveLength(0);`

src/pages/abTesting/components/AbTestCreateForm.js
```diff
@@ -0,0 +1,101 @@
...
+            validate={required}
+          />
+          <p></p>
```
jonambas: Looks like an extra `<p>`

src/actions/abTesting.js
```diff
@@ -10,3 +11,31 @@ export function listAbTests() {
     }
   });
 }
+
+export function createAbTest({ abTest, subaccount = null }) {
```
jonambas: Looks like `setSubaccountHeader` has a `null` default, so don't think you need to set those here

src/components/typeahead/TemplateTypeahead.js
```diff
@@ -29,9 +30,10 @@ export class TemplateTypeahead extends Component {
   }
 }
 
-function mapStateToProps(state) {
-  const templates = selectTemplates(state);
-
+function mapStateToProps(state, props) {
+  const templates = hasSubaccounts(state)
```
jonambas: Could we move this into a selector? Just so we can add a test for this logic - we don't typically test mapStateToProps

src/pages/abTesting/components/AbTestCreateForm.js
```diff
@@ -86,7 +86,7 @@ export class AbTestCreateForm extends Component {
 
 AbTestCreateForm.defaultProps = {};
```
jonambas: Can remove this line

PR #562, opened 2018-07-27
FE-436: Track onboarding flow forms events

bkemper:
**Onboarding Pages**
- ChoosePlan - /onboarding/plan
- SendingDomainPage - /onboarding/sending-domain
- ~SmtpOrApiPage - /onboarding/email~ (Only links)
- ~SmtpPage - /onboarding/email/smtp~ (SMTP Detials and skip link)
- ~ApiPage - /onboarding/email/api~ (Copy button, code block, and skip link)

_*** NOTE: All links change the URL which is already being tracked._

**Remaining Tasks**
- [x] Track onboarding forms
- [x] Silence validation events (per Ewan)


rajumsys:
While `initialize` event for `joinPlanForm`  working fine, I don't see  `initialize` for `joinSendingDomainForm` form. Submit and Submit success are fine though

src/helpers/analytics/utils.js
```diff
@@ -19,37 +18,3 @@ export function isWhitelistedForm(formName) {
 export function isValidEvent(event) {
   return !_.isEmpty(event);
 }
-
-/**
- * Flattens a nested object and converts to key=val
- * @param action
- * @returns {*}
- */
-function formatFieldErrors(action) {
```
rajumsys: as you're removing this, we can get rid of `flat` npm package too

rajumsys: nvm, you did that. 👍

bkemper:
> While initialize event for joinPlanForm working fine, I don't see initialize for joinSendingDomainForm form. Submit and Submit success are fine though

The reason the `@@redux-form/INITIALIZE` action isn't trigger on the sending domains form is because the form doesn't set `initialValues` in `mapStateToProps`.  I'm going to push a commit to fix this, but I think we need to rethink this event.  Instead of forcing every form to set `initialValues`, we should use the pageviews.

rajumsys:
> The reason the @@redux-form/INITIALIZE action isn't trigger on the sending domains form is because the form doesn't set initialValues in mapStateToProps. I'm going to push a commit to fix this, but I think we need to rethink this event. Instead of forcing every form to set initialValues, we should use the pageviews.

Irrespective of this event, isn't it a good idea to have initialValues? That draws a clear picture of the form initial values. 

Also, if you like initialize event in Analytics but don't want to set component's `initialValues`; alternative is to trigger redux's  `INITIALIZE` action creator manually. 

re: using this event
i'm open to re-evaluate the usefulness of this event against page view. only issue will happen on page with multiple forms but i guess that's not very common.

PR #564, opened 2018-07-30
FE-437: Track API error events

bkemper:
**How to test?**
1. Open Google Tag Manager and click "Preview"
1. For SparkPost API errors, try to [open a webhook that doesn't exist](http://app.sparkpost.test/webhooks/details/1867aae0-39ae-11e8-9d73-1505f6674400?subaccount=0)
1. For Zuora API errors, try to update payment informaiton, but only enter part of the credit card number.
1. Both events should appear in the GTM preview console


src/helpers/analytics/eventsMapper.js
```diff
@@ -0,0 +1,106 @@
...
+];
+
+export default function eventsMapper(action) {
+  const event = events.find(({ condition }) => condition(action));
+
+  if (!event) {
+    return; // skip
+  }
+
+  return _.mapValues(event.definition, (value) => _.isFunction(value) ? value(action) : value);
```
ewandennis: I might comment here describing what's going on. Something like this?

> Build an event from its definition and the incoming action

src/helpers/analytics/eventsMapper.js
```diff
@@ -0,0 +1,106 @@
...
+ *       // used in GA as more detail for the action
+ *       label: 'Mission Impossible - Fallout'
+ *     }
+ *   }
+ */
+const events = [
+  {
+    condition: all(isTypeLike(/_FAIL$/), isError('SparkpostApiError')),
+    definition: {
+      event: 'ui-api-request',
```
ewandennis: Events are intended for human consumption so lets follow the English title style used on the existing events. How about `event: 'API Request'` ?

src/helpers/analytics/eventsMapper.js
```diff
@@ -0,0 +1,106 @@
...
+      category: 'API',
+      action: getType,
+      label: getErrorMessage
+    }
+  },
+  {
+    condition: all(isTypeLike(/_FAIL$/), isError('ZuoraApiError')),
+    definition: {
+      event: 'ui-api-request',
+      category: 'API',
```
ewandennis: Should we use `category` to filter on event source? `category: 'SparkPost'` or `Zuora` ?

src/helpers/analytics/tests/middleware.test.js
```diff
@@ -0,0 +1,28 @@
...
+describe('Analytics middleware', () => {
+  const action = jest.fn();
+  const next = jest.fn((a) => a); // passthrough
+  const subject = middleware()(next);
+
+  it('pushes an event', () => {
+    const event = jest.fn();
+    eventsMapper.mockReturnValue(event);
+
+    expect(subject(action)).toEqual(action);
```
ewandennis: Part of the redux middleware contract is to chain to `next`. We should explicitly assert that in a test.

src/helpers/analytics/tests/middleware.test.js
```diff
@@ -0,0 +1,28 @@
...
+    expect(eventsMapper).toHaveBeenCalledWith(action);
```
ewandennis: Do we care if our middleware calls `eventsMapper`? That seems like an impl detail. We just need to verify that the data layer ends up with a new event.

src/helpers/analytics/tests/middleware.test.js
```diff
@@ -0,0 +1,28 @@
...
+    expect(analyticsHelpers.pushEvent).toHaveBeenCalledWith(event);
+  });
+
+  it('ignores action', () => {
```
ewandennis: I _think_ this is the negative case so we again don't care how action is processed, just that the data layer does _not_ receive an event after the middleware call.

src/helpers/analytics/middleware.js
```diff
@@ -0,0 +1,14 @@
+import { pushEvent } from '.';
+import eventsMapper from './eventsMapper';
+
+export default function middleware() {
```
ewandennis: I really like how simple this is. ❤️

src/helpers/analytics/tests/eventsMapper.test.js
```diff
@@ -0,0 +1,42 @@
+import cases from 'jest-in-case';
+import eventsMapper from '../eventsMapper';
+
+cases('eventsMapper', ({ name, ...action }) => {
```
ewandennis: We have reduced (hah) `redux-beacon`, `eventsMap.js` and `actionsToEvents.js`to just `eventsMapper.js`. We need to replicate the test coverage too. We should definitely assert white listed form +ve and -ve test cases at least.

src/helpers/analytics/tests/eventsMapper.test.js
```diff
@@ -0,0 +1,42 @@
...
+  expect(eventsMapper(action)).toMatchSnapshot();
+}, {
+  'with untracked action': {},
```
ewandennis: While an empty action technically matches the untracked case, in the spirit of data-driven design, maybe we should also assert that `eventsMapper` DTRT on a fully formed action. Maybe `{ type: 'DO_NOT_TRACK_ME' }` or something?

src/helpers/analytics/eventsMapper.js
```diff
@@ -0,0 +1,106 @@
...
+// condition helpers
+const isError = (name) => (action) => name === getErrorName(action);
+const isType = (type) => (action) => type === getType(action);
+const isTypeLike = (regex) => (action) => regex.test(getType(action));
+const isWhitelistedForm = (action) => ANALYTICS_WHITELISTED_FORMS.includes(getFormName(action));
+
+/**
+ * @example
+ *   {
+ *     // logic to map action to event
```
ewandennis: `// Condition under which we emit this event` ?

src/helpers/analytics/eventsMapper.js
```diff
@@ -0,0 +1,106 @@
...
+const getType = (action) => action.type;
+
+// condition helpers
+const isError = (name) => (action) => name === getErrorName(action);
+const isType = (type) => (action) => type === getType(action);
+const isTypeLike = (regex) => (action) => regex.test(getType(action));
+const isWhitelistedForm = (action) => ANALYTICS_WHITELISTED_FORMS.includes(getFormName(action));
+
+/**
+ * @example
```
ewandennis: Yay inline docs!

src/helpers/analytics/eventsMapper.js
```diff
@@ -0,0 +1,106 @@
...
+import { ANALYTICS_WHITELISTED_FORMS } from 'src/constants';
+import { all } from 'src/helpers/conditions/compose';
+
+// access helpers
+const getErrorMessage = (action) => _.get(action, 'payload.error.message');
+const getErrorName = (action) => _.get(action, 'payload.error.name');
+const getFormName = (action) => _.get(action, 'meta.form');
+const getType = (action) => action.type;
+
+// condition helpers
```
ewandennis: Are these likely to be reusable? If so, we should put them under `helpers/conditions/actions.js` or somesuch. Maybe not `isWhitelistedForm` though. That's pretty here-specific.

bkemper: They could be, so I'll move them.

bkemper: Done... 2704aa9c8

src/helpers/analytics/tests/middleware.test.js
bkemper: Good call!  It should be explicit.

src/helpers/analytics/tests/middleware.test.js
bkemper: Not really, I'll remove to cut down on noise.

src/helpers/analytics/tests/eventsMapper.test.js
bkemper: I added two more events that should should skipped.

src/helpers/analytics/tests/eventsMapper.test.js
bkemper: I think I covered this with the newly added tests.

bkemper:
Ewan, this PR is ready for your final review.  I addressed all your feedback and setup GTM (changes  need to submitted after final approval).

PR #566, opened 2018-08-01
FE-462 Trim whitespaces from email fields

rajumsys:


src/helpers/string.js
```diff
@@ -78,3 +78,7 @@ export function decodeBase64(str) {
     return undefined; // ignore parsing error
   }
 }
+
+export function trimWhitespaces(str) {
+  return _.isString(str) ? str.trim() : str;
```
bkemper: This is a personal preference, but I think we should avoid functions that handle and return any value.

rajumsys: sorry, not sure what you mean.

PR #567, opened 2018-08-02
FE-480: Users get stuck with spinner when returning to app with expired access and refresh tokens

bkemper:


src/actions/helpers/sparkpostApiRequest.js
```diff
@@ -76,6 +76,7 @@ const sparkpostRequest = requestHelperFactory({
           },
           // refresh token request failed
           (err) => {
+            refreshing = false;
```
jonambas: Only suggestion here is to add a comment to explain why we want to set this to false

PR #568, opened 2018-08-02
FE-452 Add datepicker and variant fields to A/B testing

jonambas:
- Uses redux-form's [FieldArray](https://redux-form.com/7.4.2/docs/api/fieldarray.md/) to generate variants.

### DIFF
--- a/src/components/typeahead/TemplateTypeahead.js
+++ b/src/components/typeahead/TemplateTypeahead.js
@@ -11,7 +11,14 @@ export class TemplateTypeahead extends Component {
   };
 
   componentDidMount() {
-    this.props.listTemplates();
+    const { results, listTemplates } = this.props;
+
+    // For redux-form FieldArrays
+    // Subsequent GETs on this list will refresh any selector using it,
+    // which also forces a form reinitialize if used in a form's initialValues.
+    if (results.length === 0) {
+      listTemplates();
+    }
   }
 
   render() {
--- a/src/components/typeahead/TemplateTypeahead.js
+++ b/src/components/typeahead/TemplateTypeahead.js
@@ -30,7 +30,7 @@ export class TemplateTypeahead extends Component {
 }
 
 function mapStateToProps(state, props) {
-  const templates = selectPublishedTemplatesBySubaccount(state);
+  const templates = selectPublishedTemplatesBySubaccount(state, props);
   return {
     results: templates,
     hasTemplates: templates.length > 0
--- a/src/pages/abTesting/components/AbTestCreateForm.js
+++ b/src/pages/abTesting/components/AbTestCreateForm.js
@@ -1,7 +1,7 @@
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
 import { Link } from 'react-router-dom';
-import { reduxForm, Field } from 'redux-form';
+import { reduxForm, Field, formValueSelector } from 'redux-form';
 import { Panel, UnstyledLink, Button, Grid } from '@sparkpost/matchbox';
 import { TextFieldWrapper } from 'src/components';
 import { TemplateTypeaheadWrapper, SubaccountTypeaheadWrapper } from 'src/components/reduxFormWrappers';
@@ -26,7 +26,8 @@ export class AbTestCreateForm extends Component {
       pristine,
       submitting,
       hasSubaccounts,
-      templates
+      templates,
+      subaccountId
     } = this.props;
 
     const disabled = pristine || submitting;
@@ -72,6 +73,7 @@ export class AbTestCreateForm extends Component {
           <Field
             name='default_variant'
             component={TemplateTypeaheadWrapper}
+            subaccountId={subaccountId}
             label={'Select this test\'s default template'}
             placeholder='Type to search'
             helpText={templates.length > 0
@@ -90,10 +92,13 @@ export class AbTestCreateForm extends Component {
 }
 
 function mapStateToProps(state) {
+  const selector = formValueSelector(formName);
+
   return {
     initialValues: {},
     hasSubaccounts: hasSubaccountsSelector(state),
-    templates: selectPublishedTemplatesBySubaccount(state)
+    templates: selectPublishedTemplatesBySubaccount(state),
+    subaccountId: selector(state, 'subaccount.id') || 0
   };
 }
 
--- a/src/pages/abTesting/components/AbTestCreateForm.js
+++ b/src/pages/abTesting/components/AbTestCreateForm.js
@@ -98,6 +98,7 @@ function mapStateToProps(state) {
     initialValues: {},
     hasSubaccounts: hasSubaccountsSelector(state),
     templates: selectPublishedTemplatesBySubaccount(state),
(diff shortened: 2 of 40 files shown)


src/components/typeahead/TemplateTypeahead.js
```diff
@@ -11,7 +11,14 @@ export class TemplateTypeahead extends Component {
   };
 
   componentDidMount() {
-    this.props.listTemplates();
+    const { results, listTemplates } = this.props;
+
+    // For redux-form FieldArrays
+    // Subsequent GETs on this list will refresh any selector using it,
+    // which also forces a form reinitialize if used in a form's initialValues.
```
jonambas: Might be a good idea to force `listTemplates` at the page level instead

src/pages/abTesting/components/AbTestCreateForm.js
```diff
@@ -85,9 +87,12 @@ export class AbTestCreateForm extends Component {
 }
 
 function mapStateToProps(state) {
+  const selector = formValueSelector(formName);
+
   return {
     initialValues: {},
-    hasSubaccounts: hasSubaccountsSelector(state)
+    hasSubaccounts: hasSubaccountsSelector(state),
+    subaccountId: selector(state, 'subaccount.id') || 0
```
rajumsys: why is this needed?  for initial load, we can probably just set in initialValues.

src/pages/abTesting/components/StatusPanel.js
```diff
@@ -16,7 +16,7 @@ const VersionSelector = ({ current, latest, id, subaccountId }) => {
     content: `View Version ${i + 1}`,
     to: `/ab-testing/${id}/${i + 1}${setSubaccountQuery(subaccountId)}`,
     component: Link,
-    visible: i + 1 !== current
+    visible: i + 1 !== Number(current)
```
rajumsys: is this casting necessary?

src/pages/abTesting/components/fields/VariantsFields.js
```diff
@@ -0,0 +1,108 @@
...
+            component={TemplateTypeaheadWrapper}
+            label='Select a default template'
+            placeholder='Type to search'
+            disabled={disabled}
+            subaccountId={subaccountId}
+          />
+        </Grid.Column>
+        <Grid.Column>
+          {
+            formValues.audience_selection === 'sample_size'
```
rajumsys: i think it'll be better readable if you put the final component name in a variable at top and here just render it; just like `MyRoute` in App.js

src/pages/abTesting/components/fields/VariantsFields.js
```diff
@@ -0,0 +1,108 @@
...
+    </Panel>
+    <FieldArray
+      name='variants'
+      component={RenderVariants}
+      formValues={formValues}
+      disabled={disabled}
+      subaccountId={subaccountId}
+    />
+
+    {/* This is a temporary hack to make sure at least some of the last typeahead is visible on screen */}
```
rajumsys: what is the permanent solution? :D

src/selectors/abTesting.js
```diff
@@ -34,11 +37,21 @@ export const selectLatestVersionNumberFromParams = createSelector(
...
+  [selectAbTestFromParams, selectTemplates],
+  (test, templates) => {
 
-    // Strip everything that is not editable
-    const values = _.omit(test, ['created_at', 'updated_at', 'id', 'status', 'version']);
+    // Strip everything that is not editable, or what will be reshaped
+    const values = _.omit(test, ['created_at', 'updated_at', 'id', 'status', 'version', 'start_time', 'end_time', 'variants', 'default_template' ]);
+
+    // Find template objects of the default template and all variants
+    const default_template = findTemplateObject(templates, test.default_template);
```
rajumsys: `defaultTemplate`?

src/pages/abTesting/components/fields/StatusFields.js
```diff
@@ -14,14 +15,23 @@ const StatusFields = ({ disabled }) => (
...
+        name='dates'
+        left
+        showPresets={false}
+        textFieldProps={{
+          helpText: 'A test may run for a maximum of 30 days',
+          label: 'When should we run this test?',
+          prefix: <AccessTime />
+        }}
+        datePickerProps={{
+          disabledDays: { before: new Date() },
```
rajumsys: we should probably fix datepicker (sometime later) to hide left arrow when we don't allow a previous date (than what is displayed). currently it shows the arrow but noop.

src/pages/abTesting/components/fields/VariantsFields.js
```diff
@@ -0,0 +1,108 @@
...
+  <Field
+    name={`${namespace}.sample_size`}
+    label='Number of messages'
+    component={TextFieldWrapper} {...props} />
+);
+
+/*
+  If you're looking at this, refer to https://redux-form.com/7.4.2/examples/fieldarrays/
+ */
+export const RenderVariants = ({ fields, formValues, disabled, subaccountId }) => (
```
rajumsys: where are these, specifically `fields` coming from?

src/pages/abTesting/components/StatusPanel.js
jonambas: Yeah, `current` comes from `version` from router params, which is a string. I could cast in the selector instead maybe

src/pages/abTesting/components/fields/VariantsFields.js
jonambas: Idk, making Popover respond to its position (like tooltips)

src/pages/abTesting/components/fields/StatusFields.js
jonambas: Ah yeah, it looks like it does the same thing on the reports too

src/pages/abTesting/components/fields/VariantsFields.js
jonambas: From `FieldArray` - see https://redux-form.com/7.4.2/docs/api/fieldarray.md/
It's a synthetic array that redux-form maintains

src/pages/abTesting/components/AbTestCreateForm.js
jonambas: There are no initialValues here as its the create form - `subaccountId` here is used only the typeahead, so it knows which templates to filter/show and for validation in another ticket. I can add a comment to make that clear

src/selectors/abTesting.js
jonambas: Think I should camel case? I'm not camel casing any of the other keys 🤔

rajumsys: no, i meant the local variable only.

src/pages/abTesting/components/StatusPanel.js
jonambas: Found a bug while investigating this - `cDU` doesn't correctly reGET ab test when changing version

src/pages/abTesting/components/AbTestCreateForm.js
rajumsys: i think i see why you did it.

PR #569, opened 2018-08-03
Hotfix/eo 208

pmcclory:


src/helpers/tests/validation.test.js
```diff
@@ -74,10 +78,10 @@ describe('Validation helpers', () => {
     const goodInput = cases[caseName].good;
     const badInput = cases[caseName].bad;
     goodInput.forEach((input) => it(`${caseName} should accept ${input}`,
-      () => expect(validations[caseName](input)).toBeUndefined()));
+      () => expect(validations[caseName](...Array.isArray(input) ? input : [input])).toBeUndefined()));
```
pmcclory: As i mentioned in the PR this is the first validator that is using additional context and not simply the value to determine if the value is valid.

Full prototype for validate is : `validate : (value, allValues, props, name) => error [optional]` according to https://redux-form.com/7.4.2/docs/api/field.md/

This seemed like kind of a hack....but seemed like the easiest way to shoehorn this checking in without retooling everything that was already there.  Happy to rework it if it offends any sensibilities though

src/helpers/tests/validation.test.js
```diff
@@ -12,7 +12,7 @@ import * as validations from '../validation';
 
 const cases = {
   required: {
-    good: ['101', 1, {}, []],
+    good: ['101', 1, {}, [[]]],
```
pmcclory: this is a little gross.  the reason for it is below.

src/helpers/validation.js
```diff
@@ -24,6 +24,10 @@ export function abTestId(value) {
   return abTestIdRegex.test(value) ? undefined : 'Must contain only letters, numbers, hyphens, and underscores';
 }
 
+export function abTestDefaultTemplate(value, allValues, props) {
+  return props.templates.includes(value) ? undefined : 'Invalid Default Template';
```
ewandennis: Since we have all form values here, lets give some actionable hints in our validation error. How about something like this?
```js
  if (props.templates.includes(value)) {
    return undefined;
  }

  if (allValues.subaccount) {
    return 'Template not available to selected subaccount';
  }

  return 'Subaccount templates not available to master account';
```

src/pages/abTesting/components/tests/AbTestCreateForm.test.js
```diff
@@ -26,12 +27,12 @@ describe('A/B Test Create Form Component', () => {
   });
 
   it('should render subaccount section if subaccounts exist', () => {
-    wrapper.setProps({ hasSubaccounts: true });
+    wrapper.setProps({ hasSubaccounts: true, templates: ['fake']});
```
ewandennis: Why are we setting the `templates` prop here? Is this test dependant on the value of `templates`?

src/pages/abTesting/components/tests/AbTestCreateForm.test.js
```diff
@@ -26,12 +27,12 @@ describe('A/B Test Create Form Component', () => {
...
     expect(wrapper.find({ name: 'subaccount' })).toMatchSnapshot();
   });
 
   it('should not render subaccount section if subaccounts do not exist', () => {
-    wrapper.setProps({ hasSubaccounts: false });
+    wrapper.setProps({ hasSubaccounts: false, templates: ['fake']});
```
ewandennis: Same question about `props.templates`

src/helpers/tests/validation.test.js
```diff
@@ -73,11 +78,12 @@ describe('Validation helpers', () => {
   Object.keys(cases).forEach((caseName) => {
     const goodInput = cases[caseName].good;
     const badInput = cases[caseName].bad;
+    const multiArg = !!cases[caseName].multiArg;
     goodInput.forEach((input) => it(`${caseName} should accept ${input}`,
-      () => expect(validations[caseName](input)).toBeUndefined()));
+      () => expect(validations[caseName](...multiArg ? input : [input])).toBeUndefined()));
```
ewandennis: eh, what can you do _shrug_

src/helpers/validation.js
pmcclory: 👍

src/pages/abTesting/components/tests/AbTestCreateForm.test.js
pmcclory: 🤔 - yea looks like this was residual from when I was mashing the keyboard trying to figure out how to write tests.  Thought it was needed, but appears to not be.  Removed in an update

PR #570, opened 2018-08-03
Feature/eo 210

ischer:


src/pages/abTesting/components/TestCollection.js
```diff
@@ -0,0 +1,111 @@
...
+
+  componentDidMount() {}
+
+  getDetailsLink = ({ id, version, subaccount_id }) => `/ab-testing/${id}/${version}${setSubaccountQuery(subaccount_id)}`
+
+  getColumns() {
+    const columns = [
+      { label: 'Name', sortKey: 'name' },
+      { label: 'Status', sortKey: 'status' },
+      { label: 'Template', sortKey: (i) => i.winning_template_id || i.default_template.template_id },
```
ischer: missing test coverage here and on the `onClick` of the actions just below in `getRowData`... not immediately sure how to get these. thoughts?

jonambas: Hm I think its fine to leave out sortKey. Same with the actions I think

Edit: got it working with this...
```js
const actionCol = shallow(wrapper.instance().getRowData(props.abTests[0]).pop());
actionCol.find('ActionList').prop('actions')[4].onClick()
expect(props.toggleDelete).toHaveBeenCalled();
```

src/reducers/abTesting.js
```diff
@@ -32,6 +34,28 @@ export default (state = initialState, action) => {
...
+    /* DELETE */
+
+    case 'DELETE_AB_TEST_PENDING':
+      return { ...state, deleteError: null };
+
+    case 'DELETE_AB_TEST_SUCCESS':
+      return { ...state };
+
+    case 'DELETE_AB_TEST_FAIL':
+      return { ...state, deleteError: action.payload };
```
jonambas: Since we don't do anything with the error payload (if we use auto global alerts from the actions), can probably get rid of the error keys

src/pages/abTesting/ListPage.js
```diff
@@ -1,102 +1,61 @@
...
+  };
+
   componentDidMount() {
     this.props.listAbTests();
   }
 
-  getDetailsLink = ({ id, version, subaccount_id }) => {
-    return `/ab-testing/${id}/${version}${setSubaccountQuery(subaccount_id)}`;
-  }
+  toggleDelete = (id, subaccount_id) => {
```
jonambas: When the modal opens, the table collection seems to reset its page (but not per page), have no idea what this could be

src/pages/abTesting/components/TestCollection.js
```diff
@@ -0,0 +1,111 @@
...
+
+const filterBoxConfig = {
+  show: true,
+  itemToStringKeys: ['name', 'id', 'status', 'test_mode'],
+  exampleModifiers: ['id', 'status', 'test_mode']
+};
+
+export class TestCollection extends Component {
+
+  componentDidMount() {}
```
jonambas: Can remove

src/pages/abTesting/DetailsPage.js
```diff
@@ -41,20 +43,51 @@ export class DetailsPage extends Component {
...
+    return {
+      breadcrumbAction: { content: 'Back to A/B Tests', component: Link, to: '/ab-testing' },
+      cancelAction: {
+        content: <span><Cancel/> Cancel Test</span>,
+        visible: status === 'scheduled' || status === 'running',
+        onClick: this.toggleCancel
+      },
+      deleteAction: {
+        content: <span><Delete/> Delete Test</span>,
+        visible: status === 'completed' || status === 'cancelled',
```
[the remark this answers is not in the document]
ischer: as of right now, these are the only states that delete is available for. we have a ticket to make delete work with all states, though. at that point, i think we can just remove this line.

src/pages/abTesting/DetailsPage.js
```diff
@@ -41,20 +43,51 @@ export class DetailsPage extends Component {
...
-      onClick: this.toggleDelete
-    },
-    test: this.props.test,
-    subaccountId: this.props.subaccountId
-  })
+  getSharedProps = () => {
+    const { status } = this.props.test;
+    return {
+      breadcrumbAction: { content: 'Back to A/B Tests', component: Link, to: '/ab-testing' },
+      cancelAction: {
```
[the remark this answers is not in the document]
ischer: cancel should be necessary in EditMode for `scheduled` tests

src/pages/abTesting/DetailsPage.js
```diff
@@ -41,20 +43,51 @@ export class DetailsPage extends Component {
...
-  //   this.setState({ showDelete: !this.state.showDelete });
-  // }
+  toggleDelete = () => {
+    this.setState({ showDeleteModal: !this.state.showDeleteModal });
+  }
+
+  handleDelete = () => {
+    const { id, subaccount_id } = this.props.test;
+
+    return this.props.deleteAbTest({ id, subaccountId: subaccount_id }).then(() => {
```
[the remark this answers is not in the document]
ischer: good catch. i'll update!

src/pages/abTesting/ListPage.js
```diff
@@ -1,102 +1,61 @@
...
-      null
-    ];
+  handleDelete = () => {
+    const { id, subaccountId } = this.state.testToDelete;
 
-    return columns;
-  }
+    return this.props.deleteAbTest({ id, subaccountId }).then(() => {
+      this.props.showAlert({ type: 'success', message: 'Test deleted' });
+      this.toggleDelete();
```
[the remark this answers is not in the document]
ischer: i was wondering the same thing. thought it was weird to continue to show a deleted test. not sure what you mean about detailsById yet, but i'll take a peek into that first, then listAbTests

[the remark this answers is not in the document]
jonambas: Nvm, not sure what I was thinking - `detailsById` is where the single gets are stored, but that wont update the list.

src/pages/abTesting/ListPage.js
ischer: huh - weird. good catch. i'll take a look into that.

src/pages/abTesting/ListPage.js
```diff
@@ -142,6 +97,21 @@ export class ListPage extends Component {
...
         }}>
         {error ? this.renderError() : this.renderCollection()}
+        <DeleteModal
+          open={this.state.showDeleteModal}
+          title='Are you sure you want to delete this test?'
+          content={<p>The test and all associated versions will be immediately and permanently removed. This cannot be undone.</p>}
+          onDelete={this.handleDelete}
+          onCancel={this.toggleDelete}
+        />
+        <ConfirmationModal
```
jonambas: Will need to pull in `isPending` for these as well

src/pages/abTesting/ViewMode.js
```diff
@@ -74,6 +96,14 @@ export class ViewMode extends Component {
           </Section.Right>
         </Section>
 
+        <ConfirmationModal
```
jonambas: `isPending` here too

src/pages/abTesting/ViewMode.js
```diff
@@ -21,21 +31,33 @@ export class ViewMode extends Component {
...
 
+  toggleOverride = () => {
+    this.setState({ showOverrideModal: !this.state.showOverrideModal });
+  }
+
+  handleOverride = () => {
+    const { id, version } = this.props.test;
+    const { subaccountId } = this.props;
+
+    return this.props.updateDraft({}, { id, subaccountId }).then(() => {
```
jonambas: Just now noticed, but guessing this is temporary until the endpoint is available?

src/pages/abTesting/DetailsPage.js
```diff
@@ -41,20 +43,53 @@ export class DetailsPage extends Component {
...
+    this.setState({ showCancelModal: !this.state.showCancelModal });
+  }
+
+  handleCancel = () => {
+    const { id } = this.props.test;
+    const { subaccountId } = this.props;
+
+    return this.props.cancelAbTest({ id, subaccountId }).then(() => {
+      this.props.showAlert({ type: 'success', message: 'Test cancelled' });
+      this.props.history.push('/ab-testing');
```
jonambas: I think here, instead of redirecting, initiate another `getAbTest({ id, version, subaccountId });`, which should automatically refresh the page

src/pages/abTesting/ViewMode.js
ischer: nope - might be a little weird, but when passing an empty object to the update draft endpoint, it "overrides" the test by creating a new draft version without the winning template.

src/pages/abTesting/ViewMode.js
ischer: 👍  when adding this, i noticed that the update draft action did not have a reducer. what happens if an action does not have a reducer? does it just not alter the state in any way?

jonambas: Yep, it just doesn't do anything. I was using redux-form `submitting` to handle loading state.

PR #573, opened 2018-08-09
FE-495 Fix AB Testing Defects

jonambas:
To verify

(1)
- Go to create page
- Select a template that is assigned to master and not shared with all
- Select a subaccount
- Attempt to submit, error should be visible

(2)
- Create a test assigned to master
- The following redirect should not include `?subaccount=0` and subaccount tag should not be visible

(3)
- Cancel a test from the list page
- Status tag should update to 'Cancelled'
- Test with an AB test assigned to a subaccount as well

### DIFF
--- a/src/selectors/templates.js
+++ b/src/selectors/templates.js
@@ -1,7 +1,6 @@
 import _ from 'lodash';
 import config from 'src/config';
 import { createSelector } from 'reselect';
-import { formValueSelector } from 'redux-form';
 import { getDomains, isVerified } from 'src/selectors/sendingDomains';
 import { selectSubaccountIdFromProps, hasSubaccounts } from 'src/selectors/subaccounts';
 import { filterTemplatesBySubaccount } from 'src/helpers/templates';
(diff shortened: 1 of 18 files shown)


src/selectors/templates.js
```diff
@@ -41,16 +43,24 @@ export const selectDomainsBySubaccount = createSelector(
...
+ * Selects subaccount id from a redux-form subaccount typeahead
+ */
+export const selectSubaccountIdFromFormTypeahead = (state, props, formName) => {
+  const selector = formValueSelector(formName);
+  return selector(state, 'subaccount.id');
+};
+
+// Selects published templates, filtered by subaccount typeahead value within redux-form
+// Used for form validation
+export const selectPublishedTemplatesBySubaccountFromTypeahead = createSelector(
```
jonambas: I really didn't see any way around have two selectors that _almost_ do the same thing. The above uses subaccount ID from `props`, while this one accepts a `formName` and uses a redux-form selector.

Tried doing this inside the component instead but (1) would need to filter on each render and (2) redux-form validator can only access values in `props` (or form values)

The above version that uses subaccount ID from `props`, is used within the typeahead itself to generate its list for both the create and details page.

jonambas: I think that looks and reads better, will give that a shot

PR #577, opened 2018-08-13
FE-502 Fix billing country and state validation

jonambas:
To reproduce:

- Open either the change plan page, onboarding plan page, or update payment modal.
- Make sure the billing address field is visible - select a new paid plan, and do not use saved credit card.
- Country initial value error:
  - Do no touch fields.
  - Attempt to submit to force validation
  - Country will show "Required"
  - State field is not displayed
- State initial value error:
  - Select a country without states (not US or CA)
  - Immediately select a country with states (US or CA)
  - States field should be visible 
  - Attempt to submit to force validation
  - State will show "Required"

### DIFF
--- a/src/pages/billing/forms/fields/tests/BillingAddressForm.test.js
+++ b/src/pages/billing/forms/fields/tests/BillingAddressForm.test.js
@@ -9,13 +9,16 @@ describe('Billing Address Form:', () => {
     formName: 'form-name',
     countryValue: 'GG',
     countries: [
-      { value: 'US', label: 'USOFA', states: [{ name: 'mrylnd', value: 'MD' }]}
+      { value: 'US', label: 'USOFA', states: [{ name: 'mrylnd', value: 'MD' }]},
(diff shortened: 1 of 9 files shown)


jonambas:
Think I need to make the same change to `BillingContactForm`

src/pages/billing/forms/fields/tests/BillingAddressForm.test.js
```diff
@@ -9,13 +9,16 @@ describe('Billing Address Form:', () => {
...
+      { value: 'CA', label: 'CANADA', states: [{ name: 'alberta', value: 'AB' }]}
     ],
-    change: jest.fn()
+    change: jest.fn(),
+    firstState: 'FirstState'
   };
 
   beforeEach(() => {
     wrapper = shallow(<BillingAddressForm {...props} />);
+    jest.clearAllMocks();
```
bkemper: This shouldn't be needed.  It is already setup.
https://github.com/SparkPost/2web2ui/blob/master/jest.config.js#L59

PR #565, opened 2018-07-31
SA-173 - Show precision in UI and better handle it

gwalls:
Web UI subtask for SA-31. The UI (and really the Metrics API) can display incorrect metrics when a time range is selected that isn't padded to the nearest precision. The API will be corrected to allow the user to specify a precision. These UI changes will ensure the user is always viewing an accurate time range for the metrics selected.

https://jira.int.messagesystems.com/browse/SA-31

API doc changes:
https://github.com/SparkPost/sparkpost-api-documentation/pull/84/files

PR #578, opened 2018-08-13
EO-209 Add A/B testing field validation & ability to schedule a draft

ischer:


src/pages/abTesting/components/fields/VariantsFields.js
```diff
@@ -4,21 +4,28 @@ import { Panel, Grid, Button } from '@sparkpost/matchbox';
...
 import styles from './VariantsFields.module.scss';
 
 export const PercentField = ({ namespace, ...props }) => (
   <Field
     name={`${namespace}.percent`}
     label='Percent of total'
+    type='number'
+    parse={Number}
     suffix='%'
+    validate={[required, minNumber(1), maxNumber(100)]}
```
rajumsys: i can't clear the value. like it has `0`. i can type `050` but can't remove the leading 0 totally. but it works if i select all and type.

rajumsys: if i put some % on default and something else on variant 1, new variant can't be added. get following error:
> TypeError: Cannot read property 'percent' of undefined

update: looks like it's breaks everytime i click add variant if i choose percentage.

src/actions/abTesting.js
```diff
@@ -65,6 +65,19 @@ export function cancelAbTest({ id, subaccountId }) {
   });
 }
 
+export function scheduleAbTest(data, { id, subaccountId }) {
```
rajumsys: looks like the convention of functions in this module is that they always receive spread the first arg. so we can write it like `scheduleAbTest({data, id, subaccountId })`

src/actions/abTesting.js
```diff
@@ -80,3 +93,15 @@ export function updateDraft(data, { id, subaccountId }) {
     }
   });
 }
+
+export function updateAbTest(data, { id, subaccountId }) {
```
rajumsys: same here, should we write it like `({data, id, subaccountId})`?

src/helpers/abTesting.js
```diff
@@ -40,3 +42,13 @@ export const findTemplateObject = (templates, { template_id, ...rest } = {}) =>
   template_object: _.find(templates, ({ id }) => id === template_id),
   ...rest
 });
+
+/**
+ * Calculates the total_sample_size of a test based upon the sample_size of default_template & variants
+ */
+export const calculateTotalSampleSize = ({ default_template, variants }) => {
+  let total = 0;
+  _.forEach(variants, (variant) => total += variant.sample_size);
```
rajumsys: `_.reduce` would be more fitting. you can pass `default_template.total_size` as initial value :).

src/helpers/tests/validation.test.js
```diff
@@ -115,6 +128,16 @@ describe('Memoized validation helpers', () => {
   });
 });
 
+describe('Number between validation', () => {
+  it('returns undefined for valid input', () => {
+    expect(validations.numberBetween(0, 1)(0.5)).toBeUndefined();
+  });
+
+  it('returns error for invalid input', () => {
```
rajumsys: one test to assert when input is `1`? you can use `cases` for these tests too.

src/components/reduxFormWrappers/RadioGroup.js
```diff
@@ -3,12 +3,19 @@ import { Grid, Radio, Label, Error } from '@sparkpost/matchbox';
...
+  meta,
+  bottomError,
+  grid = { xs: 12, sm: 12, md: 12, lg: 12, xl: 12 }
+}) {
   const { error, touched } = meta;
 
   return (
     <Radio.Group label={label}>
-      <Label>{title}{touched && error ? <Error error={error}/> : ''}</Label>
+      <Label>{title}{!bottomError && touched && error ? <Error error={error}/> : ''}</Label>
```
rajumsys: do you think if the prop name is rather error position (top/bottom) will make more sense than having `bottomError`? so we'll specify error and optionally position prop.

src/helpers/tests/validation.test.js
ischer: it didn't look like `cases` supported memoized validators with multiple arguments. i can look into adding support for this, but i wasn't sure if it was worth it.

src/pages/abTesting/components/fields/StatusFields.js
```diff
@@ -21,6 +23,7 @@ const StatusFields = ({ disabled }) => (
         name='dates'
         left
         showPresets={false}
+        validate={abTestDuration}
```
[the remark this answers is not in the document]
ischer: what does this do?

src/pages/abTesting/components/fields/VariantsFields.js
ischer: i'll take a look into this. good catch!

src/components/reduxFormWrappers/RadioGroup.js
ischer: i don't have a strong opinion on this - but went with `bottomError` so that I could have a boolean prop that maintains existing functionality wherever this was used without updating code everywhere. i feel like a string comparison here might muddy up the logic a bit here.

if you think i should update, lmk

src/helpers/abTesting.js
ischer: i'll give that a shot!

src/helpers/tests/validation.test.js
rajumsys: either way it works. just saying you have option.

src/components/reduxFormWrappers/RadioGroup.js
rajumsys: actually i misread it. it's doing exactly what i suggested.

src/pages/abTesting/components/fields/StatusFields.js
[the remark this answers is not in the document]
rajumsys: sorry, i meant to add more info but forgot. in https://github.com/SparkPost/2web2ui/pull/565/files#diff-d5b3787f0e12921036bd5c13acd24a4bR198  @gwalls added that prop which will round the value based on it's precision map. so far I understand we don't need to round here. That's only for metrics. 

By default the value of the prop is `true`.

src/helpers/validation.js
```diff
@@ -24,16 +25,32 @@ export function abTestId(value) {
...
-  if (allValues.subaccount) {
+  if (formValues.subaccount) {
     return 'Template not available to selected subaccount';
   }
   return 'Template not available to master account';
 }
 
+export function abTestDuration(value, formValues) {
+  const testDuration = getDuration(value) + (formValues.engagement_timeout || 24);
+  return testDuration > 720 ? 'Test duration + engagement timeout must be 30 days or less' : undefined;
```
rajumsys: this error message may be little cryptic for the end users. they don't see what is engagement timeout. IMO, the error message mention fields by their label. So we need to update their labels or update this error message. Having current label will make this error message too big.

we can give short labels and explain in helptext/tooltip.

src/helpers/validation.js
```diff
@@ -44,7 +44,7 @@ export function abTestDistribution(value, formValues) {
   const { default_template, variants } = formValues;
 
   if (value === 'percent') {
-    const total = _.reduce(variants, (sum, variant = { percent: 0 }) => sum + variant.percent, default_template.percent);
+    const total = _.reduce(variants, (sum, variant = { percent: 0 }) => sum + parseInt(variant.percent, 10), parseInt(default_template.percent, 10));
```
rajumsys: as we're using `_`, you can use `_.parseInt` and no need to pass the radix 10 as it's default.

rajumsys:
- From time seems good but to seems in past (if i select today that's the value ended up showing)!
screen shot 2018-08-16 at 3 52 44 pm
- i can now correctly clear all input fields (engagement timeout, percent fields) 👍

src/helpers/validation.js
```diff
@@ -53,6 +53,10 @@ export function startTimeAfterNow(value) {
   return isStartTimeAfterNow(value) ? undefined : 'Start date cannot be in the past';
 }
 
+export function startTimeAfterEndTime(value) {
```
rajumsys: will it make more sense to name it like `endTimeAfterStartTime`?

src/helpers/validation.js
```diff
@@ -53,8 +53,8 @@ export function startTimeAfterNow(value) {
   return isStartTimeAfterNow(value) ? undefined : 'Start date cannot be in the past';
 }
 
-export function startTimeAfterEndTime(value) {
-  return getDuration(value, 'minutes') > 0 ? undefined : 'Start date must be after end date';
+export function startTimeBeforeEndTime(value) {
```
rajumsys: oh you renamed anyway :). makes sense

ischer: yep! 😄  

sorry, i didn't see your original comment, but as soon as i pushed the original name i thought to myself "wait a second, this doesn't make sense" and flipped it.

PR #586, opened 2018-08-17
EO-209 cast stringified values to numbers

ischer:


src/helpers/tests/abTesting.test.js
```diff
@@ -8,32 +8,36 @@ describe('A/B testing helper', () => {
       values = {
         name: 'test_one',
         test_mode: 'bayesian',
-        confidence_level: 0.91,
+        confidence_level: '0.91',
```
rajumsys: why is this coming as string?

PR #590, opened 2018-08-22
FE-479 Add Brightback integration

jonambas:


src/reducers/brightback.js
```diff
@@ -0,0 +1,21 @@
+const initialState = {
+  valid: true,
+  url: null,
+  precancelLoading: false
+};
+
+export default (state = initialState, { type, payload, meta }) => {
+  switch (type) {
+    case 'BRIGHTBACK_PRECANCEL_PENDING':
+      return { ...state, precancelLoading: true, valid: false, url: null };
```
jonambas: Can probably get rid of this loading key, not being used

jonambas:
@ewandennis So couple things:

(1) I've moved `Brightback` into a `withBrightback` HoC, but since we are now going to be using brightback in the profile page (cancel account), this doesn't make sense anymore. Both the conditions to enable the redirect and the return urls will be different.

I think this api works:

```js
<Brightback
  condition={isDowngradeToFree && etc}
  urls={mapofurls}
  render={({ to, enabled }) =>
    <Button
       type={enabled ? 'button' : 'submit' }
       to={{enabled ? 'to' : null }} />
}/>
```

(2) Tried splitting up `ChangePlanForm` in different ways, and tbh I'm not really comfortable making a refactor this big in this ticket. All billing actions and the form's render tree require the same pieces of redux state and form values - there's no easy split. The most I think I can do here is to move `mSTP` and `mDTP` out into a "container" like templates

src/components/globalAlert/Alert.js
```diff
@@ -52,17 +52,18 @@ class Alert extends Component {
   }
 
   renderMessage() {
-    const { message, details, action = {}} = this.props;
-    const { content: actionContent, ...actionProps } = action;
+    const { message, details, action } = this.props;
```
jonambas: Will drop changes in this file, didn't realize I created a ticket for this - FE-515

src/components/brightback/Brightback.js
```diff
@@ -0,0 +1,42 @@
...
+  componentDidMount() {
+    const { precancel, data, hasBrightbackOption } = this.props;
+
+    if (hasBrightbackOption) {
+      precancel(data);
+    }
+  }
+
+  getRenderProps() {
+    const { valid, url, hasBrightbackOption, condition = true } = this.props;
```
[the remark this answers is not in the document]
jonambas: Hm nope, good point. Will add a `.isRequired` proptype to it though

src/components/brightback/Brightback.js
```diff
@@ -0,0 +1,42 @@
...
+  render() {
+    return this.props.render(this.getRenderProps());
+  }
+}
+
+const mapStateToProps = (state, props) => ({
+  valid: state.brightback.valid,
+  url: state.brightback.url,
+  data: selectBrightbackData(state, props),
+  loading: state.brightback.precancelLoading,
```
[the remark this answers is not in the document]
jonambas: Yeah, the intent was to default to current behavior if that brightback endpoint fails or hangs for whatever reason

src/selectors/brightback.js
```diff
@@ -0,0 +1,29 @@
...
+      referrer: document.referrer
+    },
+    save_return_url: `${window.location.origin}${save_return_url}`,
+    cancel_confirmation_url: `${window.location.origin}${cancel_confirmation_url}`,
+    billing_url: `${window.location.origin}${billing_url}`,
+    account: {
+      created_at: new Date(created).getTime(),
+      internal_id: customer_id
+    },
+    email
```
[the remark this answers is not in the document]
jonambas: I added it in last week because the call was returning `valid: false` without it, but seems okay without it now

src/pages/billing/forms/ChangePlanForm.js
```diff
@@ -38,6 +38,34 @@ export class ChangePlanForm extends Component {
     this.props.getPlans();
     this.props.getBillingCountries();
     this.props.fetchAccount({ include: 'billing' });
+
+    if (this.props.immediatePlanChange) {
```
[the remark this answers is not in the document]
jonambas: FE-515 for the global alert bug

Could also return `<Loading>` on immediate plan change. New route is probably the safest.

src/pages/billing/ImmediateChangePlanPage.js
```diff
@@ -0,0 +1,84 @@
...
+      .then(() => {
+        conversions.trackPlanChange({ allPlans: billing.plans, oldCode, newCode });
+        this.setState({ loading: false });
+        showAlert({ type: 'success', message: 'Subscription Updated' });
+      });
+  }
+
+  render() {
+    const { loading } = this.state;
+    return <Page breadcrumbAction={{ content: 'Back to billing', to: '/account/billing', Component: Link }}>
```
jonambas: Hm, I don't think we actually need to render a page here.

Maybe just show `<Loading>` component when loading

This seems to look okay to me: 
```js
<div style={{ display: 'flex', alignItems: 'center', height: '90vh' }}>
  <div style={{ flex: '1 0 0', textAlign: 'center'}}>
    <h1>Your subscription has been updated.</h1>
    <Button color='orange'>Back to Billing</Button>
  </div>
</div>
```

src/pages/billing/ImmediateChangePlanPage.js
```diff
@@ -0,0 +1,84 @@
...
+    });
+  }
+
+  handleImmediatePlanChange() {
+    const { billing, oldCode, immediatePlanChange: newCode, updateSubscription, showAlert } = this.props;
+    return updateSubscription({ code: newCode })
+      .then(() => {
+        conversions.trackPlanChange({ allPlans: billing.plans, oldCode, newCode });
+        this.setState({ loading: false });
+        showAlert({ type: 'success', message: 'Subscription Updated' });
```
jonambas: Probably don't need this alert since the message is show in the page

src/pages/billing/ImmediateChangePlanPage.js
```diff
@@ -0,0 +1,84 @@
...
+      search: stripImmediatePlanChange(location.search)
+    });
+  }
+
+  handleImmediatePlanChange() {
+    const { billing, oldCode, immediatePlanChange: newCode, updateSubscription, showAlert } = this.props;
+    return updateSubscription({ code: newCode })
+      .then(() => {
+        conversions.trackPlanChange({ allPlans: billing.plans, oldCode, newCode });
+        this.setState({ loading: false });
```
jonambas: Just a note, if the update fails, loading remains `true`. We could maybe pull this from redux state instead

src/pages/billing/ImmediateChangePlanPage.js
```diff
@@ -33,45 +42,58 @@ export class ImmediateChangePlanPage extends Component {
...
   }
 
+  getLoadState() {
+    return Number(this.state.loading);
+  }
+
+  renderSuccess() {
+    return <div className={styles.MessageInnards}>
+      <h1>Your subscription has been updated.</h1>
+      <Button to={BILLING_ROUTE} color='orange'>Back to Billing</Button>
```
jonambas: `PageLink` or add `component={Link}`

PR #593, opened 2018-08-23
FE-477 Hide usage report on new accounts that have not sent in first month

jonambas:


src/components/usageReport/UsageReport.js
```diff
@@ -1,6 +1,6 @@
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
-import { format } from 'date-fns';
+import { formatDateTime, formatDate } from 'src/helpers/date';
```
jonambas: Converted dates to use these, could probably get rid of `date-fns` soon

src/components/usageReport/UsageReport.js
```diff
@@ -23,29 +23,34 @@ const getPercent = (used, limit) => Math.floor((used / limit) * 100);
...
+
   render() {
-    const { subscription, usage } = this.props;
+    const { subscription, usage, accountAgeInWeeks } = this.props;
 
     if (!subscription || !usage) {
       return <PanelLoading />;
     }
 
+    if (usage.day.used === 0 && usage.month.used === 0 && accountAgeInWeeks < 4) {
```
jonambas: Felt like I needed to add this account age check. Without this the usage report would be hidden if they haven't sent anything at the beginning of every billing cycle

PR #594, opened 2018-08-23
SA-235 Changed message events advanced filter event types to be autom…

gwalls:
Changed message events advanced filter event types to be automatically populated from documentation.

This change is to support upcoming AMP events - amp_open, apm_click, and amp_initial_open. Users will now have these new message types that they can subscribe to with Webhooks or filter and search Message Events with.

Testing: I injected several of these mock AMP events that can be viewed with Message Events in the appteam account (http://localhost:3100/reports/message-events?events=amp_click&events=amp_initial_open&events=amp_open&from=2018-08-21T04%3A00%3A00Z&range=custom&to=2018-08-23T03%3A59%3A59Z)
I was also able to create a Webhook in the app and view the new events coming through.


src/pages/reports/messageEvents/components/AdvancedFilters.js
```diff
@@ -81,13 +83,16 @@ export class AdvancedFilters extends Component {
     this.setState({ search: { ...this.state.search, ...updatedSearch }});
   }
 
+  renderEvents = () => this.props.eventListing.map((event) => this.renderEvent(event));
```
ewandennis: `AdvancedFilters` is already getting large. Let me split the filters out into components.

ewandennis:
Some questions about these new events. We have some odd-shaped events without message IDs which we need to treat specially in the details view. Do these new events always have the `message_id` field?

Is there anything else odd-shaped about them compared to say `injection`, `delivery` or `bounce`?

gwalls:
The new events should look identical to their non-amp counterparts (open, click, initial_open), except for the `type` field.

PR #603, opened 2018-08-29
add 2fa status to users list

rajumsys:


src/pages/users/ListPage.js
```diff
@@ -12,11 +12,12 @@ import { selectUsers } from 'src/selectors/users';
...
 import RoleSelect from './components/RoleSelect';
 import DeleteButton from './components/DeleteButton';
+import User from './components/User';
 
 const COLUMNS = [
-  { label: 'Name', sortKey: 'name' },
+  { label: 'User', sortKey: 'name' },
   { label: 'Role', sortKey: 'access' },
-  { label: 'Email', sortKey: 'email' },
+  { label: '2 Factor Authentication', sortKey: 'tfa_enabled' },
```
bkemper: Do we want to show this column if the account has SSO turned on?

src/pages/users/ListPage.js
```diff
@@ -39,15 +40,15 @@ export class ListPage extends Component {
...
+    <User {...user} />,
     <RoleSelect
       disabled={user.isCurrentUser}
       name={user.username}
       onChange={this.handleAccessChange}
       value={user.access}
       allowSuperUser={user.access === 'superuser'}
     />,
-    user.email,
+    user.tfa_enabled ? 'Enabled' : 'Disabled',
```
bkemper: Should this be a `<Tag />`?

src/pages/users/ListPage.js
```diff
@@ -39,15 +40,15 @@ export class ListPage extends Component {
 
   // Do not allow current user to change their access/role or delete their account
   getRowData = (user) => [
-    user.name,
+    <User {...user} />,
```
bkemper: Should we pass only the user properties that we need?

PR #604, opened 2018-08-29
EO-231 

ischer:
engagement rate only shown when count_accepted present, not hidden if rate 0, fixed display of rate

### DIFF
--- a/src/helpers/abTesting.js
+++ b/src/helpers/abTesting.js
@@ -59,3 +59,11 @@ export const calculateTotalSampleSize = ({ default_template, variants }) =>
     (sum, variant) => sum + parseInt(variant.sample_size, 10),
     parseInt(default_template.sample_size, 10)
   );
+
+/**
+ * Checks if a test has any variants which have been delivered to
+ */
+export const hasTestDelivered = (abTest) => {
+  const variants = [abTest.default_template].concat(abTest.variants);
+  return !!_.find(variants, (variant = {}) => variant.count_accepted > 0);
+};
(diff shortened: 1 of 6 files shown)


src/helpers/abTesting.js
```diff
@@ -59,3 +59,11 @@ export const calculateTotalSampleSize = ({ default_template, variants }) =>
     (sum, variant) => sum + parseInt(variant.sample_size, 10),
     parseInt(default_template.sample_size, 10)
   );
+
+/**
+ * Checks if a test has any variants which have been delivered to
+ */
+export const hasTestDelivered = (abTest) => {
+  const variants = [abTest.default_template].concat(abTest.variants);
```
bkemper: Feel free to use spread here.  `const variants = [abTest.default_template, ...abTest.variants];`

src/helpers/abTesting.js
```diff
@@ -59,3 +59,11 @@ export const calculateTotalSampleSize = ({ default_template, variants }) =>
...
+  return !!_.find(variants, (variant = {}) => variant.count_accepted > 0);
```
bkemper: `Boolean` can be a nice replacement for `!!`.

src/helpers/abTesting.js
ischer: good call - definitely way better this way

src/helpers/abTesting.js
ischer: nice - yeah, much clearer.

PR #607, opened 2018-08-31
FE-508: Reinstate website auth cookie

bkemper:
**Testing Locally**
1. Log in
1. Confirm both "auth" and "website_auth" cookies were created with "Application" tab of dev tools
    - Note: verify format and token values for each cookie
1. Log out
1. Confirm both "auth" and "website_auth" cookies were deleted with "Application" tab of dev tools

TBD:
 - [x] Wait for FE-464 config bust up?
 - [x] Verify TFA
 - [ ] Verify SSO
 - [x] Deal with [website cookie expiry](https://github.com/SparkPost/webui/blob/11e2bcc7f5a8c44c218b48ceecc2cee27ac34b3d/src/assets/config/config.js#L23)
 - [x] Publish the fact that this will work across _all_ tenants unlike the previous implementation

### DIFF
--- a/public/static/tenant-config/production.js
+++ b/public/static/tenant-config/production.js
@@ -1,6 +1,15 @@
 window.SP = window.SP || {};
 window.SP.productionConfig = {
   apiBase: 'http://api.sparkpost.test/api/v1',
+  authentication: {
+    site: {
+      cookie: {
+        options: {
+          domain: '.sparkpost.test'
+        }
+      }
+    }
+  },
   cookieConsent: {
     cookie: {
       options: {
--- a/src/actions/auth.js
+++ b/src/actions/auth.js
@@ -1,5 +1,6 @@
 import { sparkpostLogin } from '../helpers/http';
 import sparkpostApiRequest from 'src/actions/helpers/sparkpostApiRequest';
+import * as websiteAuth from 'src/actions/websiteAuth';
 import { getTfaStatusBeforeLoggedIn } from 'src/actions/tfa';
 import { showAlert } from 'src/actions/globalAlert';
 import { removeHerokuToolbar } from 'src/helpers/heroku';
@@ -21,6 +22,7 @@ export function login({ authData = {}, saveCookie = false }) {
   }
 
   return (dispatch) => {
+    dispatch(websiteAuth.login(saveCookie)); // Complete the website cookie set up process
     dispatch({
       type: 'LOGIN_SUCCESS',
       payload: authData
@@ -52,6 +54,9 @@ export function authenticate(username, password, rememberMe = false) {
     return sparkpostLogin(username, password, rememberMe)
       .then(({ data = {}} = {}) => {
         const authData = { ...data, username };
+
+        // Start website auth token cookie setup process
+        dispatch(websiteAuth.authenticate(username, password, rememberMe));
 
         return Promise.all([ authData, getTfaStatusBeforeLoggedIn({ username, token: authData.access_token })]);
       })
@@ -129,8 +134,11 @@ export function ssoCheck(username) {
 }
 
 export function refresh(token, refreshToken) {
-  const newCookie = authCookie.merge({ access_token: token, refresh_token: refreshToken });
-  return login({ authData: newCookie });
+  return (dispatch) => {
+    const newCookie = authCookie.merge({ access_token: token, refresh_token: refreshToken });
+    dispatch(websiteAuth.refresh());
+    return login({ authData: newCookie });
+  };
 }
 
 export function logout() {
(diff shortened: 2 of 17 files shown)


src/actions/auth.js
```diff
@@ -129,8 +134,11 @@ export function ssoCheck(username) {
 }
 
 export function refresh(token, refreshToken) {
-  const newCookie = authCookie.merge({ access_token: token, refresh_token: refreshToken });
-  return login({ authData: newCookie });
+  return (dispatch) => {
+    const newCookie = authCookie.merge({ access_token: token, refresh_token: refreshToken });
+    dispatch(websiteAuth.refresh());
+    return login({ authData: newCookie });
```
bkemper: Now that this is a thunk, `login` will need to be `dispatch`ed.

src/actions/auth.js
```diff
@@ -21,6 +22,7 @@ export function login({ authData = {}, saveCookie = false }) {
   }
 
   return (dispatch) => {
+    dispatch(websiteAuth.login(saveCookie)); // Complete the website cookie set up process
```
bkemper: If `authenticate` succeeds and `websiteAuth.authenticate` fails (which it probably won't), this would still get called.  What about moving `websiteAuth.login` to `websiteAuth.authenticate` and `websiteAuth.refresh`?

src/actions/auth.js
```diff
@@ -53,6 +55,9 @@ export function authenticate(username, password, rememberMe = false) {
       .then(({ data = {}} = {}) => {
         const authData = { ...data, username };
 
+        // Start website auth token cookie setup process
+        dispatch(websiteAuth.authenticate(username, password, rememberMe));
```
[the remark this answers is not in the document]
ewandennis: We want to avoid singling the old SPC tenant out for special treatment. Before we deploy though, we'll let devrel and accounts know.

src/actions/auth.js
ewandennis: Good catch.

src/actions/auth.js
ewandennis: `websiteAuth.login()` is called in `auth.login()` to make sure the website cookie is only set after TFA auth is complete, same as the app auth cookie. To move it, we'd have to have the websiteAuth reducer listen for a TFA success action. Doable but a tiny bit more complex than mirroring the app auth flow.

How about we have the websiteAuth reducer clear out its auth tokens on authenticate failure so that `websiteAuth.login()` sets the cookie to a known state?

ewandennis:
@bkemper I'm rebasing to bring in your config updates.

public/static/tenant-config/production.js
```diff
@@ -1,6 +1,15 @@
 window.SP = window.SP || {};
 window.SP.productionConfig = {
   apiBase: 'http://api.sparkpost.test/api/v1',
+  authentication: {
+    site: {
+      cookie: {
+        options: {
+          domain: '.sparkpost.test'
```
bkemper: Shouldn't production configuration use the newly set default?

src/actions/websiteAuth.js
```diff
@@ -0,0 +1,62 @@
...
+    }
+    dispatch({
+      type: 'WEBSITE_AUTH_COMPLETE'
+    });
+  };
+}
+
+export function refresh() {
+  return (dispatch, getState) => {
+    const { websiteAuth: { refreshToken }} = getState();
```
avigoldman: How can the websites refresh the tokens? Or are they depend on the app to do that?

avigoldman:
Some questions:
1. What grants will the website token have?
1. How can we tell which tenant the token is from?
1. How will it work if there someone is logged into both US and EU?

src/actions/websiteAuth.js
ewandennis: You could refresh tokens yourself. You'd need to follow the procedure [triggered here](https://github.com/SparkPost/2web2ui/blame/master/src/actions/helpers/sparkpostApiRequest.js/#L51) and [implemented here](https://github.com/SparkPost/2web2ui/blame/master/src/helpers/http.js/#L9).

public/static/tenant-config/production.js
ewandennis: (This is not production config - its the config used during local dev).
Production default is set in `src/config/default.js` `authentication.site.cookie.domain`.

ewandennis:
1. What grants will website auth tokens have  @colestrode ?
1. The `website_auth` cookie doesn't specify tenant at the moment, although it could. What would you use it for @avigoldman?
1. The `website_auth` cookie is set on `.sparkpost.com` so any tenant you authenticate against under that domain will overwrite any previous token. To retain US and EU tokens concurrently would require some thought and possibly more effort on the websites' parts. Is this a common thing outside the company, given our 1 account per customer policy  @avigoldman?

avigoldman:
2. I'd need to know the tenant to make any API calls to the correct host.
3. I'm ok with only one cookie existing (.i.e. keeping the overriding behavior) since I don't want to handle that multi-account case differently from the app.

PR #612, opened 2018-09-10
FE-528 UI-driven user-level SSO control

dan2305:


src/pages/users/EditPage.js
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    // Load until we have at least one user
+    if (_.isEmpty(users)) {
+      return <Loading />;
+    }
+
+    // Error if user is not in the user list
+    if (!user) {
+      return <RedirectAndAlert to="/account/users" alert={{ type: 'error', message: 'Unknown user???' }} />;
```
bkemper: How about "Unable to find, ${username}"?

src/pages/users/EditPage.js
```diff
@@ -0,0 +1,142 @@
...
+    this.props.history.push('/account/users');
+  };
+
+  deleteUser = () => {
+    const { deleteUser } = this.props;
+    const user = this.props.match.params.id;
+
+    return deleteUser(user)
+      .then(() => {
+        this.afterDelete();
```
bkemper: This happens after `render`, so the user has already been redirected, but a success and error snackbar.

PR #648, opened 2018-10-02
FE-485 Implement template for CodeBuild infrastructure creation

rajumsys:
This PR implements cloudformation template for building AWS CodeBuild project. 

- Implements a functional appspec.yml (needed for CodeBuild)
- Implements a fake buildspec.yml which is needed for CodeDeploy. This is put here just to glue to build process so that it's attached to build artifacts. This will be fixed FE-484.
- Updates build process to include buildspec.yml in the build artifact. 

Running:
- There is a great chance we can't run it ourselves. We need someone from SRE team to run on behalf us. See the runbook for instructions how to run it. 
- Currently, we'll have to trigger a build from AWS Console. When testing this branch, we'll either have to specify branch name or last commit hash (in this branch) to make sure correct codes are there. 

I'm going to create individual template for now. Upon creation of all templates, I'll see how can we create one template to run all at once.

### DIFF
--- a/ci/appspec.yml
+++ b/ci/appspec.yml
@@ -1,0 +1,19 @@
+version: 0.0
+os: linux
+files:
+  - source: /
+    destination: /opt/msys/app/phoenix
+permissions:
+  - object: /
+    pattern: "**"
+    owner: ubuntu
+    group: ubuntu
+hooks:
+  ApplicationStop:
+    - location: scripts/stop.sh
+      timeout: 10
+      runas: ubuntu
+  ApplicationStart:
+    - location: scripts/start.sh
+      timeout: 20
+      runas: ubuntu
--- a/ci/buildspec.yml
+++ b/ci/buildspec.yml
@@ -1,0 +1,18 @@
+version: 0.2
+
+phases:
+  install:
+    commands:
+      - npm install
+  pre_build:
+    commands:
+      - npm run test-ci
+      - rm -rf node_modules
+  build:
+    commands:
+      - npm install
+      - npm run build
+
+artifacts:
+  files:
+    - 'build/**/*'
--- a/ci/codebuild.yml
+++ b/ci/codebuild.yml
@@ -1,0 +1,85 @@
+---
+AWSTemplateFormatVersion: '2010-09-09'
+Description: Create Phoenix CodeBuild Infrastructures
+
+Resources:
+  ArtifactS3Bucket:
+    Type: AWS::S3::Bucket
+    Properties:
+      BucketName: phoenix-artifacts
+      AccessControl: Private
+      VersioningConfiguration:
+        Status: Suspended
+
+  CodeBuildIAMServiceRole:
+    Type: 'AWS::IAM::Role'
+    Properties:
+      RoleName: !Sub 'CodeBuildIAMServiceRole-${AWS::StackName}'
+      AssumeRolePolicyDocument:
+        Version: 2012-10-17
+        Statement:
+          - Effect: Allow
+            Principal:
+              Service:
+                - codebuild.amazonaws.com #todo limit to certain AZ?
+            Action:
+              - 'sts:AssumeRole'
+      Policies:
+        - PolicyName: PhoenixCodeBuildRolePolicy
+          PolicyDocument:
+            Version: 2012-10-17
+            Statement:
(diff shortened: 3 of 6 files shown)


ci/buildspec.yml
```diff
@@ -0,0 +1,18 @@
...
+  install:
+    commands:
+      - npm install
+  pre_build:
+    commands:
+      - npm run test-ci
+      - rm -rf node_modules
+  build:
+    commands:
+      - npm install
```
rajumsys: need to see why `npm ci` didn't work in codebuild

ci/codebuild.yml
```diff
@@ -0,0 +1,85 @@
+---
```
bizob2828: just curious why you are putting this in the repo? this is a one time run. we have a private repo that has been housing this stuff

ci/appspec.yml
```diff
@@ -0,0 +1,19 @@
+version: 0.0
+os: linux
+files:
+  - source: /
+    destination: /opt/msys/app/phoenix
+permissions:
```
bizob2828: so this is running on an ubuntu container but it's just outputting a tarball correct? you may want to deploy this as containers as our servers are amazon linux, could hit some issues with compiled libs

ci/codebuild.yml
```diff
@@ -0,0 +1,85 @@
...
+AWSTemplateFormatVersion: '2010-09-09'
+Description: Create Phoenix CodeBuild Infrastructures
+
+Resources:
```
bizob2828: this appears to only house building, are there plans to set up a pipeline for deploying? are you also planning on disabling travis once this is done?

ci/codebuild.yml
```diff
@@ -0,0 +1,85 @@
...
+        Image: aws/codebuild/nodejs:8.11.0
+        EnvironmentVariables:
+          - Name: TEMPLATE_BUCKET
+            Value:
+              Ref: ArtifactS3Bucket
+      Source:
+        Location: https://github.com/SparkPost/2web2ui
+        Type: GITHUB
+        BuildSpec: ci/buildspec.yml
+
```
bizob2828: If you put a ```
Triggers:
  Webhook: True
```
you get PR integration, you must add the engineering group as admins

ci/codebuild.yml
rajumsys: current plan:
- keep branch in travis
- yes, codepipeline will be implemented in this epic

ci/codebuild.yml
rajumsys: oh wow, i was looking for it under `Source`. I can add it here for now, but essentially this will be in CodePipeline.

ci/appspec.yml
rajumsys: this isn't implemented yet. this file is just a placeholder. i think instead of containerizing (or uploading to s3) this, we should just serve it from S3. I'll float this to team.

ci/buildspec.yml
rajumsys: need to check default version. i've npm 6x with node8x. but it might not be default.

ci/codebuild.yml
rajumsys: ah looks i need oauth config. i'll skip it here and implement in codepipeline.

ci/buildspec.yml
bizob2828: npm ci came in 6.2 or something, it's def not what ships with 8.11.0

ci/codebuild.yml
bizob2828: i wouldn't have 2 systems building.  codebuild can build branches and update PRs with the status

ci/codebuild.yml
bizob2828: look at the samples in the codepipeline scripts repo, we've solved this.  we just use an ecrypted key

ci/codebuild.yml
rajumsys: there will be one system. i mean, in this epic we're not changing how feature/hotfix branches  are build. just changing how master branch is building and deploying (bamboo to aws). once we've that working we can (probably will) move other branches too.

rajumsys:
Moving templates to it's [own repo](https://github.com/SparkPost/ux-cloudformation). Closing this PR.

PR #654, opened 2018-10-05
Wired up auditmated to help fix security vulnerabilities

bizob2828:
by running `npm audit fix` several times(not sure why I had to do that) it went from ```found 612 vulnerabilities (466 low, 116 moderate, 27 high, 3 critical)``` to 1 high


package.json
```diff
@@ -51,7 +52,7 @@
     "@sparkpost/matchbox-icons": "^1.1.2",
     "axios": "^0.17.1",
     "bowser": "2.0.0-alpha.4",
-    "chart.js": "^2.6.0",
+    "chart.js": "^2.7.2",
```
bkemper: Shouldn't this be uninstalled?

package.json
```diff
@@ -75,32 +76,33 @@
     "react-dom": "^16.3.2",
     "react-dropzone": "^5.0.1",
     "react-helmet": "^5.2.0",
-    "react-instantsearch": "^5.0.1",
+    "react-instantsearch": "^5.3.1",
```
bkemper: Need to confirm this doesn't break our support modal.

package.json
```diff
@@ -75,32 +76,33 @@
...
     "react-load-script": "0.0.6",
     "react-loadable": "^5.3.1",
     "react-qr-svg": "^2.1.0",
     "react-recaptcha": "^2.3.8",
     "react-redux": "^5.0.5",
     "react-router-dom": "^4.1.1",
     "react-timeago": "^4.1.9",
-    "recharts": "^1.0.0-alpha.2",
+    "recharts": "^1.3.1",
```
bkemper: Need to confirm this doesn't break our reports.

package.json
```diff
@@ -75,32 +76,33 @@
...
     "redux": "^3.6.0",
     "redux-form": "^7.2.3",
     "redux-thunk": "^2.2.0",
     "reselect": "^3.0.1",
-    "source-map-explorer": "^1.5.0",
+    "source-map-explorer": "^1.6.0",
```
bkemper: Shouldn't this be a devDependency?

package.json
```diff
@@ -113,8 +115,8 @@
     "css-loader": "^1.0.0",
     "cssnano": "^4.0.3",
     "dotenv": "^6.0.0",
-    "enzyme": "^3.1.0",
-    "enzyme-adapter-react-16": "^1.0.2",
+    "enzyme": "^3.7.0",
```
bkemper: This is what is breaking all the snapshot tests.

package.json
```diff
@@ -113,8 +115,8 @@
...
+    "enzyme-adapter-react-16": "^1.6.0",
```
bkemper: I could be reading this wrong, but it looks like we are using the wrong adapter.  Looks like we should be using 16.3 since we are still on React v16.3.

https://www.npmjs.com/package/enzyme-adapter-react-16

package.json
```diff
@@ -134,14 +136,14 @@
...
+    "jest-enzyme": "^6.1.2",
     "jest-in-case": "^1.0.2",
     "jest-junit": "^5.1.0",
-    "lint-staged": "^7.0.1",
+    "lint-staged": "^7.3.0",
     "mdx-loader": "^1.1.0",
     "mdxc": "^1.1.1",
     "moment-timezone": "^0.5.14",
-    "node-sass": "^4.5.3",
+    "node-sass": "^4.9.3",
```
bkemper: @jonambas What relies on `node-sass`?

package.json
bkemper: I confirmed this can be removed.

package.json
jonambas: postcss, sass-loader I _think_

package.json
jonambas: oof 
screen shot 2018-10-09 at 10 26 16 am

bkemper:
Rebased and addressed all review feedback.  This is ready for final review.

PR #662, opened 2018-10-17
Babel 7

jonambas:


package.json
```diff
@@ -32,12 +32,25 @@
...
+      [
+        "@babel/plugin-proposal-decorators",
+        {
+          "legacy": true
+        }
+      ],
+      "@babel/plugin-proposal-function-sent",
+      "@babel/plugin-proposal-export-namespace-from",
+      "@babel/plugin-proposal-numeric-separator",
+      "@babel/plugin-proposal-throw-expressions"
```
bkemper: Are these the plugins from state-2?

PR #663, opened 2018-10-17
TR-552: List Snippets

bkemper:
To test, you will need to navigate directly to `/snippets` and the edit page has not been implemented, so links will go to a cannot be found page.

screen shot 2018-10-19 at 9 31 22 am


bkemper:
@ischer This is ready for review.

src/reducers/snippets.js
```diff
@@ -1,4 +1,22 @@
...
-;
+export default (
+  state = {
+    items: [], // cannot normalize, id is not unique, needs to consider subaccount_id
+    loading: false
+  },
+  action = {}
+) => {
+  switch (action.type) {
+    // this is ephemeral state that redux-form temporarily handles
```
ischer: why is this?

src/components/apiErrorBanner/index.js
```diff
@@ -0,0 +1 @@
+export { default } from './ApiErrorBanner';
```
ischer: is this + the actionPopover index file just added to make it so you can include the component without using components/index.js? does this help with testing or is it just preference, or something else?

bkemper: This is just a preference.  

The import is a little shorter.
```
// This is ...
import ApiErrorBanner from 'src/components/apiErrorBanner';

// Instead of, this ...
import ApiErrorBanner from 'src/components/apiErrorBanner/ApiErrorBanner';
```

Also, I find it to be helpful with hiding implementation details.  For example, the [ContentEditor](https://github.com/SparkPost/2web2ui/blob/master/src/components/contentEditor/index.js) should be async loaded by default or [SubaccountSection](https://github.com/SparkPost/2web2ui/blob/master/src/components/subaccountSection/index.js) needs to be connected to redux store, so the container is loaded by default.

src/pages/snippets/ListPage.js
```diff
@@ -0,0 +1,126 @@
...
+        }}
+      >
+        {error ? (
+          <ApiErrorBanner
+            errorDetails={error.message}
+            message="Sorry, we seem to have had some trouble loading your snippets."
+            reload={getSnippets}
+          />
+        ) : (
+          <TableCollection
```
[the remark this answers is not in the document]
ischer: ohhh, ok cool - good to know!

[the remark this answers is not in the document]
bkemper: Addressed... 146f52a8c

src/pages/snippets/ListPage.js
```diff
@@ -0,0 +1,134 @@
...
+      subaccount_id || shared_with_subaccounts
+    )
+  },
+  updateAt: {
+    label: 'Last Updated',
+    sortKey: ({ created_at, updated_at }) => updated_at || created_at
+  }
+};
+
+export const CELLS = {
```
ischer: is there any reason why this should not be a part of the `ListPage` class? delete ticket i'm working through atm will need to add an action for delete where it may be useful to have access to `this.props`.

bkemper: This component was starting to feel too large and testing felt awkward because I should test these individually which would require accessing them from `wrapper.instance()`.  I think I am going to break them out as a separate files.

src/reducers/snippets.js
bkemper: That comment was only for the "CREATE_SNIPPET" actions.  Redux form handles the storage of the form data and the submit event handler uses the `createSnippet` action to determine if the snippet was created successfully or not.

src/pages/snippets/ListPage.js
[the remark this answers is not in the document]
bkemper: I made three changes to sorting.

1. Instead of setting `sortKey: ({ field1, field2 }) => [field1, field2]`, I changed to return `field1 || field2`.  The reason is lodash `orderBy` is being used under the hood which will take the array and sort by the first value then the second value.  The problem with that is it groups the `undefined` values which will not provide the desired result.
2. The matching for `defaultSortColumn` is on `sortKey` which forced me to `defaultSortColumn={HEADERS.name.sortKey}`.
3. `orderBy` produces a sort result of lowercase then uppercase.  This produced a weird result, so I resolved it by using `toLowerCase` in the `sortKey`.

src/pages/snippets/ListPage.js
```diff
@@ -0,0 +1,126 @@
...
+    this.props.getSnippets();
+  }
+
+  Row = (data) => {
+    if (this.props.hasSubaccounts) {
+      return [
+        <Name {...data} />,
+        <Subaccount {...data} />,
+        <UpdatedAt {...data} />,
+        <Actions {...data} />
```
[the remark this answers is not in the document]
bkemper: I filed https://github.com/SparkPost/matchbox/pull/146.

bkemper:
@ischer I addressed all your feedback.  Ready for final review.

PR #664, opened 2018-10-19
FE-561: Unhandled promise rejection in unit tests

ewandennis:
This is the same old bug-in-promise-chain issue.

 - Chased down a bug in `SetupSending.test.js` mock function.
 - Added an unhandled reject handler (!) to Jest setup.

### Testing
1. Run `npm test` on master to see the unhandled rejection message fly by without useful diagnostic context.
1. Run `npm test` on FE-561 to see no message.
1. Undo the fix in `SetupSending.test.js and tun `npm test` on FE-561 to see a more useful rejection message.

Note that these messages are emitted async so we can't easily diagnose them from their location in the output stream.

FWIW, Jest _should_ be reporting unhandled rejections and failing the related tests where possible

Hint: use `CI=true npm test` to run the whole suite in about a minute with no frills output.

### DIFF
--- a/config/jest/setup.js
+++ b/config/jest/setup.js
@@ -47,3 +47,9 @@ jest.mock('moment', () => {
 
 Object.defineProperty(global.navigator, 'userAgent', { value: 'node.js', configurable: true });
 Object.defineProperty(global.navigator, 'language', { value: 'en-US', configurable: true });
+
+// Show a stack track for unhandled rejections to help
+// track them down.
+process.on('unhandledRejection', (reason) => {
+	console.log(reason)
+});
--- a/src/pages/sendingDomains/components/tests/SetupSending.test.js
+++ b/src/pages/sendingDomains/components/tests/SetupSending.test.js
@@ -28,7 +28,7 @@ describe('Component: SetupSending', () => {
         status
       },
       verifyDkimLoading: false,
-      verifyDkim: jest.fn(() => Promise.resolve()),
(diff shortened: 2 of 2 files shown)


config/jest/setup.js
```diff
@@ -47,3 +47,9 @@ jest.mock('moment', () => {
 
 Object.defineProperty(global.navigator, 'userAgent', { value: 'node.js', configurable: true });
 Object.defineProperty(global.navigator, 'language', { value: 'en-US', configurable: true });
+
+// Show a stack track for unhandled rejections to help
+// track them down.
+process.on('unhandledRejection', (reason) => {
+	console.log(reason)
```
bkemper: For me, the `console.log` message gets swallowed in watch mode.

screen shot 2018-10-19 at 8 12 00 am

bkemper:
Ugh.  This could have been surfaced by calling the event handler directly and waiting for the promise, but then we are assuming the button and event handler are wired up correctly.

```
it('should call verifyDkim on click', async () => {
  await wrapper.instance().verifyDomain();
  expect(props.verifyDkim).toHaveBeenCalledTimes(1);
});
```

config/jest/setup.js
rajumsys: in case of promise rejection, we can force fail it by using a fail test here. like `expect(true).toBe(false)` or something better than that. point is, when this code path is executed, we can cause the failure and force investigation.

PR #667, opened 2018-10-21
FE-560 Show template loading error

rajumsys:
Show message when error in loading template

Test:
- In `master`, try visiting a non-existing template edit url. You can change directly in URL. It would redirect to list page without any message. However, doing same in this branch would show a message.


rajumsys:
> Looks good. Do we need to do the same for the published page?

As published page is visited from draft page, probably it's not needed. However, I went ahead and added it. Let me know if you've any further feedback.

PR #670, opened 2018-10-22
FE-543 - Add a note about removing password upon enabling single sign on

rajumsys:
After doing this I feel like we should show this only when they check it on. so here is my thought:

Status: Unchecked
Action: Check
Note: show the note as in this PR

Status: Checked
Action: Uncheck
Note: <You'll need to reset your password....> Possibly with an action that sends password reset email directly from here rather than visiting the log in page.

### DIFF
--- a/src/pages/users/EditPage.js
+++ b/src/pages/users/EditPage.js
@@ -83,8 +83,8 @@ export class EditPage extends Component {
     }
 
     const ssoHelpText = isAccountSingleSignOnEnabled
-      ? <span>Enabling single sign-on will remove your password. <br/>To use password again, you'll need to reset it from Log In page</span>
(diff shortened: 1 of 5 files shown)


src/pages/users/EditPage.js
```diff
@@ -82,6 +82,10 @@ export class EditPage extends Component {
       });
     }
 
+    const ssoHelpText = isAccountSingleSignOnEnabled
+      ? <span>Enabling single sign-on will remove your password. <br/>To use password again, you'll need to reset it from Log In page</span>
```
ewandennis: Since this is the user edit page, the person making the change may not be the current user. How about this:

> Enabling Single sign-on will delete this user's password. If they switch back to password-based authentication, they'll need to reset their password on login.

rajumsys: Yeah makes sense. 2nd sentence, however, has same symptom as they themselves can't re-enable it. but i'll go ahead.

PR #672, opened 2018-10-23
TR-551: Delete a Snippet

ischer:


src/pages/snippets/components/SnippetCollection.js
```diff
@@ -0,0 +1,80 @@
...
+import SubaccountTableData from './SubaccountTableData';
+import UpdatedAtTableData from './UpdatedAtTableData';
+
+const filterBoxConfig = {
+  show: true,
+  exampleModifiers: ['id', 'name'],
+  itemToStringKeys: ['name', 'id', 'subaccount_id']
+};
+
+export class SnippetCollection extends Component {
```
bkemper: This `export` can be removed.  Or, `export default` this and remove the other.

src/pages/snippets/components/SnippetCollection.js
```diff
@@ -0,0 +1,80 @@
...
+
+    const columns = this.props.hasSubaccounts
+      ? [HEADERS.name, HEADERS.subaccount, HEADERS.updateAt, HEADERS.actions]
+      : [HEADERS.name, HEADERS.updateAt, HEADERS.actions];
+
+    return columns;
+  }
+
+  getRowData = (data) => {
+    const actionsData = { ...data, toggleDelete: this.props.toggleDelete };
```
bkemper: Clever!

src/pages/snippets/components/tests/SnippetCollection.test.js
```diff
@@ -0,0 +1,55 @@
...
+  });
+
+  it('should render row data properly', () => {
+    const row1 = wrapper.instance().getRowData(props.snippets[0]);
+    expect(row1).toMatchSnapshot();
+    const row2 = wrapper.instance().getRowData(props.snippets[1]);
+    expect(row2).toMatchSnapshot();
+    const row3 = wrapper.instance().getRowData(props.snippets[2]);
+    expect(row3).toMatchSnapshot();
+  });
```
bkemper: This test is not necessary.  The table data components are already tested individually.  Instead, we should add coverage for `hasSubaccounts` prop and confirm that the subaccount column is optionally rendered.

src/reducers/tests/snippets.test.js
```diff
@@ -2,10 +2,6 @@ import snippetsReducer from '../snippets';
 import cases from 'jest-in-case';
 
 describe('Snippets Reducer', () => {
-  it('returns default state', () => {
-    expect(snippetsReducer()).toMatchSnapshot();
-  });
```
bkemper: Shouldn't we test the `default` case?

src/reducers/snippets.js
```diff
@@ -1,20 +1,36 @@
...
+  switch (type) {
     // this is ephemeral state that redux-form temporarily handles
     // case 'CREATE_SNIPPET_*':
 
     case 'GET_SNIPPETS_FAIL':
-      return { error: action.payload.error, loading: false };
+      return { ...state, error: payload.error, loading: false };
     case 'GET_SNIPPETS_PENDING':
-      return { loading: true };
+      return { ...state, loading: true };
```
bkemper: To allow for a retry if loading fails, `error` will need to be set to `null`.

src/reducers/snippets.js
```diff
@@ -1,20 +1,36 @@
...
     case 'GET_SNIPPETS_SUCCESS':
-      return { items: action.payload, loading: false };
+      return { ...state, items: payload, loading: false };
+
+    case 'DELETE_SNIPPET_FAIL':
+      return { ...state, error: payload.error, deletePending: false };
```
bkemper: Sharing `error` with `GET_SNIPPETS` would display a loading error banner.  If we need to persist the error, we should name the key something different.

Also, it is important to mention that by default `sparkpostApiRequest`s display a snackbar if the action fails -- [see here](https://github.com/SparkPost/2web2ui/blob/master/src/actions/helpers/sparkpostApiRequest.js#L104-L111).

src/reducers/tests/snippets.test.js
```diff
@@ -25,4 +21,28 @@ describe('Snippets Reducer', () => {
...
+      meta: { data: { id: 'deleteMe' }}
+    },
+    'when pending': {
+      type: 'DELETE_SNIPPET_PENDING',
+      meta: { data: { id: 'deleteMe' }}
+    },
+    'when success': {
+      type: 'DELETE_SNIPPET_SUCCESS',
+      payload: [
+        { id: 'example-snippet', name: 'Example Snippet' }
```
bkemper: Should the `payload` match the `meta`?

src/reducers/tests/snippets.test.js
```diff
@@ -25,4 +21,28 @@ describe('Snippets Reducer', () => {
...
+      ],
+      meta: { data: { id: 'deleteMe' }}
+    }
```
bkemper: It would be good to provide a test case when deleting a snippet with a subaccountId.

src/actions/snippets.js
```diff
@@ -38,3 +38,15 @@ export const getSnippets = () => (
...
 );
+
+export const deleteSnippet = ({ id, subaccountId }) => (
+  sparkpostApiRequest({
+    type: 'DELETE_SNIPPET',
+    meta: {
+      method: 'DELETE',
+      url: `/snippets/${id}`,
+      headers: setSubaccountHeader(subaccountId),
+      data: { id, subaccountId }
```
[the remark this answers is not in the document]
ischer: that sounds like the better option to me - i'll switch that over

src/actions/tests/snippets.test.js
```diff
@@ -61,4 +62,24 @@ describe('Snippet Actions', () => {
...
+          id: 'test-snippet'
+        })
+      )
+    },
+    {
+      name: 'when assigned to subaccount',
+      action: () => (
+        deleteSnippet({
+          id: 'test-snippet',
+          subaccount_id: 101
```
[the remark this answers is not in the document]
ischer: good catch!

src/pages/snippets/ListPage.js
```diff
@@ -54,49 +78,16 @@ export default class ListPage extends React.Component {
...
-              exampleModifiers: ['id', 'name'],
-              itemToStringKeys: ['name', 'id', 'subaccount_id']
-            }}
-            getRowData={this.Row}
-            rows={snippets}
-            pagination
-          />
-        )}
+        {error ? this.renderError() : this.renderCollection()}
+        <DeleteModal
```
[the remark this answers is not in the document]
ischer: it probably makes more sense to move it into SnippetCollection. i'll take a look at that

src/reducers/snippets.js
ischer: does that mean i would not need to set an error here at all, unless i'm explicitly using it somewhere else outside of that snackbar? (that's the notification that pops up?)

src/reducers/tests/snippets.test.js
ischer: so, i was getting some errors about attempting to destructure (in tests only) and the other snapshots looked like they covered all of the default cases. if you'd like, i can try to take another shot at getting past that bug to get this covered explicitly

src/reducers/tests/snippets.test.js
ischer: i actually thought i removed the payload here. my bad. i shouldn't need that with delete.

src/pages/snippets/components/ActionsTableData.js
```diff
@@ -3,13 +3,19 @@ import ActionPopover from 'src/components/actionPopover';
...
         Component: PageLink,
         content: 'Edit',
-        to: `/snippets/edit/${id}${setSubaccountQuery(subaccount_id)}`
+        to: `/snippets/edit/${id}${setSubaccountQuery(subaccount_id)}`,
+        section: 1
+      },
+      {
+        content: 'Delete',
+        section: 2,
+        onClick: () => toggleDelete(id, subaccount_id)
```
bkemper: This action needs authorization and should only be `visible` for users that `hasGrants('templates/modify')`.

src/pages/snippets/components/SnippetCollection.js
ischer: need to dig deeper into this, but doing either of these seems to break stuff.

src/reducers/tests/snippets.test.js
bkemper: It would probably take an action like the Redux @@init, but don't worry about reinstating it.

PR #674, opened 2018-10-24
TR-641: Support for /labs actions

bkemper:
This is the prep work needed to support `/api/labs` endpoints.  A follow-up PR will be submitted to change the snippet actions.


bkemper:
I popped the last commit and turned this into just a prep ticket for TR-641.  The actual work for TR-641 will be in a later PR.

PR #677, opened 2018-10-25
EO-324 Polling Context

jonambas:
Usage:
```js
import { PollContext } from 'src/context/Poll';

<PollContext.Consumer>
  {({ startPolling, stopPolling }) =>(
    <button onClick={() => startPolling(action)} />
    <button onClick={() => stopPolling(action.key)} />
  )}
</PollContext.Consumer>

// Action model:
{
  // A string to represent this action
  key: 'identifier',
  
  // Time in ms between each poll
  interval: 1000,

  // Number of maximum attempts, -1 for no limit (default)
  maxAttempts: 5,

  // The function to call at every interval until duration has reached
  action: () => () 
}
```

Built for:  https://github.com/SparkPost/sparkpost-admin-api-documentation/pull/287

### DIFF
--- a/src/context/Poll.js
+++ b/src/context/Poll.js
@@ -8,10 +8,10 @@ import React, { createContext, Component } from 'react';
  *
  * @example
  * <PollContext.Consumer>
- *   {({ startPolling, stopPolling }) =>(
- *    <button onClick={() => startPolling(action)} />
+ *   {({ startPolling, stopPolling }) => [
+ *    <button onClick={() => startPolling(action)} />,
  *    <button onClick={() => stopPolling(action.key)} />
- *   )}
+ *   ]}
  * </PollContext.Consumer>
  *
  * Action model:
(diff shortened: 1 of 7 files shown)


src/context/Poll.js
```diff
@@ -0,0 +1,102 @@
...
+class Poll extends Component {
+  state = defaultContext;
+
+  componentDidMount() {
+    this.setState({
+      actions: {},
+      startPolling: this.start,
+      stopPolling: this.stop
+    });
+  }
```
bkemper: Instead of waiting for the component to render, it might be better to declare this closer to the handoff.

```
state = {
  actions: {}
}

render() {
  const value = {
    actions: this.state.actions,
    startPolling: this.start,
    stopPolling: this.stop
  }

  return (
    <PollContext.Provider value={value}>
```

src/context/tests/Poll.test.js
```diff
@@ -0,0 +1,54 @@
...
+  it('should start and finish polling', async () => {
+    const { startPolling } = wrapper.props().value;
+    startPolling(testAction);
+
+    // Check polling state
+    expect(wrapper.props().value.actions.testAction).toMatchSnapshot();
+
+    // Simulate duration has reached
+    mockMoment.diff = jest.fn().mockReturnValue(200);
+    await delay(200);
```
bkemper: Instead of delaying, we could wrap the `setTimeout`'s with a `Promise` and await the `Promise`.

PR #684, opened 2018-10-30
EO-331 Recipient Verification Results

jonambas:
###### Pre Submit
screen shot 2018-11-05 at 10 11 34 am

###### Processing / Polling State
screen shot 2018-11-05 at 10 11 41 am

###### Processing Complete
screen shot 2018-11-05 at 10 12 12 am

### DIFF
--- a/src/context/tests/Poll.test.js
+++ b/src/context/tests/Poll.test.js
@@ -29,15 +29,19 @@ describe('Poll Provider', () => {
     startPolling(testAction);
     await next();
     await next();
-
+    await next();
     expect(wrapper.state('actions').testAction).toMatchSnapshot();
   });
 
-  it('should start and stop polling when stopPolling is invoked', async () => {
+  it('should stop and not continue polling when stopPolling is invoked', async () => {
     const { startPolling, stopPolling } = wrapper.props().value;
     startPolling(testAction);
     stopPolling(testAction.key);
-    expect(wrapper.state('actions').testAction).toMatchSnapshot();
+    const actions = wrapper.state('actions').testAction;
+    expect(actions).toMatchSnapshot();
+
+    await next();
+    expect(wrapper.state('actions').testAction).toBe(actions);
   });
 
   it('should not stop if it isnt polling', async () => {
--- a/src/context/tests/Poll.test.js
+++ b/src/context/tests/Poll.test.js
@@ -10,7 +10,8 @@ describe('Poll Provider', () => {
     key: 'testAction',
     action: jest.fn(() => Promise.resolve()),
     interval: 50,
-    maxAttempts: 3
+    maxAttempts: 3,
+    maxConsecutiveErrors: 2
   };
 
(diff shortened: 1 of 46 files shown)


src/context/tests/Poll.test.js
```diff
@@ -48,8 +48,37 @@ describe('Poll Provider', () => {
   it('should not start if its already polling', async () => {
     const { startPolling } = wrapper.props().value;
     startPolling(testAction);
-    next();
+    await next();
     startPolling(testAction);
     expect(wrapper.state('actions').testAction).toMatchSnapshot();
   });
+
+  it('should increment consecutive errors', async () => {
```
rajumsys: given you aren't big fan of snapshot, can we just assert the error count rather than snapshot? :P

src/context/tests/Poll.test.js
```diff
@@ -48,8 +48,37 @@ describe('Poll Provider', () => {
...
+    const { startPolling } = wrapper.props().value;
+    testAction.action.mockReturnValue(Promise.reject('error'));
+    startPolling(testAction);
+    await next();
+    expect(wrapper.state('actions').testAction).toMatchSnapshot();
+  });
+
+  it('should reset error count on a success', async () => {
```
rajumsys: these will also be easier to follow if just errorCount is asserted than snapshot.

src/pages/recipientVerification/components/SingleAddressForm.js
```diff
@@ -21,7 +20,7 @@ export class SingleAddressForm extends Component {
     return (
       <Fragment>
         <form onSubmit={handleSubmit(this.singleAddressForm)}>
-          <p className={styles.Paragraph}>Verify an email address to determine if it is a deliverable email address or a rejected, undeliverable email address.</p>
+          <p>Verify an email address to determine if it is a deliverable email address or a rejected, undeliverable email address.</p>
```
rajumsys: can't comment in line 29. a max valid length of email address can be 254. is there any restriction in API?

rajumsys:
Mostly looks good. Left a couple of minor comments. 

We might need a file validator. I uploaded a pdf which it accepted (it correctly allows csv by default though) and processed. 

```

%ï¿½ï¿½ï¿½ï¿½ | Invalid Email
-- | --
3 0 obj | Invalid Email
<< | Invalid Email
/Filter [/FlateDecode] | Invalid Email
/Length 9971 | Invalid Email
>> stream | Invalid Email
...
```
API seems trying to process an invalid csv. So need such a check both in API and UI.

src/pages/recipientVerification/components/SingleAddressForm.js
jonambas: we started doing `eg` because people thought the example was an actual input value

rajumsys: ah, ok. let's leave it like then.

PR #687, opened 2018-11-03
Merge greenkeeper prs

rajumsys:
This closes:
- all [`greenkeeper`](https://github.com/SparkPost/2web2ui/pulls?q=is%3Apr+is%3Aopen+label%3Agreenkeeper) prs those are currently passing ( except `mdx-loader` -  #682)
- #658 
- #680 
- #585 

Also covers:
- bumping npm requirements to `6.4.1`
- fixing HMR and error overlay
- remove eslint-plugin-flowtype package

Notes: 
- ~Report summary page showing `Warning: Failed prop type: Invalid prop scale supplied to XAxis` if I use latest version. This is confusing. Their doc says `utcTime` is valid value but in source i see it's `utc`. I've now restored to original version we had but still getting same warning.~ Removed the prop and default will be used by we provide custom ticks anyway. 
- mdx-loader needs to revisit. it's current state is confusing. seems library is renamed. but if I just update it to 2.0.0, it breaks UI (notification center)
- react-dropzone removed `disablePreview ` prop hence removed. But not sure why component name is renamed from `<d>` to `<n>`. 
- downshift renamed `default*` props to `initial*`. 

**How to test?**
- All tests pass w/o any warnings
- Random app uses
- Test some key ux including login, signup
- Hot module reloading works, react error overlay (trigger some syntax error to see)
- Production bundle creation works (Travis will verify this anyway)
- Review metrics/charts


rajumsys:
@ewandennis fixed package-lock.json by running `npm install`, removed `scale` (default value is `auto`) and seems everything is fine. as this component doesn't have any tests (found out after nothing broke after i removed that prop), i've added a simple snapshot test.

rajumsys:
> tests failing on travis but passing locally. checking

it's not clear why but seems `_.identity` isn't working correctly in CI mode with coverage. So i've created a local `identity` method and seems it's working

.travis.yml
```diff
@@ -6,7 +6,7 @@ cache:
     - build # shows bundle size diff per file/chunk
     - "$HOME/.npm" # to avoid fetching modules from npm every time
 before_install: npm install -g greenkeeper-lockfile@1
-install: npm install --loglevel http # change to npm ci when travis runs npm 5.7+ (currently 5.6)
+install: npm ci --loglevel http
```
bkemper: Should we bump our npm engines constant?

https://github.com/SparkPost/2web2ui/blob/master/package.json#L7

rajumsys:
@ewandennis so far it looks like 
- This _.identity issue isn't related to anything in this PR. It exists in `master` too. Only reason it was caught in this PR because of the test. Moving test to `master` causes same problem.
- The problem was there for sometime. I've checked out a version (https://github.com/SparkPost/2web2ui/tree/cd9ff267f5130b4ee15081b29dc1cb33ddc23c7b) before our last month auditmated changes and it happens there too. 
- This might be known to some of us. `src/helpers/sortMatch.js` uses a similar implementation (use local implementation than lodash's)

These still doesn't answer why it's causing but none of the hypothesis are making sense and not sure if it's worth further investigation.

package.json
```diff
@@ -121,27 +121,27 @@
     "enzyme-adapter-react-16.3": "^1.3.0",
     "enzyme-to-json": "^3.3.1",
     "eslint": "^5.0.0",
-    "eslint-config-react-app": "^2.0.1",
+    "eslint-config-react-app": "^3.0.2",
     "eslint-config-sparkpost": "^1.3.2",
     "eslint-loader": "^2.1.0",
-    "eslint-plugin-flowtype": "^2.39.1",
+    "eslint-plugin-flowtype": "^3.1.0",
```
jonambas: we can remove this

package.json
```diff
@@ -152,7 +152,7 @@
     "prettier": "^1.14.2",
     "promise": "^8.0.1",
     "react-dev-utils": "^5.0.2",
-    "react-error-overlay": "^4.0.0",
+    "react-error-overlay": "^5.0.3",
```
jonambas: Can't really verify this works because of the hot module issue. But I think this version is meant for CRA v2, react-dev-utils v6, webpack 4, and probably a few more

rajumsys: a good point. i'll check again by disabling HMR

jonambas:
Should we specify lodash in our package?

rajumsys:
> Should we specify lodash in our package?

Given we use it, I believe we should. It's currently being available through other dependency.

PR #690, opened 2018-11-07
AC-415: enforce TFA at the account-level 

ewandennis:
Headline: Added an `EnforceTfaPanel` to `/account/settings` protected by an account UI option.

**WARNING**: Do not use a shared account for testing, since enforcing TFA kicks all other active sessions and will send a notification to all users on environments with functional email delivery.

Note: This flow offers to disable account-level SSO since it's mutually exclusive from enforced TFA.

## Testing
Give your account access:
```sh
echo '{"options": {"ui": {"tfaRequired": true}}}' | http put ENV_API_HOST:8888/api/v1/account/control x-msys-tenant:$TENANT x-msys-customer:$CID
```
Review `/account/settings` :)


src/pages/account/components/EnforceTfaPanel/TogglePanel.js
```diff
@@ -0,0 +1,43 @@
...
+
+export const TogglePanel = ({ tfaRequired, toggleTfaRequired }) => (
+  <Panel
+    title="Two-factor Authentication"
+    actions={[
+      {
+        color: 'orange',
+        content: 'Learn more',
+        component: ExternalLink,
+        to: LINKS.MANDATORY_TFA
```
nancyhabecker: this isn't defined, so the button doesn't do much. will there be a separate ticket to set `MANDATORY_TFA`?

src/pages/account/components/EnforceTfaPanel/EnforceTfaPanel.js
```diff
@@ -0,0 +1,149 @@
...
+                Enforcing two-factor authentication account-wide will have the
+                following effects:
+              </p>
+              <ul>
+                <li>
+                  All users without two-factor authentication enabled will be
+                  forced to login to the UI again and enable it on login.
+                </li>
+                <li>
+                  All users will be sent an email informing them of this change.
```
nancyhabecker: "All users without two-factor authentication enabled," technically.

Is it worth saying that API keys will not be affected? @haleysolomon

haleysolomon: Yes, I think it's worth stating that - good suggestion.

src/pages/account/components/EnforceTfaPanel/EnforceTfaPanel.js
```diff
@@ -0,0 +1,149 @@
...
+    return (
+      <React.Fragment>
+        <TogglePanel
+          tfaRequired={tfaRequired}
+          toggleTfaRequired={this.toggleTfaRequired}
+        />
+        <ConfirmationModal
+          open={disableSsoModal}
+          confirming={ssoUpdatePending}
+          title="Single sign-on must be disabled before enforcing two-factor authentication on this account. Disable?"
```
nancyhabecker: is it worth having a similar warning when an account tries to enable sso but they have tfa required? (right now, the api returns a 4xx and you get a "something went wrong." / "Cannot enable SSO with TFA required" error in the UI, which i think is fine, i just thought it was interesting that one way had a warning but the other didn't.)

src/pages/account/components/EnforceTfaPanel/EnforceTfaPanel.js
```diff
@@ -0,0 +1,149 @@
...
+    } else {
+      this.setState({ enableModal: true });
+    }
+  };
+
+  disableSso = () => {
+    const { ssoCert, ssoProvider } = this.props;
+    this.props
+      .updateAccountSingleSignOn({
+        cert: ssoCert,
```
nancyhabecker: you can just pass `enabled: false` (`cert` and `provider` will be left alone).

src/pages/account/components/EnforceTfaPanel/TogglePanel.js
ewandennis: I was just waiting for the docs update. I'm setting `MANDATORY_TFA` to `https://www.sparkpost.com/docs/my-account-and-profile/enabling-two-factor-authentication/` as per [this PR](https://github.com/SparkPost/support-docs/pull/300/files).

src/pages/account/components/EnforceTfaPanel/EnforceTfaPanel.js
ewandennis: I'm not sure I agree with this. _We_ relate users and API keys but I'm not sure accountholders do. Lets discuss.

src/pages/account/components/EnforceTfaPanel/EnforceTfaPanel.js
ewandennis: Good idea. We could just replace the single sign-on setup panel with a message to that effect.

ewandennis: On reflection, I think I went too far in adding this disable-sso-to-enforce-tfa step. Both settings appear on the same page so it'd be simplest just to let folks know the 2 are mutually exclusive and let them make the changes. Also up for discussion.

src/pages/account/components/EnforceTfaPanel/EnforceTfaPanel.js
```diff
@@ -0,0 +1,118 @@
...
+import { Panel } from '@sparkpost/matchbox';
+import PanelLoading from 'src/components/panelLoading/PanelLoading';
+import { ConfirmationModal } from 'src/components/modals';
+import TogglePanel from './TogglePanel';
+import ExternalLink from 'src/components/externalLink/ExternalLink';
+import { LINKS } from 'src/constants';
+
+export class EnforceTFAPanel extends React.Component {
+  state = {
+    enforced: false,
```
nancyhabecker: how is `enforced` used?

PR #691, opened 2018-11-07
Hotfix/tr 646

ischer:


ischer:
per comments on TR-646, the scrolling issue is no longer present (maybe this behavior changed in recent browser updates?)

src/pages/abTesting/components/Section.module.scss
```diff
@@ -6,4 +6,13 @@
   @media screen and (min-width: breakpoint(medium)) {
     margin-top: rem(-42);
   }
+
+  input[type=number]::-webkit-inner-spin-button,
+  input[type=number]::-webkit-outer-spin-button {
+    -webkit-appearance: none;
```
bkemper: We can rely on autoprefixer to add browser prefixes -- see it in action, https://autoprefixer.github.io.

```
input[type=number]::-webkit-inner-spin-button,
input[type=number]::-webkit-outer-spin-button {
  appearance: none;
}

input[type=number] {
  appearance: textfield;
}
```

bkemper: Also, what do you think about moving this to index.scss to set our standard for our whole app.

src/pages/abTesting/components/content/VariantsContent.js
```diff
@@ -2,11 +2,18 @@ import React from 'react';
 
 import { PageLink } from 'src/components';
 
-const VariantsContent = () => (
+const VariantsContent = ({ formValues }) => (
   <div>
     <p>The templates you've selected will appear here, along with their results once this test concludes.</p>
+    {
+      formValues.audience_selection === 'percent' && <p>Total percentage value for all variants must equal 100%.</p>
```
bkemper: This validation message already shows up, but in the settings section.  Should we just move this validation message to each variant percentage field?

screen shot 2018-11-07 at 1 29 30 pm

src/pages/abTesting/components/Section.module.scss
ischer: awesome! i was curious if we had a prefixer in place. glad to see that we do. I'm cool with moving it index.scss if you think it makes sense!

src/pages/abTesting/components/content/VariantsContent.js
ischer: i'll look into that! i thought it was a little weird to have where it is right now.

src/pages/abTesting/components/content/VariantsContent.js
```diff
@@ -2,11 +2,18 @@ import React from 'react';
...
+    }
     <p>If you need to create a new template, <PageLink to='/templates'>head over to the templates page</PageLink>.</p>
   </div>
 );
 
+VariantsContent.defaultProps = {
+  formValues: {}
```
bkemper: For functional components, I think it is fine to set a default argument.  Also, if we only need a single value from the object, it may be safer to only pass what we need.

src/pages/abTesting/components/Section.module.scss
bkemper: @jonambas What do you think?

jonambas: yep `index.scss` will work

src/pages/abTesting/components/content/VariantsContent.js
bkemper: Do we still need this with the new validation?

src/helpers/validation.js
```diff
@@ -46,11 +46,28 @@ export function abTestDuration(value, formValues) {
...
-    return total === 100 ? undefined : 'Total distribution must equal 100%';
+  // only validate if audience selection is percent and all fields are filled
+  if (audience_selection === 'percent') {
+    let missingValues;
+
+    if (!default_template.percent) {
+      return;
+    }
+
+    _.forEach(variants, (variant = {}) => {
```
bkemper: Could the variants array contain `undefined`?

src/helpers/validation.js
```diff
@@ -46,11 +46,28 @@ export function abTestDuration(value, formValues) {
...
+  const { default_template, audience_selection, variants } = formValues;
 
-  if (value === 'percent') {
-    const total = _.reduce(variants, (sum, variant = { percent: 0 }) => sum + parseFloat(variant.percent), parseFloat(default_template.percent));
-    return total === 100 ? undefined : 'Total distribution must equal 100%';
+  // only validate if audience selection is percent and all fields are filled
+  if (audience_selection === 'percent') {
+    let missingValues;
+
+    if (!default_template.percent) {
```
bkemper: Can condense these guards into one?

```
if ([default_template, variants].find(({ percent }) => !percent)) {
  return;
}
```

src/helpers/validation.js
```diff
@@ -46,11 +46,28 @@ export function abTestDuration(value, formValues) {
...
+      if (!variant.percent) {
+        missingValues = true;
+      }
+    });
+
+    if (missingValues) {
+      return;
+    }
+
+    const total = _.reduce(variants, (sum, variant = { percent: 0 }) => sum + parseFloat(variant.percent || 0), parseFloat(default_template.percent || 0));
```
bkemper: With the guards in place above, we shouldn't need to default to zero.

src/pages/abTesting/components/content/VariantsContent.js
ischer: missed this previously - i'll swap this out if i can. for some reason, i thought i had tried that, but i might have just had a typo.

src/helpers/validation.js
ischer: yup - right now, if the page loads with an empty variant, it'll be `{}`, however, if you add a fresh variant and don't touch it, it'll be `undefined` briefly. i'll see if i can change that behavior rather than defaulting here, though

src/helpers/validation.js
ischer: i'll see if i can condense the two checks a little cleaner. unfortunately, `find` was a little iffy since if it finds an `undefined` percent, it returns `undefined`, which is also what we get when there's no match.

src/pages/abTesting/components/content/VariantsContent.js
bkemper: I think it would be good to remove!

src/helpers/validation.js
bkemper: Yikes.  Alright, I'll make one slight change to my other proposal.

PR #692, opened 2018-11-07
FE-573 Login page submit button disabled when fields autofilled

jonambas:
Bug - this shouldn't be disabled:
chromeautofill

Repro:
1. Log out of lastpass
2. Make sure saved credentials in chrome://settings/passwords
3. Open app, let Chrome fill in fields without touching anything


ewandennis:
lgtm :shipit:

PR #702, opened 2018-11-14
TR-645 added text explaining that templates must be published 

ischer:


src/pages/abTesting/components/AbTestCreateForm.js
```diff
@@ -75,7 +75,7 @@ export class AbTestCreateForm extends Component {
             label={'Select this test\'s default template'}
             placeholder='Type to search'
             helpText={templates.length > 0
-              ? <span>We will send this template by default when the test is not running. If you need to create a new template, <UnstyledLink component={Link} to='/templates'>head over to the templates page</UnstyledLink>.</span>
+              ? <span>We will send this template by default when the test is not running. This template must be published. If you need to create a new template, <UnstyledLink component={Link} to='/templates'>head over to the templates page</UnstyledLink>.</span>
```
bkemper: Can we shorten this text to...

> Only published templates are available.  This template will be sent when the test is not running.

src/pages/abTesting/components/AbTestCreateForm.js
```diff
@@ -75,7 +75,7 @@ export class AbTestCreateForm extends Component {
...
               : <span>No available templates.  <UnstyledLink component={Link} to='/templates'>Head over to the templates page to set some up</UnstyledLink>.</span>
```
bkemper: Should we change this help text to something like...

> No published templates are available.  <PageLink to="/templates">Head over to the templates page to set some up</PageLink>.

src/pages/abTesting/components/content/VariantsContent.js
```diff
@@ -5,6 +5,7 @@ import { PageLink } from 'src/components';
 const VariantsContent = () => (
   <div>
     <p>The templates you've selected will appear here, along with their results once this test concludes. Results will be updated approximately every 15 minutes.</p>
+    <p>Selected templates must be published.</p>
```
bkemper: This might get lost if we continue to add more test to this description section.  What about changing the field label to "Select a published template"?

src/pages/abTesting/components/AbTestCreateForm.js
ischer: totally missed this spot - that sounds good to me

PR #705, opened 2018-11-16
Hotfix/AC-402

ayoung1:


src/components/billing/tests/__snapshots__/PlanPrice.test.js.snap
```diff
@@ -50,7 +52,39 @@ exports[`PlanPrice renders correctly 1`] = `
   </span>
   <span
     className="SupportLabel"
-  />
+  >
+    <br />
```
colestrode: we seem to be adding breaks in every span with class=SupportLabel, why not set the height with CSS?

colestrode:
Looks like you have failing tests, I'd clean those up first before getting a reviewer.

PR #709, opened 2018-11-21
TR-628: Scheduling A/B test should allow you to schedule immediately

ischer:


src/helpers/metrics.js
```diff
@@ -134,9 +134,9 @@ function ceilMoment(time, roundInt = 1, precision = 'minutes') {
  * @param roundToPrecision
  * @return {*}
  */
-export function getValidDateRange({ from, to, now = moment(), roundToPrecision }) {
-  // If we're not rounding, just check that 'to' is before 'now'
-  const nonRoundCondition = () => roundToPrecision ? true : to.isBefore(now);
+export function getValidDateRange({ from, to, now = moment(), roundToPrecision, preventFuture }) {
+  // If we're not rounding, check to see if we want to prevent future dates. if not, prevent past dates.
+  const nonRoundCondition = () => roundToPrecision ? true : preventFuture ? to.isBefore(now) : from.isAfter(now);
```
ischer: the only place this function is used is in the datepicker, so i don't know if it really fits here, despite having some pretty metrics-centric logic. i think i might move this to the date utils. thoughts?

src/selectors/tests/abTesting.test.js
```diff
@@ -1,12 +1,13 @@
...
+import { fn as mockMoment } from 'moment';
 
 describe('Selectors: abTesting', () => {
   let state;
   let props;
 
   beforeEach(() => {
-    const mockNow = new Date('2010-01-01T12:00:00.000Z');
-    global.Date = jest.fn(() => mockNow);
+    mockMoment.add = jest.fn().mockReturnValue({ toDate: () => '2010-01-08T12:00:00.000Z' });
```
bkemper: The return value can also be set like this.

```
mockMoment.add = jest.fn(() => ({ toDate: () => '2010-01-08T12:00:00.000Z' }));
```

src/selectors/abTesting.js
```diff
@@ -49,8 +49,8 @@ export const selectEditInitialValues = createSelector(
 
     // Reshape dates and sets defaults
     const dates = {
-      from: test.start_time ? new Date(test.start_time) : moment().add(1, 'd').toDate(),
-      to: test.end_time ? new Date(test.end_time) : moment().add(8, 'd').toDate()
+      from: test.start_time ? new Date(test.start_time) : moment().toDate(),
```
bkemper: The API request fails if I immediately submit.
screen shot 2018-11-27 at 2 39 36 pm

Also, I can't change the end time until I change start time to something other than now.

Can we pad these by a couple minutes?

src/pages/abTesting/components/fields/StatusFields.js
```diff
@@ -26,7 +26,7 @@ const StatusFields = ({ disabled }) => (
         roundToPrecision={false}
         preventFuture={false}
         fromSelectsNextHour={true}
-        validate={[abTestDuration, startTimeAfterNow, startTimeBeforeEndTime]}
```
bkemper: Shouldn't we keep this?

src/helpers/metrics.js
bkemper: Plus, this logic is getting too complex.  I'm going to file a follow-up tech debt ticket.

src/pages/abTesting/components/fields/StatusFields.js
ischer: initially, i did want to keep this, but while i was testing stuff everything out, i'd often get blocked by the validator since the updated datepicker defaults start time to now. I also updated the API so that if a start time in the past is provided, it would be set to the current time upon creation.

src/selectors/abTesting.js
ischer: the api request failure is expected, and will need https://github.com/SparkPost/ab-testing-public-api/pull/59/files to make it work.

i'll take a look into the end time thing - i didn't notice that

what do you mean by padding these by a couple minutes?

bkemper: The default value is now.  I was thinking just make the default value now + 5 minutes, so the date picker validation doesn't break if I change the end time.

src/helpers/metrics.js
ischer: I'm just going to remove the `from.isAfter(now)` check here. the A/B test api will fix it if it's in the past, and i don't want to have customers get errors if they try to change their end dates with start dates that were just a little behind by the time they got to updating the datepicker fields.

src/selectors/abTesting.js
ischer: wrt: `Also, I can't change the end time until I change start time to something other than now.` - it turns out this issue was because the `now` that the validator uses has seconds precision, while the `from` has minutes precision. since `from` is derived from `now`, it's impossible for it to be after `now` in this case. adding 5 minutes to the `from` date default would work, but only if the customer did not take 5 minutes on the form or edit a draft that already had a defined `from`.

i tried a bunch of different ways to ensure that `from` is after `now` (or at least close to it), and found edge cases in most of them that would result in being unable to change the end date. so, i'm just going remove this validation. the API will handle the past dates without this.

src/helpers/metrics.js
```diff
@@ -135,8 +135,8 @@ function ceilMoment(time, roundInt = 1, precision = 'minutes') {
  * @return {*}
  */
 export function getValidDateRange({ from, to, now = moment(), roundToPrecision, preventFuture }) {
-  // If we're not rounding, check to see if we want to prevent future dates. if not, prevent past dates.
-  const nonRoundCondition = () => roundToPrecision ? true : preventFuture ? to.isBefore(now) : from.isAfter(now);
+  // If we're not rounding, check to see if we want to prevent future dates. if not, we're good.
+  const nonRoundCondition = () => roundToPrecision ? true : preventFuture ? to.isBefore(now) : true;
```
bkemper: Isn't this the same as this...

```
validDates && from.isBefore(to) && (roundToPrecision || !preventFuture || to.isBefore(now))
```

PR #711, opened 2018-11-21
FE-580: Integration test w/ login form

PatrickSison:


ewandennis:
Looks good to me.

Missed [1 item](https://github.com/SparkPost/2web2ui/pull/711#discussion_r236611919) about using underscores in http-response filenames:

> The integration testing readme could use a small update to since it still describes the old names for mock responses:

>> A GET call to /account/something can be mocked by adding a file in the following location: http-responses/_account_something/get.js.

Then lets ship it!

PR #713, opened 2018-11-27
Hotfix/sa 353

kchu1337:


ewandennis:
Please consider using gitplus and make sure its current. We don't use `feature` or `hotfix` branch prefixes any more which gitplus will handle for you. The tool also handles safely merging to master with `--no-ff` to give us a clear commit that we can revert later if needed.

[You can install gitplus from here](https://github.com/SparkPost/gitplus).

If already installed, you can update gitplus like this: `git plus upgrade`.

src/actions/reportOptions.js
```diff
@@ -46,20 +43,48 @@ export function initTypeaheadCache() {
...
+ * It will first try to pull the data from cache, if it exists and the time range is the same.
+ * If there is no cache entry, it will make the api calls and add the results to the cache.
+ */
 export function refreshTypeaheadCache(options) {
   const params = getQueryFromOptions(options);
-  return (dispatch) => {
-    const requests = metricLists.map((list) => dispatch(list(params)));
-    return Promise.all(requests);
+  return (dispatch, getState) => {
+    const { typeAhead } = getState();
```
ewandennis: Typo here: the state key is named `typeahead`. Same for the line below.

src/actions/reportOptions.js
```diff
@@ -46,20 +43,48 @@ export function initTypeaheadCache() {
...
+    const { to: cachedTo, from: cachedFrom, cache } = typeAhead;
+    const { match, to: currentTo, from: currentFrom } = options;
+
+    if (cachedFrom === currentFrom && cachedTo === currentTo && cache[match]) {
+      dispatch({
+        type: 'UPDATE_METRICS_FROM_CACHE',
+        payload: cache[match]
+      });
+      return Promise.resolve();
+    } else {
```
ewandennis: We don't need the else clause here since we returned unconditionally from the if clause.

src/helpers/tests/metrics.test.js
```diff
@@ -29,6 +29,21 @@ describe('metrics helpers', () => {
     expect(actual).toMatchSnapshot();
   });
 
+  it('should return options with limit and match keys if given', () => {
```
ewandennis: This test is verifying 2 separate cases. Lets split them up. You can parametrise with `jest-in-case` if that helps.

Also, since its easy to verify the exact contents of `actual`, lets do that instead of deferring to a snapshot:

```js
expect(actual).toMatchObject({ match: 'foo' });
```

src/pages/reports/components/Typeahead.js
```diff
@@ -2,45 +2,76 @@ import React, { Component } from 'react';
 import Downshift from 'downshift';
 import classnames from 'classnames';
 import _ from 'lodash';
-
+import { refreshTypeaheadCache } from 'src/actions/reportOptions';
 import sortMatch from 'src/helpers/sortMatch';
 import { TextField, ActionList } from '@sparkpost/matchbox';
 import Item from './TypeaheadItem';
 import styles from './Typeahead.module.scss';
+import { withRouter } from 'react-router-dom';
```
ewandennis: What are we using `react-router` for here?

src/pages/reports/components/Typeahead.js
```diff
@@ -98,4 +130,10 @@ class Typeahead extends Component {
...
 }
 
-export default Typeahead;
+const mapStateToProps = () => ({});
+
+const mapDispatchToProps = {
+  refreshTypeaheadCache
+};
+
+export default withRouter(connect(mapStateToProps, mapDispatchToProps)(Typeahead));
```
ewandennis: You can pass `null` into `connect`() for `mapStateToProps` here. [react-redux docs ref](https://react-redux.js.org/docs/api).

src/pages/reports/components/Typeahead.js
```diff
@@ -2,45 +2,76 @@ import React, { Component } from 'react';
...
   state = {
     inputValue: '',
     matches: [],
-    calculatingMatches: false
-  }
+    calculatingMatches: false,
+    lastPattern: null
+  };
 
+  //Updates matches for Templates, Sending Domains, Subaccounts, and anything not in `TYPES_METRICS`.
```
ewandennis: `updateMatches` and `updateMatchesAsync` share a lot of logic. Can we easily refactor it into a single function?

src/pages/reports/components/Typeahead.js
```diff
@@ -2,45 +2,76 @@ import React, { Component } from 'react';
...
   updateMatches = (pattern) => {
-    let matches;
-
-    this.setState({ calculatingMatches: true });
+    const { items, selected = []} = this.props;
+    const flatSelected = selected.map(flattenItem);
+    const matches = sortMatch(items, pattern, (i) => i.value)
+      .filter(({ type, value }) => !flatSelected.includes(flattenItem({ type, value })) && !TYPES_METRICS.includes(type))
+      .slice(0, 100);
```
ewandennis: Lets move that `100` into a constant in `src/constants/index.js`

src/pages/reports/components/Typeahead.js
```diff
@@ -2,45 +2,76 @@ import React, { Component } from 'react';
...
+  // Updates the matches with the already-loaded templates, sending domains, subaccounts then
+  // makes the metrics api calls and appends to the matches.
+  updateLookAhead = (pattern) => {
+    this.setState({ calculatingMatches: true });
+    if (!pattern || pattern.length < 2) {
+      this.setState({ matches: [], calculatingMatches: false, lastPattern: null });
+      return Promise.resolve();
+    } else {
+      this.setState({ lastPattern: pattern });//This sets the last queued pattern
+      const options = { ...this.props.reportOptions, match: pattern, limit: 1000 };
```
ewandennis: This `1000` should also be a constant.

src/pages/reports/components/Typeahead.module.scss
```diff
@@ -4,6 +4,13 @@
   position: relative;
 }
 
+// Logo styles
```
ewandennis: Are these styles copied from `Loading`? If so, lets re-use them instead.

src/pages/reports/components/TypeaheadLoading.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+import styles from './Typeahead.module.scss';
+import { LoadingSVG } from '../../../components';
+
+const TypeaheadLoading = ({ isCalculating }) => {
```
ewandennis: This looks a lot like `Loading`. Any reason for the fork?

src/pages/reports/components/tests/Typeahead.test.js
```diff
@@ -22,27 +24,41 @@ describe('Component: Typeahead', () => {
...
     handleFieldChange({ target: { value: 'cros' }});
-    updateMatchesDebounced.flush(); // forces debounced calls to execute
-    expect(wrapper.state()).toMatchSnapshot();
+    updateLookAheadDebounced.flush(); // forces debounced calls to execute
+    expect(props.refreshTypeaheadCache).toHaveBeenCalledTimes(1);
+    expect(wrapper.state().calculatingMatches).toEqual(true);
+    expect(wrapper.state().lastPattern).toEqual('cros');
   });
 
   it('should produce matches on search', () => {
```
ewandennis: Since this test contains async code, we need to make the Jest wait for it complete:

```js
it('should render with matches', async () => {
  await promiseyThing;
  expect(...);
});
```

The same goes for all the async tests below.

src/reducers/typeahead.js
```diff
@@ -0,0 +1,26 @@
...
+  cache: {}
+};
+
+export default (state = initialState, action) => {
+  switch (action.type) {
+
+    case 'UPDATE_TYPEAHEAD_METRICS_CACHE': {
+      const { to, from, itemToCache } = action.payload;
+      // If the dates are different, replace the existing cache with the new item to cache
+      if (to !== state.to || from !== state.from) {
```
ewandennis: Does this logic replace the code in `ReportOptions` [here](https://github.com/SparkPost/2web2ui/pull/713/files#diff-59315dcf70ae208e2e11fb1164d4c44aL45)? 

If so, is there a reason we aren't using the `isSameDate` helper in this version?

src/reducers/typeahead.js
```diff
@@ -0,0 +1,26 @@
...
+        return { ...state, to, from, cache: itemToCache };
+      }
+      // else, append the new item to the cache
+      const cache = _.assign(state.cache, itemToCache);
```
ewandennis: Isn't this just `const cache = { ...state.cache, ...itemToCache };` ?

I prefer to use the language where feasible.

src/pages/reports/components/Typeahead.js
```diff
@@ -14,39 +14,30 @@ import { LoadingSVG } from 'src/components';
 function flattenItem({ type, value }) {
   return `${type}:${value}`;
 }
+
 const MatchesLoading = ({ isLoading }) => (isLoading) ? <LoadingSVG size='XSmall' /> : null;
 
+const staticItemTypes = ['Template', 'Subaccount', 'Sending Domain'];
```
kchu1337: Can we use SNAKE_CASE or even put it into the constants folder.

src/pages/reports/components/Typeahead.js
```diff
@@ -14,39 +14,30 @@ import { LoadingSVG } from 'src/components';
...
-
-    if (excludedItems) {
-      const isInExcludedItems = function (item) {
-        return excludedItems.some((excludedItem) => _.isEqual(item, excludedItem));
-      };
-      matches = matches.filter((item) => !isInExcludedItems(item));
+    filterItems = (pattern, itemTypes) => {
+      const { items, selected = []} = this.props;
+      const flatSelected = selected.map(flattenItem);
+      const staticItems = itemTypes ? items.filter(({ type }) => itemTypes.includes(type)) : items;
```
kchu1337: Isn't this function used for both the static and static+dynamic items? Can you rename the variable to something like ```filteredItems```?

src/pages/reports/components/tests/Typeahead.test.js
```diff
@@ -51,32 +44,20 @@ describe('Component: Typeahead', () => {
...
+    wrapper.instance().updateLookAhead('cross');
+    expect(wrapper.state().matches).toEqual(defaultCrossMatches);
   });
 
   it('should produce no matches on empty query', () => {
-    const pattern = '';
-    const result = wrapper.instance().getMatches({ pattern });
-    expect(result).toEqual([]);
+    wrapper.instance().updateLookAhead('');
+    expect(wrapper.state().matches).toEqual([]);
```
kchu1337: I think I made the same mistake but I think this is the same test as the one on line 70.

PR #719, opened 2018-11-29
AC-446: mandatory TFA for all

ewandennis:
- Removed the `tfaRequired` feature flag check so any admin can make TFA mandatory
 - Fixed an issue where you could still disable your own TFA when mandatory
 - Fixed an issue where an admin could still disable another user's TFA when mandatory

### Testing
 - Verify an account without `options.ui.tfaRequired` can set TFA to required in `/account/settings`
 - Enable mandatory TFA
 - Verify users cannot disable their own TFA
 - Verify admins cannot disable others' TFA

### DIFF
--- a/src/pages/account/AccountSettingsPage.js
+++ b/src/pages/account/AccountSettingsPage.js
@@ -5,8 +5,6 @@ import LabelledValue from 'src/components/labelledValue/LabelledValue';
 import CancellationPanel from './components/CancellationPanel';
 import SingleSignOnPanel from './components/SingleSignOnPanel';
 import EnforceTfaPanel from './components/EnforceTfaPanel';
-import { AccessControl } from 'src/components/auth';
-import { hasUiOption } from 'src/helpers/conditions/account';
 
 export function AccountSettingsPage({ currentUser }) {
   return (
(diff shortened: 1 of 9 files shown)


jmoyse:
Functionality for the enable and disable of mandatory TFA looks good to me. I checked every combination of TFA being already enabled/disabled with turning on mandatory TFA. All the flows worked as expected.

I am having problems changing another user's access from Admin to Reporting while tfa is mandatory. Trying to do so while that user has tfa enabled will give the error: `Something went wrong. View Details` `tfa_enabled can only be set to false.`. Looks like the `PUT` request to `/api/v1/users/{user}` includes a `tfa_enabled` param with the `access_level`. Not sure if this is something better handled on the Accusers side though.

I did also find that a user without TFA enabled that turns on mandatory TFA is not immediately logged out. Instead the next UI action they perform will log them out and require a new login. I know a few things changed midway through development and I'm not sure where we landed with the post mandatory TFA process. The logout seemed a bit jarring, almost like something was broken. Not the biggest issue (and this might be by design), but potentially worth filing for the backlog. 

These seem fairly edge case and the access_level issue is potentially an accusers bug, so overall it seems ready to me.

PR #721, opened 2018-12-03
EO-429 Add Sparkline component

jonambas:
Adds a reusable Sparkline component with a custom tooltip wrapper for the Signals overview page.

Some code to copy/paste for testing:
```js
import Sparkline from './components/charts/sparkline/Sparkline';

<Sparkline
  domain={[0,5]}
  onClick={(data) => console.log(data)}
  tooltipContent={(payload) => <div>test custom content {payload.value}</div>}
  data={[
  { value: 1, date: new Date() },
  { value: 3, date: new Date() },
]} />
```


ewandennis:
Yay sparklines. They look good and stack well:

screen shot 2018-12-04 at 11 43 53

(The stroke width might be a tiny bit heavy for my preference.)

The test data generator I used:
```js
const mkSeries = (size = 100) => Array.from(Array(size).keys()).map(() => ({ value: Math.random().toFixed(2), date: new Date()}));
```
I had to refer to the recharts docs to understand some of the `Sparkline` props and some I think might not be too useful to expose. A few notes:

 - `activeDot=false` looked odd to me. Maybe just force it to `true`?
 - Not setting `tooltipContent` produced a really odd tooltip. Maybe worth making that prop required or providing a default that renders Y values.
    screen shot 2018-12-04 at 11 35 19
 - I rarely find `data` to be a useful name in code since everything is data, even code. Maybe we could consider one of these options:
     - For `data`: `timeSeries` or `points`
     - for `dataKey`: `yValueKey` or just `valueKey`

On to the code review!

src/pages/signals/components/charts/sparkline/Sparkline.js
```diff
@@ -0,0 +1,50 @@
+import React, { Component } from 'react';
+import { ResponsiveContainer, LineChart, Line, Tooltip, YAxis } from 'recharts';
+import TooltipWrapper from '../tooltip/Tooltip';
```
ewandennis: Should this just be `Tooltip`?

src/pages/signals/components/charts/sparkline/Sparkline.js
```diff
@@ -0,0 +1,50 @@
...
+import _ from 'lodash';
+
+class Sparkline extends Component {
+
+  renderTooltip = (props) => {
+    const date = _.get(props, 'payload[0].payload.date');
+    const content = _.get(props, 'payload[0]', {});
```
ewandennis: Once we have `content`, can we just `_.get(content, 'payload.date')` ?

src/pages/signals/components/charts/sparkline/Sparkline.js
```diff
@@ -0,0 +1,50 @@
...
+    return <TooltipWrapper date={date} children={this.props.tooltipContent(content)} />;
+  };
+
+  render() {
+    const { data, height, width, dataKey, dot, domain, activeDot, stroke, onClick } = this.props;
```
ewandennis: Instead of `data`,  maybe `timeSeries` or `points`?
For `dataKey`, maybe `yKey`, `yValueKey`?
I sort of get `domain` but we could also say `yRange`, `yLimits` or `yBounds`.

Is `activeDot=false` necessary?
For our use cases, are `width` and `height` necessary?

src/pages/signals/components/charts/sparkline/Sparkline.js
```diff
@@ -0,0 +1,50 @@
...
+  }
+}
+
+Sparkline.defaultProps = {
+  dataKey: 'value',
+  domain: [0, 100],
+  dot: false,
+  height: 25,
+  stroke: '#000000',
+  tooltipContent: _.noop,
```
ewandennis: Using `_.noop` here produces a tooltip with just the date on it and a blank 2nd line. Maybe provide a good default or just make `tooltipContent` required.

src/pages/signals/components/charts/tooltip/Tooltip.module.scss
```diff
@@ -0,0 +1,26 @@
...
+  padding: spacing(small) spacing();
+  color: color(gray, 10);
+}
+
+.TooltipDate {
+  font-size: rem(15);
+  border-bottom: 1px solid darken(color(gray, 1), 5);
+}
+
+:global {
```
ewandennis: What is this voodoo? :-)

src/pages/signals/components/charts/sparkline/Sparkline.js
jonambas: Name collision with recharts tooltip

PR #725, opened 2018-12-05
AC-456 adds default plan for downgrade to free

colestrode:
I wasn't sure what to call the plan config, so I modeled it off of what we do in the accusers API where we call it `defaults.plan`

### DIFF
--- a/src/selectors/tests/__snapshots__/brightback.test.js.snap
+++ b/src/selectors/tests/__snapshots__/brightback.test.js.snap
@@ -8,7 +8,7 @@ Object {
   },
   "app_id": "9N0rWBvKGR",
   "billing_url": "http://phoenix.test/billing_url",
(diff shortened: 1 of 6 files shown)


src/selectors/tests/__snapshots__/brightback.test.js.snap
```diff
@@ -8,7 +8,7 @@ Object {
   },
   "app_id": "9N0rWBvKGR",
   "billing_url": "http://phoenix.test/billing_url",
-  "cancel_confirmation_url": "http://phoenix.test/cancel_confirmation_url",
+  "cancel_confirmation_url": "http://phoenix.test/cancel_confirmation_url?immediatePlanChange=undefined",
```
colestrode: I'll add a test config so this value isn't undefined

colestrode:
Works for me, let's ship it 👯‍♂️

PR #726, opened 2018-12-10
EO-434 Add bar chart

jonambas:
Demo code: 

```js
import BarChart from './components/charts/barchart/BarChart';
import moment from 'moment';
import _ from 'lodash';

// fake data
const randInt = (max) => Math.floor(Math.random() * Math.floor(max));
const data = _.range(90).map((n, i) => ({
  bar: randInt(25),
  foo: randInt(25),
  baz: randInt(25),
  abc: randInt(25),
  date: moment().subtract(90 - i, 'day').toISOString()
}));

// class state setup
state = {
  selected: ''
}

handleBarClick = ({ payload }) => {
  this.setState({ selected: payload.date });
}

// render
<BarChart timeSeries={data}
  yDomain={[0,100]}
  yKeys={[
    { key: 'bar', fill: '#B157CE' },
    { key: 'foo', fill: '#28C0C4' },
    { key: 'baz', fill: '#FF9502' },
    { key: 'abc', fill: '#50D1F3' }
  ]}
  gap={0.3}
  onClick={this.handleBarClick}
  selected={this.state.selected}
  xAxisProps={{ interval: 88, tickFormatter: (tick) => moment(tick).format('M/D') }}
/>

<BarChart timeSeries={data}
  yDomain={[0,50]}
  yKey='foo'
  onClick={this.handleBarClick}
  selected={this.state.selected}
  gap={0.3}
  xAxisProps={{ interval: 88, tickFormatter: (tick) => moment(tick).format('M/D') }}
/>
```

### DIFF
--- a/src/pages/signals/components/charts/barchart/BarChart.js
+++ b/src/pages/signals/components/charts/barchart/BarChart.js
@@ -65,7 +65,7 @@ class BarChart extends Component {
     const { gap, height, timeSeries, tooltipContent, width, xKey, xAxisProps, yDomain, yAxisProps } = this.props;
 
     return (
-      <ResponsiveContainer height={height} width={width}>
+      <ResponsiveContainer height={height} width={width} className='SignalsBarChart'>
         <RechartsBarchart barCategoryGap={gap} data={timeSeries} margin={{ top: 5, left: 5, right: 18, bottom: 5 }}>
           {this.renderBackgrounds()}
           <CartesianGrid vertical={false} />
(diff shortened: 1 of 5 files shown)


src/pages/signals/components/charts/barchart/BarChart.scss
```diff
@@ -0,0 +1,9 @@
+:global {
```
jonambas: Use a wrapper class so these rules dont leak

src/pages/signals/components/charts/barchart/BarChart.js
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    return this.renderBar({ key: yKey, fill });
+  }
+
+  renderBackgrounds = () => {
+    const { xKey, selected, onClick } = this.props;
+
+    return (
+      <Bar
```
jonambas: This is a little hacky. I'm using another bar to render the background behind the main bars. 

`{...background}` is what sets its coordinates and size

PR #730, opened 2018-12-11
AC-454 (WIP): Added field and redux flow for error and promo state on UI

PatrickSison:


jmoyse:
Everything looks like its working from what I can tell. 

I tested by creating a new account and first upgraded to the 50k plan using the `julian` promo to test the free to paid path. The discount was applied and looks to be the correct amount.
I upgraded the account to 150K and used the same promo code and it was correctly applied and upgraded. The invoice numbers look a bit off. The total invoiced seems to include an extra charge of the discount amount from the free to paid upgrade ($1.35).  This is probably an accusers side issue. I can cut a ticket to see what the problem is. 

I also noticed moving to the 2M plan kept the discount amount even though the promo didn't apply to that plan. Resulted in a $134 discount on a $900 plan. Again this is probably an accusers issue.

Promo code validation looks correct before and during the purchase. And it seems to correctly prevent the purchase with a bad promo code.

Only path I'm not able to check is the paid->free->paid since that takes a month to have an account go on cool down.

Unrelated to this ticket, but the reading the invoice with discounts is crazy confusing. Might be worth having UI or UX people look it over at some point.

Not seeing any problems with anything on your end though. lgtm

PR #734, opened 2018-12-13
TR-845: Fix cannot read property 'trim' of null

bkemper:


src/components/contentEditor/ContentEditor.js
```diff
@@ -79,15 +79,15 @@ class ContentEditor extends React.Component {
   }
 
   render() {
-    const { readOnly, isAmpLive, contentOnly } = this.props;
```
ischer: we'll need to keep `contentOnly` here for the content validator check. templates require one of html or text, while snippets will require html, text, or amp_html. removing `contentOnly` would allow users to try submitting with just amp_html which will throw an error

bkemper: Ahhh, I thought AMP was going live for templates and snippets.  I'll fix this.

ischer: it will be, but the validation for templates and snippets are different. it confused me at first, as well.

bkemper: Fixed.

bkemper:
I removed that last commit causing the problem.

PR #741, opened 2018-12-21
AC-471: Promo codes in onboarding plan selection

PatrickSison:


src/pages/onboarding/ChoosePlan.js
```diff
@@ -104,12 +134,13 @@ export class OnboardingPlanPage extends Component {
         <Grid>
           <Grid.Column>
             <Panel title='Select A Plan'>
-              <PlanPicker disabled={submitting} plans={plans} />
+              <PlanPicker disabled={submitting} plans={plans} onChange={this.onPlanSelect}/>
```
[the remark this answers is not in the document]
ewandennis: Yeah, this looks possible if you paste your promo code into the form field:

screen shot 2019-01-02 at 17 21 45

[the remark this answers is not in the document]
PatrickSison: I had implemented the disable on conditional, which is supposed to both the picker and the submit, but it only seems to disable the submit properly which means it's likely an issue with the plan picker. I will look into disabling the plan picker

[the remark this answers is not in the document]
PatrickSison: Nevermind. Didn't realize it was a different variable

src/components/billing/PlanPrice.js
```diff
@@ -34,7 +34,7 @@ const PlanPrice = ({ plan, showOverage = false, showIp = false, showCsm = false,
     <span className='notranslate'>
       <span className={styles.MainLabel} {...rest}>
         <strong>{plan.volume.toLocaleString()}</strong> emails/month
-        {priceInfo.price > 0 && <span> at {discountAmount !== priceInfo.price && <s className={styles.DiscountedLabel}>${priceInfo.price}</s>}<strong>${discountAmount.toLocaleString()}</strong>/{priceInfo.intervalShort}</span>}
+        {priceInfo.price > 0 && <span> at {discountAmount !== priceInfo.price && (<span><s className={styles.DiscountedLabel}>${priceInfo.price}</s>&nbsp;</span>)}<strong>${discountAmount.toLocaleString()}</strong>/{priceInfo.intervalShort}</span>}
```
ewandennis: Lets try to avoid the non-breaking space hack. We could either:

 - Put a normal space inside the `<s>` tags
 - Add a little padding to the `DiscountedLabel` style: `padding-right: rem(8);`

src/pages/billing/helpers/promoCodeValidate.js
```diff
@@ -1,9 +1,10 @@
 import { verifyPromoCode } from 'src/actions/billing';
 import { change } from 'redux-form';
 
-export default function promoCodeValidate(values, dispatch) {
+const promoCodeValidate = (formName) => (values, dispatch, props, field) => {
```
ewandennis: Nice generalisation. 

Missing unit tests.

src/components/billing/PlanPrice.js
ewandennis: Lets add a test for promotional pricing display.

src/components/planPicker/PlanPicker.module.scss
```diff
@@ -69,7 +69,7 @@
 
 .List {
   position: absolute;
-  z-index: 1;
```
ewandennis: Do we know why we have an explicit `z-index` here? I don't see an issue in either place it's used but its worth understanding.

src/pages/billing/helpers/promoCodeValidate.js
```diff
@@ -1,9 +1,10 @@
...
   const { promoCode, planpicker } = values;
-  if (!promoCode) {
+  if (!promoCode || field === 'planpicker') {
```
ewandennis: Are we tangling up 2 separate cases here? It seems odd to do `change(formName, 'promoCode', '')` when `!promoCode`. Lets separate them if we can.

src/pages/onboarding/ChoosePlan.js
```diff
@@ -104,12 +135,13 @@ export class OnboardingPlanPage extends Component {
         <Grid>
           <Grid.Column>
             <Panel title='Select A Plan'>
-              <PlanPicker disabled={submitting} plans={plans} />
+              <PlanPicker selectedPromo={billing.selectedPromo} disabled={disableSubmit} plans={plans} onChange={this.onPlanSelect}/>
               <AccessControl condition={not(isAws)}>
+                {!selectedPlan.isFree && this.renderPromoCodeField()}
```
ewandennis: I doubt we should tackle this here but worth discussing with the team. Someone _could_ game our no-promo-codes-on-downgrade policy by visiting `/onboarding/plan` when already on a paid plan. Then they could enter a promo code and downgrade since we don't check here.

A few options for the future:
 - Stop people revisiting `/onboarding/` pages after some condition; maybe just `/onboarding/plan` iff their already paying.
 - Remove the no-promo-code-on-downgrade policy.

PR #743, opened 2019-01-02
EO-465 Add new feature nav item tag

jonambas:
screen shot 2019-01-02 at 12 23 25 pm

### DIFF
--- a/src/components/navigation/components/Item.js
+++ b/src/components/navigation/components/Item.js
@@ -1,10 +1,17 @@
 import React, { Component } from 'react';
+import PropTypes from 'prop-types';
 import { Link } from 'react-router-dom';
 import { WindowSizeContext } from 'src/context/WindowSize';
 import { Tag } from '@sparkpost/matchbox';
 import { ChevronLeft } from '@sparkpost/matchbox-icons';
 import classnames from 'classnames';
 import styles from './Item.module.scss';
+
+const TAGS = {
+  beta: { label: 'BETA' },
+  new: { color: 'blue', label: 'NEW' },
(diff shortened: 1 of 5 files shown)


src/components/navigation/components/Item.js
```diff
@@ -39,13 +39,14 @@ export class Item extends Component {
   renderItem = ({ mobile }) => {
     const {
       beta,
```
ewandennis: Do we want to tidy up the `beta`, `newFeature` and `labs` props while we're here? How about a `tag` prop. Something like `propTypes.oneOf(['beta', 'labs', 'new'])` ?

That way, we avoid the inevitable situation where someone sets 2 flags and gets a surprise. We also gain a tiny bit of readability.

src/components/navigation/components/Item.js
```diff
@@ -79,12 +80,16 @@ export class Item extends Component {
       releaseTag = <Tag color="blue">LABS</Tag>;
     }
 
+    if (newFeature) {
```
ewandennis: With an enum `tag` prop, we could collapse these 3 clauses into a table lookup:

```jsx
const TAGS = {
  beta: { color: 'orange', label: 'BETA' },
  new: { color: 'orange', label: 'NEW' },
  labs: { color: 'blue', label: 'LABS' }
};

if (tag) {
  const tagSpec = TAGS[tag];
  releaseTag = <Tag color={tagSpec.color}>{tagSpec.label}</Tag>;
}
```

ewandennis: On a stylistic note, do we like sharing orange with both `beta` and `new` tags?

PR #744, opened 2019-01-02
AC-412 make onboarding accurately represent new free plan

James-Burba:


src/components/billing/PlanPrice.js
```diff
@@ -33,9 +33,10 @@ const PlanPrice = ({ plan, showOverage = false, showIp = false, showCsm = false,
   return (
     <span className='notranslate'>
       <span className={styles.MainLabel} {...rest}>
-        <strong>{plan.volume.toLocaleString()}</strong> emails/month
+        <strong>{plan.volume.toLocaleString()}</strong> emails
+        {plan.isFree ? ' for 30 days' : '/month'}
```
PatrickSison: This will show for 30 days even after it has switched to the free 500 plan. It should probably check the code for the plan as well

James-Burba: oh, good call

PatrickSison: Sorry, another nitpick. Should we indicate that after the 30 days, it drops down to 500 /month?

PR #756, opened 2019-01-09
EO-501: Multiply spam trap hits ratio by 100 for percent formatting

bkemper:


src/selectors/tests/__snapshots__/signals.test.js.snap
```diff
@@ -231,14 +220,14 @@ Object {
...
         "date": "2018-01-03",
-        "health_score": 0.98,
-        "injections": 35000,
+        "health_score": null,
+        "injections": null,
         "weights": Array [],
       },
     ],
-    "empty": false,
+    "empty": true,
```
jonambas: Thanks

PR #761, opened 2019-01-11
EO-507 Add health score component weight timeseries chart

jonambas:


src/pages/signals/HealthScorePage.js
```diff
@@ -39,18 +40,29 @@ export class HealthScorePage extends Component {
...
-      const last = _.last(data);
-      this.setState({ selectedDate: last.date });
+    // Select last date in time series
+    if (dataSetChanged && !selectedDataByDay) {
+      selectedDataByDay = _.last(data);
+      this.setState({ selectedDate: selectedDataByDay.date });
+    }
+
+    // Select first component weight type
+    if (dataSetChanged || prevState.selectedDate !== selectedDate) {
```
bkemper: When switching dates in the component bar chart, the component switches.

health_score

Instead, how about something like this to only change the selected component if selectedComponent is not defined or does not exist for the selected date?

```
const selectedComponentExists = _.find(
  (_.find(data, ['date', selectedDate]) || {}).weights,
  ['weight_type', selectedComponent]
);

if (!selectedComponent || !selectedComponentExists) {
  const firstComponentType = _.get(selectedDataByDay, 'weights[0].weight_type');
  this.setState({ selectedComponent: firstComponentType });
}
```

jonambas: Hmm, not sure that's needed. I think it'll always exist on any date because of the `fillByDate`, I'll just remove `prevState.selectedDate !== selectedDate`

PR #763, opened 2019-01-11
EO-511 / EO-483 - Display cohort weights, Fix spam trap y axis bugs

jonambas:


src/pages/signals/components/charts/divergingBar/DivergingBar.js
```diff
@@ -37,14 +37,15 @@ class DivergingBar extends Component {
   }
 
   renderYTick = ({ payload, ...props }) => {
-    const { data, selected, yKey } = this.props;
+    const { data, selected, yKey, yLabel } = this.props;
```
bkemper: I guess this could have been the same as `yLabel = () => payload.value`.

PR #771, opened 2019-01-15
Sa 365- Cursor Based Pagination For Events UI

kchu1337:
Switch to cursor based pagination in new events page.
Users can go forward and back a page as well as return to the first page.

Major Impacting Changes:
* `src/components/collection/Pagination.js`
  * PerPage and SaveToCSV rendering has been moved into its own component `CollectionControls`
  * This affects `src/components/collection/Collection.js`, which affects `src/components/collection/TableCollection.js`
  * TableCollection is imported in many files.

* `src/actions/helpers/sparkpostApiRequest.js`
  * Added an `extra` key in the dispatch. Will now show up in the `action` param in the reducers
  * Possible effects in reducers. However most reducers use the destructuring `{ type, payload, meta }`
  * Those that do not destructure `action` seem to only use its component values and not the entire object itself. 

Major Isolated Changes:
* `src/pages/reports/messageEvents/MessageEventsPage.js`
  * Removed old pagination
  * Added forward, back, and first page buttons with disabled and enabled logic for the first 2.
  * Still contains Per Page and Save as CSV buttons. Both will show regardless of data length.
  * Shows Page **X**/Y right of the pager. 
  * Added logic for changing the page. Uses a cache. 
  * Handles the reload logic for search filter change. This was previously handled in `src/pages/reports/messageEvents/components/MessageEventsSearch.js`
  * Cursor paging logic in it's own component. 

* `src/reducers/messageEvents.js` 
  * Added logic for saving next page links and caching current results. This happens in both the initial call and subsequent page changes.
  * `GET_MESSAGE_EVENTS` handles calls the reload the page. This happens when search filters change, per page changes, or firs page button is clicked.  Removes all items in cache and creates new cache & links.
  *  `GET_MESSAGE_EVENTS_PAGE` handles page changes. It saves the next link and caches the results.
  * Added logic for changing page with both new api calls and loading from a cache.

- [x] API request passes the next cursor onto the reducer
- [x] Reducer saves the next link and caches results. Does this in both initial load and after loading a new page. 
- [x] Functional Forward and Previous buttons. Loads from cache if available, else, makes new api call.
- [x] Functional First Page button that makes a new api call rather than load from cache. 
- [x] Shows a 'Page **X**/Y' right of the pager
- [x] Functional Per Page and Save CSV buttons. Per Page buttons will refresh the window it with a new api call and return the current page back to 1.


src/pages/reports/messageEvents/components/CursorPaging.js
```diff
@@ -0,0 +1,28 @@
...
+    <Button className = {styles.RewindButton}
+      key='rewind'
+      onClick={handleFirstPage}
+    >
+      <FirstPage/>
+    </Button>
+    <Pager>
+      <Pager.Previous
+        onClick={() => handlePageChange(currentPage - 1)}
+        disabled={previousDisabled}/>
```
ewandennis: ~This component knows the current page. Surely it could disable the previous page button without a `previousDisabled` prop.~

This isn't great since the events could arrive out of order and after a delay.

src/pages/reports/messageEvents/components/CursorPaging.js
```diff
@@ -0,0 +1,28 @@
...
+      <Pager.Next
+        onClick={() => handlePageChange(currentPage + 1)}
+        disabled = {nextDisabled}/>
```
ewandennis: ~This component knows the current page, items per page and total item count. It could disable the next page button without a `nextDisabled` prop. `const nextDisabled = Math.ceil(totalCount / perPage) >= numPages`~

Not a great idea to short cut the latest data from the API.

src/pages/reports/messageEvents/tests/MessageEventsPage.test.js
```diff
@@ -54,6 +72,39 @@ describe('Page: Message Events tests', () => {
     });
   });
 
+  it('should correctly disable previous button', () => {
+    wrapper.setState({ currentPage: 1 });
+    expect(instance.isPreviousDisabled()).toEqual(true);
```
ewandennis: When unit testing components, we should be testing final output where possible. That means our expectations should check the rendered results, and not implementation details if at all possible.

For instance, instead of checking the result of a method you wrote on your component, you could test whether the component rendered `CursorPaging` with its `previousDisabled` prop correctly set: 

```js
expect(wrapper.find(CursorPaging).prop('previousDisabled').toEqual(true);
```

The same logic applies to all of our tests. Avoid testing internal state (e.g. `wrapper.state('whatever')` or `instance.whatever`) and instead test the component's rendered output (e.g. `wrapper.find(...)...`). 

This isn't always possible but it's a good general rule.

I recommend getting familiar with [enzyme](https://airbnb.io/enzyme/docs/api/shallow.html) for querying rendered output of React components.

src/pages/reports/messageEvents/tests/MessageEventsPage.test.js
```diff
@@ -54,6 +72,39 @@ describe('Page: Message Events tests', () => {
...
+  });
+
+  it('should correctly disable next button', () => {
+    wrapper.setState({ currentPage: 3 });
+    expect(instance.isNextDisabled()).toEqual(true);
```
ewandennis: `expect(wrapper.find(CursorPaging).prop('nextDisabled').toEqual(true);`

src/pages/reports/messageEvents/tests/MessageEventsPage.test.js
```diff
@@ -54,6 +72,39 @@ describe('Page: Message Events tests', () => {
...
+  });
+
+  it('should correctly handle changing per page', () => {
+    const { getMessageEvents, search } = props;
+    instance.handlePerPageChange(100);
+    expect(wrapper.state().perPage).toEqual(100);
```
ewandennis: I would remove this expectation that tests component internal state. The fact that the `getMessageEvents` action is called with the correct `perPage` value is enough.

src/pages/reports/messageEvents/tests/MessageEventsPage.test.js
```diff
@@ -54,6 +72,39 @@ describe('Page: Message Events tests', () => {
...
+    expect(getMessageEvents).toHaveBeenCalledWith({ ...search, perPage: 100 });
+  });
+
+  it('should correctly handle changing page', () => {
+    const { changePage, linkByPage, cachedResultsByPage } = props;
+    instance.handlePageChange(1);
+    const { currentPage } = wrapper.state();
+    expect(currentPage).toEqual(1);
```
ewandennis: Again, we don't need to raise expectations against internal state.

src/pages/reports/messageEvents/tests/MessageEventsPage.test.js
```diff
@@ -64,11 +115,6 @@ describe('Page: Message Events tests', () => {
     expect(wrapper).toMatchSnapshot();
   });
 
-  it('renders banner correctly with too many (1000) records', () => {
```
ewandennis: Heh. Removing this restriction is the big payoff :)

src/pages/reports/messageEvents/tests/MessageEventsPage.test.js
```diff
@@ -54,6 +72,39 @@ describe('Page: Message Events tests', () => {
...
+    const { currentPage } = wrapper.state();
+    expect(currentPage).toEqual(1);
+    expect(changePage).toHaveBeenCalledWith({ linkByPage, currentPage: 1, cachedResultsByPage });
+  });
+
+  it('should correctly handle first page reload', () => {
+    const { getMessageEvents, search } = props;
+    instance.handleFirstPage();
+    const { perPage } = wrapper.state();
+    expect(wrapper.state().currentPage).toEqual(1);
```
ewandennis: Please remove `state` expectations.

src/pages/reports/messageEvents/components/CursorPaging.js
```diff
@@ -0,0 +1,28 @@
...
+    </Pager>
+    <span className={styles.PageDisplay}
+    >Page: <strong className={styles.Bold} style={{ fontSize: '100%' }}>{currentPage}</strong> / {Math.ceil(totalCount / perPage)}</span>
```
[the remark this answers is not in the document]
kchu1337: Doesn't seem to change text size anyways so I am removing it.

ewandennis:
Merged into SA-358 branch

PR #789, opened 2019-01-22
EO-502 Signals styling feedback

jonambas:


src/pages/signals/components/HealthScoreOverview.js
```diff
@@ -160,6 +160,7 @@ class HealthScoreOverview extends React.Component {
                   facet={facet.key}
                   id={props[facet.key]}
                   subaccountId={props.sid}
+                  truncate
```
bkemper: This prop should be set for subaccount as well for all overview tables.

jonambas: Only health score overview has that column

PR #793, opened 2019-01-23
TR-938: Do not wait for supporting data to render template form

bkemper:
**What changed?**

- Template form will load immediately (previously, it would show a spinner while loading sending domains which could be 30+ seconds for some of our accounts with a large number of sending domains)
- Let API validate sending domain in the "From Email" field

screen shot 2019-01-31 at 3 10 39 pm
This is the error that users should receive if they provide an invalid sending domain.


bkemper:
Rebased.

PR #804, opened 2019-01-31
FE-585: Prevent translation on preview to avoid React from breaking

PatrickSison:


src/pages/templates/components/PreviewPanel.js
```diff
@@ -46,12 +46,15 @@ export default class PreviewPanel extends Component {
       : this.props[contentType];
 
     return (
-      <div className={styles.PreviewPanel}>
-        <Tabs selected={selectedTabIndex} tabs={tabs} />
-        <div className={styles.PreviewPanelWrapper}>
-          <PreviewFrame content={content} key={contentType} />
+      <div className="notranslate">
+        <div className={styles.PreviewPanel}>
```
kchu1337: I think you can use `classnames` instead of using another `<div>` tag. If you made a conscious decision not to use `classnames`, then it's fine

```suggestion
        <div className={classnames(styles.PreviewPanel, 'notranslate')}>
```

PR #805, opened 2019-01-31
TR-985: From email populated with sandbox domain for new templates

bkemper:


bkemper:
Rebased.

bkemper:
This was merged.

PR #806, opened 2019-02-04
EO-562 Display disposable and role validation

jonambas:


src/reducers/tests/__snapshots__/recipientVerificationLists.test.js.snap
```diff
@@ -30,8 +30,10 @@ Object {
   "latestJobLoading": false,
   "singleResults": Object {
     "email": "invalid@address.com",
-    "reason": undefined,
-    "valid": undefined,
+    "singleResults": Object {
```
jonambas: Fix these tests!

PR #807, opened 2019-02-04
FE-615: Storybook: app-centric UI patterns

ewandennis:
## What I Did
 - Installed and configured [Storybook](https://storybook.js.org)
    - Config: `.storybook/`
    - Stories: `src/stories/`
 - Sketched some `ListPage` stories as an example of a formal list page concept

## Why
We have lots of UI patterns but we tend to cut and paste them across instances. That means the usual copypasta issues but also no formal recognition of TheRightWay to present a piece of UI. With a library of high-level patterns, we can more quickly and easily present new features and revise those we already have.

## How To Test / Play
 - `npm run storybook`
 - The storybook is at http://localhost:9001/
 - Stories are under `src/stories`
 - [Docs for writing stories here](https://storybook.js.org/basics/writing-stories/)


.storybook/config.js
```diff
@@ -0,0 +1,12 @@
+import { configure } from '@storybook/react';
+
+import '../src/critical.scss';
+import '../src/index.scss';
+
+const req = require.context('../src/stories', true, /.stories.js$/);
```
bkemper: Are stories source?

bkemper:
The only suggestion I have is if we can, we should probably move `stories` out of `/src`.

PR #809, opened 2019-02-05
FE-608: Update list of supported browsers

bkemper:
Run `npx browserslist` to spotcheck the browserslist.

### DIFF
--- a/scripts/tests/__snapshots__/browsersSnapshotGen.test.js.snap
+++ b/scripts/tests/__snapshots__/browsersSnapshotGen.test.js.snap
@@ -3,19 +3,19 @@
 exports[`BrowsersSnapshotGenerator returns list of supported browser 1`] = `
 Object {
   "desktop": Object {
-    "chrome": ">66.0.0",
-    "edge": ">16.0.0",
-    "firefox": ">52.0.0",
-    "opera": ">52.0.0",
-    "safari": ">11.0.0",
+    "chrome": ">58.0.0",
+    "edge": ">17.0.0",
+    "firefox": ">60.0.0",
+    "opera": ">56.0.0",
+    "safari": ">11.1.0",
   },
   "mobile": Object {
     "android": ">4.4.3",
     "baidu": ">7.12.0",
     "blackberry": ">7.0.0",
-    "chrome": ">67.0.0",
-    "firefox": ">60.0.0",
-    "ios": ">11.0.0",
+    "chrome": ">71.0.0",
+    "firefox": ">64.0.0",
+    "ios": ">11.3.0",
     "operamini": ">all.0.0",
     "operamobile": ">12.1.0",
     "qqandroid": ">1.2.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -3512,21 +3512,30 @@
       }
     },
     "browserslist": {
-      "version": "4.0.0",
-      "resolved": "https://registry.npmjs.org/browserslist/-/browserslist-4.0.0.tgz",
-      "integrity": "sha512-fJq4izbUYHHNdQd/5Mco31HeL8U8dg5sSaj5boaDP17+aAe41CrxSZbQifIjaWw27iIilmy48z9PrVtelNJhbw==",
+      "version": "4.4.1",
+      "resolved": "https://registry.npmjs.org/browserslist/-/browserslist-4.4.1.tgz",
+      "integrity": "sha512-pEBxEXg7JwaakBXjATYw/D1YZh4QUSCX/Mnd/wnqSRPPSi1U39iDhDoKGoBUcraKdxDlrYqJxSI5nNvD+dWP2A==",
       "dev": true,
       "requires": {
-        "caniuse-lite": "^1.0.30000859",
-        "electron-to-chromium": "^1.3.50",
-        "node-releases": "^1.0.0-alpha.10"
+        "caniuse-lite": "^1.0.30000929",
+        "electron-to-chromium": "^1.3.103",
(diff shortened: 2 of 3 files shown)


scripts/tests/__snapshots__/browsersSnapshotGen.test.js.snap
```diff
@@ -3,19 +3,19 @@
 exports[`BrowsersSnapshotGenerator returns list of supported browser 1`] = `
 Object {
   "desktop": Object {
-    "chrome": ">66.0.0",
-    "edge": ">16.0.0",
-    "firefox": ">52.0.0",
-    "opera": ">52.0.0",
-    "safari": ">11.0.0",
+    "chrome": ">58.0.0",
```
rajumsys: is it supposed to be 58 or 68?

ewandennis: 68 came out in 2018-07. Thats a little tight. Here's the Chrome usage spread over the last few weeks:
screen shot 2019-02-05 at 14 36 03

scripts/tests/__snapshots__/browsersSnapshotGen.test.js.snap
```diff
@@ -3,19 +3,19 @@
...
+    "edge": ">17.0.0",
+    "firefox": ">60.0.0",
```
ewandennis: Here's the Firefox spread, a chunk of folk on v56 that we're about to disown:

screen shot 2019-02-05 at 14 38 16

scripts/tests/__snapshots__/browsersSnapshotGen.test.js.snap
```diff
@@ -3,19 +3,19 @@
...
+    "opera": ">56.0.0",
+    "safari": ">11.1.0",
```
ewandennis: A few Safari 11.0 users too:
screen shot 2019-02-05 at 14 40 59

scripts/tests/__snapshots__/browsersSnapshotGen.test.js.snap
```diff
@@ -3,19 +3,19 @@
 exports[`BrowsersSnapshotGenerator returns list of supported browser 1`] = `
 Object {
   "desktop": Object {
-    "chrome": ">66.0.0",
-    "edge": ">16.0.0",
-    "firefox": ">52.0.0",
-    "opera": ">52.0.0",
-    "safari": ">11.0.0",
+    "chrome": ">58.0.0",
+    "edge": ">17.0.0",
```
ewandennis: Edge share is small, few very few <17.

scripts/tests/__snapshots__/browsersSnapshotGen.test.js.snap
```diff
@@ -3,19 +3,19 @@
...
+    "firefox": ">60.0.0",
+    "opera": ">56.0.0",
```
ewandennis: Opera share is small, mostly >=v57

scripts/tests/__snapshots__/browsersSnapshotGen.test.js.snap
bkemper: Disown is a strong word.  I'll identify the impact and if we must we can override it.

scripts/tests/__snapshots__/browsersSnapshotGen.test.js.snap
rajumsys: i meant whether it's a typo; not as a suggestions to use it.

bkemper: Unfortunately, it is not a typo.

```

sparkpost/2web2ui FE-608$ npx browserslist "> 1%"
and_chr 71
and_uc 11.8
chrome 71
chrome 70
chrome 58
edge 17
firefox 64
firefox 63
ie 11
ios_saf 12.0-12.1
ios_saf 11.3-11.4
op_mini all
safari 12
```

bkemper:
A couple of topics to hash out.

**Not Used By Babel**

The react-app preset [ignores our browserslist](https://github.com/facebook/create-react-app/blob/master/packages/babel-preset-react-app/create.js#L89) and [sets its own targets](https://github.com/facebook/create-react-app/blob/master/packages/babel-preset-react-app/create.js#L84-L86).  All that discussion about not supporting IE and I think we have been minus polyfills.  Should we change our preset?

**Measuring Impact**

Our browserslist is used by autoprefixer, babel, and `scripts/browsersSnapshotGen.js`.  All three provide a way to detect change when changing your browserslist.  Autoprefixer provides `npx autoprefixer --info` to generate a list of at-rules, selectors, and properties needed for the targets.  Babel provides a [debug option](https://babeljs.io/docs/en/babel-preset-env#debug) to generate a list of plugins for the targets.  Our script has a snapshot test.  All three have different consequences.  If a property is missing from autoprefixer, a user might not get a properly styled DOM element, but our app won't break.  If a plugin is missing, a error will be thrown once that line of code is reached and the user will not be able to complete their task.  If our target browsers is too strict, errors will be reported to Sentry, but as warnings and not errors.  Are there any other considerations?

**Maintenance**

I fear maintaining a minimum version for each browser.  The beauty of "> 1%" is to leverage caniuse database, so you can set it and forget it.  An alternative could be to relax our 1% to something really low like .25% which doesn't change our bundle size right now and configure a [performance budget](https://webpack.js.org/configuration/performance/) to control bundle size moving forward.  Which is better?

bkemper:
Closing for now.

PR #810, opened 2019-02-05
FE-574: Right aligned popover containers do not resize when hidden

PatrickSison:
Bug is when window is dragged to smaller width and is horizontally scrollable (shouldn't be). Fixed it so it no longer scrolls

### DIFF
--- a/src/components/navigation/components/AccountDropdown.js
+++ b/src/components/navigation/components/AccountDropdown.js
@@ -55,7 +55,7 @@ export class AccountDropdown extends Component {
 
   render() {
     return (
-      <Popover left trigger={this.renderActivator()} open={this.state.open} onClose={this.toggleDropdown}>
+      <Popover popoverId='popover-portal-2' left trigger={this.renderActivator()} open={this.state.open} onClose={this.toggleDropdown}>
         <ActionList
           className={styles.AccountList}
           onClick={this.toggleDropdown}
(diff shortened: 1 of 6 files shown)


src/index.scss
```diff
@@ -11,6 +11,15 @@ html {
   @media screen and (min-width: breakpoint(larger)) { font-size: rem(20); }
 }
 
+#popover-portal {
+  position: absolute;
```
jonambas: Some weird behavior, check vertical scroll on large pages and clicking underneath this div

I think it needs `pointer-events: none`, and `height: 100%`

src/components/navigation/components/AccountDropdown.js
```diff
@@ -55,7 +55,7 @@ export class AccountDropdown extends Component {
 
   render() {
     return (
-      <Popover left trigger={this.renderActivator()} open={this.state.open} onClose={this.toggleDropdown}>
+      <Popover portalId='popover-portal' left trigger={this.renderActivator()} open={this.state.open} onClose={this.toggleDropdown}>
```
jonambas: This sucks but we should maybe abstract out `Popover` into something like:
```js
const AppPopover = (props) => <Popover portalId='popover-portal' {...props} />
```

So we don't have to add an ID everywhere

jonambas: ^ to follow that up, all other popovers have the same issue, eg. those action list popovers in list pages

jonambas: OR maybe we fix the issue on the matchbox side so this wont be necessary?

jonambas:
Moving the fix into matchbox

PR #814, opened 2019-02-06
EO-574 Display recommended actions for selected day

jonambas:
This PR:
-  Adds day selection into the engagement recency and spam trap details pages. Similar to day selection in health score.
- Displays recommended actions based on the day selected on all details pages

--- 

Difficult to verify locally / staging. I can provide reviewer with a way to view some data.

### DIFF
--- a/src/pages/signals/EngagementRecencyPage.js
+++ b/src/pages/signals/EngagementRecencyPage.js
@@ -23,6 +23,35 @@ import cohorts from './constants/cohorts';
 import styles from './DetailsPages.module.scss';
 
 export class EngagementRecencyPage extends Component {
+  state = {
+    selectedDate: null
+  }
+
+  componentDidMount() {
+    const { selected } = this.props;
+
+    if (selected) {
+      this.setState({ selectedDate: selected });
+    }
+  }
+
+  componentDidUpdate(prevProps) {
+    const { data } = this.props;
+    const { selectedDate } = this.state;
+
+    const dataSetChanged = prevProps.data !== data;
+    let selectedDataByDay = _.find(data, ['date', selectedDate]);
+
+    // Select last date in time series
+    if (dataSetChanged && !selectedDataByDay) {
+      selectedDataByDay = _.last(data);
+      this.setState({ selectedDate: selectedDataByDay.date });
+    }
+  }
+
+  handleDateSelect = (node) => {
+    this.setState({ selectedDate: _.get(node, 'payload.date') });
+  }
 
   getYAxisProps = () => ({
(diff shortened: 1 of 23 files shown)


src/pages/signals/EngagementRecencyPage.js
```diff
@@ -23,6 +23,35 @@ import cohorts from './constants/cohorts';
 import styles from './DetailsPages.module.scss';
 
 export class EngagementRecencyPage extends Component {
+  state = {
+    selectedDate: null
+  }
+
+  componentDidMount() {
+    const { selected } = this.props;
```
ewandennis: Where is the `selected` prop set? `withEngagementRecencyDetails` doesn't seem to set it. If its unused, lets remove it.

...or wire it up like the Health Score page using `getSelectedDateFromRouter` from `src/selectors/signals`.

src/pages/signals/EngagementRecencyPage.js
```diff
@@ -53,6 +82,8 @@ export class EngagementRecencyPage extends Component {
 
   renderContent = () => {
     const { data = [], loading, gap, empty, error } = this.props;
+    const { selectedDate } = this.state;
+    const selectedData = _.find(data, ['date', selectedDate]) || {};
```
ewandennis: Do we need the default `{}` here?

(I am assuming the API returns a full set of records for the selected period)

ewandennis: `selectedDate` and `selectedData` are bit collidy. How about `s/selectedData/selectedRecord/`?

src/pages/signals/SpamTrapPage.js
```diff
@@ -23,13 +23,40 @@ import styles from './DetailsPages.module.scss';
 
 export class SpamTrapPage extends Component {
   state = {
+    selectedDate: null,
     calculation: 'relative'
   }
 
+  componentDidMount() {
+    const { selected } = this.props;
```
ewandennis: Same question here about `selected`. It's not wired up in the container.

src/pages/signals/SpamTrapPage.js
```diff
@@ -59,7 +86,8 @@ export class SpamTrapPage extends Component {
 
   renderContent = () => {
     const { data = [], loading, gap, empty, error } = this.props;
-    const { calculation } = this.state;
+    const { calculation, selectedDate } = this.state;
+    const selectedData = _.find(data, ['date', selectedDate]) || {};
```
ewandennis: Same question about the default `{}`. Do we know if `data` is sparse?

jonambas: Seems like this is here for initial render before loading or empty are set

PR #816, opened 2019-02-07
AC-560: Hide Help Center, Alerts, and Signals from users without grants

PatrickSison:
- [x] Contrary to the AC, this currently hides the whole help panel, not just the support ticket form.
 - [x] /dashboard is not available to`subaccount_reporting` users so their default redirect should be the summary report

### DIFF
--- a/src/components/navigation/components/Top.js
+++ b/src/components/navigation/components/Top.js
@@ -32,7 +32,7 @@ export class Top extends Component {
       <div className={styles.RightSideWrapper}>
         <NotificationCenter />
         <AccessControl condition={hasGrants('support/manage')}>
-          <HelpOutline className={styles.SupportIcon} onClick={this.openSupportPanel} size={22} />}
+          <HelpOutline className={styles.SupportIcon} onClick={this.openSupportPanel} size={22} />
         </AccessControl>
         <AccountDropdown />
       </div>
--- a/src/components/navigation/components/Top.js
+++ b/src/components/navigation/components/Top.js
@@ -10,6 +10,10 @@ import AccountDropdown from './AccountDropdown';
 import NotificationCenter from 'src/components/notifications/NotificationCenter';
 import styles from './Top.module.scss';
 import { DEFAULT_REDIRECT_ROUTE } from 'src/constants';
+import { hasGrants } from 'src/helpers/conditions';
+import { AccessControl } from 'src/components/auth';
+import { isSubaccountUser } from 'src/helpers/conditions/user';
+import not from 'src/helpers/conditions/not';
 
 export class Top extends Component {
   renderMobile = () => (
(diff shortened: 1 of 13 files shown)


src/config/routes.js
```diff
@@ -265,27 +265,31 @@ const routes = [
   {
     path: '/signals',
     component: signals.OverviewPage,
+    condition: hasGrants('signals/manage'),
```
kchu1337: Can you add this grant to the list in the comments on ~line 60? Also change `Angular UI` on line 60

src/config/routes.js
```diff
@@ -675,6 +679,7 @@ const routes = [
   {
     path: '/alerts',
     component: AlertsPage,
+    condition: hasGrants('alerts/manage'),
```
kchu1337: Same here with `alerts/manage`

src/helpers/conditions/user.js
```diff
@@ -9,6 +9,8 @@ export const hasRole = (role) => ({ currentUser }) => currentUser.access_level =
 
 export const isAdmin = any(hasRole('admin'), hasRole('superuser'));
 
+export const isSubaccountUser = ({ currentUser }) => currentUser.access_level && currentUser.access_level.includes('subaccount');
```
kchu1337: Can do a nested destructure here. Also note that if `access_level` does not exist, it will return `undefined`. It's still falsy, but the other functions return the boolean `false`. Shouldn't be an issue though. 
```suggestion
export const isSubaccountUser = ({ currentUser : {access_level} }) => access_level && access_level.includes('subaccount');
```

src/config/routes.js
ewandennis: Those 2 comments are both out of date and unnecessary. Removing in favour of a note pointing to the source of truth: `@sparkpost/access`.

ewandennis:
Contrary to the AC, this currently hides the whole help panel, not just the support ticket form.

Is there a way to test with a subaccount_reporting user in `tst` today?

src/components/navigation/components/Top.js
```diff
@@ -29,7 +31,9 @@ export class Top extends Component {
       <Link to={DEFAULT_REDIRECT_ROUTE} className={styles.Logo}><SparkPost.Logo type='halfWhite' /></Link>
       <div className={styles.RightSideWrapper}>
         <NotificationCenter />
-        <HelpOutline className={styles.SupportIcon} onClick={this.openSupportPanel} size={22} />
+        <AccessControl condition={hasGrants('support/manage')}>
```
nancyhabecker: There's a "Get Help" link in the user menu that opens the support panel as well.

ewandennis: This wholesale removal of the support panel wasn't in the AC. I have instead hidden just the support form.

src/config/supportIssues.js
```diff
@@ -121,7 +134,7 @@ const augmentIssuesList = function () {
   return _.map(supportIssues, (supportIssue) => ({
     ...supportIssue,
     messageLabel: supportIssue.messageLabel || defaultMessageLabel,
-    condition: supportIssue.condition || idConditionsMap[supportIssue.id] || defaultCondition
+    condition: supportIssue.condition || defaultCondition
```
kchu1337: We can remove `hasGrants` from the default condition and individual conditions and just add it at the end. 
```suggestion
    condition: all(supportIssue.condition || defaultCondition, hasGrants('support/manage'))
```
We could also revert every change and do

```suggestion
    condition: all(supportIssue.condition || idConditionsMap[supportIssue.id] || defaultCondition, hasGrants('support/manage'))
```

OR even make changes in `selectors/support` instead of here.

PR #819, opened 2019-02-12
FE-619 Hydrate redux store with website auth

kchu1337:
See Jira ticket for testing. 
My approach to fixing:
1. Hydate redux store with websiteAuth in `AuthenticationGate`. 
2. Remove login dispatch and have the entire authenticate and save cookie process in one dipatch.
3. Add getting websitecookie from cookie store, and change save method.

### DIFF
--- a/src/actions/websiteAuth.js
+++ b/src/actions/websiteAuth.js
@@ -4,7 +4,11 @@ import config from 'src/config';
 const { authentication } = config;
 const { site: siteCfg } = authentication;
 
-function successAction({ data = {}}) {
+export function successAction({ data = {}, saveCookie = false }) {
+  if (saveCookie) {
+    siteCookie.save(data);
+  }
+
   return {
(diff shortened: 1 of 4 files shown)


src/actions/websiteAuth.js
```diff
@@ -34,22 +38,11 @@ export function authenticate(username, password, rememberMe = false) {
...
-    dispatch({
-      type: 'WEBSITE_AUTH_COMPLETE'
-    });
-  };
-}
-
 export function refresh() {
   return (dispatch, getState) => {
-    const { websiteAuth: { refreshToken }} = getState();
+    const { websiteAuth: { refresh_token }} = getState();
```
kchu1337: If we do something like 
```suggestion
    const  { refresh_token } = siteCookie.get();
```
We don't even need to save website auth in the redux store.

PR #820, opened 2019-02-13
FE-625 handle invalid dates and default to day range for invalid custom range

rajumsys:
- Upon component error, we redirect to landing page. Still support reload for any app level exceptions
- i've also kept the previous fix to use default value for date when invalid date is encountered.


src/helpers/tests/reports.test.js
```diff
@@ -54,9 +55,21 @@ describe('report helpers', () => {
       expect(reports.parseSearch(search)).toMatchSnapshot();
     });
 
-    it('Should parse search with no empty value', () => {
+    it('should parse search with no empty value', () => {
       expect(reports.parseSearch('')).toMatchSnapshot();
```
ewandennis: `expect(reports.parseSearch('')).toEqual({ options: {}})` ?

src/helpers/reports.js
```diff
@@ -47,16 +48,20 @@ export function parseSearch(search) {
...
 
-  if (to) {
-    options.to = new Date(to);
+  if (to && toTime.isValid()) {
+    options.to = toTime.toDate();
   }
 
   if (range) {
-    options = { ...options, ...getRelativeDates(range) };
+    const effectiveRange = options.from && options.to ? range : 'day';
```
ewandennis: Could we please include tests to cover this new logic?

src/helpers/tests/reports.test.js
```diff
@@ -1,5 +1,6 @@
 import * as reports from '../reports';
 import * as dateHelpers from 'src/helpers/date';
+import cases from 'jest-in-case';
```
ewandennis: Some nice new explicit assertions here. ❤️ 

It would be even better if we could replace some of the existing snapshot tests with assertions. That would make the `parseSearch` contract much clearer. At the moment, the snapshots just document the current behaviour, they don't define it in any useful manner.

src/helpers/reports.js
```diff
@@ -47,16 +48,20 @@ export function parseSearch(search) {
     options.metrics = (typeof metrics === 'string') ? [metrics] : metrics;
   }
 
-  if (from) {
-    options.from = new Date(from);
```
ewandennis: This change moves from browser-defined date string parsers to Moment's ISO 8601 parser. Browser date parsing is variable ([MDN reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#Parameters) so we _might_ confuse some people who edit URLs to fit their browser's date formats.

Hopefully thats minimal but I guess we'll no longer detect it since the app no longer breaks in that situation.

src/components/errorBoundaries/ErrorBoundary.js
```diff
@@ -23,7 +18,16 @@ export default class ErrorBoundary extends Component {
     ErrorTracker.report('error-boundary', error);
   }
 
+  handleCtaClick = () => {
+    window.location.replace(DEFAULT_REDIRECT_ROUTE); //deliberately using native location to avoid any potential problem with react router itself
+  }
+
   render() {
+    const primaryAction = {
+      content: this.props.ctaLabel || 'Go to Dashboard',
```
ewandennis: Enterprise and reporting folks don't have a dashboard so we should say something safer such as `Back to safety` here.

src/App.js
```diff
@@ -18,7 +18,7 @@ import {
 } from 'react-router-dom';
 
 const App = () => (
-  <ErrorBoundary>
+  <ErrorBoundary onCtaClick={() => window.location.reload(true)} ctaLabel='Reload Page'>
```
ewandennis: I don't usually worry too much about inline functions in props. If we can be sure that `App` will not re-render, say, each time the route changes, this is probably fine.

src/components/errorBoundaries/tests/ErrorBoundary.test.js
```diff
@@ -35,4 +36,18 @@ describe('Component: ErrorBoundary', () => {
...
+    wrapper.setProps({ ctaLabel: 'To Safety' });
+    wrapper.setState({ hasError: true });
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('uses custom action when passed', () => {
+    const mockFn = jest.fn();
+    wrapper.setProps({ ctaLabel: 'Reload Page', onCtaClick: mockFn });
+    wrapper.setState({ hasError: true });
+    expect(wrapper).toMatchSnapshot();
```
ewandennis: I don't think this snapshot is necessary, since we're just verifying the action function is intact.

src/components/errorBoundaries/tests/ErrorBoundary.test.js
```diff
@@ -35,4 +36,18 @@ describe('Component: ErrorBoundary', () => {
     expect(wrapper.state('hasError')).toEqual(true);
     expect(ErrorTracker.report).toHaveBeenCalledWith('error-boundary', error);
   });
+
+  it('renders custom cta label when passed', () => {
+    wrapper.setProps({ ctaLabel: 'To Safety' });
+    wrapper.setState({ hasError: true });
+    expect(wrapper).toMatchSnapshot();
```
ewandennis: Can we avoid the snapshot here? They're large in this component and difficult to review.

`expect(wrapper.find('EmptyState').prop('primaryAction').content).toEqual(props.ctaLabel);`

src/components/errorBoundaries/ErrorBoundary.js
```diff
@@ -23,7 +18,16 @@ export default class ErrorBoundary extends Component {
     ErrorTracker.report('error-boundary', error);
   }
 
+  handleCtaClick = () => {
+    window.location.replace(DEFAULT_REDIRECT_ROUTE); //deliberately using native location to avoid any potential problem with react router itself
```
ewandennis: Can we test that clicking the primary action navigates to `/landing` ?

src/App.js
rajumsys: I see your point and I'll be surprised if it's re-enders on each route change. But regardless, I can move it out to a function and reference here. at minimum, we won't have to think about the issue with re-renders.

src/components/errorBoundaries/ErrorBoundary.js
rajumsys: i wanted to then i thought it's like testing EmptyState component. but I can add a test separately just for this method.

src/components/errorBoundaries/ErrorBoundary.js
rajumsys: Yeah, that's why I redirected to landing then dasbhoard but wasn't sure about copy.

src/helpers/reports.js
rajumsys: i have done lot of back and forth with this. we can omit format and strict check. that'll leave it to browser's implementation of parsing date from string. but moment keeps yelling at me that it'll be removed in upcoming versions. i can support multiple formats though (using `INPUT_DATETIMES` constant we've)

src/components/errorBoundaries/ErrorBoundary.js
ewandennis: > Go Back

src/components/errorBoundaries/ErrorBoundary.js
ewandennis: I meant that we should verify `ErrorBoundary` behaviour: clicking its primary action causes navigation to `DEFAULT_REDIRECT_ROUTE`. We shallow render so `EmptyState` is not implicated:

```js
  it('navigates to default landing page on primary action click', () => {
    window.location.replace = jest.fn();
    wrapper.setState({ hasError: true });
    wrapper.find('EmptyState').prop('primaryAction').onClick();
    expect(window.location.replace).toHaveBeenCalledWith(DEFAULT_REDIRECT_ROUTE);
  });
```

rajumsys: - I've added test to verify component is rendering using the function we've passed - https://github.com/SparkPost/2web2ui/pull/820/files#diff-db341567afbadcb446dc8a0b61124c1dR51

- then verified onCtaClick is calling `location.replace` in https://github.com/SparkPost/2web2ui/pull/820/files#diff-db341567afbadcb446dc8a0b61124c1dR54

 IMO, by this test we're aiming to test whether EmptyState component is invoking the function. But in unit test we're testing our implementations only. I believe above two tests asserts the behavior of our implementations. let me know if you still think we need this test.

PR #823, opened 2019-02-14
FE-626: Testing Documentation TLC

bkemper:
Key points to hit:
 - [x] Why: the goals of unit/integration testing
    - define interface contracts / prove logic correctness
    - help refactors / regressions
    - hilight architectural issues / needs / anti-patterns
 - [x] What: how to decide what tests we need
    - account for each piece of logic within the subject
    - use tools to gauge _logic_ coverage
 - [x] How: the main aspects of unit testing a component
    - input: props, events
    - output: DOM
    - assertions: try to avoid snapshots. Use enzyme instead.
    - Prefer `wrapper.find('MyTastyThing')` to `wrapper.find(MyTastyThing)`
    - Call out enzyme-jest matchers
 - [x] stub out deps only when necessary
 - [x] verbosity: refer to tool docs rather than repeating in line
 - [x] know your tools


bkemper:
Do we want to add [jest-extended](https://github.com/jest-community/jest-extended)?

docs/testing/README.md
```diff
@@ -1,27 +1,49 @@
-# Testing Introduction
+# Testing
 
-This section of our docs is about how we test our code. We'll explain the types of testing we do and the tools we use to do it, as well as some examples of how to write tests for our most common cases.
+This section of our documentation provides guidance on how we test our code.  It is important to note that writing tests is not just a task to confirm your code works as expected, but an exercise to validate your code design.  If your code is hard to test, consider a refactor.
```
rajumsys: > If your code is hard to test, consider a refactor. 💯

docs/testing/README.md
```diff
@@ -1,27 +1,49 @@
...
+ * Explicit, Not Implicit - avoid using [manual mocks](https://jestjs.io/docs/en/manual-mocks)
+ * Assert Outputs Not Internals
+
+
+## Patterns
+
+These are common patterns for testing our basic file types.
+
+ * [Action Creators](patterns/action-creators.md)
+ * [Component Call To Action](patterns/component-call-to-action.md)
```
ewandennis: I understand the intent behind the term "call to action" but it clashes with the established term "CTA".

How about:
- Component Function Calls
- Component Function Props

docs/testing/patterns/component-inspection.md
```diff
@@ -0,0 +1,20 @@
...
+```js
+it('renders a disabled button', () => {
+  const wrapper = shallow(<MyComponent disabled={true} />);
+  const button = wrapper.find('button');
+
+  expect(button).toHaveProp('disabled', true);
+})
+```
+
+Do not snapshot every component test case.  Instead, to test `MyComponent` renders a disabled button when `disabled` prop is provided, [`find`](https://airbnb.io/enzyme/docs/api/ShallowWrapper/find.html) the button and inspect its disabled prop value.
```
ewandennis: Should we pull these tasty notes into a single testing antipatterns list? That might be easier to consume for more experienced folk.

docs/testing/patterns/component-state-changed.md
```diff
@@ -0,0 +1,35 @@
...
+
+```js
+it('opens modal when button is clicked', () => {
+  const wrapper = shallow(<MyComponent />);
+  wrapper.find('button').simulate('click');
+  expect(wrapper.find('Modal')).toHaveProp('isOpen', true);
+})
+```
+
+Do not inspect the state of `MyComponent`.  Instead, to test `MyComponent` opens `Modal` when `button` is clicked, [`simulate`](https://airbnb.io/enzyme/docs/api/ShallowWrapper/simulate.html) a click of `button` and assert `isOpen` prop is `true`.
```
ewandennis: A good one for an antipattern list.

docs/testing/README.md
```diff
@@ -1,27 +1,49 @@
...
+ * [Coveralls](https://coveralls.io/) - This is a service used to provide [test coverage analysis](https://github.com/SparkPost/2web2ui/pull/834#issuecomment-466549343) in your pull requests.
 
-   * [Action Creators](testing-actions.md)
-   * [Helpers](testing-helpers.md)
-   * [Selectors](testing-selectors.md)
-   * [Components](testing-components.md)
 
-1. [**Mocking with Jest**](mocking.md)  
-Mocking dependencies and stubbing out methods is pretty different in Jest, so this section goes over common mocking strategies and conventions.
+## Layers
```
ewandennis: Maybe we could take some load off the reader by outlining where our tests are and how to run them.

## Where Are Our Tests?
 - Any file under `src/` with a `.test.js` extension is treated as a test. ([Full test file spec here](https://github.com/SparkPost/2web2ui/blame/master/jest.config.js#L36))
 - By convention, unit tests live in a `tests/` directory beside the items under test.
    - e.g. The unit tests for `src/pages/auth/AuthPage.js` are in `src/pages/auth/tests`
 - Our integration tests live under `src/__integration__`

## How Do I Run Our Tests?
 - To run the whole test suite: `npm test`
 - To run the suite in CI mode (faster, uglier output): `CI=true npm test`
 - To run just a subset of tests: `npm test -- <paths or regex>`
    - e.g. To run just integration tests: `npm test -- src/__integration__`

docs/testing/patterns/dependency-injection.md
```diff
@@ -0,0 +1,33 @@
...
+
+it('should return the right string', () => {
+  const result = two({ localOne: jest.fn(() => 'x') });
+
+  expect(result).toEqual('x - z');
+})
+```
+
+#### Private Functions (or any functions in the same file as the one you're testing)
+
```
bkemper: I need to rework this...

docs/testing/patterns/component-state-changed.md
bkemper: Good, I'm going to pull some of those anti-patterns in to our Principles.

PR #824, opened 2019-02-14
FE-621 Simplify tenant configs

rajumsys:
Includes tenant configs in the bundle

**todo**
- [x] cleanup some config generator
- [x] remove `script` tag for config from index.html or another way to avoid 404 on production. this is needed on development but not on prod. 
- [x] update/fix tests
- [x] update doc

**how to test**
- run development mode and see if it works as usual.
- create a prod build and serve it using (serve or live-server). use a production hostname like `app.sparkpost.com` by overriding `/etc/hosts` so that hostname -> tenant map works. access the site locally using the above hostname. 
- all tests pass


public/index.html
```diff
@@ -55,7 +55,7 @@
       window.addEventListener('load', handleAppLoad);
     </script>
 
-    <script type="text/javascript" src="%PUBLIC_URL%/static/tenant-config/%CONFIG_CACHE_ID%/production.js"></script>
+    <%= process.env.NODE_ENV === 'development' ? '<script type="text/javascript" src="%PUBLIC_URL%/static/tenant-config/%CONFIG_CACHE_ID%/production.js" />' : '' %>
```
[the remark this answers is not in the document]
rajumsys: That's actually good idea. personally wanted to avoid webpack config changes as much as possible (it's always intimidating). i'll give it another shot.

scripts/generateConfigs/index.js
```diff
@@ -1,26 +1,28 @@
...
       }
 
-      writeContent(config.host, content);
+      allConfigs[config.host] = content;
     });
-  })
-}
+  });
+
+  return JSON.stringify(allConfigs);
```
ewandennis: Could we stringify at the point of use in `webpack.config.js`? That would match the other variable we define. It would make our snapshot a little easier to read too.

src/config/index.js
```diff
@@ -1,9 +1,7 @@
-import defaultConfig from './default';
-import testConfig from './test-config';
 import _ from 'lodash';
+import defaultConfig from './default';
+import envConfig from './env';
 
-const prodConfig = _.get(window, 'SP.productionConfig', {});
```
ewandennis: An unexpected side-effect of this is that we can no longer access app.sparkpost.test using a bundle, since that config is not available in production builds.

Not a huge issue - we can still use the traditional /etc/hosts trick to test any tenant.

docs/config.md
```diff
@@ -9,7 +9,8 @@ import config from 'src/config'
 The configuration module under `src/config/` merges variables from these locations in order:
 
  - `src/config/default.js`
- - `SP.productionConfig` from the current tenant's config file (see below)
```
ewandennis: These docs could use another pass to take the latest `config/env/` setup into account.

rajumsys: Yes, it'll need update. Didn't want to update doc before finalizing approach.

scripts/generateConfigs/index.js
rajumsys: actually i didn't want to stringify at all (and then parse again). it's a regression from my prev commit though. i'm stringifying but not parsing again. So removed stringification completely.

Other variable (`SUPPORTED_BROWSERS`) used as it is (stringified) so that's fine for that one.

src/config/index.js
rajumsys: why is that? if we're using app.sparkpost.test, it'll load `dev-config` as long as env is not production.

public/index.html
```diff
@@ -55,8 +55,6 @@
       window.addEventListener('load', handleAppLoad);
     </script>
 
-    <script type="text/javascript" src="%PUBLIC_URL%/static/tenant-config/%CONFIG_CACHE_ID%/production.js"></script>
```
bkemper: Is this file isn't being loaded and setting the `window.SP` values then won't `trackAppLoadError` not work?

config/webpack.config.js
```diff
@@ -641,7 +642,8 @@ module.exports = function(webpackEnv) {
           logLevel: 'info'
         }),
         new webpack.DefinePlugin({
-          SUPPORTED_BROWSERS: JSON.stringify(snapshotGenerator())
+          SUPPORTED_BROWSERS: JSON.stringify(snapshotGenerator()), //Refer to docs/browser-support-sentry-issue.md for more info
+          TENANT_CONFIGS: generateConfigs()
```
bkemper: What will do this do to our bundle size?

bkemper: Could the [`Dotenv`](https://webpack.js.org/plugins/environment-plugin/#dotenvplugin) plugin make it easier when working with different environments?

public/index.html
rajumsys: that's a good catch. i'll cover this.

config/webpack.config.js
rajumsys: bundle size will be + ~4kB

we could. but not sure, how much effort will needed to handle tenant configurations with that. given the simplicity of current implementation, probably adding another dep is unnecessary. also we've added config files in scm but their motivation is 12factor app (which essentially decouples configs from codes). let me know if you have reasons for it.

bkemper: Just thought it might make the handling of different environments a little easier.

rajumsys: yeah, definitely appreciate the thought. but i was not thinking to add another library for just a couple of lines of codes. 

btw, i'll look into this anyway to see if somehow helps with https://github.com/SparkPost/2web2ui/pull/824#discussion_r259973234

public/index.html
rajumsys: I exposed final config again in `SP.productionConfig`. I don't have other idea without significant refactoring. Let me know if you've further ideas.

config/webpack.config.js
ewandennis: While our bundle will grow, our critical path request count will reduce by 1. For the payload size, I don't think its worth worrying about.

src/config/index.js
```diff
@@ -1,9 +1,11 @@
...
 import _ from 'lodash';
+import defaultConfig from './default';
+import envConfig from './env';
+
+const mergedConfig = _.merge({}, defaultConfig, envConfig);
 
-const prodConfig = _.get(window, 'SP.productionConfig', {});
-const maybeTestConfig = process.env.NODE_ENV === 'test' ? testConfig : {};
-const mergedConfig = _.merge({}, defaultConfig, prodConfig, maybeTestConfig);
+window.SP = {
```
bkemper: I think this is fine as long as we are comfortable with losing coverage for the 404 error case and understand that values will not be present if an error occurs while importing config or before it.

PR #830, opened 2019-02-19
AC-573: Add subaccount to users list page

ayoung1:
**What I did**
- Added a subaccount id column to the users list page
- Added 'subaccount' to search terms

**How to test**
- Create a user with a linked subaccount in Dynamobd
- Load the users list page and verify your user has been picked up
- Verify subaccount search feature, keyword : subaccount

**Todo**
- [ ] Verify user list page includes subaccount_id data

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1912,7 +1912,7 @@
     },
     "ansi-escapes": {
       "version": "3.1.0",
-      "resolved": "http://registry.npmjs.org/ansi-escapes/-/ansi-escapes-3.1.0.tgz",
+      "resolved": "https://registry.npmjs.org/ansi-escapes/-/ansi-escapes-3.1.0.tgz",
       "integrity": "sha512-UgAb8H9D41AQnu/PbWlCofQVcnV4Gs2bBJi9eZPxfU/hgglFh3SMDMENRIqdr7H6XFnXdoknctFByVsCOotTVw==",
       "dev": true
     },
@@ -7403,7 +7403,7 @@
     },
     "finalhandler": {
       "version": "1.1.1",
-      "resolved": "https://registry.npmjs.org/finalhandler/-/finalhandler-1.1.1.tgz",
+      "resolved": "http://registry.npmjs.org/finalhandler/-/finalhandler-1.1.1.tgz",
       "integrity": "sha512-Y1GUDo39ez4aHAw7MysnUD5JzYX+WaIj8I57kO3aEPT1fFRL4sr7mjei97FgnwhAyyzRYmQZaTHb2+9uZ1dPtg==",
       "dev": true,
       "requires": {
@@ -9114,7 +9114,7 @@
         },
         "readable-stream": {
           "version": "2.3.6",
-          "resolved": "https://registry.npmjs.org/readable-stream/-/readable-stream-2.3.6.tgz",
+          "resolved": "http://registry.npmjs.org/readable-stream/-/readable-stream-2.3.6.tgz",
           "integrity": "sha512-tQtKA9WIAhBF3+VLAseyMqZeBjW0AHJoxOtYqSUZNJxauErmLbVm2FW1y+J/YA9dUrAC39ITejlZWhVIwawkKw==",
           "dev": true,
           "requires": {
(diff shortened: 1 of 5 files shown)


src/pages/users/ListPage.js
```diff
@@ -21,6 +22,16 @@ const COLUMNS = [
   null
 ];
 
+const COLUMN_SUBS = [
```
ewandennis: I recommend setting the widths of these columns to avoid crushing the smaller ones. (My huge subaccount reporting user email address on the appteam account shows that up nicely ;) ). How about this: 

```js
const COLUMN_SUBS = [
  { label: 'User', sortKey: 'name', width: '40%' },
  { label: 'Role', sortKey: 'access', width: '11%' },
  { label: 'Subaccount', sortKey: 'subaccount_id', width: '15%' },
  { label: 'Two Factor Auth', sortKey: 'tfa_enabled', width: '8%' },
  { label: 'Last Login', sortKey: 'last_login', width: '14%' },
  null
];
```

ewandennis: I also moves the "subaccount" column just after "role" in the above since they are strongly related. Just a thought.

src/pages/users/ListPage.js
```diff
@@ -54,6 +65,7 @@ export class ListPage extends Component {
     user.access,
```
ewandennis: I reckon we should format the role name since we don't want to present separate "reporting" and "subaccount_reporting" roles. How about this:

```js
function formatRole(role) {
  if (role === SUBACCOUNT_REPORTING_ROLE) {
    return 'reporting';
  }
  return role;
}
```

with this in `src/constants/index.js`:
```js
export const SUBACCOUNT_REPORTING_ROLE = 'subaccount_reporting';
```

src/pages/users/ListPage.js
```diff
@@ -54,6 +65,7 @@ export class ListPage extends Component {
...
     user.tfa_enabled ? <Tag color={'blue'}>Enabled</Tag> : <Tag>Disabled</Tag>,
     user.last_login ? <TimeAgo date={user.last_login} live={false} /> : 'Never',
+    user.subaccount_id ? <SubaccountTag id={user.subaccount_id} /> : <p>null</p>,
```
ewandennis: If you're happy with pulling a full list of subaccounts while loading this page, you _could_ present subaccount names. We could even do it opportunistically by showing ids first, then if and when the subaccount load completes, we update the view to show names.

If that sounds like too much for an 'S' story, so be it. We could use it in other places too though... ;)

package-lock.json
```diff
@@ -5478,7 +5478,7 @@
         },
         "readable-stream": {
           "version": "2.3.6",
-          "resolved": "https://registry.npmjs.org/readable-stream/-/readable-stream-2.3.6.tgz",
```
ewandennis: Please remove your updates to this file from this PR. They're caused by an issue with NPM [which can be resolved like this](https://npm.community/t/some-packages-have-dist-tarball-as-http-and-not-https/285/58).

src/pages/users/tests/ListPage.test.js
```diff
@@ -16,9 +16,11 @@ describe('Page: Users List', () => {
       },
       loading: false,
       listUsers: jest.fn(),
+      hasSubaccounts: true,
```
ewandennis: Please include tests for both sides of this condition. I also recommend writing your assertions against the rendered output of the component using Enzyme, instead of writing a new snapshot. This particular test suite overuses snapshots which makes it difficult to review and verify behaviour.

For instance, you could use this test to verify the the `hasSubaccounts` prop controls rendering of the subaccount column:

```js
  // This assumes `hasSubaccounts=false` in the wrapper as a precondition
  it('should render subaccount info when the account uses them', () => {
    wrapper.setProps({ hasSubaccounts: true });
    expect(wrapper.find('TableCollection').prop('columns')).toContainEqual(expect.objectContaining({
      label: 'Subaccount'
    }));
  });
```

src/pages/users/ListPage.js
ewandennis: Lets just return `null` (to render nothing) if a user has no `subaccount_id` assigned. Explicitly rendering `<p>null</p>` looks a little odd.

PR #831, opened 2019-02-20
AC-564: Invite subaccount reporting users

ewandennis:
# What I Did
 - Enabled optional subaccount assignment for reporting users
 - Updated the role description copy on the user invite page

# How To Test
 - Use an account with subaccounts
 - Flip the `subaccount_reporting` feature flag (see docs/feature-flags.md)
 - Manage users -> Invite User
 - Note the new subaccount checkbox and typeahead when selecting the new role
 - Checkout the user edit page too ;)

# Todo
 - [x] Prototype review feedback
 - [x] Complete implementation
 - [x] Feature flag
 - [x] Check the user edit page for pre-release corner cases, since it shares `RoleRadioGroup`


src/pages/users/components/RoleRadioGroup.js
```diff
@@ -1,40 +1,95 @@
...
-      {...rest}
-    />
-  );
+  render() {
+    const { disabled, ...rest } = this.props;
+
+    const roles = this.renderRoles();
+    const options = roles.map((role) => ({ ...role, disabled }));
+
+    return <RadioGroup title="Role" grid={{ xs: 12, sm: 12, md: 6 }} options={options} {...rest} />;
```
PatrickSison: Not a change for this PR, but when we add more roles, it looks ok in 2 columns, but maybe change to single column  with reporting at the bottom since it's bundled together. It starts to be a bit dense when the roles are in 2 columns

src/pages/users/components/EditForm.js
```diff
@@ -23,7 +24,8 @@ export const EditForm = ({
         <Field
           name="access"
           disabled={user.isCurrentUser}
-          allowSuperUser={currentUser.access === 'superuser'}
+          allowSuperUser={currentUser.access === ROLES.SUPERUSER}
+          allowSubaccountAssignment={false}
```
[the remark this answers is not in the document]
ewandennis: The only change to `EditForm` in this PR is to disable subaccount assignment since it's an invite-time-only thing. AC-563 does the rest.

src/pages/users/components/RoleRadioGroup.js
ewandennis: Yeah, the first prototype used 1 column on larger screens. We can switch to that when we add roles.

PR #832, opened 2019-02-20
AC-563 show subaccount information on user page for subaccount_reporting users

James-Burba:
To test - create a subaccount and invite a user using the users/invite endpoint with an x-msys-subaccount header and access_level=subaccount_reporting. The edit user page for that user should hide the role selection radio buttons and include the name and ID of the subaccount associated with the user.


src/pages/users/EditPage.js
```diff
@@ -56,18 +56,31 @@ export class EditPage extends Component {
     }
   }
 
+  componentDidUpdate() {
+    const { user, subaccount, getSubaccount } = this.props;
+    if (user && user.access === 'subaccount_reporting' && (!subaccount || subaccount.id !== user.subaccount_id)) {
```
PatrickSison: Nitpicking, but I would put the check on current loaded subaccount logic on a new line for clarity

src/pages/users/EditPage.container.js
```diff
@@ -15,7 +16,9 @@ const mapStateToProps = (state, props) => {
     accountSingleSignOn: state.accountSingleSignOn,
     currentUser: state.currentUser,
     isAccountSingleSignOnEnabled: state.accountSingleSignOn.enabled,
+    loading: state.users.loading || (state.subaccounts && state.subaccounts.getLoading),
     loadingError: state.users.error,
+    subaccount: state.subaccounts.subaccount,
```
PatrickSison: Line 19 I'm assuming is making a check to see if state.subaccounts exists before checking getLoading, but the same check isn't made on line 21.

src/pages/users/EditPage.js
James-Burba: Agreed

src/pages/users/EditPage.container.js
James-Burba: yeah, from what I'm reading apparently redux will always keep the default subaccount state so the check isn't necessary? I removed it and am unable to get it to crash at least, so I'll take it out?

PatrickSison: Yeah, I think that it's fine. subaccounts should exist at the very least within the state

PR #834, opened 2019-02-21
TR-1088: Improve domain setup UX for auto verified accounts

bkemper:
### What Changed

For accounts with the auto verify domains option enabled, the re-verify buttons have been removed from the sending domain edit page and a banner was added only for "verified" domains.

screen shot 2019-03-01 at 3 30 25 pm

_Please note that domain status is also displayed on the list page and on the subaccount sending domains tab.  No changes are needed.  They both display ready for values when a domain is verified._ 

In addition, removed "Ready For" tags from edit page for sending domains that are "pending" or "blocked".  This was not in the original scope of this ticket, but is a bug that needed to be fixed.

screen shot 2019-03-01 at 1 02 56 pm

### How To Test

_Please note that we cannot fully test this ticket until AC-619 is released._

1. Create a free account, create a sending domain, and confirm that it is unverified.  (for sanity, click the verify buttons to confirm they fail)
2. Upgrade the account to a paid plan and enable the account option ([see instructions](https://jira.int.messagesystems.com/browse/TR-1088?focusedCommentId=488868&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-488868)).  Then confirm that the auto verify stamp icons are displayed and the verify buttons are not displayed.


bkemper:
I need to tweak the logic to cover this case.  We should not be showing "Ready For" when status is "pending" compliance check.

screen shot 2019-02-26 at 11 15 50 am

src/pages/billing/forms/tests/ChangePlanForm.test.js
```diff
@@ -57,7 +55,6 @@ describe('Form Container: Change Plan', () => {
       updateSubscription: jest.fn(() => Promise.resolve()),
       isAws: false
     };
-    accountConditions.isAws = jest.fn(() => false);
     wrapper = shallow(<ChangePlanForm {...props} />);
     instance = wrapper.instance();
     submitSpy = jest.spyOn(instance.props, 'handleSubmit');
```
bkemper: This mock is not necessary.  I found it while removing `__mocks__/account.js`.

src/pages/sendingDomains/components/StatusDescription.js
```diff
@@ -2,23 +2,26 @@ import React, { Fragment } from 'react';
...
+  const { subaccount_id, is_default_bounce_domain } = domain;
 
   return <Fragment>
     <Panel.Section>
       <LabelledValue label={<StatusTooltipHeader />}>
-        { status === 'verified' && <div><VerifiedIcon/> <strong>Verified</strong></div> }
-        { status !== 'verified' && <DomainStatusTag status={status} /> }
-        <div><ReadyFor {...readyFor} bounceDefault={is_default_bounce_domain} subaccount={subaccount_id} /></div>
+        <StatusLabel status={status} />
+        {status === 'verified' && (
```
bkemper: This is a bug fix that I included.

src/pages/sendingDomains/components/DomainStatus.js
```diff
@@ -19,11 +19,17 @@ export const DomainStatus = ({ domain, onShareDomainChange }) => {
         <VerificationHelp status={status} />
       </SendingDomainSection.Left>
       <SendingDomainSection.Right>
+        {hasAutoVerifyEnabled && status === 'verified' && (
```
jonambas: 🤔 Wondering if we should show this banner when pending. Or we also pull this banner into the create page?

src/pages/sendingDomains/components/tests/DomainStatus.test.js
```diff
@@ -1,14 +1,32 @@
...
+      {...props}
+    />
+  );
+
+  it('renders section', () => {
+    expect(subject()).toMatchSnapshot();
+  });
+
+  it('renders section with auto-verify banner', () => {
+    expect(subject({ hasAutoVerifyEnabled: true })).toMatchSnapshot();
```
jonambas: can we target `Banner`

src/pages/sendingDomains/EditPage.js
```diff
@@ -119,6 +124,7 @@ export class EditPage extends Component {
 }
 
 const mapStateToProps = (state) => ({
+  hasAutoVerifyEnabled: hasAutoVerifyEnabledSelector(state),
```
jonambas: Since this is only being passed through to DomainStatus, why not use `<ConditionSwitch>` inside the component?

src/pages/sendingDomains/components/SetupInstructionPanel.js
```diff
@@ -0,0 +1,39 @@
...
+  isAutoVerified = false,
+  isVerified = false,
+  isVerifying = false,
+  onVerify,
+  recordType,
+  verifyButtonIdentifier
+}) => (
+  <Panel
+    actions={[
+      !isAutoVerified && {
```
jonambas: If you move to `ConditionSwitch`, might make sense to pull `connect` down into this component too. Meh.

src/pages/sendingDomains/EditPage.js
bkemper: `hasAutoVerifyEnabledSelector` does a little more than just look at access control state which would make it hard to use either of those components.

src/pages/sendingDomains/components/DomainStatus.js
bkemper: I'll submit a follow-up ticket, but since this feature is used by a small number of accounts, I wanted to limit the amount of logic and banners.

PR #839, opened 2019-02-26
SA-512 Fix dropped filters for events search

kchu1337:
### What Changed
 - Added `events` and `recipients` to whitelist of filters. 
 - Added `events` and `recipients` as constants and change usage in other places to use the imported constants. 

### How To Test
 
1. Go to the Events Search page. 

1. Use recipients and event types as filters. 

1. Click the share button to grab URL.

1. Enter the URL in a new page/tab. The resulting page should correctly show all the filters.


src/constants/index.js
```diff
@@ -103,6 +103,12 @@ export const EVENTS_SEARCH_FILTERS = {
   ab_tests: { placeholder: 'list of full words in ab test ids (e.g. newsletter, fire-sale)', label: 'AB Test IDs' },
   ab_test_versions: { placeholder: 'list of version numbers; requires AB Test ID filter(e.g. 4)', label: 'AB Test Versions' }
 };
+
+export const ADDITIONAL_EVENTS_FILTERS = {
```
ewandennis: Should we use `ADDITIONAL_EVENTS_FILTERS` to initialise our redux state too? I think we'd use it [here](https://github.com/SparkPost/2web2ui/blame/master/src%2Freducers%2FmessageEvents.js#L18-L19) like this:

`...getEmptyFilters(ADDITIONAL_EVENTS_FILTERS)`

src/pages/reports/messageEvents/components/ActiveFilters.js
```diff
@@ -23,7 +22,7 @@ export class ActiveFilters extends Component {
     const activeFilters = _.flatMap(nonEmptyFilterTypes,({ value, label, itemToString }, typeIndex) =>
```
ewandennis: Looks like we're not using `itemToString` and the `filterTypes` entries don't include it so lets not bother extracting it.

src/helpers/tests/messageEvents.test.js
```diff
@@ -48,7 +51,7 @@ describe('messageEvents helpers', () => {
   });
 
   const testCases = {
-    'parses correctly when all param exists': { searchText: '?from=2018-03-23T03:02:32Z&range=hour&to=2018-03-23T04:02:32Z&filter1=foo&filter2=bar' },
+    'parses correctly when all param exists': { searchText: '?from=2018-03-23T03:02:32Z&range=hour&to=2018-03-23T04:02:32Z&filter1=foo&filter2=bar&additionalFilter=loren' },
```
ewandennis: Lets include separate tests for each class of event filter: `EVENTS_SEARCH_FILTERS` and `ADDITIONAL_EVENTS_FILTERS`.

src/constants/index.js
ewandennis: Maybe we could also take the opportunity to name these classes of filters to match their usage? How about

`ADDITIONAL_EVENTS_FILTERS` -> `EVENTS_FILTERS` - since these are the easiest and most obvious to use
`EVENTS_SEARCH_FILTERS` -> `ADVANCED_EVENTS_FILTERS` - since these are only available with the query builder

src/pages/reports/messageEvents/components/MessageEventsSearch.js
```diff
@@ -83,8 +84,8 @@ export class MessageEventsSearch extends Component {
             <Grid.Column xs={12} md={5}>
               <TextField
                 labelHidden
-                label="Recipient Email(s)"
-                placeholder="Filter by recipient email address"
+                label={recipients.label}
```
ewandennis: Since we're hiding the label with `labelHidden`, lets not bother setting the `label` prop.

PR #844, opened 2019-02-28
AC-600 Onboarding copy changes to position as test account

jmoyse:
Small PR to change the naming used for the free account teir in the ChoosePlan and PlanPrice components. The subtext for free accounts when selecting a plan has been removed. The 
`50,000 emails/month for free` has been replaced with `Test Account`.

### DIFF
--- a/src/components/billing/PlanPrice.js
+++ b/src/components/billing/PlanPrice.js
@@ -16,8 +16,8 @@ const PlanPrice = ({ plan, showOverage = false, showIp = false, showCsm = false,
     ? 'First dedicated IP address is free'
     : null;
 
-  const currentFreePlan = plan.code && plan.code.includes('free500-1018');
-  const planTitle = currentFreePlan ? 'Test Account' : `${plan.volume.toLocaleString()} emails`;
+  const currentFreePlan = plan.code && ['free500-1018', 'free15K-1018'].includes(plan.code);
+  const planTitle = currentFreePlan ? 'Test Account' : `${plan.volume.toLocaleString()}`;
 
   const displayCsm = showCsm && plan.includesCsm;
 
@@ -34,7 +34,7 @@ const PlanPrice = ({ plan, showOverage = false, showIp = false, showCsm = false,
   return (
     <span className='notranslate'>
       <span className={styles.MainLabel} {...rest}>
-        <strong>{planTitle}</strong>
+        <strong>{planTitle}</strong> {!currentFreePlan && 'emails'}
         {!plan.isFree && '/month'}
         {priceInfo.price > 0 && <span> at {discountAmount !== priceInfo.price && (<s className={styles.DiscountedLabel}>${priceInfo.price}</s>)}<strong>${discountAmount.toLocaleString()}</strong>/{priceInfo.intervalShort}</span>}
       </span>
--- a/src/components/billing/PlanPrice.js
+++ b/src/components/billing/PlanPrice.js
@@ -12,9 +12,7 @@ const PlanPrice = ({ plan, showOverage = false, showIp = false, showCsm = false,
 
   const is30DayPlan = plan.code && plan.code.includes('free15K');
 
-  const overage = plan.isFree
-    ? 'Full-featured developer account'
-    : plan.overage ? `$${plan.overage.toFixed(2)}/ thousand extra emails. ` : null;
+  const overage = !plan.isFree && plan.overage ? `$${plan.overage.toFixed(2)}/ thousand extra emails. ` : null;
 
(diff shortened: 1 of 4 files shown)


src/components/billing/PlanPrice.js
```diff
@@ -32,13 +30,15 @@ const PlanPrice = ({ plan, showOverage = false, showIp = false, showCsm = false,
     discountAmount = discountAmount * ((100 - selectedPromo.discount_percentage) / 100);
   }
 
+  const planName = plan.isFree ? 'Test Account' : plan.volume.toLocaleString();
+  const planVolume = is30DayPlan ? ' for 30 days' : '/month';
```
James-Burba: calling this planVolume seems confusing, since the planName is the actual volume and this isn't. Also `for 30 days` is never displayed so we don't need to keep it.

src/components/billing/PlanPrice.js
```diff
@@ -32,13 +30,15 @@ const PlanPrice = ({ plan, showOverage = false, showIp = false, showCsm = false,
...
+
   return (
     <span className='notranslate'>
       <span className={styles.MainLabel} {...rest}>
-        <strong>{plan.volume.toLocaleString()}</strong> emails
-        {is30DayPlan ? ' for 30 days' : '/month'}
+        <strong>{planName}</strong> {!plan.isFree && ' emails'}
```
James-Burba: I think this can be simplified. we only have 2 cases - the free plan and a paid plan. So in the case of the paid plan we could just say
```
<strong>{plan.volume.toLocaleString()}</strong> emails/month
```
and for a free plan we could put
```
<strong>Test Account</strong>
```

James-Burba: At the very least `{!plan.isFree && ' emails'}{!plan.isFree && planVolume}` can just be `{!plan.isFree && ' emails/month'}`

src/components/billing/PlanPrice.js
```diff
@@ -10,11 +10,7 @@ const PlanPrice = ({ plan, showOverage = false, showIp = false, showCsm = false,
 
   const priceInfo = getPlanPrice(plan);
 
-  const is30DayPlan = plan.code && plan.code.includes('free15K');
-
-  const overage = plan.isFree
-    ? 'Full-featured developer account'
-    : plan.overage ? `$${plan.overage.toFixed(2)}/ thousand extra emails. ` : null;
+  const overage = !plan.isFree && plan.overage ? `$${plan.overage.toFixed(2)}/ thousand extra emails. ` : null;
```
PatrickSison: Do we want to remove the secondary text entirely? It looks fine on the `/onboarding/plan` because of the description underneath, but if you're on a current free plan and you are on `/account/billing/plan`, it looks a bit strange with no other information besides 'Test Account'. It could also be a later change if needed

jmoyse: Agreed. I think it has a lot of extra white space and looks a bit out of a place with everything else. I sent screenshots out and mentioned the white-space and missing text. This is what I was told to go with 🤷‍♀️. There might be additional changes coming down the pipeline, so who knows?

src/components/billing/PlanPrice.js
```diff
@@ -3,23 +3,22 @@ import _ from 'lodash';
...
-  const overage = plan.isFree
-    ? 'Full-featured developer account'
-    : plan.overage ? `$${plan.overage.toFixed(2)}/ thousand extra emails. ` : null;
+  const overage = plan.isFree ? 'Full-featured developer account' : plan.overage ? `$${plan.overage.toFixed(2)}/ thousand extra emails. ` : null;
 
   const ip = plan.includesIp
     ? 'First dedicated IP address is free'
     : null;
 
+  const currentFreePlan = plan.code && plan.code.includes('free500-1018');
```
PatrickSison: Should this include the 15K plan until it's phased out?

src/components/billing/PlanPrice.js
```diff
@@ -35,10 +34,9 @@ const PlanPrice = ({ plan, showOverage = false, showIp = false, showCsm = false,
   return (
     <span className='notranslate'>
       <span className={styles.MainLabel} {...rest}>
-        <strong>{plan.volume.toLocaleString()}</strong> emails
-        {is30DayPlan ? ' for 30 days' : '/month'}
+        <strong>{planTitle}</strong> {!currentFreePlan && 'emails'}
+        {!plan.isFree && '/month'}
```
PatrickSison: Should the logic be consolidated and just check? If old free plans can show volume, I think it should be able to see `<volume> emails/month`, then on test account, it just omits `emails/month`

PR #845, opened 2019-02-28
EO-588: Edit Alerts

21042jim:
EO-588: Alerts Edit Page

### What Changed
 - Added page to Edit Alerts
    > Loads alerts get action to populate AlertForm
    > Performs alerts update action on form submit
    > Form exists within an Edit Page Container

### How To Test
 - Standard method


src/pages/alerts-new/EditPage.js
```diff
@@ -0,0 +1,116 @@
...
+import { Link } from 'react-router-dom';
+import { Page } from '@sparkpost/matchbox';
+import withAlertsEdit from './containers/EditPage.container';
+import AlertForm from './components/AlertForm';
+import { Loading, DeleteModal, ApiErrorBanner } from 'src/components';
+
+class EditPage extends Component {
+  state = {
+    showDeleteModal: false,
+    alertToDelete: {}
```
jonambas: Can remove this since we aren't using this object

src/reducers/alerts.js
```diff
@@ -20,6 +21,17 @@ export default (state = initialState, { type, payload, meta }) => {
...
+      /* GET */
+
+    case 'GET_ALERT_PENDING':
+      return { ...state, getPending: true, getError: null };
+
+    case 'GET_ALERT_FAIL':
+      return { ...state, getPending: false, getError: payload };
+
+    case 'GET_ALERT_SUCCESS':
+      return { ...state, get: payload, getPending: false };
```
jonambas: Might be better to call this `alert: payload`, it might be confused with the get action.

You'll need to change all references to this though, `state.alerts.get` -> `state.alerts.alert`

src/pages/alerts-new/EditPage.js
```diff
@@ -0,0 +1,116 @@
...
+      alert_metric,
+      assignTo,
+      alert_subaccount,
+      email_addresses,
+      facet_name,
+      facet_value,
+      criteria_comparator,
+      criteria_value,
+      enabled
+    } = values;
```
jonambas: Are you only destructuring these values out to  restructure them into the APIs format?

You can use `.` in redux form field names to nest them ie:

```js
<Field name='threshold.error.comparator' />
// evaluates to values.threshold.error.comparator
```

I think if we do that, we'd only need to pick out `id` right?

src/pages/alerts-new/EditPage.js
```diff
@@ -0,0 +1,116 @@
...
+        alert_metric,
+        assignTo,
+        alert_subaccount,
+        email_addresses,
+        facet_name,
+        facet_value,
+        threshold: {
+          error: {
+            comparator: criteria_comparator,
+            target: Number(criteria_value)
```
jonambas: You can use the `parse` prop on a `<Field />` to cast its value to a number, ie

(its either `parse` or `normalize`, don't remember)

```js
<Field parse={Number} />
```

src/pages/alerts-new/EditPage.js
```diff
@@ -0,0 +1,116 @@
...
+          }
+        },
+        enabled
+      }
+    });
+
+    showAlert({ type: 'success', message: 'Update Successful' });
+    getAlert({ id: this.props.match.params.id });
```
jonambas: in your container, you can avoid this deep reference like:
```js
// map state to props
{
  id: props.match.params.id,
  ...
}
```

and it'll be available through `this.props.id`

src/pages/alerts-new/EditPage.js
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  renderForm() {
+    return (
+      <AlertForm alert={this.props.alert} submitText = 'Update Alert' formName = 'EditAlertForm' newAlert = {false}
+        toggleDelete={this.toggleDelete} onSubmit = {this.update}/>
```
jonambas: Dont think you need to pass through `toggleDelete` anymore

src/pages/alerts-new/components/AlertForm.js
```diff
@@ -0,0 +1,198 @@
...
+    );
+  }
+}
+const mapStateToProps = (state, props) => {
+  const alertValues = formatEditValues(state, state.alerts.get);
+  const selector = formValueSelector(formName);
+
+  return {
+    disabled: props.pristine || props.submitting,
+    submitText: props.submitting ? 'Submitting...' : (props.newAlert ? 'Create Alert' : 'Update Alert'),
```
jonambas: Let's put this `submitText` somewhere in render instead of mapStateToProps

src/pages/alerts-new/components/AlertForm.js
```diff
@@ -0,0 +1,198 @@
...
+    alert_metric: selector(state, 'alert_metric'),
+    facet_name: selector(state, 'facet_name'),
+    assignTo: selector(state, 'assignTo'),
+    alert_subaccount: selector(state, 'alert_subaccount'),
+    enabled: selector(state, 'enabled'),
+    email_addresses: selector(state, 'email_addresses'),
+    initialValues: alertValues
```
jonambas: I think I would turn `formatEditValues` into a selector instead of a helper. You could do the same for the redux form selectors too, and put everything into a `selectAlertsForm` selector

src/pages/alerts-new/EditPage.js
jonambas: Also since you access `get` from connect inside the form, you don't need to pass through `alert`

jonambas: Can remove `submitText` and `formName` as well since that is handled inside the form

jest.config.js
```diff
@@ -1,68 +0,0 @@
-module.exports = {
```
jonambas: what happened here?

package-lock.json
```diff
@@ -5094,7 +5094,7 @@
     },
     "ansi-escapes": {
       "version": "3.1.0",
-      "resolved": "https://registry.npmjs.org/ansi-escapes/-/ansi-escapes-3.1.0.tgz",
+      "resolved": "http://registry.npmjs.org/ansi-escapes/-/ansi-escapes-3.1.0.tgz",
```
jonambas: There are some comments in this thread that might help resolve this:
https://npm.community/t/npm-install-downgrading-resolved-packages-from-https-to-http-registry-in-package-lock-json/1818/7

src/pages/alerts-new/components/AlertForm.js
jonambas: I think you might need to check `newAlert` so that the form doesn't populate on the create page.
`initialValues: props.newAlert ? alertValues : null`

src/pages/alerts-new/components/AlertForm.js
```diff
@@ -0,0 +1,222 @@
...
+                        component={SelectWrapper}
+                        options={getOptions(FACETS)}
+                        disabled={submitting}
+                        validate={required}
+                      />
+                    }
+                    component={TextFieldWrapper}
+                    onChange={removeFacetValue()}
+                    disabled={submitting || checkFacet()}
+                    placeholder={facet_name === 'ALL' ? 'No facet selected' : ''}
```
jonambas: on initial load, this placeholder isn't showing up and isn't disabled if the facet name is 'ALL'

jonambas: I fixed this locally by adding a default to `facet_name` in the selector. It was coming back as `undefined` until the field was touched

```js
export function formatEditValues(state, { ...  facet_name = 'ALL', ... }) 
```

src/pages/alerts-new/components/AlertForm.js
```diff
@@ -0,0 +1,222 @@
...
+                    validate={validateFacet()}
+                  />
+                </div>
+              </Grid.Column>
+            </Grid>}
+            {isSignals && <br/>}
+            <label>Criteria</label>
+            <Grid>
+              <Grid.Column xs={5} md={3}>
```
jonambas: Can we make this match the same grid as above `<Grid.Column xs={6} md={4}>`

src/pages/alerts-new/components/tests/AlertForm.test.js
```diff
@@ -0,0 +1,178 @@
...
+      expect(wrapper.find({ name: 'facet_value' })).not.toExist();
+      expect(wrapper).toMatchSnapshot();
+      wrapper.setProps({ alert_metric: 'signals_health_threshold' });
+      expect(wrapper.find({ name: 'facet_value' }).props().connectLeft.props.name).toEqual('facet_name');
+      expect(wrapper).toMatchSnapshot();
+    });
+
+    it('should clear facet_value and validation when facet_name is set to ALL and signals threshold', () => {
+      wrapper.setProps({ alert_metric: 'signals_health_threshold', facet_name: 'ALL', facet_value: 'something' });
+      expect(wrapper.find('span')).toMatchSnapshot();
```
jonambas: I don't see any spans on the form, what were you trying to capture with these?

src/pages/alerts-new/components/tests/AlertForm.test.js
```diff
@@ -0,0 +1,178 @@
...
+  });
+
+  describe('account/subaccount props', () => {
+    it('should show subaccount typeahead when assignTo is set to subaccount', () => {
+      wrapper.setProps({ assignTo: 'subaccount' });
+      expect(wrapper.find({ name: 'subaccount' })).toExist();
+      expect(wrapper).toMatchSnapshot();
+      wrapper.setProps({ assignTo: 'master' });
+      expect(wrapper.find({ name: 'subaccount' })).not.toExist();
+      expect(wrapper).toMatchSnapshot();
```
jonambas: We don't need to snapshot every single case if the assertions you make above these lines cover what's expected to change

src/pages/alerts-new/constants/facets.js
```diff
@@ -0,0 +1,6 @@
+export const FACETS = {
+  ALL: 'NONE',
```
jonambas: Maybe `No Facet` instead of `NONE`

src/selectors/alerts.js
```diff
@@ -0,0 +1,11 @@
+import { selectSubaccountFromId } from './subaccounts';
+import getAssignTo from 'src/pages/alerts-new/helpers/getAssignTo';
+import _ from 'lodash';
+
+export function formatEditValues(state, { id, name, alert_metric, alert_subaccount, email_addresses, facet_name, facet_value, threshold, enabled }) {
```
jonambas: I showed you this but, `alert_subaccount` isn't being returned for me, but `subaccount_id` is.

src/pages/alerts-new/components/AlertForm.js
```diff
@@ -0,0 +1,222 @@
...
+                        name='threshold.error.comparator'
+                        component={SelectWrapper}
+                        options={getOptions(COMPARATOR)}
+                        disabled={submitting}
+                        validate={required}
+                      />
+                    }
+                    disabled={submitting}
+                    prefix={alert_metric !== 'signals_health_threshold' ? 'Above' : ''}
+                    suffix={alert_metric !== 'signals_health_threshold' ? '%' : ''}
```
jonambas: This is displaying `%` for monthly sending limit`

src/pages/alerts-new/components/AlertForm.js
```diff
@@ -0,0 +1,222 @@
...
+            />
+            <Grid>
+              <Grid.Column xs={1} md={1}>
+                <label>Enabled</label>
+              </Grid.Column>
+              <Grid.Column xs={1} md={1}>
+                <div>
+                  <Field
+                    name='enabled'
+                    checked={enabled}
```
jonambas: Remove `checked={checked}`, you can also remove the selector that pulls in the `enabled` form value.

And add `type='checkbox'`, this should fix that console error. I think you'll be able to remove the `parse` prop too

src/pages/alerts-new/components/tests/AlertForm.test.js
```diff
@@ -0,0 +1,178 @@
...
+  });
+
+  describe('enabled props', () => {
+    it('should only show comparator select when alert_metric is set to signals_health_threshold', () => {
+      wrapper.setProps({ enabled: false });
+      expect(wrapper.find({ name: 'enabled' }).props().checked).toEqual(false);
+      expect(wrapper).toMatchSnapshot();
+      wrapper.setProps({ enabled: true });
+      expect(wrapper.find({ name: 'enabled' }).props().checked).toEqual(true);
+      expect(wrapper).toMatchSnapshot();
```
jonambas: Too many snapshots 😆

src/pages/alerts-new/components/AlertForm.js
jonambas: This might be because I was looking at an old alert

21042jim: cant hurt

src/pages/alerts-new/EditPage.js
```diff
@@ -15,6 +16,14 @@ export class EditPage extends Component {
     this.props.getAlert({ id: this.props.id });
   }
 
+  componentDidUpdate() {
+    const { error } = this.props;
+    if (error) {
+      this.props.history.push('/alerts-new');
+      this.props.showAlert({ type: 'error', message: `Alert ${this.props.id} does not exist` });
```
jonambas: This might not always be a 404, I'd say something generic like 'Unable to load alert'

src/pages/alerts-new/EditPage.js
```diff
@@ -1,15 +1,100 @@
...
+  }
+
+  renderForm() {
     return (
-      <Page title='Create an Alert' breadcrumbAction={{ content: 'Back to Alerts', to: '/alerts', component: Link }} >
+      <AlertForm newAlert = {false} onSubmit = {this.handleUpdate}/>
+    );
+  }
 
+  renderError() {
```
jonambas: Since we're redirecting to the list page on error, we can remove this all together

PR #846, opened 2019-02-28
Tenant config full snapshot

rajumsys:
In FE-621, we've simplified tenant config but given unit tests stubbed, we don't have a full picture of generated configs. So in this PR, I've added another test module to take a snapshot of a full tenant config. This will enable us to detect any unwanted changes to final configs should we make any changes in config generation script. 

### What Changed
 - Includes snapshot of full config

### How To Test
 - Run tests
 - Manually inspect


rajumsys:
Once #850 is merged, I'll move this test to func test directory.

PR #848, opened 2019-03-01
AC-590: Changed billingInfo to utilize billing endpoint 

PatrickSison:
### What Changed
 - Removed usage of GET on /account?includes=billing
- Added usage of GET on /account/billing
- Updated tests
- Added mock request for /account/billing

### How To Test
 - Open /account/billing page
- Verify that /account request does not contain query for 'includes=billing'
- Verify that a billing request is made

### DIFF
--- a/src/actions/account.js
+++ b/src/actions/account.js
@@ -23,6 +23,17 @@ export function getPlans({ meta = {}} = {}) {
   });
 }
 
+export function getBillingInfo({ meta = {}} = {}) {
+  return sparkpostApiRequest({
+    type: 'GET_BILLING',
+    meta: {
+      method: 'GET',
+      url: '/v1/account/billing',
+      ...meta
+    }
+  });
+}
+
 export function update(data) {
   return sparkpostApiRequest({
     type: 'UPDATE_ACCOUNT',
--- a/src/pages/billing/SummaryPage.js
+++ b/src/pages/billing/SummaryPage.js
@@ -1,7 +1,7 @@
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
 import { Page } from '@sparkpost/matchbox';
-import { fetch as fetchAccount, getPlans } from 'src/actions/account';
+import { fetch as fetchAccount, getPlans, getBillingInfo } from 'src/actions/account';
 import { list as getSendingIps } from 'src/actions/sendingIps';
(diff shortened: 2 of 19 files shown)


src/actions/account.js
```diff
@@ -23,6 +23,17 @@ export function getPlans({ meta = {}} = {}) {
   });
 }
 
+export function getBillingInfo({ meta = {}} = {}) {
```
bkemper: Is `meta` needed?  I can't find a place that it is being used.

src/pages/billing/SummaryPage.js
```diff
@@ -51,4 +52,4 @@ const mapStateToProps = (state) => ({
   invoices: state.invoices.list
 });
 
-export default connect(mapStateToProps, { getInvoices, getSendingIps, getPlans, fetchAccount })(BillingSummaryPage);
```
bkemper: Should `loading` consider `state.account.billingLoading`?

src/reducers/account.js
```diff
@@ -48,6 +48,15 @@ export default (state = initialState, { type, meta, payload }) => {
...
       return { ...state, createError: payload, createLoading: false };
 
+    case 'GET_BILLING_SUCCESS':
+      return { ...state, billing: payload, billingLoading: false };
+
+    case 'GET_BILLING_FAIL':
+      return { ...state, billingError: payload, billingLoading: false };
+
+    case 'GET_BILLING_PENDING':
+      return { ...state, loading: true };
```
bkemper: This needs to be changed from `loading` to `billingLoading`.

src/reducers/account.js
```diff
@@ -48,6 +48,15 @@ export default (state = initialState, { type, meta, payload }) => {
     case 'CREATE_ACCOUNT_FAIL':
       return { ...state, createError: payload, createLoading: false };
 
+    case 'GET_BILLING_SUCCESS':
+      return { ...state, billing: payload, billingLoading: false };
```
bkemper: The account endpoint with `?include=billing` would only include a `billing` object for paid accounts.  There are a number of places that we check if `account.billing` is falsy.  The billing endpoint will return an empty object in this case which will break those falsy checks.  I recommend setting `billing` when empty to `null` to maintain the current behavior.

PR #854, opened 2019-03-05
AC-629: Hides SSO checkbox if SSO is not configured

PatrickSison:
### What Changed
 - Hid SSO checkbox on user Edit form if SSO is not configured

### How To Test
 - Check account with SSO enabled to see if checkbox is visible and enabled (appteam works)
 - Check account without SSO configured to see if checkbox is hidden

### To Do
- []


src/pages/users/components/EditForm.js
```diff
@@ -42,16 +39,18 @@ export const EditForm = ({
...
-        <Field
-          component={CheckboxWrapper}
-          disabled={!isAccountSingleSignOnEnabled}
-          helpText={ssoHelpText}
-          label="Enable single sign-on authentication for this user"
-          name="is_sso"
-          type="checkbox"
-        />
-      </Panel.Section>
+      {isAccountSingleSignOnEnabled &&
```
ewandennis: This pops in after the page loads because the page doesn't take single sign-on state into account. We can improve that like this:
 - In `pages/users/EditPage.container.js`:
    - Include `state.accountSingleSignOn.loading` as part of the component's `loading` prop [here](https://github.com/SparkPost/2web2ui/blame/master/src%2Fpages%2Fusers%2FEditPage.container.js#L19).
 
Lets pre-empt a bug too:
 - In the `pages/users/EditPage.container.js`, include `state.accountSingleSignOn.error` in `loadingError` so we can report SSO state loading errors

src/pages/users/components/tests/EditForm.test.js
```diff
@@ -18,7 +18,7 @@ describe('Component: EditForm', () => {
     expect(subject()).toMatchSnapshot();
   });
 
-  it('should disable single sign-on checkbox and display instructions', () => {
+  it('should hide single sign-on checkbox if single sign-on is not configured', () => {
     expect(subject({ isAccountSingleSignOnEnabled: false })).toMatchSnapshot();
```
ewandennis: As discussed, lets make this an explicit assertion.

PR #855, opened 2019-03-05
AC-591 : Proprietary user roles

ayoung1:
### What Changed
 - Two new user roles added -- Developer and Email
     > Gated behind the UI option `developer_and_email_roles`

### How To Test
 - Verify that the new roles are not visible while on an account without the new UI option
     > appteam has this option

### To Do
- [ ] Verify roles are visible when flag is enabled
- [ ] Confirm new descriptions are correct


src/pages/users/components/tests/RoleRadioGroup.test.js
```diff
@@ -32,6 +33,18 @@ describe('Component: RoleRadioGroup', () => {
     expect(options(subject({ disabled: true }))).toEqual(expect.arrayContaining([expect.objectContaining({ disabled: true })]));
   });
 
+  cases('it should render developer roles', ({ showDeveloperRoles }) => {
+    expect(subject({ showDeveloperRoles })
```
PatrickSison: This test is needed, but could be changed to check that the proper roles were added instead of just checking if the flag to control them is set as a prop. You could do something like this:
```
    expect(subject({ showDeveloperRoles })
      .find('RadioGroup[title="Role"]')
      .prop('options'))
      .toEqual(
        expect.arrayContaining([
          expect.objectContaining({ value: ROLES.DEVELOPER }),
          expect.objectContaining({ value: ROLES.EMAIL })
        ]))
```
This would check the array of roles to see if those specific roles were added because of the flag change. You may need to break it into 2 separate tests though because it's harder to check negative case using Jest-in-case with these assertions.

src/pages/users/components/RoleRadioGroup.js
```diff
@@ -87,7 +105,8 @@ RoleRadioGroup.defaultProps = {
 const mapStateToProps = (state) => ({
   selectedRole: formValueSelector(FORMS.INVITE_USER)(state, 'access'),
   hasSubaccounts: hasSubaccounts(state),
-  useSubaccountChecked: formValueSelector(FORMS.INVITE_USER)(state, 'useSubaccount')
+  useSubaccountChecked: formValueSelector(FORMS.INVITE_USER)(state, 'useSubaccount'),
+  showDeveloperRoles: isAccountUiOptionSet('developer_and_email_roles', false)(state)
```
jonambas: I believe this will need to be wrapped in `selectCondition`

https://github.com/SparkPost/2web2ui/blob/master/docs/access-control.md#condition-helpers

src/pages/users/components/tests/RoleRadioGroup.test.js
```diff
@@ -32,6 +33,25 @@ describe('Component: RoleRadioGroup', () => {
...
+          expect.objectContaining({ value: ROLES.DEVELOPER }),
+          expect.objectContaining({ value: ROLES.EMAIL })
+        ])
+      );
+  });
+
+  it('should not render developer roles when the UI flag `developer_and_email_roles` is disabled', () => {
+    expect(subject({ showDeveloperRoles: false })
+      .find('RadioGroup[title="Role"]')
+      .prop('showDeveloperRoles'))
```
PatrickSison: You can use the same test as the above one, but prepend `.toEqual` with `not` so it's `.not.toEqual`, which will just make sure that it doesn't contain both those roles

PR #860, opened 2019-03-07
EO-614 Allow signals tooltips to overflow beyond chart containers

jonambas:
### What Changed
- Disables recharts positioning by setting `position={{ x: 0, y: 0 }}` on `<Tooltip />`
- Custom tooltip wrapper now calculates its own position

### How To Test
- Point your openresty upstreams to staging, or use `.fake` signals actions
- Visit all signals pages with charts
- Verify there are no z-index issues
- Verify tooltips switch to the left of curser if tooltip overflows past browser's right viewport edge (happens on diverging bar chart in health score)

### To Do
- [x] Address feedback
- [x] Maybe redo with hooks because code would look a lot cleaner

### DIFF
--- a/src/helpers/geometry.js
+++ b/src/helpers/geometry.js
@@ -1,0 +1,3 @@
+export function getBoundingClientRect(node) {
+  return node.current.getBoundingClientRect();
+}
--- a/src/pages/signals/HealthScorePage.js
+++ b/src/pages/signals/HealthScorePage.js
@@ -88,7 +88,7 @@ export class HealthScorePage extends Component {
               <Fragment>
                 <BarChart
                   gap={gap}
-                  onClick={this.handleDateSelect}
+                  onClick={handleDateSelect}
                   selected={selectedDate}
                   timeSeries={data}
                   tooltipContent={({ payload = {}}) => (
@@ -104,7 +104,7 @@ export class HealthScorePage extends Component {
                 <BarChart
                   gap={gap}
                   height={190}
-                  onClick={this.handleDateSelect}
+                  onClick={handleDateSelect}
                   selected={selectedDate}
                   timeSeries={data}
                   tooltipContent={({ payload = {}}) => (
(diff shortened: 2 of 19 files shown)


src/pages/signals/components/charts/tooltip/tests/Tooltip.test.js
```diff
@@ -16,7 +17,7 @@ describe('Signals Tooltip Component', () => {
         }
       }]
     };
-    wrapper = shallow(<Tooltip {...props}/>);
+    wrapper = mount(<Tooltip {...props}/>);
```
jonambas: `createRef` does not work on shallow mounted components because `this` loses its reference

src/pages/signals/HealthScorePage.js
```diff
@@ -104,58 +104,60 @@ export class HealthScorePage extends Component {
...
             {panelContent || (
-              <Fragment>
-                <BarChart
-                  gap={gap}
-                  onClick={this.handleDateSelect}
-                  selected={selectedDate}
-                  timeSeries={data}
-                  tooltipContent={({ payload = {}}) => (
-                    <TooltipMetric label='Health Score' value={`${roundToPlaces(payload.health_score * 100, 1)}`} />
+              <div className='LiftTooltip'>
```
jonambas: Move this into BarChart instead of the page

src/pages/signals/components/charts/tooltip/Tooltip.js
```diff
@@ -1,16 +1,43 @@
...
   const content = _.get(props, 'payload[0]', {});
   const date = _.get(content, 'payload.date');
+  const wrapper = useRef(null);
+
+  const [positionedOnRight, setPositionedOnRight] = useState(true);
+  const [position, setPosition] = useState({ left: 0, top: 0 });
+
+  // Calculates Tooltip position
+  useLayoutEffect(() => {
+    const rect = getBoundingClientRect(wrapper);
```
jonambas: I couldn't really find a way to mock the ref, so just created an external helper.

src/helpers/geometry.js
```diff
@@ -0,0 +1,3 @@
+export function getBoundingClientRect(node) {
+  return node.current.getBoundingClientRect();
```
jonambas: I should maybe add in a default return here in case !node.current.getBoundingClientRect

PR #862, opened 2019-03-08
TR-1118: Some recipient list upload errors handled incorrectly

bkemper:
### What Changed
When creating or updating a recipient list, fixed the handling if Papaparse is unable able to read the file provided.  Specifically, `RecipientListForm` expects an array of errors from `parseRecipientListCsv`, but `error` callback was rejecting with an `Error`.  Changed it to return an array with a static message.

screen shot 2019-03-07 at 11 04 17 pm

### How To Test

1. Navigate to /lists/recipient-lists/create
2. Download the example recipient list
3. Fill out the form, provide a label, identifier, and select example recipient list from your downloads directory
4. Rename the example recipient list file in your downloads directory
5. Click "Create Recipient List" button

You should expect to get a "Unable to read your file" error.

### DIFF
--- a/src/pages/recipientLists/helpers/csv.js
+++ b/src/pages/recipientLists/helpers/csv.js
@@ -62,8 +62,8 @@ const parseRecipientListCsv = (file) => new Promise((resolve, reject) => {
     header: true,
     skipEmptyLines: true,
     complete: (result) => parseRawRecords(result, resolve, reject),
-    error: (err) => reject(err)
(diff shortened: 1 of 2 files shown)


ewandennis:
You can force it by choosing a file on the upload form, renaming the CSV file, the hitting submit.

Could we please have a test to cover the Papaparse error case? Since we're there, we could also eradicate a few more snapshot tests if you have headspace.

bkemper:
Rebased.

PR #863, opened 2019-03-08
Update bowser to the latest version 🚀

kchu1337:
## What Changed

To make testing easier, added extra data with each error reported to Sentry.  Bowser was upgraded -- here is the [diff](https://github.com/lancedikson/bowser/compare/943adfb..28bff84) and change log (primarily fixes and documentation).

```
### 2.1.2 (March 6, 2019)
- [FIX] Fix buggy `getFirstMatch` reference

### 2.1.1 (March 6, 2019)
- [ADD] Add detection of PlayStation 4 (#291)
- [ADD] Deploy docs on GH Pages (#293)
- [FIX] Fix files extensions for importing (#294)
- [FIX] Fix docs (#295)
```

## How To Test?

Bowser is only used in our `errorTracker`.  It is used to parse the UserAgent, check if browser is on list of supported browsers, and if not supported, may mark errors reported to Sentry as warnings.  To test:

- Login
- Visit http://app.sparkpost.test/webhooks/details/0
- Find error in Sentry and confirm error has correct `isSupportedBrowser` value

Then repeat for with an unsupported browser using Sauce Labs.

### DIFF
--- a/src/helpers/errorTracker.js
+++ b/src/helpers/errorTracker.js
@@ -108,6 +108,12 @@ export function getEnricherOrDieTryin(store, currentWindow) {
 
     return {
       ...data,
+      extra: { // extra data that doesn't need to be searchable
+        ...data.extra,
+        isApiError: apiError,
+        isChunkFailure: chunkFailure,
+        isSupportedBrowser
+      },
       request,
       level: !fromOurBundle || apiError || chunkFailure || !isSupportedBrowser ? 'warning' : 'error',
       tags: { // all tags can be easily searched and sent in Slack notifications
--- a/src/helpers/tests/__snapshots__/errorTracker.test.js.snap
+++ b/src/helpers/tests/__snapshots__/errorTracker.test.js.snap
@@ -2,6 +2,11 @@
 
 exports[`.getEnricherOrDieTryin by default 1`] = `
 Object {
+  "extra": Object {
+    "isApiError": false,
+    "isChunkFailure": false,
+    "isSupportedBrowser": true,
+  },
   "level": "warning",
   "logger": "test",
   "request": undefined,
@@ -19,6 +24,11 @@ exports[`.getEnricherOrDieTryin with api error 1`] = `
 Object {
   "exception": Object {
     "type": "SparkpostApiError",
+  },
+  "extra": Object {
+    "isApiError": false,
+    "isChunkFailure": false,
+    "isSupportedBrowser": true,
   },
   "level": "warning",
   "logger": "test",
@@ -42,6 +52,11 @@ Object {
       },
     ],
   },
+  "extra": Object {
+    "isApiError": false,
+    "isChunkFailure": true,
+    "isSupportedBrowser": true,
+  },
   "level": "warning",
   "logger": "test",
   "request": undefined,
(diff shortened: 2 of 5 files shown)


src/helpers/errorTracker.js
```diff
@@ -108,6 +108,12 @@ export function getEnricherOrDieTryin(store, currentWindow) {
 
     return {
       ...data,
+      extra: { // extra data that doesn't need to be searchable
+        ...data.extra,
+        isApiError: apiError,
```
ewandennis: Luckily ran into an API error on tst. 🆗

src/helpers/errorTracker.js
```diff
@@ -108,6 +108,12 @@ export function getEnricherOrDieTryin(store, currentWindow) {
 
     return {
       ...data,
+      extra: { // extra data that doesn't need to be searchable
+        ...data.extra,
```
ewandennis: What do we begin capturing with this? Are we going to find potentially unwanted new fields in our error reports or is this a useful-in-future type thing?

src/helpers/errorTracker.js
```diff
@@ -108,6 +108,12 @@ export function getEnricherOrDieTryin(store, currentWindow) {
...
+        isApiError: apiError,
+        isChunkFailure: chunkFailure,
```
ewandennis: Forced a chunk loading error by serving a local build with a missing chunk. 🆗

src/helpers/errorTracker.js
```diff
@@ -108,6 +108,12 @@ export function getEnricherOrDieTryin(store, currentWindow) {
...
+        isSupportedBrowser
```
ewandennis: We'll have to verify this on staging.

src/helpers/tests/__snapshots__/errorTracker.test.js.snap
```diff
@@ -20,6 +25,11 @@ Object {
   "exception": Object {
     "type": "SparkpostApiError",
   },
+  "extra": Object {
+    "isApiError": false,
```
ewandennis: Oink?

We put an API error in but the snapshot shows `isApiError: false`.

src/helpers/errorTracker.js
bkemper: > What do we begin capturing with this?

I can't think of another use case at this moment.  We should be careful to not abuse this.  This was the easiest way for me to test the error level logic.

> Are we going to find potentially unwanted new fields in our error reports or is this a useful-in-future type thing?

Nope, this is to preserve the current behavior of passing "session:duration" along to Sentry.

src/helpers/tests/__snapshots__/errorTracker.test.js.snap
bkemper: Fixed.  The test data was incorrect.

bkemper: Verified it is now working correctly -- https://sentry.io/organizations/sparkpost/issues/858131365/?project=232588&statsPeriod=14d&utc=false

bkemper:
@ewandennis This PR is ready for final review.

bkemper:
Rebased.

PR #866, opened 2019-03-11
SD-747 Allow IP Editing from a new page

rajumsys:
Currently pool reassignment can be done on ip pools page from the list. As a foundation of more IP related operation, we're moving it to a new page in this ticket/PR. 

### What Changed
 - Individual IP editing happens from their own screen. You can click of IP from ip pools page and the edit page will be opened. 

### How To Test
 - Try re-assigning IP from ip edit page

### Todo

- [x] Tests for IPList component
- [x] Add missing tests (if any) due to refactor

### Notes
- Refactor some old codes to reduce snapshots in favor of inline assertions.


src/pages/ipPools/components/IPForm.js
```diff
@@ -0,0 +1,63 @@
...
+      </Panel>
+    );
+  }
+}
+
+const formName = 'ipForm';
+const mapStateToProps = (state, props) => ({
+  currentPool: selectCurrentPool(state),
+  currentIp: selectIpForCurrentPool(state, props),
+  reAssignPoolsOptions: getReAssignPoolsOptions(state, props),
```
ewandennis: I think we should keep the select option construction logic inside this component. It can form whatever set of options it needs close to the point of use. Selectors are meant as filters on redux state.

src/selectors/ipPools.js
```diff
@@ -78,3 +85,18 @@ export const shouldShowIpPurchaseCTA = createSelector(
 export const selectFirstIpPoolId = createSelector(
   [getIpPools], (ipPools) => _.get(ipPools, '[0].id')
 );
+
+
+export const getReAssignPoolsOptions = createSelector(
```
ewandennis: We use a `select` prefix on selectors by convention.

src/selectors/ipPools.js
```diff
@@ -78,3 +85,18 @@ export const shouldShowIpPurchaseCTA = createSelector(
...
+  [getIpPools, selectCurrentPool], (pools, currentPool) => pools.map((pool) => ({
+    value: pool.id,
+    label: (pool.id === currentPool.id) ? '-- Change Pool --' : `${pool.name} (${pool.id})`
```
ewandennis: Maybe `-- Select a new pool --` ?

src/pages/ipPools/tests/EditIPPage.test.js
```diff
@@ -0,0 +1,69 @@
...
+  });
+
+  it('loads data after mount', () => {
+    expect(props.listPools).toHaveBeenCalledTimes(1);
+    expect(props.getPool).toHaveBeenCalledTimes(1);
+  });
+
+  it('renders loader when loading is true', () => {
+    wrapper.setProps({ loading: true });
+    expect(wrapper.exists('Loading')).toBe(true);
```
ewandennis: You can use the enzyme `toExist ` matcher here:

```js
    expect(wrapper.find('Loading')).toExist();
    expect(wrapper.find('Page')).not.toExist();
```

src/pages/ipPools/tests/EditIPPage.test.js
```diff
@@ -0,0 +1,69 @@
...
+  it('renders errors', () => {
+    const err = new Error('API Failed');
+    wrapper.setProps({ error: err });
+    expect(wrapper.exists('ApiErrorBanner')).toBe(true);
+    expect(wrapper.find('ApiErrorBanner')).toMatchSnapshot();
+  });
+
+  describe('onUpdateIp', () => {
+    it('updates ip with ip pool data', async () => {
+      await wrapper.instance().onUpdateIp({ ip_pool: 'foo' });
```
ewandennis: Instead of using `instance()`, we could call the `onSubmit` handler from `IPForm` directly. That way, we don't depend on the name of a handler and we explicitly verify that the handler we set on the form does the job we want.

```
      await wrapper.find('IPForm').prop('onSubmit')({ ip_pool: 'foo' });
```

You will have to set `displayName` on `IPForm` to make that work. Otherwise, you could use `wrapper.childAt(0)` if required.

src/pages/ipPools/tests/EditIPPage.test.js
```diff
@@ -0,0 +1,69 @@
...
+      expect(props.updateSendingIp).toHaveBeenCalledWith('1.1.1.1', 'foo');
+      expect(props.showAlert).toHaveBeenCalledTimes(1);
+    });
+    it('throws on error', async () => {
+      const err = new Error('API Failed');
+      props.updateSendingIp.mockReturnValue(Promise.reject(err));
+      await expect(wrapper.instance().onUpdateIp({ ip_pool: 'bar' })).rejects.toThrow(err);
```
ewandennis: Similar note here: prefer `wrapper.find` over `wrapper.instance`.

src/pages/ipPools/components/tests/PoolForm.test.js
```diff
@@ -47,48 +47,44 @@ describe('PoolForm tests', () => {
...
   it('should show help text when editing default pool', () => {
     wrapper.setProps({ pool: { id: 'default', name: 'Default' }});
-    expect(wrapper).toMatchSnapshot();
+    expect(wrapper.find('Field').prop('helpText')).toMatch('You cannot change the default IP pool\'s name');
   });
 
   it('should not render signing_domain if editing default pool', () => {
     wrapper.setProps({ pool: { id: 'default', name: 'Default' }});
-    expect(wrapper).toMatchSnapshot();
+    expect(wrapper.exists('Field[name="signing_domain"]')).toBe(false);
```
ewandennis: `expect(wrapper.find(...)).not.toExist()` is useful here

src/pages/ipPools/components/tests/PoolForm.test.js
```diff
@@ -47,48 +47,44 @@ describe('PoolForm tests', () => {
 
   it('should show help text when editing default pool', () => {
     wrapper.setProps({ pool: { id: 'default', name: 'Default' }});
-    expect(wrapper).toMatchSnapshot();
+    expect(wrapper.find('Field').prop('helpText')).toMatch('You cannot change the default IP pool\'s name');
```
ewandennis: This whole file is so much better and the snapshots are ~650 lines shorter \o/.

src/pages/ipPools/EditIPPage.js
```diff
@@ -0,0 +1,95 @@
+import React, { Component } from 'react';
```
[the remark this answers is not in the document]
rajumsys: we have `IPForm` which is actually I started naming. But it's inconsistent even in this scope (pages/ipPools dir)

src/pages/ipPools/EditIPPage.js
```diff
@@ -0,0 +1,95 @@
...
+      return this.renderError();
+    }
+
+    return <IPForm onSubmit={this.onUpdateIp}/>;
+  }
+
+  render() {
+    const { loading, currentPool, currentIp } = this.props;
+
+    if (loading || !currentIp || !currentPool) {
```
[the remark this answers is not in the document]
rajumsys: `loading` is false on initial mount as componentDidMount isn't triggered yet which eventually make it true. So during first render, it'll continue `render` method which will fail where we try to access `currentPool` or `currentIp`. that's why those checks. Though later two aren't really loading state, but as rendering `Loading` component is the only use case here, i guess it doesn't harm to combine them all under `loading` prop.

src/selectors/ipPools.js
rajumsys: Is that only style we are aiming for? I see bunch of selectors with `get` prefix

src/selectors/ipPools.js
rajumsys: the label is unchanged from previous implementation but i'm fine with suggested one.

src/pages/ipPools/components/IPForm.js
rajumsys: I'm not sure if I understand the reasoning. This is filtering data from redux store. Also having it a selector means, it won't recompute unless concerned args are changed.

src/pages/ipPools/components/PoolForm.js
```diff
@@ -2,97 +2,65 @@ import React, { Component, Fragment } from 'react';
...
-      </Fragment>
-    );
-  }
+
+    return (<Panel title='Sending IPs'>
+      <Panel.Section>
+        <p>
+          {!ips.length && <span>No IP in this pool. </span>}
+          Add dedicated IPs to this pool by moving them from their current pool{purchaseCTA}.
+          {!!ips.length && <span> Click on existing Sending IP to modify.</span>}
```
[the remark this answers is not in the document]
rajumsys: at this point, i don't think either makes any difference as we all know what are they. moreover, `!!` is consistent with with `!` ;). anyway, we don't need it anymore.

src/pages/ipPools/tests/EditIPPage.test.js
rajumsys: this is a `onUpdateIp` suite where i'm asserting it's working as expected when invoked; not testing if it's invoked correctly.

rajumsys:
I've addressed most of the feedbacks. However, did some refactoring on Ip pool edit page as it's simpler now and which I missed during first pass. As I found, all the properties already coming from ip list endpoint, we don't need to single pool GET call. So made some changes around that.

I've fixed broken tests but there are missing tests which I'll add tomorrow.

src/pages/ipPools/components/IpForm.js
```diff
@@ -0,0 +1,63 @@
...
+      </Panel>
+    );
+  }
+}
+
+const formName = 'ipForm';
+const mapStateToProps = (state, props) => ({
+  currentPool: selectCurrentPool(state, props),
+  currentIp: selectIpForCurrentPool(state, props),
+  reAssignPoolsOptions: selectReAssignPoolsOptions(state, props),
```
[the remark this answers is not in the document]
ewandennis: If the calculated result might be large, lets use `_.memoize()` in the component.

src/pages/ipPools/EditPage.js
```diff
@@ -103,6 +93,30 @@ export class EditPage extends Component {
...
+      ? <Fragment>, or by <UnstyledLink to="/account/billing" component={Link}>purchasing new
+        IPs</UnstyledLink></Fragment>
+      : null;
+
+    return (<Panel title='Sending IPs'>
+      <Panel.Section>
+        <p>
+          {!ips && <span>There are no IPs in this pool. </span>}
+          Add dedicated IPs to this pool by moving them from their current pool{purchaseCTA}.
+          {ips && <span> Click on existing Sending IP to modify.</span>}
```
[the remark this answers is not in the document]
rajumsys: I think it's fine now as we'll have other editing options coming anyway. if you're suggesting to use your provided text now and then update it with other, that's doable but just given we're going to add that swiftly after this, probably we're good either way. let me know.

src/pages/ipPools/tests/EditIPPage.test.js
rajumsys: I've added a test that verifies the submission of form. I, however, just wanted to verify if submitting the form invokes `this.onUpdateIp` method but spying/stubbing that doesn't work as expected. So, I've verified if `updateSendingIp` was called.

src/pages/ipPools/EditPage.js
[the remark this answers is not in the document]
ewandennis: 👍

PR #869, opened 2019-03-11
EO-619 Create signals v2 dashboard pages and routes

jonambas:
### What Changed
- Creates three new pages, one for each leaderboard
- Nests all new page routes under a new signals nav item

### How To Test
- Verify new routes
  - Log into appteam, or add the `feature_signals_v2` UI option to your account
  - Verify new nav item renders and each page renders its leaderboard
- Verify current experience is not affected
  - Log into an account without the UI option
  - Verify the current signals nav item works, and the new nav item does not render
  - Verify new routes 404

### To Do
- [ ] Address feedback

### DIFF
--- a/src/config/navItems.js
+++ b/src/config/navItems.js
@@ -43,10 +43,32 @@ export default [
     ]
   },
   {
+    // TODO Remove when signals v2 is released
     label: 'Signals',
     to: '/signals',
     icon: Signal,
     tag: 'new'
+  },
+  {
+    // Duplicated, but will not render without feature_signals_v2 UI option
+    label: 'Signals',
+    to: '/signals',
+    icon: Signal,
+    tag: 'new',
+    children: [
+      {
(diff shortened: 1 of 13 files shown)


src/config/navItems.js
```diff
@@ -43,11 +43,33 @@ export default [
...
   },
   {
+    // TODO Remove when signals v2 is released
     label: 'Signals',
     to: '/signals',
     icon: Signal,
     tag: 'new'
   },
+  {
+    // Duplicated, but will not render without feature_signals_v2 UI option
```
ewandennis: I had to satisfy myself that this was true.

[This line](https://github.com/SparkPost/2web2ui/blame/master/src%2Fselectors%2FnavItems.js#L35) in `selectors/navItems.js` removes nav items which have children which are all filtered due to their `condition`.

PR #871, opened 2019-03-12
AC-591 : Snapshot Update

ayoung1:
### What Changed
 - Snapshot updated

### To Do
- [ ] Verify Tests


src/pages/users/components/tests/__snapshots__/RoleRadioGroup.test.js.snap
```diff
@@ -16,6 +16,13 @@ exports[`Component: RoleRadioGroup should render the RadioGroup 1`] = `
   allowSuperUser={false}
   clearFields={[MockFunction]}
   developer_and_email_roles={false}
+  grid={
```
PatrickSison: After the merge with master, it undid the changes you had in RoleRadioGroup.js with the description and removing the Grid. If you remove the grid again and fix the description, you can revert the snapshot

PR #877, opened 2019-03-14
fixed validation bug

21042jim:
EO-634: Fix redux-form validation bug

[Bug 1]

Help text does not render correctly when you do not have valid domains to use
 

[Bug 2]

Investigate a maximum depth exceeded error on input change or assignTo change
Seems to be related to field validation on edit only

### DIFF
--- a/src/pages/alerts-new/components/AlertForm.js
+++ b/src/pages/alerts-new/components/AlertForm.js
@@ -45,6 +45,27 @@ export class AlertForm extends Component {
     }
   }
 
+  validateFacet = (value, allValues) => {
+    const { facet_name } = allValues;
+    const { domains, ipPools } = this.props;
+
+    if (facet_name === 'ALL') {
+      return undefined;
+    }
+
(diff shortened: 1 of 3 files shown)


src/pages/alerts-new/components/AlertForm.js
```diff
@@ -45,14 +45,34 @@ export class AlertForm extends Component {
...
+      const domainsArray = domains.map(({ domain }) => domain) || [];
+      return domainsArray.includes(value) ? undefined : 'This is not a valid sending domain for this account';
+    }
+
+    if (facet_name === 'ip_pool' && value && value !== '') {
+      const ipPoolsArray = ipPools.map(({ id }) => id) || [];
+      return ipPoolsArray.includes(value) ? undefined : 'This is not a valid ip pool for this account';
+    }
+
+    return required;
```
jonambas: I think this needs to be `return required(value)`

jonambas:
`z-index: z-index(overlay);` in MultiFacet.module.scss

PR #878, opened 2019-03-14
AC-647: Removed subaccount columns for subaccount users

PatrickSison:
### What Changed
 - Hid subaccount columns for any with subaccount_reporting access level
   - Hid in templates list page
   -  Hid in snippets list page
 - Right aligned icon in ActionPopover to be consistent with other lists
    - ActionPopover used in users list and snippets list

### How To Test
 - Check that subaccounts columns show up in list for non subaccount users
- Check that the column gone if user is a subaccount reporting user

### To Do
- [ ] Address feedback

### DIFF
--- a/src/pages/snippets/ListPage.js
+++ b/src/pages/snippets/ListPage.js
@@ -5,7 +5,7 @@ import { Templates } from 'src/components/images';
 import Loading from 'src/components/loading';
 import PageLink from 'src/components/pageLink';
 import SnippetCollection from './components/SnippetCollection';
-
+import { ROLES } from 'src/constants';
 export default class ListPage extends React.Component {
   componentDidMount() {
     this.props.getSnippets();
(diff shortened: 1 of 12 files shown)


src/pages/snippets/ListPage.js
```diff
@@ -23,11 +23,12 @@ export default class ListPage extends React.Component {
   }
 
   renderCollection() {
-    const { snippets, hasSubaccounts, canCreate } = this.props;
+    const { snippets, hasSubaccounts, canCreate, userAccessLevel } = this.props;
     return (
       <SnippetCollection
         canCreate={canCreate}
         hasSubaccounts={hasSubaccounts}
+        canViewSubaccounts={userAccessLevel !== ROLES.SUBACCOUNT_REPORTING}
```
jonambas: Wondering if `canViewSubaccounts` should be a selector instead to avoid evaluating in multiple places. But I guess that can be part of an access control rework.

PR #887, opened 2019-03-18
SA-518 Updated save as CSV button in Events UI to save first *5000* results instead of only 1 page

kchu1337:
### What Changed
 - Changed 'Save as CSV' button in the Events UI to make an events API call and save the results 

### How To Test
 - Navigate to reports/message-events and click the 'Save as CSV' button in the bottom. It should save 5000 results.
 -Also check that other pages' per page and save as CSV did not break. 

### To Do
- [x] Possibly increase limit from 1k to 5k.

### Implementation notes
 - Created a new download helper with methods for 'formatCSV' and 'download'. 'formatCSV' is copied from `src/components/collection/SaveCSVButton.js` and 'download' is copied from `src/pages/billing/components/InvoiceHistory.js`
 - Removed `CollectionControls` and changed `Pagination` and `MessageEventsPage` to have 2 components for per page and downloadCSV. 
 - Made new action/reducer pair for getting csv data using a new API call. 
 - Updated `MessageEventsPage` to know when to download after button has been clicked.

### Other notes
- Possible change/remove the downloading test from `InvoiceHistory` as its duplicative of the download helper test.


src/reducers/messageEvents.js
```diff
@@ -38,13 +40,26 @@ export default (state = initialState, { type, payload, meta, extra }) => {
       const hasMorePagesAvailable = Boolean(next);
       const linkByPage = [currentUrlParams, nextUrlParams ];
       const cachedResultsByPage = [ payload ];
-      return { ...state, linkByPage, totalCount, cachedResultsByPage, loading: false, events: payload, hasMorePagesAvailable };
+      return { ...state, eventsCSV: [], linkByPage, totalCount, cachedResultsByPage, loading: false, events: payload, hasMorePagesAvailable };
```
rajumsys: is resetting `eventsCSV` intentional?

src/helpers/downloading.js
```diff
@@ -0,0 +1,29 @@
+import _ from 'lodash';
+import Papa from 'papaparse';
+
+export function formatToCsv({ data: rows, returnBlob }) {
+  // we are doing this because certain keys are objects/array which papa parse doesn't stringify
+  const mappedRows = _.map(rows, (row) => _.mapValues(row, (value) => _.isObject(value) || _.isArray(value) ? JSON.stringify(value) : value));
+  const csvData = Papa.unparse(mappedRows);
+  return (returnBlob)
```
rajumsys: brace is not needed

src/reducers/messageEvents.js
kchu1337: The 5k events is like 2.5MB. I just thought it would be good to remove it from redux store when we don't need it.

rajumsys: ok. that's fine. but i was thinking it would be more useful if we can clear it after download. because using this approach, it won't be cleared until user refreshes/paginates.

kchu1337: I was also considering that, but then I would need to make a new action/reducer and I wasn't how it would interact with the async call. I can certainly add that functionality if it makes a big performance improvement. This was just a quick way to add some minor improvement.

rajumsys: well, not sure how much performance gain it'll be without some profiling but i guess at least that amount of data will be in memory. 

the only reason of that suggestion is that you're resetting it via other action. if that action is not triggered it's not getting reset. so if you're doing for performance improvement, IMO, we should do it always; not on some other action. let me know your thought.

kchu1337: I just couldn't figure out how to have the action trigger on `success`. If i uses a `promise...then`, the action dispatches during the `action_pending` stage.

src/pages/reports/messageEvents/MessageEventsPage.js
```diff
@@ -121,14 +140,19 @@ export class MessageEventsPage extends Component {
...
+            <PerPageButtons
+              totalCount={totalCount}
+              data={events}
+              onPerPageChange={this.handlePerPageChange}
+              perPageButtons={DEFAULT_PER_PAGE_BUTTONS}
+              perPage={perPage}
+              saveCsv={true}
+            />
+            <Button onClick={this.getCSV} disabled={eventsCSVLoading}>
+              {(eventsCSVLoading) ? 'Saving CSV...' : 'Save as CSV'}
```
rajumsys: Should we say something else like `Preparing CSV to Download` or something similar? this doesn't really save it but rather prepares and then asks user to save.

PR #889, opened 2019-03-18
SD-748 - IP Auto Warmup

rajumsys:
### What Changed
     - Allow ip warmup and stage to be selected/modified

### How To Test
- [ enable this option to your account](https://github.com/SparkPost/2web2ui/blob/master/docs/feature-flags.md#give-access-to-your-feature): `ip_auto_warmup`. it's enabled on `appteam` on staging. remember to submit whole `ui` stanza with new/updated values. 
- enable/disable: No special instructions. Just check the box on/off. 
- stage: Use control endpoint to set higher stage. See https://msys.slack.com/archives/CFFCXGAE7/p1553099114072900


src/pages/ipPools/components/IpForm.js
```diff
@@ -13,13 +14,32 @@ import {
...
   render() {
-    const { ip, pool, pools, handleSubmit, submitting, pristine } = this.props;
+    const { ip, pool, pools, ipAutoWarmupEnabled, handleSubmit, maxStages, submitting, pristine } = this.props;
     const reAssignPoolsOptions = pools.map((currentPool) => ({
       value: currentPool.id,
       label: (currentPool.id === pool.id) ? '-- Select a new pool --' : `${currentPool.name} (${currentPool.id})`
     }));
 
+    const confirmationModalText = ipAutoWarmupEnabled
+      ? 'Enabling Auto IP Warmup will limit the amount of traffic that is able to be sent over this IP based on the warmup stage. Additional traffic will be distributed amongst other IPs in the same pool or the designated overflow pool.'
```
ewandennis: `<clippy>Passive voice. Consider revising</clippy>` :)
Instead of `...is able to be...`, how about `...you can...`.

`Additional traffic...` sounds like _some_ additional traffic but I think the intent is to describe _all_ traffic above the limit. How about `Remaining traffic...` or `All additional traffic...`

src/pages/ipPools/components/IpForm.js
```diff
@@ -13,13 +14,32 @@ import {
...
+      : 'Disabling Auto IP Warmup will remove the volume restrictions from this IP, if this IP is not properly warmed, this can have negative consequences on deliverability and sender reputation.';
```
ewandennis: I think this is 2 sentences:
> Disabling Auto IP Warmup will remove the volume restrictions from this IP. If this IP is not properly warmed, this can have negative consequences on deliverability and sender reputation.

Maybe also refer to the reader explicitly with "your" to make the point clear:

> Disabling Auto IP Warmup will remove the volume restrictions from this IP. If this IP is not properly warmed, this can have negative consequences on your deliverability and sender reputation.

src/pages/ipPools/components/IpForm.js
```diff
@@ -13,13 +14,32 @@ import {
...
+
+    const stageOptions = _.map(_.range(maxStages), (i) => ({
```
ewandennis: I imagine we can get away with `Array.prototype.map()` here. 🤷‍♂️

src/pages/ipPools/components/IpForm.js
```diff
@@ -39,12 +59,50 @@ export class IpForm extends Component {
             </LabelledValue>
           </Panel.Section>
 
+          <Panel.Section actions={[{ content: <UnstyledLink to="https://www.sparkpost.com/docs/deliverability/ip-warm-up-overview/" external>What is Auto Warmup?</UnstyledLink>, onClick: _.noop, color: 'orange' }]}>
```
ewandennis: Panel.Section actions can be `Button`-y.
[Button supports `to` and `external` which make it act like a `Link`](https://github.com/SparkPost/matchbox/blame/master/packages%2Fmatchbox%2Fsrc%2Fcomponents%2FButton%2FButton.js#L22).

```jsx
          <Panel.Section
            actions={[
              {
                content: 'What is Auto Warmup?',
                to: 'https://www.sparkpost.com/docs/deliverability/ip-warm-up-overview/',
                external: true,
                color: 'orange'
              }
            ]}
          >
```

src/pages/ipPools/components/IpForm.js
```diff
@@ -39,12 +59,50 @@ export class IpForm extends Component {
...
+            <LabelledValue label='Auto IP Warmup'>
+              <Field
+                name="auto_warmup_enabled"
+                component={CheckboxWrapper}
+                onChange={() => {
+                  this.setState({ warningModal: true });
```
ewandennis: Do we want this warning to show when we check the box? The language on the modal reads like confirming it will trigger the enable action which doesn't actually happen until form submit.

If the intent is to warn when checking/unchecking the box, we should change the modal language to note that change is affected when the form is submitted. For instance, the primary button should likely say "Got it", "Ok"  or similar, and we wouldn't need a cancel button.

We have form for this type of warning In other cases (e.g. those on the SSO and mandatory TFA account settings), we show the message on submit, and the chosen action on the modal submits the form or doesn't.

src/pages/ipPools/components/IpForm.js
```diff
@@ -39,12 +59,50 @@ export class IpForm extends Component {
...
             <Button submit primary disabled={submitting || pristine}>
               {submitting ? 'Saving' : 'Update Sending IP'}
             </Button>
           </Panel.Section>
         </form>
+
+        <ConfirmationModal
+          open={this.state.warningModal}
+          title={`Are you sure you want to ${ipAutoWarmupEnabled ? 'enable' : 'disable'} Auto IP Warmup?`}
+          content={<p>{confirmationModalText}</p>}
```
ewandennis: Almost all the language for this modal is in this location, which makes it nice and readable even with the conditionals. Can we move `confirmationModalText` in here too?

src/pages/ipPools/components/IpForm.js
```diff
@@ -53,15 +111,19 @@ export class IpForm extends Component {
 IpForm.defaultProps = {
   pools: [],
   pool: {},
-  ip: {}
+  ip: {},
+  maxStages: 20
```
ewandennis: Is this just a constant? If so, we should move it under `/src/constants`. However, based on how this is used, maybe we just need a list of constants for each warmup stage.

src/pages/ipPools/components/IpForm.js
```diff
@@ -39,12 +59,50 @@ export class IpForm extends Component {
...
+                disabled={submitting}
+              />
+            </LabelledValue>
+            {ipAutoWarmupEnabled &&
+            <Fragment>
+              <LabelledValue label='Warmup Stage'>
+                <Field
+                  name='auto_warmup_stage'
+                  component={SelectWrapper}
+                  options={stageOptions}
```
ewandennis: The support doc we link to has a stage for each day across 2 weeks and the wireframe shows volumes in each stage. We're showing `stage 1` thru `20`. Can we include the volumes for each stage? If so, do we need the `Stage N` prefix at all?

src/pages/ipPools/components/IpForm.js
```diff
@@ -39,12 +59,50 @@ export class IpForm extends Component {
...
           </Panel.Section>
         </form>
+
+        <ConfirmationModal
+          open={this.state.warningModal}
+          title={`Are you sure you want to ${ipAutoWarmupEnabled ? 'enable' : 'disable'} Auto IP Warmup?`}
+          content={<p>{confirmationModalText}</p>}
+          onCancel={this.revertAutoWarmupToggling}
+          onConfirm={() => this.setState({ warningModal: false })}
+          confirmVerb={ipAutoWarmupEnabled ? 'Yes, I want to turn ON Auto IP Warmup' : 'Yes, I want to turn OFF Auto IP Warmup'}
```
ewandennis: Yeah, this is the language that confuses me. If I click `Yes, I want to... turn ON Auto IP Warmup` that doesn't actually turn it on.

src/pages/ipPools/components/IpForm.js
```diff
@@ -38,13 +74,48 @@ export class IpForm extends Component {
               />
             </LabelledValue>
           </Panel.Section>
-
+          <AccessControl condition={configFlag('featureFlags.ip_auto_warmup')}>
```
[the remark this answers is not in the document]
rajumsys: seems this isn't an account option, rather it's a tenant level flag. api is also not released on production. also seems we're using signing_domain using this same method.

src/pages/ipPools/components/IpForm.js
```diff
@@ -1,25 +1,61 @@
...
       value: currentPool.id,
       label: (currentPool.id === pool.id) ? '-- Select a new pool --' : `${currentPool.name} (${currentPool.id})`
     }));
 
+    const confirmationModalText = isAutoWarmupEnabled
+      ? 'Enabling Auto IP Warmup will limit the amount of traffic that you can send over this IP based on the warmup stage. Remaining traffic will be distributed amongst other IPs in the same pool or the designated overflow pool.'
+      : 'Disabling Auto IP Warmup will remove the volume restrictions from this IP. If this IP is not properly warmed, this can have negative consequences on your deliverability and sender reputation.';
+
+    const stageOptions = IP_WARMUP_STAGES.map((stage) => ({
+      label: `${stage.name} (Volume: ${stage.volume})`,
```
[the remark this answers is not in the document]
rajumsys: Team doesn't want to mention day. 

Number formatting is a good idea.

src/pages/ipPools/components/tests/IpForm.test.js
```diff
@@ -35,7 +47,87 @@ describe('IP Form tests', () => {
...
+    expect(wrapper.find('Button').prop('disabled')).toBe(true);
+  });
+
+  it('changes submit button text when saving', () => {
+    wrapper.setProps({ submitting: true });
+    expect(wrapper.find('Button').shallow().text()).toEqual('Saving');
+  });
+
+  describe('auto_warmup_enabled', () => {
+    it('shows confirmation modal on submit to enable ip auto warmup', async () => {
```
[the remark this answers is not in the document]
rajumsys: noe. oversight.

rajumsys:
> After re-assigning an IP, I was confused to be sent to the edit page for the pool the IP used to be in. I would expect to stay on the IP edit page, like the non-reassignment case. 

I've done this by redirecting to new pool -> ip path. but it comes at a cost of not redirecting to pool page after ip is not found in the pool (like someone hits an old url with old pool & ip combination). doesn't seem big issue but if it is, we might have to revert getting ip pool and do redirect after ip is not found in that response.

rajumsys:
Changed the behavior of reassignment. it redirects to path with updated pool in path. 
this comes at a cost that i can no longer redirect when an ip is not found. might not be a big deal as many other edit page are like that.

src/config/default.js
```diff
@@ -48,7 +48,8 @@ const config = {
     allow_default_signing_domains_for_ip_pools: false,
     allow_mailbox_verification: true,
     allow_anyone_at_verification: false,
-    has_signup: false
+    has_signup: false,
+    ip_auto_warmup: true
```
rajumsys: this has to be `false` in default config

ewandennis:
> > After re-assigning an IP, I was confused to be sent to the edit page for the pool the IP used to be in. I would expect to stay on the IP edit page, like the non-reassignment case.
> 
> I've done this by redirecting to new pool -> ip path. but it comes at a cost of not redirecting to pool page after ip is not found in the pool (like someone hits an old url with old pool & ip combination). doesn't seem big issue but if it is, we might have to revert getting ip pool and do redirect after ip is not found in that response.

Redirecting after reassignment makes sense.
We should definitely handle the case where we navigate to a non-existent pool/ip route. [The templates edit page has an example](https://github.com/SparkPost/2web2ui/blame/master/src%2Fpages%2Ftemplates%2FEditPage.js#L27).

PR #892, opened 2019-03-20
EO-628 Health Score Gauge

jonambas:
This PR isn't opened against master. Will need to do some cleanup with EO-632 when that is ready.

### What Changed
- Added new components to support these mocks: https://sparkpost.invisionapp.com/share/H8R0LV8E5NA#/352294734_Health_Dashboard

### How To Test
- Point your local upstreams to staging
- Log into appteam or enable the `feature_signals_v2` UI option
- Navigate to the new Health Score page from the main side nav

### To Do
- [ ] Address Feedback
- [ ] Lift the GET health score request up to the page component from the summary table
  - May hold on this to avoid ballooning this PR

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5237,7 +5237,7 @@
     },
     "ansi-escapes": {
       "version": "3.1.0",
-      "resolved": "https://registry.npmjs.org/ansi-escapes/-/ansi-escapes-3.1.0.tgz",
+      "resolved": "http://registry.npmjs.org/ansi-escapes/-/ansi-escapes-3.1.0.tgz",
       "integrity": "sha512-UgAb8H9D41AQnu/PbWlCofQVcnV4Gs2bBJi9eZPxfU/hgglFh3SMDMENRIqdr7H6XFnXdoknctFByVsCOotTVw==",
       "dev": true
     },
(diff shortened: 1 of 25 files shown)


src/helpers/signals.js
```diff
@@ -28,3 +29,28 @@ export const getFriendlyTitle = ({ prefix, facet, facetId, subaccountId }) => {
...
+
+export const getDoD = (current, before) => {
+  if (_.isNil(current) || _.isNil(before)) {
+    return null;
+  }
+
+  return ((current - before) / before * 100);
+};
+
+export const getCaretProps = (value, reverse) => {
```
jonambas: Might just move this into the MetricDisplay component

src/selectors/signals.js
```diff
@@ -308,7 +309,8 @@ export const selectHealthScoreOverviewData = createSelector(
         normalizedHistory.reduce((total, { health_score }) => total + health_score, 0) / normalizedHistory.length,
         1
       ),
-      WoW: _.isNil(WoW) ? null : roundToPlaces(WoW * 100, 0)
+      WoW: _.isNil(WoW) ? null : roundToPlaces(WoW * 100, 0),
+      current_DoD: getDoD(_.last(filledHistory).health_score, filledHistory[filledHistory.length - 2].health_score)
```
PatrickSison: This key is `current_DoD` but in CurrentHealthGuage.js, it's expecting the key to be `DoD`

PatrickSison:
Will the EO-632 put a grid on the 2 sections for that and this ticket? On mobile view, the 2 panels could potentially stack vertically

package-lock.json
```diff
@@ -5237,7 +5237,7 @@
     },
```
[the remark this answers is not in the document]
jonambas: Nope, will fix

src/selectors/signals.js
jonambas: Oh good catch, one of the downsides of using sparse datasets

PR #894, opened 2019-03-20
SD-749: Add overflow pool 

rajumsys:
This PR adds ability to set overflow pool for an IP Pool. There seems to have a misunderstanding even with the prototype. I understood it will be on Sending IP but turns out it'll be on IP Pool. Harold updated the ticket's AC. 

As it's in IP Pool object/page, we don't have any blocker with two other ip warmup fields that we've added in #889 PR. 

### What Changed
 - Added overflow pool in ip pool editing form
 - Added the field under feature flag

### How To Test
 - [ enable this option to your account](https://github.com/SparkPost/2web2ui/blob/master/docs/feature-flags.md#give-access-to-your-feature): `ip_auto_warmup`. it's enabled on `appteam` on staging. remember to submit whole `ui` stanza with new/updated values. 
 - Open an IP Pool and you will see the drop down. You can select another pool (as overflow pool) if it has IP. 

### To Do
- [x] Tests
- [x] ~Confirm Community pool label, value and when it should be shown (no pool with ip vs no pool with warmed ip)~ We won't show it in UI at all. Also hide overflow pool when no pools to show.
- [x] ~Should Default pool have a overflow pool and can default pool can be chosen as overflow pool?~ 
   Default pool can't have overflow but other pools can have

### DIFF
--- a/scripts/generateConfigs/tenants/staging.js
+++ b/scripts/generateConfigs/tenants/staging.js
@@ -61,8 +61,7 @@ const stagingTenants = {
       allow_anyone_at_verification: true,
       allow_default_signing_domains_for_ip_pools: true,
       has_signup: true,
-      ip_auto_warmup: true,
-      ip_auto_warmup_shared_pool_access: true
+      ip_auto_warmup: true
     },
     host: 'app-staging.sparkpost.com',
     nextHost: 'phoenix-next-stg.sparkpost.com',
--- a/scripts/generateConfigs/tenants/staging.js
+++ b/scripts/generateConfigs/tenants/staging.js
@@ -61,7 +61,8 @@ const stagingTenants = {
       allow_anyone_at_verification: true,
       allow_default_signing_domains_for_ip_pools: true,
       has_signup: true,
-      ip_auto_warmup: true
+      ip_auto_warmup: true,
+      ip_auto_warmup_shared_pool_access: true
     },
     host: 'app-staging.sparkpost.com',
     nextHost: 'phoenix-next-stg.sparkpost.com',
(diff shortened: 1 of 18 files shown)


scripts/generateConfigs/tenants/staging.js
```diff
@@ -22,7 +22,8 @@ const stagingTenants = {
   stagingmtas: {
     apiBase: 'https://api-staging-mtas.sparkpost.com/api',
     featureFlags: {
-      allow_anyone_at_verification: true
+      allow_anyone_at_verification: true,
+      ip_auto_warmup: true
```
jonambas: Should we remove these feature flags if we are using `isAccountUiOptionSet`?

src/pages/ipPools/components/PoolForm.js
```diff
@@ -41,6 +63,19 @@ export class PoolForm extends Component {
...
             }
+
+            {!editingDefault && Boolean(overflowPools.length) &&
+              <AccessControl condition={isAccountUiOptionSet('ip_auto_warmup', false)}>
+                <Field
+                  name='auto_warmup_overflow_pool'
+                  label='Overflow Pool'
+                  component={SelectWrapper}
+                  options={overflowPools}
+                  helpText='If you are using the automated IP warmup feature, the overflow pool will be used when volume threshold for this pool has been reached.'
```
jonambas: Does this make sense?

```suggestion
                  helpText='With automatic IP warmup enabled, this pool will be used for excess sending volume to protect your IPs.'
```

scripts/generateConfigs/tenants/staging.js
rajumsys: yes, not needed anymore.

src/pages/ipPools/components/PoolForm.js
rajumsys: either way sounds good to me.

scripts/generateConfigs/tenants/staging.js
rajumsys: yeah, i thought i've removed all.

PR #895, opened 2019-03-20
Force isSupportedBrowser to boolean so always sent to Sentry

bkemper:


src/helpers/tests/errorTracker.test.js
```diff
@@ -23,6 +20,15 @@ describe('.breadcrumbCallback', () => {
 });
 
 cases('.getEnricherOrDieTryin', ({ currentWindow = {}, data = {}, state = {}}) => {
+  mockBowser.getParser = jest.fn((userAgent) => ({
+    satisfies: jest.fn(() => {
+      // return undefined like bowser when browser is unsupported
+      // see, https://github.com/lancedikson/bowser/blob/700732f7b3f490cb61f9f4d34142d0a9ea5a38d3/index.d.ts#L177
+      if (userAgent === 'unsupported') { return; }
```
kchu1337: You can simplify to something like:
` (userAgent === 'unsupported') ? undefined : true)`
but meh, it doesn't really matter.

PR #901, opened 2019-03-25
FE-640 Fix Webhooks form submitting state

kchu1337:
### What Changed
 - Moved variables derived from redux form props out of `mapStateToProps` and into the render function.
 - Added tests
 - Changed `EditTab` to explicitly pass the newWebhook prop as false to `WebhookForm`. 

### How To Test
 - Navigate to webhooks page. Try to create a new webhook. When you click the `create webhook` button, it should start submitting. The button should now be disabled and say `Submiting`. The button should also be disabled when no changes have occurred. 
 - Do the same with updating page.

### DIFF
--- a/src/pages/webhooks/components/tests/WebhookForm.test.js
+++ b/src/pages/webhooks/components/tests/WebhookForm.test.js
@@ -83,6 +83,7 @@ describe('Webhooks Form Component', () => {
     expect(wrapper.find('EventsRadioGroup').prop('disabled')).toBe(true);
     expect(wrapper.find('EventCheckBoxes').prop('disabled')).toBe(true);
     expect(wrapper.find('AuthDropDown').prop('disabled')).toBe(true);
+    expect(wrapper.find('AuthFields').prop('disabled')).toBe(true);
     expect(wrapper.find('ActiveField').prop('disabled')).toBe(true);
   });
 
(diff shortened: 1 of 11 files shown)


src/pages/webhooks/components/tests/WebhookForm.test.js
```diff
@@ -75,10 +75,25 @@ describe('Webhooks Form Component', () => {
...
+      wrapper.setProps({ newWebhook: true });
+      expect(wrapper.find('Button').props().children).toEqual('Create Webhook');
+    });
+    it('should render correct submit text and be disabled when submitting', () => {
+      wrapper.setProps({ submitting: true });
+      expect(wrapper.find('Button').props().disabled).toEqual(true);
+      expect(wrapper.find('Button').props().children).toEqual('Submitting...');
+    });
+    it('should disable submit button when pristine', () => {
+      expect(wrapper.find('Button').props().disabled).toEqual(true);
```
rajumsys: you can use `toBe(true)`.

src/pages/webhooks/components/tests/WebhookForm.test.js
```diff
@@ -75,10 +75,25 @@ describe('Webhooks Form Component', () => {
...
+    });
+    it('should enable submit button when not pristine and not submitting', () => {
+      wrapper.setProps({ pristine: false });
+      expect(wrapper.find('Button').props().disabled).toEqual(false);
```
rajumsys: `toBe(false)`

src/pages/webhooks/components/tests/WebhookForm.test.js
```diff
@@ -75,10 +75,25 @@ describe('Webhooks Form Component', () => {
...
       expect(wrapper.find('Button').props().children).toEqual('Update Webhook');
     });
+    it('should render correct submit text when creating a new webhook', () => {
+      wrapper.setProps({ newWebhook: true });
+      expect(wrapper.find('Button').props().children).toEqual('Create Webhook');
+    });
+    it('should render correct submit text and be disabled when submitting', () => {
+      wrapper.setProps({ submitting: true });
+      expect(wrapper.find('Button').props().disabled).toEqual(true);
+      expect(wrapper.find('Button').props().children).toEqual('Submitting...');
```
rajumsys: this is fine just fyi, you could also write `prop('children')`

rajumsys:
Looks good. @kchu1337 when I `Select individual Events`, it didn't disable the individual event checkboxes.

PR #906, opened 2019-03-26
EO-650 Remove sending domain typeahead

jonambas:
### What Changed
- Removes the sending domain typeahead from the alerts form
- Adds a domain validator to the facet_value field

### How To Test
- Add the `feature_alerts` UI option to your account, or log into appteam
- Go to `/alerts-new/create`, or an alerts edit page
- Select a health score metric, and set facet name to 'Sending Domain'

### To Do


src/pages/alerts-new/components/AlertForm.js
```diff
@@ -244,8 +232,6 @@ const mapStateToProps = (state, props) => {
   const selector = formValueSelector(formName);
 
   return {
-    domains: selectDomainsBySubaccount(state, formName),
-    domainsLoading: state.sendingDomains.listLoading,
```
21042jim: i think you can also remove subaccount here since you removed it as a prop.

PR #908, opened 2019-03-27
SD-800 Fix IP Autowarmup issues

rajumsys:
Address SD-800 to fix some UI issues

### What Changed
- Fix issue to set overflow pool during pool creation
- Show overflow pool always (except default) in editing
- Add placeholder overflow and unsetting overflow pool
- Fix logics for listing overflow pools 

### How To Test
- Assign overflow pool during creation
- Unset overflow pool

### DIFF
--- a/src/__integration__/__snapshots__/billing.test.js.snap
+++ b/src/__integration__/__snapshots__/billing.test.js.snap
@@ -6,7 +6,6 @@ Array [
     Object {
       "data": Object {
         "name": "newPool",
-        "signing_domain": undefined,
       },
       "headers": Object {
         "Authorization": "very-real-access-token",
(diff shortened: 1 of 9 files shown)


PatrickSison:
Looks good. The only thing that is a bit weird is the dropdown having "Select an overflow Pool" meaning the same as it being an IP only pool, rather than saying "None" or something like that. I get it being there for placeholder text for creation on selection and matchbox/select doesn't have some placeholder prop. Maybe a change later on?

PatrickSison:
API handles it, but if you try to assign an overflow pool's overflow pool, it gives you the error. Possible in the future, we could disable the dropdown if it is an overflow pool (but I imagine it's a bit messy to grab that condition)

PR #910, opened 2019-03-27
AC-686 Updating /account/sending-domains permission

jmoyse:
Small fix to have the `account/sending-domains` route viewable with the `sending_domains/view` instead of `sending_domains/manage`. This should match the functionality of `account/tracking-domains`


PatrickSison:
I think we don't want to add this in yet since the idea is that we clean up the list page for sending domains first to check for permissions, then open it up to read permissions. Things like the `Add a domain` on the page would be viewable since there was only a route level check

jmoyse:
> I think we don't want to add this in yet since the idea is that we clean up the list page for sending domains first to check for permissions, then open it up to read permissions. Things like the `Add a domain` on the page would be viewable since there was only a route level check

Yeah this is more of a work in progress. Either we match the existing functionality for email users (links are clickable but error out) or make none of them clickable. Either way the current state is that you can attempt to edit your tracking domains as an email user and it errors out. I'll bring up a discussion to figure out the best course of action.

PR #911, opened 2019-03-28
AC-675: Disabled brightback for dev environment

PatrickSison:
### What Changed
 - Added enabled flag to brightback config
- Added enabled:false to dev brightback config
- Brightback does not prep or send when downgrading from paid to free plan

### How To Test
 - Using a paid account, downgrade from paid to free on all environments

### To Do
- [ ] Double check on staging and prod
- [ ] Address any feedback

### DIFF
--- a/src/components/brightback/Brightback.js
+++ b/src/components/brightback/Brightback.js
@@ -37,6 +37,6 @@ const mapStateToProps = (state, props) => ({
   valid: state.brightback.valid,
   url: state.brightback.url,
   data: selectBrightbackData(state, props),
-  enabled: configFlag('brightback.enabled')(state)
+  enabled: configFlag('brightback.enabled')()
 });
 export default withRouter(connect(mapStateToProps, { prepBrightback })(Brightback));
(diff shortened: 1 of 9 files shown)


PatrickSison:
Not sure if the default config should be disabled w/ test configs enabling vs disabling w/ prod enabling

rajumsys:
> Not sure if the default config should be disabled w/ test configs enabling vs disabling w/ prod enabling
Does it work (or supposed to) in env other than prod? if no, then let's disable by default and enable only in prod. 

Otherwise LGTM

PatrickSison:
Rebased

src/components/brightback/Brightback.js
```diff
@@ -34,6 +36,7 @@ Brightback.propTypes = BrightbackPropTypes;
 const mapStateToProps = (state, props) => ({
   valid: state.brightback.valid,
   url: state.brightback.url,
-  data: selectBrightbackData(state, props)
+  data: selectBrightbackData(state, props),
+  enabled: configFlag('brightback.enabled')(state)
```
rajumsys: does it need `state` as arg? i don't see how it's being used.

PatrickSison: It doesn't need it. Removed it

PR #917, opened 2019-03-29
EO-579 Add Complaints by cohort page

jonambas:
Note: No data is returned from the new signals endpoints from staging. You'll have to point to production. Also, this UI is hidden until 4/30.

### What Changed
- Adds Complaints By Cohort page. Page functionality is mostly copied from other pages.

### How To Test
- Log into appteam or add the `feature_signals_v2` UI option to your account.
- Visit an engagement recency details page
- Click on the complaints tabs

### To Do


src/helpers/tests/date.test.js
```diff
@@ -316,7 +316,12 @@ describe('Date helpers', () => {
 
   describe('getDateTicks', () => {
     it('returns an array of start, end and middle days', () => {
-      expect(getDateTicks('14days')).toMatchSnapshot();
+      jest.spyOn(Date, 'now').mockImplementation(() => '2018-02-02T04:20:00-04:00');
```
bkemper: Instead of mock `Date` which is used by `moment`, I think it is better to make `now` and argument with a default.

PR #924, opened 2019-04-01
EO-661 Fix current health score dashboard

jonambas:
### What Changed
- Created a new health score action for the dashboard that always gets the account aggregated current score (`-1`)

### How To Test
- [Point your local upstreams to staging](https://confluence.int.messagesystems.com/display/ENG/Configuring+Openresty+Upstreams)
  - Data available in account 108 (spc)
- Log into appteam or enable the `feature_signals_v2` UI option
- Navigate to the new Health Score page from the main side nav
- Verify the top two panels contain valid data when changing the filters:
  - Date ranges (should update)
  - Facet filters (should not update)
  - Subaccount (should not update)

### DIFF
--- a/src/pages/signals/dashboards/HealthScoreDashboard.js
+++ b/src/pages/signals/dashboards/HealthScoreDashboard.js
@@ -1,7 +1,7 @@
-import React, { Component } from 'react';
+import React, { useEffect } from 'react';
 import { connect } from 'react-redux';
 import { list as getSubaccounts } from 'src/actions/subaccounts';
-import { getInjections } from 'src/actions/signals';
+import { getCurrentHealthScore, getInjections } from 'src/actions/signals';
 import { Grid } from '@sparkpost/matchbox';
 import Page from '../components/SignalsPage';
 import HealthScoreOverview from '../containers/HealthScoreOverviewContainer';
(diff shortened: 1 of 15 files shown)


src/pages/signals/dashboards/HealthScoreDashboard.js
```diff
@@ -10,48 +10,47 @@ import DateFilter from '../components/filters/DateFilter';
...
-        </div>
-        <HealthScoreOverview subaccounts={subaccounts} />
-      </Page>
-    );
-  }
+  // Gets injections and current score for gauge and timeseries only when dates change
+  useEffect(() => {
+    getCurrentHealthScore({ relativeRange });
+    getInjections({ relativeRange });
+  }, [relativeRange, getCurrentHealthScore, getInjections]);
```
21042jim: from [https://reactjs.org/docs/hooks-effect.html](url), 

> useEffect(() => {
  document.title = `You clicked ${count} times`;
}, [count]); // Only re-run the effect if count changes

will the functions change? seems like just the state var `relativeRange` would change

src/selectors/signals.js
```diff
@@ -404,6 +405,44 @@ export const selectHealthScoreOverviewData = createSelector(
   })
 );
 
+export const selectCurrentHealthScoreDashboard = createSelector(
```
21042jim: since we are refactoring, maybe we should just put a getInjections selector here?

src/pages/signals/dashboards/HealthScoreDashboard.js
jonambas: No, but we have a lint plugin from react that automatically adds any vars used in effects as dependancies.

PR #932, opened 2019-04-04
FE-653: 5XX errors attempt to retry before throwing error

PatrickSison:
As recently there have been 502/503 errors with reaching auth, the app should more gracefully handle a 5XX. Instead of stopping upon a 500 error, will reattempt to make request

### What Changed
 - 500 Errors reattempt request
 - After max retries, throws regular 500 error

### How To Test
 - Using a local API (such as accusers), throw 500 errors
 - Request should reattempt

### To Do
- [x] Add a little bit of delay in between requests with setTimeout ?
- [x] Add retry to auth requests
- [x] Add testing for retrying on 5XX
- [ ] Address any feedback


src/pages/billing/components/Banners.js
```diff
@@ -81,8 +81,9 @@ export const EnterpriseBanner = () => (
 );
 
 export const FreePlanWarningBanner = ({ account = {}, accountAgeInDays = 0, ageRangeStart = 0, ageRangeEnd = 30 }) => {
-  const { subscription = {}, pending_subscription } = account;
-  if (pending_subscription || !subscription.code.includes('free15K')) {
+  const { subscription, pending_subscription } = account;
```
ewandennis: This looks like an unrelated piece of defensiveness. Is there a bug related to receiving an account subscription object without a `code` field?

PatrickSison: The issue was that the page was being rendered before even if account didn't load. w/ the new axios interceptor change, it no longer errors out (although it does stay on a loading page now). I can remove

src/helpers/tests/axiosInterceptors.test.js
```diff
@@ -0,0 +1,53 @@
...
+      config: requestConfig,
+      response: {
+        status: 404,
+        message: 'not found'
+      }
+    };
+  });
+  describe('5XX errors', () => {
+    cases('should retry if it is a 5XX error', async ({ retries = 0 }) => {
+      err.response.status = 500;
```
ewandennis: Should we test a few of the actual codes the API returns to us? `502` and 503` are common.

ewandennis:
Tested with a local accusers service using this middleware:

```js
const Random5xxMiddleware = (req, res, next) => {
  if (Math.random() < 0.3) {
    return res.singleError(
      new Error(`Unwarranted splodyboom`),
      503,
      'Boom'
    );
  }
  next();
};
```

PR #933, opened 2019-04-05
TR-1201: Tag elements for Pendo feature tracking

bkemper:
See [TR-1201](https://jira.int.messagesystems.com/browse/TR-1201) for more information.

### What Changed
Tagging elements with a name attribute, so they can be easily tracked in Pendo.

### How To Test
Good question.


src/components/collection/FilterBox.js
```diff
@@ -9,6 +9,7 @@ export default function CollectionFilterBox(props) {
   return (
     <Panel sectioned>
       <TextField
+        name="collection-filter-box"
```
rajumsys: IMO, this can come via prop with a default value. so we can override where we use it? like if we want to have it `sending-domains-filter-box`, from sending domain page we can do that if it's a prop.

src/components/collection/PerPageButtons.js
```diff
@@ -17,6 +17,7 @@ const PerPageButtons = ({ data, perPage, perPageButtons = DEFAULT_PER_PAGE_BUTTO
         <Button
           className={classnames(perPage === buttonAmount && styles.Selected)}
           key={buttonAmount}
+          name="collection-per-page-button"
```
rajumsys: looks like now each button will have same name. how will we differentiate which button is clicked?

src/components/collection/FilterBox.js
rajumsys: btw, i'm not strongly convinced overriding it needed. probably same name is fine as on Pendo this will be usually referenced with page (path). So should be fine unless we've more than one filterbox.  so i'll leave it up to you.

bkemper: I didn't put much thought into what this component could be.  I only focused on our current use case and in naming  it this way, the UX team could easily see the use of this component across the app or per page.

src/components/collection/PerPageButtons.js
bkemper: I think we should avoid classes as much as possible unless there is a specific style we want to track.  This button group is functioning the same as a radio button except the value is in the contents.  We could use `[name="collection-per-page-button"]:contains('100')` selector if we wanted to measure the use of a specific button.  Right now, I am just measuer the group on the TR pages.

bkemper:
@rajumsys All good questions!  Please let me know if you are not sold and I should make a change.

src/components/collection/PerPageButtons.js
rajumsys: i guess we wanted to avoid classes for style. but anyway, yes content selector should be fine.

bkemper:
I can show you how we can tag content in Pendo.

bkemper:
Rebased

src/components/modals/ConfirmationModal.js
```diff
@@ -30,7 +30,13 @@ export default class ConfirmationModal extends Component {
     return (
       <div>
         {content}
-        <Button onClick={onConfirm} primary className={styles.Confirm} disabled={confirming}>
+        <Button
+          className={styles.Confirm}
+          disabled={confirming}
+          name={`confirm-${confirmVerb.toLowerCase()}-button`}
```
bkemper: @rajumsys There are two examples, `IpForm` and `TrackingDomainRow`, where the `confirmVerb` prop value is not a single word.  I'm thinking about changing the name to a static value, "confirmation-modal-confirm-button", or should I replace spaces with hyphens before lower casing?

rajumsys: IMO static value is fine.

rajumsys:
> I can show you how we can tag content in Pendo.

i'm somewhat familiar with it. i was curious whether we should have some tagging implementation as a part of the verification process in addition to seeing the attributes from console.

PR #939, opened 2019-04-08
AC-701: Enabled reinitialize

PatrickSison:
[Link to ticket](https://jira.int.messagesystems.com/browse/AC-701)
### What Changed
 - Enabled reinitialize on redux form to allow form to reset on finishing loading country info
    - Fixes issue with `State` field not loading even though United States is defaulted country.

### How To Test
 - Check that `State` field appears on initial form
    - Affects `/account/billing/plan`, and not `/onboarding/plan`

### To Do
- [ ] Address any feedback

### DIFF
--- a/src/pages/billing/components/CardSection.js
+++ b/src/pages/billing/components/CardSection.js
@@ -1,0 +1,49 @@
+import React from 'react';
+import { Panel } from '@sparkpost/matchbox';
+import CardSummary from './CardSummary';
+import PaymentForm from '../forms/fields/PaymentForm';
+import BillingAddressForm from '../forms/fields/BillingAddressForm';
+
+const FORMNAME = 'changePlan';
+
+class CardSection extends React.Component {
+
+
+  render() {
+    const { countries, selectedPlan, account, useSavedCC, submitting, handleCardToggle } = this.props;
+    if (selectedPlan.isFree) {
+      return null; // CC not required on free plans
+    }
+
+    if (account.billing && useSavedCC) {
+      return (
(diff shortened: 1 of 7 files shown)


src/pages/billing/forms/ChangePlanForm.js
```diff
@@ -1,4 +1,4 @@
-/* eslint max-lines: ["error", 200] */
+/* eslint max-lines: ["error", 220] */
```
ewandennis: This is an indication that this component has grown to be a little unruly. How about we take the opportunity to do a little cleanup?

We could pull `renderCCSection()` out into a local component in a separate file.

We could also move `mapStateToProps()` and `mapDispatchToProps()` into a "container". Take a look at the [user edit page container](https://github.com/SparkPost/2web2ui/blame/master/src%2Fpages%2Fusers%2FEditPage.container.js) for an example of that pattern. Its also a good idea to include `propTypes` on the component if you move the redux connection stuff to a separate component. 

We could get away with just 1 of these changes to keep this component manageable.

src/pages/billing/components/CardSection.js
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
+import { Panel } from '@sparkpost/matchbox';
+import CardSummary from './CardSummary';
+import PaymentForm from '../forms/fields/PaymentForm';
+import BillingAddressForm from '../forms/fields/BillingAddressForm';
+
+const FORMNAME = 'changePlan';
+
+class CardSection extends React.Component {
```
kchu1337: I think you can make this a functional component. Or even possibly move the ```useSavedCC``` state into this component.

PR #940, opened 2019-04-08
More Follow-up for FE-635: Configure smtpAuth defaults

bkemper:
This is more follow-up for FE-635.

### What Changed
Need to by default enable smtpAuth and set the username to identifier.

### How To Test
SMTP Settings page


bkemper:
I don't see the enabled flag being used.  Possible we can remove it.

PR #947, opened 2019-04-10
EO-623 Custom Time Ranges

jonambas:
This is a large PR. If you are reviewing, let me walk you through the changes. When reviewing code, it's probably best to work each commit separately.

#### Signals
  - Signals `DateFilter` uses DatePicker
  - Adds support for custom `to` and `from` in all signals actions and selectors
  - Various labels and headers that referenced 'Current' are now replaced with a custom time range
  - Filters moved into panels to be more consistent with the UI

#### DatePicker
  - Adds a 7 day minimum validator and adds a `validate` prop to the datepicker
  - Now correctly hides next/previous buttons

#### Other
  - Removes bar view in Summary tables
  - Bumps Matchbox (Textfield focus behavior fix)
  - DatePicker style improvements

--- 

### How To Test

### Things to Verify
- Where to verify:
  - All details pages
  - All V1 summary tables
  - All V2 dashboards (with summary tables)
- What to verify:
  - The requests themselves & the data displayed
  - X axis tick marks
  - Any text referencing date with 'Current'
  - Date validation

### To Do
- [ ] Address any feedback


src/pages/signals/components/filters/DateFilter.js
```diff
@@ -1,34 +1,62 @@
...
-    changeSignalOptions({ relativeRange: event.currentTarget.value });
-  }
+export function DateFilter({ signalOptions, changeSignalOptions, left, now = new Date() }) {
+  // Start of day is used to prevent unwanted side effects when now changes
+  const relativeNow = getStartOfDay(moment(now).subtract(1, 'day')).toString();
+  const [dates, setDates] = useState({ to: null, from: null });
+  const { relativeRange, from, to } = signalOptions;
+
+  // Creates dates for the datepicker from signal options
+  useEffect(() => {
```
jonambas: Now that I'm looking at this, I don't think dates need to be stored in state at all 🤔

ewandennis:
Interaction / visual review first:
 - `/signals` is missing some intro copy:
    Screen Shot 2019-04-11 at 15 34 57
 - The new order of filters feels jarring, coming from the current layout. I see why `DateFilter` would be first. Maybe worth experimenting with original `SubaccountFilter` -> `FacetFilter` ordering?
 - Using a validator on the `DatePicker` is economical on our time but the experience is a bit rough. Letting me drag over a 3 day window but berating me for trying to select it feels like punishment. Maybe we have to live with it given the effort required but a prototype might have surfaced this much earlier.
 - The line or bar chart selector is missing from the table header on `/signals/health-score`
    Screen Shot 2019-04-11 at 15 50 24
 - some info tooltips are missing from `/signals` and `/signals/health-score` but since they show the same content, I guess thats not a loss :)
 - On `/signals/health-score` is looks like the `CurrentHealthGauge` always shows the score for the end day of the selected range. Thats not in the AC but seems logical. Intended?
 - Moving focus away from the filter text field refreshes the page even if the filter text hasn't changed.

-----------------

 - Unrelated: the (i) and (!) icons are very similar.
    Screen Shot 2019-04-11 at 15 55 35

src/pages/signals/containers/ComplaintsByCohortDetailsContainer.js
```diff
@@ -5,30 +5,33 @@ import { getComplaintsByCohort } from 'src/actions/signals';
 import { selectComplaintsByCohortDetails, getSelectedDateFromRouter } from 'src/selectors/signals';
```
ewandennis: Can we reduce this lot to a single `withDetails` HoC that takes a wrapped component, list of actions and a selector?

This is the most obvious piece of duplication I see in Signals. Most of the rest manages complexity and code duplication pretty well.

src/pages/signals/components/filters/FacetFilter.js
```diff
@@ -47,26 +48,25 @@ export class FacetFilter extends React.Component {
...
           />
-        </div>
-        {facet && (
-          <div className={styles.FacetFilter}>
+          {facet && (
             <TextField
               onChange={this.handleSearchChange}
               onKeyPress={onEnter(this.handleFacetSearch)}
-              placeholder="Search here"
+              onBlur={this.handleFacetSearch}
```
ewandennis: I think triggering the search on blur might be the reason we run a refresh cycle even if the search params have not changed.

Maybe we should add a test in `handleFacetSearch` checking whether `props.signalOptions.facetSearchTeam !== this.state.searchTerm`.

src/pages/signals/components/filters/FacetFilter.module.scss
```diff
@@ -1,7 +1,35 @@
 @import '~@sparkpost/matchbox/src/styles/config.scss';
 
+// Simulating connectLeft behavior
```
ewandennis: Nice. This is helpful.

src/pages/signals/dashboards/components/CurrentHealthGauge/CurrentHealthGauge.js
```diff
@@ -48,11 +51,15 @@ export function CurrentHealthGauge(props) {
...
+    ? `Health Score for ${formatDate(props.filters.to, FORMATS.DATE)}`
+    : 'Current Health Score';
+
   return (
     <Panel sectioned>
       <div className={styles.Content}>
         <h2 className={styles.Header}>
-          Current Health Score – Overall
+          {title}
           {' '}
```
ewandennis: Hehe. Add `margin-left: 0.5em;` to `.TooltipIcon` maybe?

src/pages/signals/dashboards/components/CurrentHealthGauge/CurrentHealthGauge.js
```diff
@@ -48,11 +51,15 @@ export function CurrentHealthGauge(props) {
     return { value: _.isNil(value) ? 'n/a' : `${value}%`, ...getCaretProps(value) };
   }
 
+  const title = _.get(props, 'filters.relativeRange') === 'custom'
```
ewandennis: This new condition will need a little test coverage.

src/pages/signals/components/dataCells/SparklineDataCell.js
```diff
@@ -12,7 +12,7 @@ const StyledDot = (props) => (
 );
 
 const SparklineDataCell = ({ data, dataKey, label, onClick = () => {}, relative }) => {
-  const currentDate = _.last(data).date;
+  const currentDate = _.get(_.last(data), 'date');
```
ewandennis: I think this came up earlier. Do we expect to receive records from the API which are _missing_ these fields (`date` in this case) or are the values expected to sometimes be `null`?

jonambas:
> /signals is missing some intro copy:

Will add back

> The new order of filters feels jarring, coming from the current layout. I see why DateFilter would be first. Maybe worth experimenting with original SubaccountFilter -> FacetFilter ordering?

Yep I can revert the order, not sure what make sense here. I initially changed the order because not all accounts have subaccounts.

> Using a validator on the DatePicker is economical on our time but the experience is a bit rough. Letting me drag over a 3 day window but berating me for trying to select it feels like punishment. Maybe we have to live with it given the effort required but a prototype might have surfaced this much earlier.

Not sure what the best solution is here. Our other date pickers need the validation too. Could experiment with different styles on the selected/hovered range while selecting.

> The line or bar chart selector is missing from the table header on /signals/health-score

Should be part of the AC. Most felt the bar view was useless.

> some info tooltips are missing from /signals and /signals/health-score but since they show the same content, I guess thats not a loss :)

Tooltips are actually missing from the spam trap and engagement tables, let me see if I can find a place for them

> On /signals/health-score is looks like the CurrentHealthGauge always shows the score for the end day of the selected range. Thats not in the AC but seems logical. Intended?

Yep intended, the header text changes so hopefully that is clear.

> Moving focus away from the filter text field refreshes the page even if the filter text hasn't changed.

Intended because the only way to submit the search was to press enter (not obvious). I can maybe check if value has changed before refreshing though.

src/pages/signals/components/EngagementRecencyOverview.js
```diff
@@ -7,21 +7,18 @@ import SummaryTable, { Column } from 'src/components/summaryTable';
 import { setSubaccountQuery } from 'src/helpers/subaccounts';
```
ewandennis: Groundhog day :) I imagine we could make a single overview component and provide a simple open-for-extension to override it for the health score case. The test suites look like direct copy and paste too :/

src/helpers/date.js
```diff
@@ -155,12 +156,17 @@ export const parseDate = (str) => moment(str, FORMATS.INPUT_DATES, true);
 export const parseTime = (str) => moment(str, FORMATS.INPUT_TIMES, true);
 export const parseDatetime = (...args) => moment(args.join(' '), FORMATS.INPUT_DATETIMES, true);
 
-export const fillByDate = ({ dataSet, fill = {}, now, relativeRange } = {}) => {
-  const { from, to } = getRelativeDates(relativeRange, { now });
+export const fillByDate = ({ dataSet, fill = {}, from, now, relativeRange, to } = {}) => {
+  let dates = { from, to };
+
+  if (relativeRange && relativeRange !== 'custom') {
```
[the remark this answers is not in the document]
jonambas: I think a lot of this logic can be removed if `from` & `to` is calculated in the signalOptions action, let me see if I can do that

src/actions/signals.js
```diff
@@ -15,14 +15,21 @@ const ORDER_BY_MAPPING = {
...
   orderBy,
   relativeRange,
-  subaccount
+  subaccount,
+  to
 }) => {
-  const { from , to } = getRelativeDates(relativeRange, { now: moment().subtract(1, 'day') });
+  let dates = { from, to };
+
+  if (relativeRange !== 'custom') {
```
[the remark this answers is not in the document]
jonambas: This was here because initial signalOptions state did not include `to` and `from`, I think I have a solution you will like

src/pages/signals/components/filters/DateFilter.js
```diff
@@ -14,32 +14,35 @@ const OPTIONS = [
...
-
-  function handleChange(updates) {
-    changeSignalOptions(updates);
+    changeSignalOptions(options);
   }
 
+  // On mount, sets dates from a 90 day default
+  useEffect(() => {
+    handleChange({ relativeRange: DEFAULT_RANGE });
+  }, []); // eslint-disable-line react-hooks/exhaustive-deps
```
ewandennis: Before this change, you could set a custom range on 1 signals page and it would be used by the other signals pages. Now each page resets to 90 days on mount. FWIW, our other aggregate reports share date settings.

Its caused by setting `relativeRange` on mount unconditionally, which clobbers any date options set on previous pages.

We could just conditionally initialise:
```js
useEffect(() => {
  if (!signalOptions.relativeRange) {
    handleChange({ relativeRange: DEFAULT_RANGE });
  }
}, []);
```

jonambas: Oh hey, I had this solution locally for a bit. The pages still double loads because it attempts to render without and dates set.

Solutions:
(1) Check `!signalOptions.relativeRange` in every container
(2) Create initial dates in the reducer

I went with 2

PR #952, opened 2019-04-15
SA-560 Changed summary report typeahead to use metrics template endpoint

kchu1337:
### What Changed
The summary report typeahead no longer uses the `/templates` endpoint but the `/metrics/templates` endpoint. 

### How To Test
 - Search for a template in the summary page typeahead. The template should appear in the typeahead dropdown
 -  Can also test that accounts with >50k templates have all of them available. 

### To Do
- [x] Address review feedback

### DIFF
--- a/src/actions/metrics.js
+++ b/src/actions/metrics.js
@@ -42,6 +42,11 @@ export function fetchMetricsIpPools(params = {}) {
   const path = 'ip-pools';
   return fetch({ type, path, params });
 }
+export function fetchMetricsTemplates(params = {}) {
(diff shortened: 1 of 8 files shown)


src/actions/metrics.js
```diff
@@ -42,6 +42,11 @@ export function fetchMetricsIpPools(params = {}) {
   const path = 'ip-pools';
   return fetch({ type, path, params });
 }
+export function fetchMetricsTemplates(params = {}) {
```
jonambas: Wow, we don't have tests for these

PR #954, opened 2019-04-17
AC-692: Updated Test account to include new codes

PatrickSison:
### What Changed
 - Added new free plan code into the list of plans considered 'Test Account'

### How To Test
 - With new free plan (to be rolled out in [AC-683](https://jira.int.messagesystems.com/browse/AC-683)), should show as 'Test Account' in PlanPicker on /account/billing/plan

### To Do
- [ ] Address feedback

### DIFF
--- a/src/components/billing/PlanPrice.js
+++ b/src/components/billing/PlanPrice.js
@@ -16,7 +16,7 @@ const PlanPrice = ({ plan, showOverage = false, showIp = false, showCsm = false,
     ? 'First dedicated IP address is free'
     : null;
 
-  const currentFreePlan = plan.code && ['free500-1018', 'free15K-1018'].includes(plan.code);
+  const currentFreePlan = plan.code && ['free500-1018', 'free15K-1018', 'free500-0419', 'free500-SPCEU-0419'].includes(plan.code);
   const planTitle = currentFreePlan ? 'Test Account' : `${plan.volume.toLocaleString()}`;
 
   const displayCsm = showCsm && plan.includesCsm;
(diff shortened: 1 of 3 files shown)


src/components/billing/tests/PlanPrice.test.js
```diff
@@ -81,6 +81,14 @@ describe('PlanPrice', () => {
     expect(wrapper).toMatchSnapshot();
   });
 
+  it('renders test account with current free plan', () => {
+    plan.monthly = 0;
+    plan.isFree = true;
+    plan.code = 'free500-0419';
```
bkemper: Let composition be your friend, it will save you from an ugly mutant.

```
const currentFreePlan = {
  ...plan,
  monthly: 0,
  isFree: true,
  code: 'free500-0419'
};
wrapper.setProps({ plan: currentFreePlan });
```

src/components/billing/PlanPrice.js
```diff
@@ -16,7 +16,7 @@ const PlanPrice = ({ plan, showOverage = false, showIp = false, showCsm = false,
     ? 'First dedicated IP address is free'
     : null;
 
-  const currentFreePlan = plan.code && ['free500-1018', 'free15K-1018'].includes(plan.code);
+  const currentFreePlan = plan.code && ['free500-1018', 'free15K-1018', 'free500-0419', 'free500-SPCEU-0419'].includes(plan.code);
```
bkemper: Aren't there other free plans?  (e.g. ccfree, free1, etc.)  Should they be included in this list?  Should we maintain a list at all?  Or could we consider all codes that contain "free" as free plans?

PatrickSison: There are other free plans, but currently only the most recent ones are considered 'Test Account', while older ones are considered grandfathered 'Free accounts'

PR #955, opened 2019-04-17
AC-712 Updated preview button to use transmission grant

PatrickSison:
### What Changed
 - Templates view page uses `transmissions/modify` grant to check if it can send

### How To Test
 - Using an email or reporting user, open up a template Edit/View page
   - You would need UI option `developer_and_email_roles` if you need to create role
- Button should now say 'Preview' for email and reporting (previously would show Preview & Send for email users)

### To Do
- [ ] Address feedback

### DIFF
--- a/src/pages/templates/containers/EditPage.container.js
+++ b/src/pages/templates/containers/EditPage.container.js
@@ -15,6 +15,7 @@ const FORM_NAME = 'templateEdit';
 const mapStateToProps = (state, props) => {
   const template = selectTemplateById(state, props).draft;
   const canModify = hasGrants('templates/modify')(state);
+  const canSend = hasGrants('transmissions/modify')(state);
 
   return {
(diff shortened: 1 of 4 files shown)


src/pages/templates/containers/EditPage.container.js
```diff
@@ -15,6 +15,7 @@ const FORM_NAME = 'templateEdit';
 const mapStateToProps = (state, props) => {
   const template = selectTemplateById(state, props).draft;
   const canModify = hasGrants('templates/modify')(state);
+  const canSend = hasGrants('transmissions/modify')(state);
```
bkemper: Just a note, `hasGrants` should be used with `AccessControl` component or `selectCondition` selector.  There are other places in templates and snippets that do it this way.  Leave it for now, I'll fix it up on future tickets.

PR #956, opened 2019-04-17
EO-689 Shift behavior by cohorts to display data from 3 days ago 

jonambas:
### What Changed
- Engagement cohort behavioral charts no longer display **empty** data between now and 3 days ago

### Notes
- This PR only adjust what is displayed in the charts, and does not modify the requests or DateFilter
- This PR does not shift `from`, only `to`
- DateFilter does not represent its displayed dates. It is agnostic of its page/location, and we would need to consider some UX to make that change.
  - navigation between charts with custom dates (would need to store the displayed and intended dates separately in redux, and switch between those depending on location)
  - disabling last 3 days in datepicker depending on location
  - adjusting both `from` and `to` when selecting custom dates feels like a bad interaction to me

### How To Test
- Visit V2 engagement details pages
  - Play around with the date filter
  - Check chart data and x axis ticks

### To Do
- [ ] Address any feedback

### DIFF
--- a/src/selectors/signals.js
+++ b/src/selectors/signals.js
@@ -116,15 +116,20 @@ export const selectEngagementRateByCohortDetails = createSelector(
     // Rename date key
     const normalizedHistory = _.get(match, 'history', []).map(({ dt: date, ...values }) => ({ date, ...values }));
 
+    // Engagement behavior charts do not contain data within the last 3 days (including today)
+    const shouldTrimTo = moment().diff(to, 'days') < 3;
+
     const filledHistory = fillByDate({
       dataSet: normalizedHistory,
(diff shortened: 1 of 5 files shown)


src/selectors/signals.js
```diff
@@ -108,6 +108,12 @@ export const selectEngagementRecencyDetails = createSelector(
   }
 );
 
+// Engagement behavior charts do not contain data within the last 3 days (including today)
+const excludeLastNDays = (date, n) =>
+  moment().diff(date, 'days') < n ? moment().subtract(3, 'days') : date;
```
bkemper: Minor change to use `n` as the number to subtract.

PR #957, opened 2019-04-17
EO-696 Properly namespace global reports css

jonambas:
### What Changed
- Global CSS classes in the summary and engagement charts are now nested under class wrappers.
- CSS should no longer leak between summary, engagement, and signals.
- No actual changes to signals charts, report charts were leaking CSS.

### How To Test

(1) Health Score Reference Lines
- Visit V2 health score dashboard or details page & hard refresh (Reference lines should have color)
- Visit summary report and return to health score  (Reference lines should **still** have color)

(2) Engagement Report
- Visit engagement report & hard refresh (axis lines should not be black)
- Edit: Wasn't part of the AC, but I noticed while fixing the first issue. I added some CSS here to make it more consistent with the summary chart

### To Do
- [ ] Address any feedback


jonambas:
What's rendered after visiting summary:
(Y axis tick colors and reference lines)
Screen Shot 2019-04-17 at 3 32 02 PM

What's expected:
Screen Shot 2019-04-17 at 3 32 17 PM

src/pages/reports/engagement/components/async/EngagementChart.scss
```diff
@@ -0,0 +1,35 @@
...
+    fill: color(gray, 5);
+    font-size: rem(13);
+    font-weight: 500;
+  }
+
+  .recharts-cartesian-axis-line {
+    stroke: color(gray, 7);
+  }
+
+  .recharts-text.recharts-cartesian-axis-tick-value {
```
kchu1337: Isn't this the same as line 12?

src/pages/reports/engagement/components/async/EngagementChart.js
```diff
@@ -6,7 +6,7 @@ import { Empty, PanelLoading } from 'src/components';
 import { formatNumber, formatPercent } from 'src/helpers/units';
 import { safeRate } from 'src/helpers/math';
 import PercentLabel from '../PercentLabel';
-import styles from './EngagementChart.module.scss';
+import './EngagementChart.scss';
```
bkemper: This still leaves us vulnerable to a leak.  We should avoid using global imports and instead locally import a class with global style like `EngagementChart.module.scss`.

PR #960, opened 2019-04-18
AC-711: Update 'Email' role to 'Templates'

PatrickSison:
User feedback shows that `email` is confusing for the role instead of actual intended functions of role. The role will likely be changed API side eventually to reflect changes.

### What Changed
 - User List Page now shoes 'templates' instead of 'email' role
 - RoleRadioGroup (found in User edit form and User invite) now shows 'Templates' instead of 'Email'
 - Pulled formatter to helper function

### How To Test
 - Check user invite form says `Templates` as role
 - Check user edit form says `Templates` instead of `email` as role
 - Check list page shows templates for new templates user or older `email` users

### To Do
- [ ] Address feedback


src/pages/users/ListPage.js
```diff
@@ -61,19 +61,12 @@ export class ListPage extends Component {
...
-    return role;
-  }
-
   // Do not allow current user to change their access/role or delete their account
   getRowData = (user) => {
     const { hasSubaccounts, isSubAccountReportingLive } = this.props;
     const data = [
       <User name={user.name} email={user.email} username={user.username} />,
-      this.formatRole(user.access),
+      formatRole(user.access),
```
ewandennis: Unfortunately, `Collection` uses the objects passed through the `rows` prop to provide sorting and filtering.

Sorting is broken because the objects contain `{role: 'email'}` so `email` users appear between `developer` and `heroku` users, but we now render the role column as `templates`.

The filter box is broken too. Its supports a `key:value` syntax and the user has no way to know that they need to type `role:email` to find `templates` users.

Users isn't a public API - maybe we should be fixing this at the API level. If we absolutely can't, we should transform the user list coming from the API, either in the users reducer on `LIST_USERS_SUCCESS`.

I expect we'd have to take a close look at all other places we use a user object to make sure we don't break those.

src/reducers/currentUser.js
```diff
@@ -12,7 +13,8 @@ const initialState = {
 export default (state = initialState, { type, payload, meta }) => {
   switch (type) {
     case 'GET_CURRENT_USER_SUCCESS':
-      return { ...state, ...payload, cookie_consent: !!payload.cookie_consent };
+
+      return { ...state, ...payload, access_level: formatRole(payload.access_level), cookie_consent: !!payload.cookie_consent };
```
ewandennis: Direct use of `access_level` looks limited to `helpers/conditions` and none use `email` or `templates so 👍.

src/reducers/users.js
```diff
@@ -41,7 +42,7 @@ export default (state = initialState, action) => {
 
       return fp.set(['entities', username], {
         ...state.entities[username],
-        access: access_level,
+        access: formatRole(access_level),
```
ewandennis: Direct use of `access` looks limited to `users/EditPage` so 👍.

src/helpers/userRoles.js
```diff
@@ -0,0 +1,21 @@
+import { ROLES } from 'src/constants';
+
+export function formatRole(role) {
+  switch (role) {
+    case ROLES.SUBACCOUNT_REPORTING:
```
ewandennis: This `reporting` transform is not reverse by `unformatRole`. Should it?

src/pages/users/ListPage.js
ewandennis: Since we format in the reducer, do we still need this?

src/helpers/userRoles.js
PatrickSison: No, but this is actually another issue where checks are made for subaccount_reporting, but now are immediately changed to reporting

src/constants/index.js
```diff
@@ -131,10 +131,20 @@ export const ANALYTICS_WHITELISTED_FORMS = [
 export const ROLES = {
   ADMIN: 'admin',
   DEVELOPER: 'developer',
-  EMAIL: 'email',
+  TEMPLATES: 'email',
```
bkemper: We should leave a comment, just so our future selves know why we left the id 'email'

src/pages/users/ListPage.js
```diff
@@ -73,7 +73,7 @@ export class ListPage extends Component {
     const { hasSubaccounts, isSubAccountReportingLive } = this.props;
     const data = [
       <User name={user.name} email={user.email} username={user.username} />,
-      this.formatRole(user.access),
```
bkemper: Looks like `formatRole` can be deleted now.

src/pages/users/ListPage.js
```diff
@@ -140,7 +140,7 @@ export class ListPage extends Component {
           rows={this.props.users}
           filterBox={{
             show: true,
-            keyMap: { role: 'access' },
+            keyMap: { role: 'roleLabel' },
```
bkemper: We can remove the `keyMap` now that the label is being provided with the dataset.

src/pages/users/components/RoleRadioGroup.js
```diff
@@ -24,8 +24,8 @@ const DEVELOPER_ROLE = {
 };
 
 const EMAIL_ROLE = {
-  label: <strong>Email</strong>,
-  value: ROLES.EMAIL,
+  label: <strong>Templates</strong>,
```
bkemper: Shouldn't we use the new labels map for these?

src/pages/users/tests/__snapshots__/ListPage.test.js.snap
```diff
@@ -306,7 +306,7 @@ Array [
     name="test-name"
     username="test-username"
   />,
-  "admin",
+  undefined,
```
bkemper: I'm assuming the list of users doesn't have a `roleLabel` prop.

src/reducers/users.js
```diff
@@ -8,7 +8,7 @@ export const initialState = {
   invite: {}
 };
 
-const reduceUsers = fp.reduce((result, user) => ({ ...result, [user.username]: user }), {});
+const reduceUsers = fp.reduce((result, user) => ({ ...result, [user.username]: { ...user }}), {});
```
bkemper: This might have been accidentally committed.

PatrickSison: Yup. Removing

src/pages/users/tests/__snapshots__/ListPage.test.js.snap
PatrickSison: Yeah, it doesn't. The selector adds the label. I could do the same check out where if there's no roleLabel, fall back on access

src/pages/users/ListPage.js
```diff
@@ -61,19 +60,12 @@ export class ListPage extends Component {
...
-    return role;
-  }
-
   // Do not allow current user to change their access/role or delete their account
   getRowData = (user) => {
     const { hasSubaccounts, isSubAccountReportingLive } = this.props;
     const data = [
       <User name={user.name} email={user.email} username={user.username} />,
-      this.formatRole(user.access),
+      user.roleLabel || user.access,
```
bkemper: This default value is not needed since the selector always provides a label.

PR #971, opened 2019-04-25
TR-1205: Do not lose focus on content tab after save

bkemper:
Refer to [TR-1205](https://jira.int.messagesystems.com/browse/TR-1205) for AC.

### What Changed

- After saving a template draft, we were refetching the template.  I believe this was a safeguard, but it was causing a full page reload which would cause the user to lose focus on the current content tab and their place in the content editor.  The fix is to not refetch after save.
- Bonus: Disabled delete and duplicate secondary actions when saving.

### How To Test

1. Open a draft template
2. Switch to "Text" or "AMP HTML" tab 
3. Click "Save as Draft" button

Expected Result: The primary and secondary should all disable temporarily while saving and user should not lose the focus on content tab.

### DIFF
--- a/src/pages/templates/EditPage.js
+++ b/src/pages/templates/EditPage.js
@@ -38,13 +38,13 @@ export default class EditPage extends Component {
     });
   }
 
-  handleSave = (values) => {
-    const { update, match, getDraft, showAlert, getTestData, subaccountId } = this.props;
-    return update(_.omit(values, 'published'), subaccountId).then(() => {
-      getDraft(match.params.id, subaccountId);
-      getTestData({ id: match.params.id, mode: 'draft' });
-      showAlert({ type: 'success', message: 'Template saved' });
-    });
+  handleSave = ({ published, ...values }) => { // must omit published value
+    const { showAlert, subaccountId, update } = this.props;
+
+    return (
+      update(values, subaccountId)
+        .then(() => showAlert({ type: 'success', message: 'Template saved' }))
+    );
   }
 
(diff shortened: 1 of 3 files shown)


src/pages/templates/EditPage.js
```diff
@@ -38,13 +38,13 @@ export default class EditPage extends Component {
...
-    return update(_.omit(values, 'published'), subaccountId).then(() => {
-      getDraft(match.params.id, subaccountId);
-      getTestData({ id: match.params.id, mode: 'draft' });
-      showAlert({ type: 'success', message: 'Template saved' });
-    });
+  handleSave = ({ published, ...values }) => { // must omit published value
+    const { showAlert, subaccountId, update } = this.props;
+
+    return (
+      update(values, subaccountId)
```
jonambas: Page title won't change if name is updated.

This is fine, but you might need to replace the stored template data with form data on update.

jonambas: 1) I think this is a bigger problem we need to solve on other edit pages as well. 

2) I don't think AC is accurate, focus is always lost because saving requires a click outside (we don't do cmd + s). What you can do is save a ref to tab index, and after update (and refetch) set tab index to the ref.

bkemper: To summarize our discussion yesterday, updating the page title is not a problem since it only displays id which is provided by the request url and the AC is to not lose focus on the tab not the editor itself.

bkemper:
Rebased.

PR #979, opened 2019-04-30
SA-605 Fix byte conversion from 1000 to 1024

kchu1337:
### What Changed
 - Changed Byte conversion factor from 1000 to 1024. 
 - Memoized Y-axis-formatter function because it is being called every mouseover. 
 - Changed the chart CSS to improve readability.

### How To Test
 - Go to reports-> summary, select `Delivery Message Volume` as a metric. Compare the chart values to the values returned in the JSON response. It should be dividing by 1024 instead of 1000. 
 - Change various time intervals to make sure graph is legible. Also hover over various chart data points. 

### To Do
- [x] Address feedback

### DIFF
--- a/src/helpers/chart.js
+++ b/src/helpers/chart.js
@@ -42,8 +42,7 @@ function getLineChartFormatters(precision) {
   return formatters;
 }
 
-const formatYAxisPercent = _.memoize((v) => `${roundToPlaces(v, v < 1 ? 3 : 1)}%`
-);
+const formatYAxisPercent = _.memoize((v) => `${roundToPlaces(v, v < 1 ? 3 : 1)}%`);
 
 export {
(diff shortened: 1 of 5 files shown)


src/helpers/chart.js
```diff
@@ -42,9 +42,8 @@ function getLineChartFormatters(precision) {
   return formatters;
 }
 
-function formatYAxisPercent(v) {
-  return `${roundToPlaces(v, v < 1 ? 3 : 1)}%`;
-}
+const formatYAxisPercent = _.memoize((v) => `${roundToPlaces(v, v < 1 ? 3 : 1)}%`
+);
```
PatrickSison: Sorry, smallest nitpick ever. Can you move closing parentheses up?

PR #980, opened 2019-04-30
EO-714 Add mb Provider to Alerts

21042jim:
### What Changed
 - Added onFocus event support to MultiFacet component
 - Added mb provider groups to typeahead facet

### How To Test

### Things to Verify
- Where to verify:
  - /alerts-new Create and Edit pages facet name/values
- What to verify:
  - onFocus support for ip pools and mb provider
  - mb provider group typeahead

### To Do
- [ ] Address any feedback


src/pages/alerts-new/components/MultiFacet.js
```diff
@@ -16,59 +16,62 @@ import _ from 'lodash';
...
-      matches: matches.slice(begin, maxNumberOfResults + begin)
+      matches: filterMatch
     });
   }, 300);
 
+  handleFocus = (stateChange) => {
+    const stateType = _.get(stateChange, 'type', '');
+
+    if (stateType === '__autocomplete_change_input__' || stateType === '__autocomplete_unknown__') {
+      this.setState({
```
kchu1337: Might want to insert a `this.updateMatches.cancel();`. If you type something, and then delete it <300ms later, it will start the debounced `updateMatches`, then change matches here, then change matches from the debounced `updateMatches`.

src/pages/alerts-new/components/MultiFacet.js
```diff
@@ -16,59 +16,62 @@ import _ from 'lodash';
...
     this.updateMatches.cancel();
   }
 
-  handleInputValueChange = (value) => {
-    this.updateMatches(value);
-  }
-
   handleStateChange = (changes, downshift) => {
-    // Highlights first item in list by default
+    const input = _.get(changes, 'inputValue', '');
```
kchu1337: Can also use native destructuring like `const { inputValue: input = '' } = changes`. But it's up to your personal preference.

src/pages/alerts-new/components/MultiFacet.js
```diff
@@ -16,59 +16,62 @@ import _ from 'lodash';
...
+    const filterMatch = (updateMatches[0] === nextValue) ? _.filter(updateMatches, nextValue) : updateMatches;
 
     this.setState({
-      matches: matches.slice(begin, maxNumberOfResults + begin)
+      matches: filterMatch
     });
   }, 300);
 
+  handleFocus = (stateChange) => {
+    const stateType = _.get(stateChange, 'type', '');
```
kchu1337: Same here with destructuring

src/pages/alerts-new/components/AlertForm.js
```diff
@@ -174,8 +176,8 @@ export class AlertForm extends Component {
                         validate={required}
                       />
                     }
-                    component={facet_name === 'ip_pool' ? MultiFacetWrapper : TextFieldWrapper}
-                    items={(facet_name === 'ip_pool') ? ipPools : null}
+                    component={(facet_name === 'ip_pool' || facet_name === 'mb_provider') ? MultiFacetWrapper : TextFieldWrapper}
+                    items={facet_name === 'ip_pool' ? _.map(ipPools, 'id') : facet_name === 'mb_provider' ? _.keys(MB_PROVIDERS) : null}
```
kchu1337: So this shows keys like 'yahoo_uk` rather then 'Yahoo UK'. Not sure if intended.

src/pages/alerts-new/components/MultiFacet.js
```diff
@@ -16,59 +16,62 @@ import _ from 'lodash';
...
-    const matches = sortMatch(
-      itemsArray,
+    const updateMatches = sortMatch(
+      this.props.items,
       nextValue
     );
+
     // note, exclude the first match if it is the current input value
-    const begin = matches[0] === nextValue ? 1 : 0;
+    const filterMatch = (updateMatches[0] === nextValue) ? _.filter(updateMatches, nextValue) : updateMatches;
```
kchu1337: Could also do `updateMatches.shift()`, but it's up to your preference. Although `_.filter`  can remove more than 1 if there are duplicates.

src/pages/alerts-new/components/MultiFacet.js
21042jim: i think i prefer referencing input as opposed to inputValue.input everywhere so i will probably just leave it

src/pages/alerts-new/components/AlertForm.js
21042jim: this came up in EO-713 as well. we are going to open a separate ticket to address it. ip pools does not have a friendly name like mb provider, so we will need to either remove redux or hack the mapping together from friendly name to id in the action before it is sent out.

src/pages/alerts-new/components/MultiFacet.js
21042jim: doing updateMatches.shift actually caused an error in InputMenu. Apparently updateMatches.shift() returns the string element instead of an array with the single string element. So the InputMenu component could no longer do a map on items.

src/pages/alerts-new/components/MultiFacet.js
21042jim: added this, thanks.

src/pages/alerts-new/components/tests/AlertForm.test.js
```diff
@@ -58,34 +58,41 @@ describe('Alert Form Component', () => {
...
+      wrapper.setProps({ facet_name: 'ip_pool', facet_value: 'blah' });
       expect(wrapper.find({ name: 'facet_value' }).prop('component').name).toEqual('MultiFacetWrapper');
       expect(wrapper.find({ name: 'facet_value' }).props()).toMatchSnapshot();
     });
 
-    it('should clear facet_value and validation when facet_name is set to ALL and signals threshold', () => {
-      wrapper.setProps({ alert_metric: 'signals_health_threshold', facet_name: 'ALL', facet_value: 'something' });
+    it('should clear facet_value and validation when facet_name is set to master and signals threshold', () => {
+      wrapper.setProps({ facet_value: 'something' });
       expect(wrapper).toMatchSnapshot();
```
kchu1337: You can mock the `change` prop and check that it is called when `facet_name` changes.

PR #987, opened 2019-05-06
EO-737 Add friendly label to tables for Mailbox Providers

kchu1337:
### What Changed
 - Added a function that converts the id of the mailbox provider to a more friendly label.
 - Moved the Mailbox Providers constants to the generic constants file.

### How To Test
- Log into appteam or account 108 for data.
- Go to signals Health Score or Engagement Recency.
- Change facets to `Mailbox Providers`. Check that the cells in the table show the friendly name. 

### To Do
- [x] Review Feedback
- [ ] ~Maybe add friendly page header labels~


src/constants/index.js
```diff
@@ -139,3 +139,23 @@ export const ROLES = {
 };
 
 export const SUBACCOUNT_ROLES = [ROLES.SUBACCOUNT_REPORTING];
+
+export const MB_PROVIDERS = {
```
bkemper: I think avoiding acronyms and abbreviations will help readability for next engineer — `MAILBOX_PROVIDERS`.

src/pages/signals/components/dataCells/FacetDataCell.js
```diff
@@ -1,11 +1,12 @@
 import React from 'react';
 import PageLink from 'src/components/pageLink';
 import { setSubaccountQuery } from 'src/helpers/subaccounts';
+import { MB_PROVIDERS } from 'src/constants';
 import styles from './DataCell.module.scss';
 import classnames from 'classnames';
 
 const FacetDataCell = ({ dimension, facet, id, name, subaccountId, truncate }) => {
-  let label = id;
+  let label = friendlyLabel(facet, id);
```
bkemper: The list of conditions to write a friendly label is growing, but I don't think a switch is the solution.  If it was, both the master account and name with id cases could have easily been refactored into it.  Maybe a new component?  Label after all is used as a child to PageLink.

src/pages/signals/components/dataCells/FacetDataCell.js
```diff
@@ -1,32 +1,31 @@
...
+          <div className={classnames(styles.PaddedCell, truncate && styles.OverflowCell)}>
+            Master & All Subaccounts
+          </div>
+        );
+      } else if (id === 0) {
+        label = 'Master Account';
+      }
+      break;
+    case 'mb_provider':
+      label = MAILBOX_PROVIDERS[id];
```
bkemper: What if the id doesn't match a mailbox provider?  Shouldn't we show id instead of nothing?

PR #993, opened 2019-05-09
AC-743: Removed obsolete link

PatrickSison:
[AC-743](https://jira.int.messagesystems.com/browse/AC-743)
### What Changed
 - Removed support ticket link 

### How To Test
 - Chrome dev tools, block request for `api.sparkpost.test/api/v1/account` or `*/api/v1/account`
 - Attempt to create account
- Message should read `Something went wrong. Please try again in a few minutes.`

### To Do
- [ ] Address feedback

### DIFF
--- a/src/constants/index.js
+++ b/src/constants/index.js
@@ -154,4 +154,4 @@ export const PLAN_TIERS = {
   premier: 'Premier'
 };
 
-export * from './users';
+export { ROLES, SUBACCOUNT_ROLES, ROLE_LABELS } from './users';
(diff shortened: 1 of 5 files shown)


PatrickSison:
Links are removed and we'll add more actionable content after talking w/ support

src/constants/index.js
```diff
@@ -175,3 +153,5 @@ export const PLAN_TIERS = {
   starter: 'Starter',
   premier: 'Premier'
 };
+
+export * from './users';
```
bkemper: To avoid an overwrite we should be explicit here.

export { ROLES, ROLE_LABELS, SUBACCOUNT_ROLES } from './users';

PatrickSison: 👍

PR #995, opened 2019-05-09
TR-1284: Create next template editor page

bkemper:
Refer to [TR-1284](https://jira.int.messagesystems.com/browse/TR-1284) for AC.  Please note this is a subtask.

### What Changed

- New page for next template editor

Screen Shot 2019-05-10 at 5 34 15 PM

### How To Test
 - Visit http://app.sparkpost.test/templates/edit/amp/next


bkemper:
@rajumsys This PR is ready for your 👁 👁 when you have time.

src/context/RouterContext.js
```diff
@@ -0,0 +1,26 @@
+import React, { createContext, useMemo } from 'react';
+import { withRouter } from 'react-router';
+import qs from 'qs';
+
+// see, https://github.com/ReactTraining/react-router/issues/6430
+// note, maintainers do not recommend using __RouterContext from react-router
+const RouterContext = createContext();
+
+export const ProviderComponent = ({ children, ...routerProps }) => {
+  const value = useMemo(() => ({
```
rajumsys: is `useMemo` useful w/o args?

src/pages/templates/EditAndPreviewPage.js
```diff
@@ -0,0 +1,25 @@
...
+import { Grid, Page } from '@sparkpost/matchbox';
+import useRouter from 'src/hooks/useRouter';
+
+const EditAndPreviewPage = () => {
+  const router = useRouter();
+
+  return (
+    <Page
+      breadcrumbAction={{ content: 'Back Link', to: '/templates' }}
+      title={router.requestParams.id}
```
rajumsys: `.name` (show template name)

src/pages/templates/EditAndPreviewPage.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import { Grid, Page } from '@sparkpost/matchbox';
+import useRouter from 'src/hooks/useRouter';
+
+const EditAndPreviewPage = () => {
+  const router = useRouter();
```
rajumsys: Is this a pattern we intend to implement going forward?

src/pages/templates/EditAndPreviewPage.js
bkemper: In TR-1284, I fetch the template and set the title to the template name.

src/pages/templates/EditAndPreviewPage.js
bkemper: I think so.  The benefit verse the HOC is readability and a single place to abstract common router needs, such as, requestParams.  Let's see how it performs on this prototype.

src/context/RouterContext.js
bkemper: I think Kent said it best.

> Most of the time, this wont be a huge deal, especially if your context value doesn't change very often, but it's normally pretty easy to memoize the value with React.useMemo anyway so may as well do it all the time for your context value and avoid the problem.
> — https://kentcdodds.com/blog/how-to-optimize-your-context-value

src/pages/templates/EditAndPreviewPage.js
rajumsys: ah, that's right. name isn't available yet.

src/context/RouterContext.js
rajumsys: haha. it's ok

_thinking about optimization w/o measuring :)_

src/hooks/useRouter/useRouter.js
```diff
@@ -0,0 +1,9 @@
+import { useContext } from 'react';
+import RouterContext from 'src/context/RouterContext';
+
+// see, https://github.com/ReactTraining/react-router/issues/6430
+// watch, https://www.youtube.com/watch?v=3yiialslPbc
+// note, custom hooks like this one are recommended to make mocking in tests easier
+const useRouter = () => useContext(RouterContext);
```
rajumsys: a quick question: could we just use the context in the component? what's the benefit of using this custom hook when it doesn't use any other hooks/effects?

PR #996, opened 2019-05-10
EO-664 Release Signals V2

jonambas:
### What Changed
- Removes V1 Overview Page
- Removes V2 access control checks
- Removes summary table headers that were only used in V1

### How To Test
- Verify V2 signals nav items, without the flag
- Verify engagement recency tabs renders without flag
- Verify back links on details pages

### To Do
- [ ] Address any feedback


src/config/routes.js
```diff
@@ -230,40 +230,30 @@ const routes = [
...
-    layout: App,
-    title: 'Signals',
-    supportDocSearch: 'signals'
-  },
-  // Signals V2 Dashboards
   {
     path: '/signals/health-score',
     component: signals.HealthScoreDashboard,
-    condition: all(hasGrants('signals/manage'), hasUiOption('feature_signals_v2')),
+    condition: hasGrants('signals/manage'),
```
rajumsys: feels like we should now break routes into multiple modules. too big :)

jonambas:
@rajumsys sometimes locally it will 401 if you've clicked a link to a details page with a subaccount that doesn't exist (appteam data is pulled from another account)

Good call on the redirect, i will add one as a precaution

PR #997, opened 2019-05-10
EO-754 Add recommended actions messages for Signals Engagement Rate by cohort

kchu1337:
### What Changed
 - Added recommended actions messages for engagement rate by cohort. Values from here: https://docs.google.com/spreadsheets/d/1AOcX1eQ6flweFV0O3vFTaYyG0h_QC5e7LHLvS7Mjdsg/edit#gid=92043718.
The total engagement rate error message is unique because it requires a facet specific link. 
- Added component that generates the recommended actions messages. 
- Added tests.

### How To Test
- Log into appteam or account 108 for data ( production account 108 works best because it has actual data)
- Visit v2 Engagement Recency dashboard
- Click on one of the accounts
- Click on `Engagement Rate` tab.
- Check that all recommendations are the correct ones. 
### To Do
- [ ] Address review feedback


src/pages/signals/constants/engagementRateByCohortContent.js
```diff
@@ -0,0 +1,109 @@
...
+//Separated because it needs a custom link
+export const contentTotalEngagement = {
+  condition: ({ p_total_eng }) => p_total_eng < 0.15,
+  content: (
+  <>
+    An overall low engagement rate may indicate you emails are being sent to the spam folder. Drill into your Health Score to find the issue.
+  </>
+  ),
+  type: 'bad',
+  link: (sid) => `https://www.sparkpost.com/signals/signals/health-score/sid/${sid}`
```
jonambas: This should go in-app right? Would need to pass a new flag in here that uses a normal page link instead of ExternalLink

kchu1337: There's actually a lot more than that wrong with it lol. But I added a kinda weird way to get around the ExternalLink vs Link.

src/pages/signals/components/Actions.js
```diff
@@ -9,8 +9,9 @@ import styles from './Actions.module.scss';
 const Action = ({ content, link, type = 'bad' }) => {
```
jonambas: You should have access to `facet` and `facetId` here, along with any keys specified in `engagementRateByCohortContent.js`

src/pages/signals/components/Actions.js
```diff
@@ -9,8 +9,9 @@ import styles from './Actions.module.scss';
...
   let iconMarkup;
 
-  const linkMarkup = link && (
-    <ExternalLink to={link} className={styles.Link}>Learn More</ExternalLink>
+  const linkMarkup = link && (React.isValidElement(link)
+    ? link
```
jonambas: Let's use the `PageLink` component here

src/pages/signals/components/Actions.js
```diff
@@ -9,8 +9,9 @@ import styles from './Actions.module.scss';
 const Action = ({ content, link, type = 'bad' }) => {
   let iconMarkup;
 
-  const linkMarkup = link && (
-    <ExternalLink to={link} className={styles.Link}>Learn More</ExternalLink>
+  const linkMarkup = link && (React.isValidElement(link)
```
jonambas: Check a new flag instead, something like `internal: true`

src/pages/signals/constants/engagementRateByCohortContent.js
```diff
@@ -0,0 +1,115 @@
...
+//Separated because it needs a custom link
+export const contentTotalEngagement = {
+  condition: ({ p_total_eng }) => p_total_eng < 0.15,
+  content: (
+  <>
+    An overall low engagement rate may indicate you emails are being sent to the spam folder. Drill into your Health Score to find the issue.
+  </>
+  ),
+  type: 'bad',
+  link: (facet, facetId) => (
```
jonambas: Like we discussed, let's use a flag to make this internal. and we can keep `link` as a function:
```js
link: (facet, facetId) => `/signals/health-score/${facet}/${facetId}`,
internal: true

//
const linkMarkup = link && internal
  ? <PageLink to={link(facet, facetId)}>Learn More</PageLink>
  : <ExternalLink to={link} className={styles.Link}>Learn More</ExternalLink>
```

src/pages/signals/constants/engagementRateByCohortContent.js
```diff
@@ -0,0 +1,115 @@
...
+    <Link to={`/signals/health-score/${facet}/${facetId}`} className={styles.Link}>
+      Learn More
```
jonambas: I'd double check with Avi about this text, should this be "View Health Score"?

src/pages/signals/constants/engagementRateByCohortContent.js
```diff
@@ -0,0 +1,115 @@
...
+  <>
+    You have healthy engagement rates! Keep up the great work!
+  </>
+  ),
+  type: 'good'
+};
+
+//Separated because it needs a custom link
+export const contentTotalEngagement = {
+  condition: ({ p_total_eng }) => p_total_eng < 0.15,
```
jonambas: Doc says `5%`

kchu1337: You're right. I probably copied it and forgot to change it. Thanks

src/pages/signals/constants/engagementRateByCohortContent.js
kchu1337: I would rather keep link as a string so that the actions component always has link as a string. If that is what you are referring to?

jonambas: Yeah you could do that, and `link` wouldn't include facet and facet id? just `/signals/health-score/`?

kchu1337: I was thinking of leaving the total engagement as it's separate conditional. But are you saying you would want it inside the content array and pass link as a function? If you anticipate the link in `Actions` to be internal in other situations, then I can make that change but I do not want to add code to the general use `Actions` component for a specific use-case.

jonambas: Hmm, yeah that's fair. I just wanna make sure we keep as much react code as we can out of the content map, otherwise its no different from creating a custom component.

I think you should make `<Action>` support either (1) internal and external links OR (2) internal links and custom link component

src/pages/signals/constants/engagementRateByCohortContent.js
jonambas: yep `View Health Score`

src/pages/signals/components/actionContent/EngagementRateByCohortActions.js
```diff
@@ -5,12 +5,13 @@ import { content, contentTotalEngagement, contentGood } from '../../constants/en
...
   let actions = [];
 
-  //First check total engagement because it needs to pass along subaccount id to generate a link
+  //First check total engagement because it needs to pass along facet info to generate a link
   if (contentTotalEngagement.condition(engagementByCohort)) {
     actions.push({
       content: contentTotalEngagement.content,
       type: contentTotalEngagement.type,
-      link: contentTotalEngagement.link(facet, facetId)
+      link: contentTotalEngagement.link(facet, facetId),
```
jonambas: 👍

PR #1008, opened 2019-05-15
EO-764 Adjust signals subtitles, Remove other charts header

jonambas:
### What Changed
- Removed title from OtherChartsHeader, renamed to Divider
- Adjusted page subtitles to be more clear

### How To Test
- Visit all signals details pages
- Verify page subtitles with different facets, and with/without a subaccount

### To Do
- [ ] Address feedback


src/helpers/signals.js
```diff
@@ -14,23 +15,34 @@ const translateSubaccount = (id) => {
   return `Subaccount ${id}`;
 };
 
-export const getFriendlyTitle = ({ prefix, facet, facetId, subaccountId, dimension = true }) => {
-  if (!prefix) {
-    return null;
+const translateFacet = (facet, id) => {
+  switch (facet) {
```
PatrickSison: Not sure what the styling should be, but should the ID have some separation from the facet label? Seeing `IP Pool subaccounts` kind of flows as one thing and seems a bit weird. Might be unnecessary fix

jonambas: Yeah, wasn't sure either. The idea was to keep it similar to what were doing with subaccounts - "Subaccount {id}"

PR #1009, opened 2019-05-15
AC-735: Warn account they are on an old plan

PatrickSison:
### What Changed
 - Message shows in `/account/billing/plan` confirmation (right side) if on an old plan changing to a newer plan.

### How To Test
 - On an account that has an old plan, check that plans selection shows message
 - On an account that has an old plan, check that when no new plan selected, does not show message
 - On any account that has a new plan, check that message never shows.

### To Do
- [ ] Address feedback

### DIFF
--- a/src/pages/billing/components/Confirmation.js
+++ b/src/pages/billing/components/Confirmation.js
@@ -58,7 +58,7 @@ export class Confirmation extends React.Component {
           <Warning className={styles.icon} size={32}/>
         </div>
         <div className={styles.content}>The current plan you are on is no longer available.
-      If you switch to a different plan, you will not be able to switch back to this one.</div>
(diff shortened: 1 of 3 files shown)


src/pages/billing/components/Confirmation.js
```diff
@@ -58,7 +58,7 @@ export class Confirmation extends React.Component {
           <Warning className={styles.icon} size={32}/>
         </div>
         <div className={styles.content}>The current plan you are on is no longer available.
-      If you switch to a different plan, you will not be able to switch back to this one.</div>
+      If you switch to the new plan, you will not be able to switch back to your current one.</div>
```
rajumsys: can we say `if you switch to the selected plan....`

rajumsys:
LGTM

PR #1014, opened 2019-05-16
EO-766 Refactored tests and fixed file names

kchu1337:
### What Changed
 - Changed file names to match formatting standards in our codebase. 
 - Changed some ComplaintByCohortActions tests to be assertion tests rather than snapshots. 
 - Changed EngagementRateByCohortActions test to only have full engagement recency on the specific cohort rather than all cohorts. This is to catch errors for engagement recency conditionals. 

### How To Test
 - Npm run test
 - Nothing customer facing, just code cleanup


src/pages/signals/components/actionContent/tests/ComplaintsByCohortActions.test.js
```diff
@@ -5,98 +5,152 @@ import ComplaintsByCohortActions from '../ComplaintsByCohortActions';
...
-    'render Recently Engaged bad value': {
-      ...fullEngagementRecency,
-      p_14d_fbl: 0.0051
+    'Total complaint rate cohort': {
+      complaintsByCohort: { p_total_fbl: 0.0051 },
+      recencyByCohort: fullEngagementRecency
+    }
+  });
+
+  cases('should render warning value actions for', ({ complaintsByCohort, recencyByCohort }) => {
```
jonambas: You could combine these cases with the cases above if you want. But i'm okay if you don't want to refactor, since the other action tests follow this same pattern.

eg:
```js
cases('should render actions for', ({ complaintsByCohort, recencyByCohort, expected }) => {
    wrapper = shallow(...);
    expect(wrapper.prop('actions')).toHaveLength(1);
    expect(wrapper.prop('actions')[0].type).toEqual(expected);
}, {
  'bad new cohort value': {
      complaintsByCohort: { p_new_fbl: 0.0026 },
      recencyByCohort: { c_new },
      expected: 'bad'
  },
  'warning new cohort value': {
      complaintsByCohort: { p_new_fbl: 0.0011 },
      recencyByCohort: { c_new },
      expected: 'warning'
  }
}
```

I think you could also pull in the good case test as well

PR #1016, opened 2019-05-17
UX-80 Signals Prototype

jonambas:
### Whats changed
- Reverts SparkPost/2web2ui#1015
- Adds a default component edge instead of a button when no components selected

### To verify
- Verify no page load without UI flag `prototype_signals_details`
- Verify prototype loads with flag (use appteam)
- Verify code splitting

### To do
- [ ] Should I not do this in a revert branch?


jonambas:
Kapture 2019-05-17 at 12 41 32

jonambas:
Interviews completed!

PR #1018, opened 2019-05-17
EO-767 Fix alerts page bugs with populating values and request body

kchu1337:
### What Changed
 - renamed `master and all subaccounts` in alertPage from `ALL` to `all`. `getAssignTo.js` has the value as `all`. 
- Fixed a bug where facet value initial value gets overwritten.
- Prevents unnecessary params from being passed in the request body. 
- Changed signals health WOW/DOD to always be positive and greater than.

### How To Test
 - Go to alerts page.
 - Check that alerts data is populating properly by clicking on an alert with a single subaccount and a facet. 
 - Check that an alert with `All master and subaccounts` is properly initialized. 
 - When creating/updating an alert, 
       - Monthly sending limit alerts do not pass facet_name, facet_value or alert_subaccount in the request body
       - `All master and subaccounts` alerts do not pass facet_name or facet_value in the request body. 
  - Signals WOW, Signals DOD, and Monthly Sending Limit only pass on `gt` as the comparator value.

### To Do
- [x] Maybe implement the changes in EO-770. If not, will create a new EO ticket to implement that.


src/pages/alerts/helpers/formatActionData.js
```diff
@@ -1,28 +1,47 @@
 import _ from 'lodash';
```
bkemper: This is a general comment, but we should not use lodash unless we need it — there are native equivalents for `map`, `split`, and `trim`.

src/pages/alerts/helpers/formatActionData.js
```diff
@@ -1,28 +1,47 @@
...
 
 export default function (values) {
-  const { email_addresses, alert_metric, threshold, assignTo, subaccount = '' } = values;
+  const formattedValues = { ...values }; //So values stays immutable
```
bkemper: This function is very difficult to read.  In the future, I highly recommend a more functional approach of composing values together instead of mutating a single object.

PR #1019, opened 2019-05-20
FE-662: sign up flow func tests

ewandennis:
### What Changed
 - Added functional tests for pages `/join` and the `/onboarding/*` pages
 - Removed the old `__integration__/registration.test.js` since these new tests supersede them.

Note: I did _not_ add tests for page transitions that are just links, such as skipping sending domain setup. Instead I focussed on page functionality.

### How To Test
 - `npm test -- __func__/signup`

### To Do
- [x] Address feedback


src/__func__/signup/choosePaidPlan.test.js
```diff
@@ -0,0 +1,12 @@
+import signupFlow, { freePlanPredicate } from './helpers';
+
+test('Signup flow: choose paid plan', async () => {
+  const page = await signupFlow();
+  await page.signup();
+  await page.choosePlan({ planFilter: (plan) => !freePlanPredicate(plan), submit: false });
+  await page.fillBillingForm();
+  await page.simulate('form', 'submit');
```
rajumsys: we're testing here form is submitted which is nice. but should we also find the submit button and simulate click which will cover clicking the button and submit of form as well?

src/__func__/signup/setupRest.test.js
```diff
@@ -0,0 +1,6 @@
+import mountRoute from 'src/__testHelpers__/mountRoute';
+
+test('Signup flow: REST setup', async () => {
+  const page = await mountRoute('/onboarding/email/api');
+  expect(page.mockApiCalls).toMatchSnapshot();
```
rajumsys: Unlike other tests, this isn't interacting beyond visiting the page. how about if we 
- Visit `/onboarding/email`
- Simulate click on `REST` 
- Assert route and api calls?

src/__func__/signup/setupSmtp.test.js
```diff
@@ -0,0 +1,6 @@
+import mountRoute from 'src/__testHelpers__/mountRoute';
+
+test('Signup flow: SMTP setup', async () => {
+  const page = await mountRoute('/onboarding/email/smtp');
+  expect(page.mockApiCalls).toMatchSnapshot();
```
rajumsys: Similarly here
- Visit `/onboarding/email`
- Click on 'SMTP'
- Assert

src/__func__/signup/setupRest.test.js
ewandennis: Yes, thats what I mean by not testing links. The main reason is that `jsdom` does not support navigation. That might seem like a big limitation but the inability to verify that the browser can navigate to a route using an anchor tag doesn't sound too big a loss. We could assert that a link exists (which I would do in a component test) or that a route functions (maybe a functional test).

So maybe we need to assert that a path exists from each link-driven route to the expected next route...?

src/__func__/signup/choosePaidPlan.test.js
ewandennis: Hmm. Maybe we should _only_ be clicking buttons. Submitting the form directly isn't very user-centric, is it?

ewandennis: Update: form submit buttons are another part of `jsdom`'s missing navigation capability.

We'll stick with a form submit event for now. We could also assert existence of a submit button I imagine.

I'll also look for a way to enable `jsdom`'s "unimplemented feature" warnings.

ewandennis: `jsdom` doesn't warn when you attempt to navigate through anchors :( It will warn if you use `window.location` however.

src/__func__/signup/setupRest.test.js
rajumsys: >So maybe we need to assert that a path exists from each link-driven route to the expected next route...?

I'm fine with it given the routing limitation

src/__testHelpers__/mountRoute.js
```diff
@@ -31,7 +31,7 @@ export default async function mountRoute(route, { authenticated = true } = {}) {
   const axiosMock = axios.create();
 
   if (authenticated) {
-    await store.dispatch(authenticate('test-username', 'password'));
+    await store.dispatch(authenticate('test-username', 'test-password'));
```
rajumsys: > fix a clash with upstream API call mocking
(your commit message)

what it means and what's the issue here?

ewandennis: You saw the the Travis `pr` check was failing, yeah? That was because a previously merged PR changed the `axios` mock response for `/api/v1/authenticate` and I failed at the time to change `mountRoute()` to match. 

So when Travis merged this branch into master, the tests which rely upon `mountRoute()` doing authentication for them failed.

PR #1020, opened 2019-05-20
TR-1208 Create Recipients List

rajumsys:
### What Changed
 - Moved redux form decoration to Create/Edit page (parent component) so after upload recipients can be managed from parent component
- Update create page according to mock
- Update edit page briefly (more to do on relevant ticket)
Screen Shot 2019-05-20 at 10 24 14 AM
- Unlike past, CSV will be uploaded as soon as file is selected

### How/What to test
- Create a new recipient list with upload
  -- Upon upload you should see the recipients preview below
- Edit list should work as usual 

### Notes
There are some improvements needed on Edit page, which i'll do in the relevant ticket that's coming next.

### DIFF
--- a/src/pages/recipientLists/CreatePage.js
+++ b/src/pages/recipientLists/CreatePage.js
@@ -1,7 +1,7 @@
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
-
 import { Link, withRouter } from 'react-router-dom';
+import { change, getFormValues, reduxForm } from 'redux-form';
 
 import { Page } from '@sparkpost/matchbox';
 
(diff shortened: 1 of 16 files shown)


rajumsys:
@bkemper This is ready for your 👀

src/pages/recipientLists/components/RecipientsCollection.js
```diff
@@ -0,0 +1,49 @@
...
+      getRowData={this.getRowData}
+      pagination={true}
+    />);
+  };
+
+  render() {
+    const { hasCsv } = this.props;
+
+    return (<Panel title='Preview Contacts'>
+      {!hasCsv && this.renderNoCsv()}
```
bkemper: Should we use a ternary?

src/pages/recipientLists/components/RecipientsCollection.js
```diff
@@ -0,0 +1,49 @@
...
+      rows={recipients}
+      getRowData={this.getRowData}
+      pagination={true}
+    />);
+  };
+
+  render() {
+    const { hasCsv } = this.props;
+
+    return (<Panel title='Preview Contacts'>
```
bkemper: Do we need a Panel?  This might make it hard to add the filter box.

src/pages/recipientLists/CreatePage.js
```diff
@@ -1,47 +1,96 @@
...
+        <RecipientListForm formName={formName}/>
+        <RecipientsCollection hasCsv={!!csv} recipients={recipients}/>
+      </form>
     </Page>;
   }
 }
 
+const valueSelector = formValueSelector(formName);
+const mapStateToProps = (state, props) => ({
+  initialValues: {},
```
bkemper: Probably can be removed.

src/pages/recipientLists/components/tests/RecipientsCollection.test.js
```diff
@@ -0,0 +1,42 @@
...
+    props = {
+      recipients: []
+    };
+
+    wrapper = shallow(<RecipientsCollection {...props} />);
+  });
+
+  it('renders correctly when hasCsv is false', () => {
+    wrapper.setProps({ hasCsv: false });
+    expect(wrapper.find('p').text()).toEqual('Once you upload a CSV, recipients will be previewed here.');
```
bkemper: `toHaveText` is shorthand matcher for `.text()).toEqual(`

src/pages/recipientLists/components/RecipientListForm.js
```diff
@@ -1,59 +1,37 @@
...
   }
 
+  // Fills in identifier field based on Name
+  handleIdFill = (e) => {
+    const { editMode, autofill, formName } = this.props;
+    if (editMode) {
+      return;
+    }
+
+    autofill(formName, 'id', slugify(e.target.value));
```
bkemper: 👍 for use of `autofill`

bkemper:
@rajumsys I'm finished my initial pass.  Looks good.  I'll wait for your replies to my comments.

src/pages/recipientLists/components/RecipientsCollection.js
rajumsys: Which element you suggest to use? I couldn't match design with anything else!

rajumsys:
> The "Identifier" field needs help text to let the user know that this field cannot be changed after creation.

Good point. I'll add that. 

> The original request was to redirect users to the list page after creation — assuming they have already prepped the list enough and don't need to make additional changes.

I've intentionally modified that. Regardless of what's intended, redirecting lo list page has least value to me. Also our normal pattern is that we redirected to the newly created resource page. Let me know if you prefer the original behavior. 

> The success alert should also include the number of users that were rejected if any.

👍 

> The "Save As CSV" button should produce a file that can be uploaded.

I wanted to hide the button but forgot. I don't see any value of it, at least on create page as we just uploaded a CSV. 

> Sort on the collection does not work.

Gonna fix it.

src/pages/recipientLists/components/RecipientsCollection.js
bkemper: I was just thinking something like...

```
return (
  <>
    <h1>Preview Contacts</h1>
    <TableCollection />
  </>
);
```

bkemper:
> I wanted to hide the button but forgot. I don't see any value of it, at least on create page as we just uploaded a CSV.

Yep, and I forgot that we have a separate ticket to address downloading.  I'll add a note to it.

src/pages/recipientLists/components/RecipientsCollection.js
rajumsys: Ok, I've modified to be tablecollection w/o panel but those empty messages with a panel. 
_also added some test coverage_

src/pages/recipientLists/CreatePage.js
```diff
@@ -1,47 +1,100 @@
...
         Component: Link,
-        to: '/lists/recipient-lists' }}>
-
-      <RecipientListForm onSubmit={this.createRecipientList} />
+        to: '/lists/recipient-lists'
+      }}>
 
+      <form onSubmit={handleSubmit(this.createRecipientsList)}>
+        <RecipientListForm formName={formName}/>
+        {csv && <RecipientsCollection hasCsv={!!csv} recipients={recipients}/>}
```
bkemper: I think `hasCsv` can be removed.

src/pages/recipientLists/components/tests/RecipientListForm.test.js
```diff
@@ -6,85 +6,61 @@ import { RecipientListForm } from '../RecipientListForm';
...
-      ],
-      ...formValues
-    };
-    const onSubmit = jest.fn();
-    const wrapper = shallow(<RecipientListForm onSubmit={onSubmit} {...props} />);
-    await wrapper.instance().preSubmit(formValuesWithCsv);
-    expect(onSubmit).toHaveBeenCalledWith(parsedFormValues);
+  it('autofills id on name change on create mode', () => {
+    wrapper.find('Field[name="name"]').simulate('change', { target: { value: 'FOo bAr' }});
+    // expect(wrapper.find('Field[name="id"]').value).toEqual('foo-bar');
```
bkemper: Do we need this?

rajumsys: oops! not sure how it remained there :)

PR #1023, opened 2019-05-21
FE-680: Ad blocker message

PatrickSison:
### What Changed
 - If SparkpostAPIError is missing response (when blocked request), show new message

### How To Test
 - Block request or go to `/reports/summary` w/ ad blocker
 - Error message should address adblock or network issue
 
### To Do
- [ ] Address feedback

### DIFF
--- a/src/actions/helpers/sparkpostApiError.js
+++ b/src/actions/helpers/sparkpostApiError.js
@@ -4,12 +4,10 @@ import _ from 'lodash';
 function SparkpostApiError(error) {
   const apiError = _.get(error, 'response.data.errors[0]', {});
 
-  if (!error.response) {
-    error.message = 'You may be having network issues or an adblocker may be blocking part of the app.';
-  }
   this.name = 'SparkpostApiError';
-  this.message = apiError.description || apiError.message || error.message;
   this.stack = error.stack; // must manually assign prototype value
+  this.message = apiError.description || apiError.message ||
+    (error.response ? error.message : 'You may be having network issues or an adblocker may be blocking part of the app.');
 
   // Intentionally assigning additional properties
   Object.assign(this, error);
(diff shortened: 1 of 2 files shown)


src/actions/helpers/sparkpostApiError.js
```diff
@@ -4,6 +4,9 @@ import _ from 'lodash';
 function SparkpostApiError(error) {
   const apiError = _.get(error, 'response.data.errors[0]', {});
 
+  if (!error.response) {
```
PatrickSison: We could also change this question to look at `error.message` to see if it's 'Network Error'

src/actions/helpers/sparkpostApiError.js
```diff
@@ -4,6 +4,9 @@ import _ from 'lodash';
...
+    error.message = 'You may be having network issues or an adblocker may be blocking part of the app.';
+  }
   this.name = 'SparkpostApiError';
   this.message = apiError.description || apiError.message || error.message;
```
kchu1337: You could move the conditional here. Something like 
```suggestion
   this.message = apiError.description || apiError.message ||
    (error.response ? error.message : 'You may be having network issues or an adblocker may be blocking part of the app.');
```

src/actions/helpers/tests/sparkpostApiError.test.js
```diff
@@ -23,7 +23,7 @@ describe('SparkpostApiError', () => {
 
   it('returns error message', () => {
     const error = createTestError();
-    expect(error).toHaveProperty('message', 'Oh no!');
```
kchu1337: Could you keep this test by adding a response code? And then make the adblock message its own test.

src/actions/helpers/sparkpostApiError.js
kchu1337: Doesn't really make too much of a difference. Up to you.

PatrickSison: I prefer your way since it's better than just overwriting errorMessage

PR #1033, opened 2019-05-28
EO-780 - Replace with RV "request access" page with the "add RV" page

kchu1337:
### What Changed
 - Changed the page to self-request adding Recipient Validation. 
 - Changed the flag from UI option to account option
 - There are 3 possible outcomes: 
   1. Self-Service and on paid plan: Can click a button to enable RV.
   2. Self-Service and on free plan: Can click a button to redirect to billing page
   3. Manually-Billed: No button at all. 

### How To Test
 - Test the 3 situations above. 
   1. Self-Service and on paid plan: Click the enable RV button and it should reload to unlocked RV page. 
   2. Self-Service and on free plan: Click the upgrade plan button and get redirected to billing page. 
   3. Manually-Billed: No button exists at all. 

### To Do
- [x] Double check the upgrade account put endpoint with RV as an option. Will get implemented by the Accounts team.
- [x] Rebase after AC-770 to use existing RV_tier constant.

### DIFF
--- a/src/pages/recipientValidation/components/RVDisabledPage.js
+++ b/src/pages/recipientValidation/components/RVDisabledPage.js
@@ -1,5 +1,5 @@
 import { Component } from 'react';
-import { Page, Panel, Table, Button, UnstyledLink } from '@sparkpost/matchbox';
+import { Page, Panel, Table, Button } from '@sparkpost/matchbox';
 import React from 'react';
 import styles from './RVDisabledPage.module.scss';
 import { currentPlanSelector } from 'src/selectors/accountBillingInfo';
@@ -66,13 +66,13 @@ export class RVDisabledPage extends Component {
     return rows;
   };
 
-  getActionButton = () => {
+  renderActionButton = () => {
     const { isSelfServeBilling, isFree } = this.props;
     if (!isSelfServeBilling) {
       return null;
     }
     return (isFree)
-      ? (<UnstyledLink component={Button} to={'/account/billing'} primary>Upgrade your plan</UnstyledLink>)
+      ? (<Button primary to={'/account/billing'}>Upgrade your plan</Button>)
       : (<Button primary onClick={this.enableRV}>Enable Recipient Validation</Button>);
   };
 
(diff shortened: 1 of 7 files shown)


src/pages/recipientValidation/components/RVDisabledPage.module.scss
```diff
@@ -0,0 +1,26 @@
+@import '~@sparkpost/matchbox/src/styles/config.scss';
+
+.Header {
+  color: color(orange);
+  font-size: rem(15);
+  margin-bottom: rem(100);
```
jonambas: margin bottom doesnt work on `strong` which is an inline element, you can use `display: block` here and i think remove the brs

src/pages/recipientValidation/components/RVDisabledPage.js
```diff
@@ -1,23 +1,123 @@
...
 
+const priceData = [
+  ['5,000', '10,000', '$0.008'],
+  ['10,000', '50,000', '$0.006'],
+  ['50,000', '100,000', '$0.004'],
+  ['100,000', '250,000', '$0.003'],
+  ['250,000', '750,000', '$0.0015'],
+  ['750,000', '1,000,000', '$0.00100']
+];
+const priceDataEdge = [
```
PatrickSison: At this point, would it need to remain in an array? Since it would statically be 2 things? Also we want to consolidate our constants since we're working on the same thing?

src/pages/recipientValidation/components/RVDisabledPage.js
```diff
@@ -1,23 +1,112 @@
...
+          efficient way to verify that addresses are valid before you send. We run emails through a series of checks
+          to catch many common problems, including syntax errors and non-existent mailboxes.
+          </p>
+          <p>
+          We have a monthly pay-as-you-go plan using tiered pricing. The more you validate, the less you pay per message.
+          </p>
+        </div>
+        <Panel className = {styles.Table}>
+          <Table >
+            <tbody className={styles.Table}>
```
PatrickSison: Is the styling needed for tbody? Seems to look the same, but I might be missing something

PR #1038, opened 2019-05-30
FE-664: mock API response middleware

ewandennis:
### What Changed
 - Added the ability to specify custom API mock responses for any test, using the common `use(...)` pattern.
 - Changed the default API mock responses to behave like a free account, to cue us up for billing func tests.

### How To Test
 - See the 3rd FE-664 PR for example usage

This will not be merged on its own. We'll merge the whole of FE-664 together.


src/__testHelpers__/http-responses/default/v1/account/get.js
```diff
@@ -16,11 +16,13 @@ export default ({ params }) => {
     status_updated: '2018-04-13T03:07:48.627Z',
     updated: '2018-06-12T02:16:30.576Z',
     subscription: {
-      code: '50K-0817',
```
PatrickSison: Should there be paid version of account retrieval?

PatrickSison: Nevermind, it's in main PR

PR #1032, opened 2019-05-27
TR-1357 Add campaign nav item with children and show if enabled for account

rajumsys:
Groups templates, a/b test & snippets under `Campaigns` menu. Refactored the navigation to
- Fix issue with opening parent on child link upon reload. This was working but given we're not changing route for templates, snippets & a/b tests the current logic didn't work. 
- Used hooks to get a gentle introduction to hooks

Screen Shot 2019-05-27 at 12 03 48 AM

### How to test
- Enable feature flag for the account.
```
{ 
	"options": 
	{
		"ui": {
			"templatesV2": true
		}
	} 
}
````
- This will reveal `Campaigns` menu
- Use the app normally. 

### To Do
- [x] Update icon once https://github.com/SparkPost/matchbox/pull/170 is merged
- [x] Address feedback

PR #1035, opened 2019-05-29
TR-1358 - Templates list page

rajumsys:
Implements templates list view based on mock. As discussed, using `templatesV2` directory and `templatesv2` new route to namespace all v2 work and minimize cleanup/diff for release. This PR is based on #1032; merging that first would reduce diff. 

### What Changed
 - Created a new directory `templatesV2` to keep new designs. 
- In order to minimize cleanup later, refactored routes to support both templates and templatesV2 routing. Any routes missing in v2, will fallback to v1 (/templates). 

### How To Test
 - Checkout to branch
 - Visit `Templates` from `Campaigns` menu. It should take you to new list page. Rest should work as is. 
 - Note: If you click breadcrumbs (like from template edit page), it'll take to old templates route. Those can be fixed after `next` editing paths and breadcrumb links are updated in editing PRs. 

### To Do
- [x] Displaying separate rows for draft & publish
- [x] Implement delete
- [x] Update description 
- [x] Address review feedback

- [x] Change base branch to `master` before merging

Notes:
- Brian used `/next` suffix for all of his routing. As we discussed this, in a later PR, i'll modify that and bring under `/templatev2`

PR #1043, opened 2019-05-31
Templatesv2 routes

rajumsys:
This is based on TR-1358. 

Move new template works under `/templatesV2` directory 

This is done for a few reasons including:
1) To avoid naming files like `ListPageV2` 
2) to minimize number of changes required during release (& removal of v1)
3) to avoid unintentionally affecting current template works while working on v2
4) to wire up routes correctly within v2

### What Changed
 - Moved EditAndPreviewPage to `/templatesV2` directory
 - Update routes and remove unused routes
 - Move related components

### How To Test
- Add necessary feature flag - https://github.com/SparkPost/2web2ui/pull/1032
- Make sure existing templates are working fine
- Make sure new template routes (/templatesv2) are working in
 
### Todo

- [x]  Move tests


rajumsys:
> I found a discrepancy between old and new templates list pages.
> I have some published templates with draft content. They show as published in v1 but only as draft items in v2. Shouldn't they also show as published items?
Will demo this so we can bottom it out.

Fixed that in TR-1358 branch (https://github.com/SparkPost/2web2ui/pull/1035/commits/73a1e132ba55cf21fd43dfe020bcdb398d029422). 

> src/config/routes/index.js includes an unused Fullscreen import on line 46 which could be cleaned up.

Fixed in https://github.com/SparkPost/2web2ui/pull/1043/commits/77fba117c8d9ac2faf4a5e580cdc47ce0aef3d27

PR #1047, opened 2019-06-05
EO-802 Add in a date-cutover divider for Health Score V2

kchu1337:
### What Changed
 - Added vertical reference line between June 6th and June 7th with label.

### How To Test
 - Change system date to after June 7th,
 - Go to health score dashboard and select a date that encompasses June 7th. Make sure the divider line is there. 
 - Click on a facet to go to heath score page. Make sure the divider line is there too. 

### To Do
- [x] Check with TPM to get design checked off.

PR #1052, opened 2019-06-07
EO-802 Re-add divider line

kchu1337:
### What Changed
 - Redo https://github.com/SparkPost/2web2ui/pull/1047
 - Change the label text to be left aligned and to match other axis text
 
### How To Test
 - see https://github.com/SparkPost/2web2ui/pull/1047


src/pages/signals/constants/healthScoreV2.js
```diff
@@ -0,0 +1,24 @@
+import styles from 'src/pages/signals/dashboards/components/HealthScoreChart/HealthScoreChart.module.scss';
+
+const V2Date = '2019-06-08';
+
+export const newModelLine = [
+  {
```
PatrickSison: A little bit weird that it's 2 separate labels. Is there a possible way to keep them within the same label?

PR #1055, opened 2019-06-10
AC-770: Recipient validation pricing

PatrickSison:
### What Changed
 - Added recipient validation to billing summary

### How To Test
 - On an account w/ recipient validation enabled, navigate to billing summary page
- See that usage calculates price
 - Click on `How is this calculated?`
 - See that usage calculation is broken down by tier

### To Do
- [x] Consolidate load conditions
- [x] Remove redundant price calculation (?)
- [x] Add testing for modal content
- [x] Check code coverage
- [ ] Address feedback


src/pages/billing/helpers/totalRecipientValidationCost.js
```diff
@@ -0,0 +1,14 @@
+import { RECIPIENT_TIERS } from 'src/constants';
+import { formatCurrency } from 'src/helpers/units';
+
+const totalRVCost = (volume = 0) => {
+  let totalCost = 0;
+
+  RECIPIENT_TIERS.forEach(({ volumeMax, volumeMin, cost }) => {
```
kchu1337: You can use `.reduce` here instead.

src/pages/billing/components/tests/RecipientValidationModal.test.js
```diff
@@ -0,0 +1,24 @@
...
+          timestamp: '2019-06-10T00:00:00.000Z'
+        }
+      }
+    };
+    wrapper = shallow(< RecipientValidationModal {...props} />);
+  });
+
+  it('should render properly', () => {
+    expect(wrapper).toMatchSnapshot();
+  });
```
kchu1337: Maybe add an assertion to check which rows are bolded? It's just that the snapshot is long and hard to read/verify.

src/pages/billing/components/RecipientValidationModal.js
```diff
@@ -0,0 +1,67 @@
...
+import { Close } from '@sparkpost/matchbox-icons';
+import { formatCurrency, formatFullNumber } from 'src/helpers/units';
+import styles from './RecipientValidationModal.module.scss';
+import cx from 'classnames';
+import { RECIPIENT_TIERS } from 'src/constants';
+import _ from 'lodash';
+
+export default ({ onClose, usage }) => {
+  const volumeUsed = _.get(usage, 'recipient_validation.month.used', 0);
+  let totalCost = 0;
```
kchu1337: Let's use the helper function you wrote. The extra calculations time should be trivial.

src/pages/billing/SummaryPage.js
```diff
@@ -22,10 +22,11 @@ export class BillingSummaryPage extends Component {
...
     this.props.getSendingIps();
     this.props.getInvoices();
+    this.props.getUsage();
   }
 
   render() {
-    const { loading, account, billingInfo, sendingIps, invoices, accountAgeInDays } = this.props;
+    const { loading, account, billingInfo, sendingIps, invoices, accountAgeInDays, hasRecipientValidation } = this.props;
 
     if (loading) {
```
kchu1337: Can we move the `usageLoading` check here instead?

src/pages/billing/components/BillingSummary.js
```diff
@@ -55,9 +62,24 @@ export default class BillingSummary extends Component {
...
+      </Panel.Section>
+    );
+  };
+
   render() {
-    const { account, currentPlan, canChangePlan, canUpdateBillingInfo, canPurchaseIps, invoices, isAWSAccount, accountAgeInDays } = this.props;
+    const { account, currentPlan, canChangePlan, canUpdateBillingInfo, canPurchaseIps, invoices, isAWSAccount, accountAgeInDays, hasRecipientValidation } = this.props;
+    const { rvUsage, usageLoading } = account;
     const { show } = this.state;
+    const showRecipientValidation = hasRecipientValidation && rvUsage && !usageLoading;
```
kchu1337: Left a comment above about loading. I feel like it might be jarring if the recipient validation modal link suddenly appears when it's finished loading

src/actions/account.js
```diff
@@ -23,6 +23,16 @@ export function getPlans({ meta = {}} = {}) {
   });
 }
 
+export function getUsage() {
```
kchu1337: Could also add a test for this but it's kinda self-explanatory. Up to you.

src/constants/index.js
```diff
@@ -155,4 +155,15 @@ export const PLAN_TIERS = {
   premier: 'Premier'
 };
 
+export const RECIPIENT_TIERS = [
```
kchu1337: I would rename to `RECIPIENT_VALIDATION_TIERS` so its really obvious what it is.

src/pages/billing/components/RecipientValidationModal.js
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
+import { Panel, Grid } from '@sparkpost/matchbox';
+import { Close } from '@sparkpost/matchbox-icons';
+import { formatCurrency, formatFullNumber } from 'src/helpers/units';
+import styles from './RecipientValidationModal.module.scss';
+import cx from 'classnames';
+import { RECIPIENT_TIERS } from 'src/constants';
+import _ from 'lodash';
+
+export default ({ onClose, usage }) => {
```
kchu1337: Also, this is destructuring `usage` wheras you are passing down `rvUsage`. You could also just pass down volume used instead of the whole usage object.

src/pages/billing/SummaryPage.js
PatrickSison: Moved it into the selector since it acts similarly to pulling billing info

PR #1057, opened 2019-06-10
TR-1348 - Template settings (using hooks)

rajumsys:
This PR attempts to use React Hook for form based on [this implementation](https://github.com/DJanoskova/React-validator-demo)n. This isn't ready, opened for early feedbacks and that's why no tests yet. Once the approach is agreed upon, i'll write tests. For now, just look at hooks implementation. Avoid non-hooks related feedback. I will create another PR for regular review once hooks is finalized. 

### What's changed
- Implements template settings update (except subaccount assignment)

### How to use:
- Campaigns -> Templates 
- Click a template
- Click Settings. Try to update settings fields. 

### Notes:
- Uses hook for settings form
- Does not support nested fields like `content.from`. They're flattened in `useTemplateSettings` hook and then appropriate structure is created before making API request
- Does not have state vars like `submitting` yet.
- Subaccount panel is not displayed/implemented as it depends on redux form. We'll have to refactor/re-implement that based on hooks. Same as for From Email field.
- Uses `useDispatch` hook to wrap action creator with store and invokes `dispatch` on it. Example `showAlert`. 
- **There might not be any distinction in the v2 editor about draft & published version yet.** 
- I've copied some files from v1 to v2 dirs. But to avoid dups, I've pointed "v1" files to corresponding "v2" files. This way, we will have less diff and would be able to drop `templates` dir without affecting v2. I'll follow this pattern in other tickets/PRs. 
- I see toggling doesn't work nicely with pristine. If you toggle back to prev state, `isPristine` is becoming false. 

This branch is based on hooks. But just in case this design isn't good to go and/or takes long time/effort, i've [`TR-1348` ](https://github.com/SparkPost/2web2ui/compare/TR-1348?expand=1) branch which implements same features using old way (redux).


src/pages/templatesV2/components/settings/Form.js
```diff
@@ -0,0 +1,156 @@
...
+const SettingsForm = (props) => {
+  const { settings, draft, domainsLoading, domains, updateDraft, subaccountId, showAlert, history } = useEditorContext();
+
+  const formHook = useForm(settings);
+  const { values, isPristine, isValid } = formHook;
+
+  const submitForm = (e) => {
+    e.preventDefault();
+    //prepare values
+    const settingValues = _.merge({}, draft, {
```
rajumsys: probably this method to re-structure settings should be in `useTemplateSettings` hook.

src/pages/templatesV2/EditAndPreviewPage.js
```diff
@@ -31,8 +32,13 @@ const EditAndPreviewPage = () => {
       breadcrumbRedirectsTo={`/${routeNamespace}`}
       title={draft.name}
     >
-      <EditNavigation primaryArea={<PrimaryArea />} />
-      <Contents />
+      <div className={styles.EditorNav}>
```
rajumsys: This part (fixing nav bar) is moved out to TR-1470  ( #1062)

src/hooks/useForm/useForm.js
```diff
@@ -0,0 +1,56 @@
...
+export function useForm(defaultValues) {
+  const formHandler = useState(defaultValues);
+  const [values, setValues] = formHandler;
+  const [errors, setErrors] = useState({});
+  const [mounted, setMounted] = useState(false);
+  const [isValid, setIsValid] = useState(false);
+  const [isPristine, setIsPristine] = useState(true);
+
+  // initial mounted flag
+  useEffect(() => setMounted(true), []);
```
rajumsys: why do we need this?

src/hooks/useForm/useForm.js
```diff
@@ -0,0 +1,56 @@
...
+  useEffect(() => {
+    setValues(defaultValues);
+  }, [defaultValues, setValues]);
+
+  useEffect(() => { //detect if form is valid
+    setIsValid(mounted && !Object.values(errors).length);
+  }, [errors, mounted, values]);
+
+  useEffect(() => { //detect if form is pristine
+    setIsPristine(_.isEqual(defaultValues, values));
```
rajumsys: > effect depends on setIsPristine

Didn't get that. Are you suggesting to add `setIsPristine` in dependency list? If so, i've seen this in a few places but don't understand sure how it helps.

rajumsys:
Though I believe it's getting into a better shape, it is still going to take a lot of time specially given several issues (like no default values etc.), i need unit tests rather than manual tests to be confident on the implementation. For the sake of time, I'm going to close this PR and open one with the redux based implementation.

PR #1059, opened 2019-06-11
EO-812: Added injections column to health score overview

PatrickSison:
### What Changed
 - Added injections column to health score page

### How To Test
 - [Point openresty to prd](https://confluence.int.messagesystems.com/pages/viewpage.action?spaceKey=ENG&title=Configuring+Openresty+Upstreams)
- Check the health score page and see the new column after `Current Score`
 - Select a custom date before 7 Jun 2019 (new model date)
 - Column should not be sortable and it should render `- - -` instead

### To Do
- [ ] Address feedback

### DIFF
--- a/src/pages/signals/components/HealthScoreOverview.js
+++ b/src/pages/signals/components/HealthScoreOverview.js
@@ -96,15 +96,15 @@ class HealthScoreOverview extends React.Component {
 
   render() {
     const {
-      data, error, facet, loading, signalOptions, subaccounts, tableName, totalCount, header, to
+      data, error, facet, loading, signalOptions, subaccounts, tableName, totalCount, header
     } = this.props;
+    const { to } = signalOptions;
 
     const subaccountFilter = _.get(signalOptions, 'subaccount.id');
     const isCustomRange = signalOptions.relativeRange === 'custom';
-
     const noFacetSelected = facet.key === 'sid';
     const noSubaccountFilter = subaccountFilter === undefined;
-    const afterNewModel = moment(V2Date, 'YYYY-MM-DD').diff(to) <= 0;
+    const afterNewModel = to && moment(V2Date, 'YYYY-MM-DD').diff(to) <= 0;
 
     // Filter out account aggregate, there is no way to do it via api
     // This is done here to preserve pagination functionality
@@ -158,10 +158,9 @@ class HealthScoreOverview extends React.Component {
           <Column
             dataKey="history"
             label="Daily Health Score"
-            width='20%'
+            width={afterNewModel ? '20%' : '30%'}
             component={({ history, ...filteredData }) => {
               const id = filteredData[facet.key];
-
               return (
                 <SparklineDataCell
                   data={history}
@@ -183,21 +182,22 @@ class HealthScoreOverview extends React.Component {
               <NumericDataCell value={current_health_score} />
             )}
           />
-          <Column
-            align="right"
-            dataKey="current_total_injection_count"
-            label={isCustomRange ? 'Injections' : 'Current Injections'}
-            sortable={afterNewModel}
-            width="15%"
-            component={({ current_total_injection_count }) => (
-              <NumericDataCell value={afterNewModel ? current_total_injection_count : null} />
-            )}
-          />
+          {afterNewModel &&
+            <Column
+              align="right"
+              dataKey="current_total_injection_count"
+              label={isCustomRange ? 'Injections' : 'Current Injections'}
+              sortable
+              width="15%"
+              component={({ current_total_injection_count }) => (
+                <NumericDataCell value={current_total_injection_count} />
+              )}
+            />}
           <Column
             align="right"
             dataKey="WoW"
             label={<WoWHeaderCell/>}
-            width="10%"
+            width={afterNewModel ? '12.5%' : '15%'}
             component={({ WoW }) => (
               <WoWDataCell value={WoW} />
             )}
@@ -206,7 +206,7 @@ class HealthScoreOverview extends React.Component {
             align="right"
             dataKey="average_health_score"
             label="Average Score"
-            width="12.5%"
+            width={afterNewModel ? '12.5%' : '15%'}
             component={({ average_health_score }) => (
               <NumericDataCell value={average_health_score} />
             )}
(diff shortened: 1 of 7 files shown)


src/pages/signals/components/HealthScoreOverview.js
```diff
@@ -94,14 +95,15 @@ class HealthScoreOverview extends React.Component {
...
-      data, error, facet, loading, signalOptions, subaccounts, tableName, totalCount, header
+      data, error, facet, loading, signalOptions, subaccounts, tableName, totalCount, header, to
     } = this.props;
 
     const subaccountFilter = _.get(signalOptions, 'subaccount.id');
     const isCustomRange = signalOptions.relativeRange === 'custom';
 
     const noFacetSelected = facet.key === 'sid';
     const noSubaccountFilter = subaccountFilter === undefined;
+    const afterNewModel = moment('2019/06/7').diff(to) <= 0;
```
rajumsys: I guess this is why getting the following console warning
```
console.js:39 Deprecation warning: value provided is not in a recognized RFC2822 or ISO format. moment construction falls back to js Date(), which is not reliable across all browsers and versions. Non RFC2822/ISO date formats are discouraged and will be removed in an upcoming major release. Please refer to http://momentjs.com/guides/#/warnings/js-date/ for more info.
Arguments: 
[0] _isAMomentObject: true, _isUTC: false, _useUTC: false, _l: undefined, _i: 2019/06/7, _f: undefined, _strict: undefined, _locale: [object Object]
Error
    at Function.createFromInputFallback (http://localhost:3100/static/js/7.chunk.js:118714:89)
    at configFromString (http://localhost:3100/static/js/7.chunk.js:120776:11)```

Comment (rajumsys) [reply src/pages/signals/components/HealthScoreOverview.js]: Can we use a valid date (for moment) and also put it in a constant somewhere (this module is fine too).

File: src/pages/signals/components/HealthScoreOverview.js [hunk 2]
```diff
@@ -180,11 +182,21 @@ class HealthScoreOverview extends React.Component {
               <NumericDataCell value={current_health_score} />
             )}
           />
+          <Column
+            align="right"
+            dataKey="current_total_injection_count"
+            label={isCustomRange ? 'Injections' : 'Current Injections'}
```
rajumsys: This is probably fine but just making sure, ticket is saying it should be just `Current Injections`?

src/pages/signals/components/HealthScoreOverview.js
```diff
@@ -180,11 +182,21 @@ class HealthScoreOverview extends React.Component {
...
+            sortable={afterNewModel}
+            width="15%"
+            component={({ current_total_injection_count }) => (
+              <NumericDataCell value={afterNewModel ? current_total_injection_count : null} />
```
rajumsys: Should we hide this column completely in that case? I mean we're not communicating in UI that it's only viewable in certain condition. So people might be wondering why it's not showing any value. But if we hide, they will know at least the column is not their (but i guess it'll still be a question).

src/pages/signals/components/tests/HealthScoreOverview.test.js
```diff
@@ -93,6 +94,12 @@ describe('HealthScoreOverview', () => {
     expect(wrapper.find('Column[dataKey="current_health_score"]').prop('label')).toEqual('Score');
   });
 
+  it('render injections properly when before new model date', () => {
+    const wrapper = subject({ signalOptions: { relativeRange: 'custom' }, to: '2019/06/05' });
+    expect(wrapper.find('Column[dataKey="current_total_injection_count"]').prop('label')).toEqual('Injections');
+    expect(wrapper.find('Column[dataKey="current_total_injection_count"]').prop('sortable')).toEqual(false);
```
rajumsys: how about a test showing where `sortable` will be `true`?

PatrickSison: I was going based off of similarly the Current Score vs Score which is based on the selected date. I'll ask Avi in the morning

src/pages/signals/components/HealthScoreOverview.js
PatrickSison: I was going based off of similarly the Current Score vs Score which is based on the selected date. I'll ask Avi in the morning

src/pages/signals/components/HealthScoreOverview.js
PatrickSison: I was going based off of similarly the Current Score vs Score which is based on the selected date. I'll ask Avi in the morning

src/pages/signals/components/HealthScoreOverview.js
PatrickSison: Only issue with this is the width will have to go to another column or multiple. Currently there's no autowidth, so the same condition would need to apply to 1 or several more columns

PatrickSison: I'll ask about this one as well

src/pages/signals/components/HealthScoreOverview.js
PatrickSison: We're going ahead with `Injections` if it's a custom range and `Current Injections` if it's the most recent data

src/pages/signals/components/tests/HealthScoreOverview.test.js
```diff
@@ -94,19 +94,11 @@ describe('HealthScoreOverview', () => {
...
-    expect(wrapper.find('Column[dataKey="current_total_injection_count"]').prop('sortable')).toEqual(false);
-  });
-
-  it('render injections properly when after model date', () => {
-    const wrapper = subject({ signalOptions: { relativeRange: 'custom' }, to: new Date('2019/06/07') });
-    expect(wrapper.find('Column[dataKey="current_total_injection_count"]').prop('label')).toEqual('Injections');
-    expect(wrapper.find('Column[dataKey="current_total_injection_count"]').prop('sortable')).toEqual(true);
+  it('renders injections column if after V2 release date', () => {
+    const wrapper = subject({ signalOptions: { relativeRange: 'custom', to: new Date('2019/06/08') }});
+    expect(wrapper).toMatchSnapshot();
```
rajumsys: Previous inline explicit assertion was better but I can see why you did snapshot though. 

We need a test for `Current Injection` code path.

PR #1060, opened 2019-06-11
TR-1373 Editor actions

rajumsys:
This implements the actions in the editor interface. The actions are on the right side of the bar. The action panel has one primary action and a couple of secondary actions in the drop down. They change based on template status (draft/published)

### What Changed
- Implements save and publish, edit draft, view published actions
- ~Makes editors read only if rendering published mode~ Moved out to #1068 

### How To Test
- Visit Campaigns -> Templates
- Click on a template. From the actions list on the right side, try different actions

### Notes
- The green (circular checkbox next to `published` text) is little different than the green in mock. It's fine for now. The green is from `sparkpost colors` (run on slack)
- `ActionList` wasn't used mostly because in that each item is wrapped in `<a>` tag which cause problem if we have other contents like `ConfirmationModal`
- Unlike past implementation of Duplicate (it was just redirecting to `/templates/create/:id`), new UX will be duplicate modal. It'll be implemented in TR-1512.
- Though it'll redirect correctly to draft and published mode, there is no difference in editors. That's fixed in #1068 
- Having each action in own component increased diff size mostly due to component and test boiler plates. Let me know if you think that's too redundant. 

### To Do
- [x] ~Finalize the loading state.~ Update: Lynn suggested to hide the dropdown upon click any item. This also isn't working nicely with confirmation modal. So I'll have to think about an alternative implementation and come back to it later. 
- [x] Font color of action items probably needs to be little lighter (as per mock) - I'll fix that as soon as VPN works and I can access app or along with addressing review feedback.
- [x] Write tests

Lmk If anything isn't clear, I can walk through implementation and codes.


src/pages/templatesV2/components/editorActions/SaveDraft.js
```diff
@@ -0,0 +1,24 @@
...
+
+export default ({ className, onClick }) => {
+  const { draft, content, updateDraft } = useEditorContext();
+
+  const handleClick = useCallback(() => {
+    if (onClick) {
+      onClick();
+    }
+
+    updateDraft({ id: draft.id, content }, draft.subaccount_id);
```
PatrickSison: Will this have some feedback like a loading, last saved, or snackbar alert? Either now or some future iteration?

src/pages/templatesV2/components/editorActions/DraftModeActions.js
```diff
@@ -0,0 +1,34 @@
...
+import { ArrowDropDown } from '@sparkpost/matchbox-icons';
+import SaveAndPublish from './SaveAndPublish';
+import ViewPublished from './ViewPublished';
+import SaveDraft from './SaveDraft';
+import styles from './Actions.module.scss';
+
+export default () => {
+  const [open, setOpen] = useState(false);
+
+  return (<div>
```
PatrickSison: If you replace with `Button.Group` instead of `div`, it looks a bit more joined which may closer to what you're trying to achieve.

src/pages/templatesV2/components/editorActions/SaveAndPublish.js
```diff
@@ -0,0 +1,46 @@
...
+    {children && <UnstyledLink onClick={showModal}>{children}</UnstyledLink>}
+    {!children && <UnstyledLink onClick={showModal}><CheckCircleOutline/>&nbsp;&nbsp;Save and Publish</UnstyledLink>}
+    <ConfirmationModal
+      title='Are you sure you want to publish your template?'
+      content={<p>Once published, your template will be available for use in email campaigns and A/B tests.</p>}
+      confirming={isDraftPublishing}
+      isPending={isDraftPublishing}
+      open={open}
+      confirmVerb='Save and Publish'
+      onCancel={hideModal}
```
PatrickSison: This may be too broad of a change, but should the cancel button be moved to left?

src/pages/templatesV2/components/editorActions/SaveAndPublish.js
```diff
@@ -0,0 +1,46 @@
...
+  return <div className={className}>
+    {children && <UnstyledLink onClick={showModal}>{children}</UnstyledLink>}
+    {!children && <UnstyledLink onClick={showModal}><CheckCircleOutline/>&nbsp;&nbsp;Save and Publish</UnstyledLink>}
+    <ConfirmationModal
+      title='Are you sure you want to publish your template?'
+      content={<p>Once published, your template will be available for use in email campaigns and A/B tests.</p>}
+      confirming={isDraftPublishing}
+      isPending={isDraftPublishing}
+      open={open}
+      confirmVerb='Save and Publish'
```
PatrickSison: Should there be a cancel verb to match the mockup?

src/pages/templatesV2/components/editorActions/SaveDraft.js
rajumsys: We're still not sure tbh. we're going to work with autosave and at that time we'll say `saved` on the left side of these button dropdown. So we can ask Lynn if another feedback will be needed.

src/pages/templatesV2/components/editorActions/SaveAndPublish.js
rajumsys: This is how our our confirmation modal is designed. I guess Lynn prefers what you suggested but once UX folks modifies modal, it'll be applied for all cases.

src/pages/templatesV2/components/editorActions/SaveAndPublish.js
rajumsys: You mean `Keep Editing`, right? Lynn designed that thinking these two buttons will be next to each other. After seeing/hearing our confirm/cancel are at distance, she decided no.

src/pages/templatesV2/components/editorActions/DraftModeActions.js
rajumsys: Was not aware of Button.Group. I'll look into it

src/pages/templatesV2/components/editorActions/SaveAndPublish.js
PatrickSison: 👍

src/pages/templatesV2/components/editorActions/SaveAndPublish.js
PatrickSison: 👍

src/pages/templatesV2/components/editorActions/SaveDraft.js
PatrickSison: 👍

PR #1061, opened 2019-06-11
TR-1445 Create Templates

rajumsys:
https://jira.int.messagesystems.com/browse/TR-1445
Implements the create template. See the ticket for attached mockup. 

New UX for template creation is that it asks for minimal required fields before creating a draft. Once it's created, it's redirected to edit page where contents can be edited. Also unblocking them by allowing `sparkpostbox.com` where they don't have verified sending domain. 

### What Changed
- Copied `FromEmailWrapper.js` to `templatesV2` directory and point to this file from `templates`. This way, we still share but do not dup same thing.
- Uses `sparkpostbox` domain if the account does not have verified sending domain. 

### How to Test
- Visit `Campaigns -> Create New Template`. Test it using account with verified domains and also with account without verified domain. 

### Notes

- Uses redux
- Modified `slugify` to avoid leading & trailing dashes. 
- Need better copy for sparkpostbox warning
- Because `components` directory is becoming too big (too many files), i've started creating relevant directory inside it to group things like `create`, `settings` etc.


src/pages/templatesV2/components/create/CreateForm.js
```diff
@@ -0,0 +1,114 @@
...
+            name='content.from.email'
+            component={FromEmailWrapper}
+            placeholder='example@email.com'
+            label='From Email'
+            validate={[required, emailOrSubstitution]}
+            domains={domains}
+            helpText={this.fromEmailHelpText()}
+          />
+        </Grid.Column>
+        <Grid.Column xs={12} lg={5}>
```
PatrickSison: Styling is a bit weird when it goes below to smaller width. Old templates has Subaccount lined up with fields above it.
Screen Shot 2019-06-12 at 11 08 21 AM
Screen Shot 2019-06-12 at 11 07 59 AM

src/pages/templatesV2/components/create/CreateForm.js
```diff
@@ -0,0 +1,114 @@
...
+
+  fromEmailHelpText() {
+    const { domains, domainsLoading, subaccountId, fromEmail } = this.props;
+
+    if (domainsLoading) {
+      return null;
+    }
+
+    if (fromEmail && fromEmail.split('@')[1] === config.sandbox.domain) {
+      return 'You are using sandbox domain which has certain restrictions!';
```
PatrickSison: I'm being pedantic, but should this be `You are using a sandbox domain...` or `You are using the sandbox domain...`?

rajumsys: `the` makes sense as sandbox domain is proper here :).

src/helpers/tests/string.test.js
```diff
@@ -39,6 +39,12 @@ describe('slugify', () => {
   it('should remove invalid characters', () => {
     expect(slugify('Example #1')).toEqual('example-1');
   });
+
+  it('should remove leading & trailing dashes', () => {
```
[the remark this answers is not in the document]
rajumsys: That's fine. i can add that that.

src/pages/templatesV2/components/create/CreateForm.js
rajumsys: That's true. Let me see if I can fix that.

rajumsys: I'm not sure if `Panel.Section` is cool without a panel. But seems adding a panel.section fixes it and `Subaccount` component also uses a Panel.Section.

PR #1065, opened 2019-06-17
FE-687/UX-124 Add Calibre as a static asset, Add font-face sheet

jonambas:
### What Changed
- Hosts fonts
- Conversion from gotham > calibre is in another ticket

### How To Test
- This should not change the bundle, but you can verify the fonts work by adding this CSS anywhere
```css
@import url('/static/fonts/calibre/calibre.css');

* {
  font-family: Calibre !important;
}
```
- Can check network/sources tab in chrome dev tools to confirm after that

### To Do
- [x] Feedback


rajumsys:
Looks good to me. 

Questions:
- why this change? Not sure if it looks better than the other. 
- What you meant by it won't change bundle size? Are you referring to net change after removing existing one?

jonambas:
@rajumsys The fonts aren't imported yet, so no change to existing UI. And we are changing because the website updated last week. The fonts won't look good until we update font sizes (another ticket)

jonambas:
Merged into FE-689 #1085

PR #1068, opened 2019-06-18
TR-1533 Fix content editing modes

rajumsys:
Renders editor in read only mode when viewing a published template; editable when viewing a draft. 

### What Changed
 - Render editors based on template status. Draft is editable, locked in published version
 - Add draft/published indicate at top bar
 - Fields in `Template Settings` tab are disabled when viewing published template
   - Delete button should be active
 - To distinguish (and deep link), modified route to add version
 - List and create pages are updated to take to new routes
 - Updated List page to use `list_name` to preserve original name in order to show correct messaging after delete. 

### How To Test
From Campaigns -> Templates
 - Open a draft template. You should be able to edit and top indicator should show draft
- Open a published template. Editing areas should be read only and indicated it's a published template like the image below. Template Settings should also be in read only mode
Screen Shot 2019-06-18 at 6 51 29 PM
-  Create a template. After template creation, it should take you to `/edit/<id>/draft/content`
- Modify a draft (w/o publish) and see the two versions are rendered correctly

### DIFF
--- a/src/pages/templatesV2/EditAndPreviewPage.js
+++ b/src/pages/templatesV2/EditAndPreviewPage.js
@@ -1,16 +1,17 @@
 import React from 'react';
+import { FileEdit, CheckCircle } from '@sparkpost/matchbox-icons';
+
 import { RedirectAndAlert } from 'src/components/globalAlert';
 import FullPage from 'src/components/fullPage';
 import Loading from 'src/components/loading';
 import EditNavigation from './components/EditNavigation';
 import links from './constants/editNavigationLinks';
 import useEditorContext from './hooks/useEditorContext';
-
-import { routeNamespace } from './constants/routes';
 import styles from './EditAndPreviewPage.module.scss';
+import { routeNamespace } from './constants/routes';
 
 const EditAndPreviewPage = () => {
-  const { currentNavigationIndex, draft, hasDraftFailedToLoad, isDraftLoading } = useEditorContext();
+  const { currentNavigationIndex, draft, hasDraftFailedToLoad, isDraftLoading, isPublishedMode } = useEditorContext();
   const Contents = links[currentNavigationIndex].render;
   const PrimaryArea = links[currentNavigationIndex].renderPrimaryArea;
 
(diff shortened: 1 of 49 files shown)


src/pages/templatesV2/hooks/useEditorContent.js
```diff
@@ -1,18 +1,19 @@
...
   // hydrate when loaded
   useEffect(() => {
-    if (draft.content) {
-      setState(draft.content);
+    const activeContent = isPublishedMode ? published : draft;
+    if (activeContent.content) {
+      setState(activeContent.content);
     }
-  }, [draft.content]);
+  }, [draft, draft.content, isPublishedMode, published, published.content]);
```
rajumsys: probably we don't need separate `draft` and `draft.content` and same for published.

src/pages/templatesV2/hooks/useEditorContent.js
```diff
@@ -1,18 +1,19 @@
...
   const [state, setState] = useState({});
   const setContent = (nextState) => {
     setState({ ...state, ...nextState }); // merge-in
   };
 
   // hydrate when loaded
   useEffect(() => {
-    if (draft.content) {
-      setState(draft.content);
+    const activeContent = isPublishedMode ? published : draft;
```
rajumsys: i'll revert this. based on TR-1539, we don't need to update published content at any moment.

src/pages/templatesV2/EditAndPreviewPage.js
```diff
@@ -27,10 +28,13 @@ const EditAndPreviewPage = () => {
     return <Loading />;
   }
 
+  const primaryArea = isPublishedMode ? <div>PUBLISHED <CheckCircle size={17} className={styles.GreenColor} /> </div> : <div>DRAFT <FileEdit size={17} /></div>;
```
PatrickSison: Screen Shot 2019-06-24 at 10 18 21 AM
Would it be possible to get this aligned? If not, it's not a big deal.

PR #1071, opened 2019-06-19
AC-736: Cancel accounts

PatrickSison:
### What Changed
 - Button to cancel plan on billing 
 - Button to renew plan if plan if pending cancellation
 - Pulled out brightback config

### How To Test
 - Navigate to `/account/billing`
 - Click on `Cancel Account`
 - Go through Brightback
 - Routed back and plan should be cancelled
 - Option should exist to renew plan
 - Should re-enable account
 - Downgrading to free should still work with brightback

### Notes
 - Currently staging and uat links will show downgrade link (which is the sandbox link).
   - Production should show the main cancellation link

### To Do
- [ ] Fix issue with new account info not being pulled after renewing account
- [ ] Address feedback


src/pages/billing/ImmediateCancelPlanPage.js
```diff
@@ -0,0 +1,77 @@
...
+  render() {
+    const { loading } = this.state;
+
+    if (loading === LOAD_STATE.PENDING) {
+      return <Loading />;
+    }
+
+    return <div className={styles.MessageBlock}>
+      {loading === LOAD_STATE.FAILURE && this.renderError()}
+      {loading === LOAD_STATE.SUCCESS && this.renderSuccess()}
```
ewandennis: `renderSuccess` doesn't actually render anything. Instead it triggers side effects. If this component renders in success state more than once, we'll get multiple alerts and multiple redirects.

I recommend separating the alert and redirect logic from the page display. More generally, we should avoid triggering side effects during render.

You could follow the pattern from `ImmediateChangePlanPage` or just remove the success logic from `render()` entirely and move the `showAlert()` and `history.push()` into the success clause of `handlePlanCancellation`.

src/pages/billing/ImmediateCancelPlanPage.js
```diff
@@ -0,0 +1,77 @@
...
+
+const BILLING_ROUTE = '/account/billing';
+
+export const LOAD_STATE = {
+  PENDING: 1,
+  SUCCESS: 2,
+  FAILURE: 3
+};
+
+export class ImmediateCanelPlanPage extends Component {
```
ewandennis: Minor: typo

src/pages/billing/SummaryPage.js
```diff
@@ -1,32 +1,50 @@
...
-    this.props.getInvoices();
-    this.props.getUsage();
+    const { fetchAccount, getBillingInfo, getPlans, getSendingIps, getInvoices, getUsage, brightbackEnabled, prepBrightback, brightbackData } = this.props;
+    fetchAccount();
+    getBillingInfo();
+    getPlans();
+    getSendingIps();
+    getInvoices();
+    getUsage();
+    if (brightbackEnabled) {
```
ewandennis: The `Brightback` component is designed to take care of preparing Brightback and insulating other components from having to deal with it. `Brightback` accepts a render prop which will pass an `enabled` flag and `to` link. 

As it is, Brightback logic has now leaked into `SummaryPage` and `BillingSummary`.

You can see `Brightback` used in `Confirmation` and you could also use it directly in `BillingSummary`. I will leave some sample code as a comment there.

src/pages/billing/components/BillingSummary.js
```diff
@@ -76,27 +76,48 @@ export default class BillingSummary extends Component {
...
+
+    const brightbackReady = brightback.valid && brightback.url;
+
+    if (!pending_cancellation) {
+      changePlanActions.push({
+        content: 'Cancel Plan',
+        to: brightbackReady || '/account/billing/plan/cancel',
+        color: 'orange',
+        disabled: cancelLoading,
+        component: brightbackReady ? undefined : Link
```
ewandennis: Lets build a simple component which renders a link and uses `Brightback`. Then we can use it in the `Cancel Plan` action as the `component` prop, and we can remove all BB logic from `SummaryPage`:

```jsx
import Brightback from 'src/components/brightback/Brightback';

const CANCEL_PLAN_ROUTE = '/account/billing/plan/cancel';

const CancelPlanButton = (props) => {
  const MaybeBB = ({ enabled, to }) => <Link {...props} to={enabled ? to : CANCEL_PLAN_ROUTE } />;
  return <Brightback
    config={config.brightback.downgradeToFreeConfig}
    condition={true}
    render={MaybeBB} />;
};
```

then in `render()`:

```jsx
    if (!pending_cancellation) {
      changePlanActions.push({
        content: 'Cancel Plan',
        color: 'orange',
        to: CANCEL_PLAN_ROUTE,
        disabled: cancelLoading,
        component: CancelPlanButton
      });
```

PatrickSison: That would be a lot better. Had trouble trying to render the brightback button.

PatrickSison: I think this runs into the same issue I was running into before. The way that buttons are rendered in the Panel and the way that brightback can render don't feed nicely into each other. Brightback is never mounted so it never calls its precancel

src/pages/account/components/tests/CancellationPanel.test.js
```diff
@@ -3,17 +3,27 @@ import { shallow } from 'enzyme';
...
   });
 
-  it('opens support ticket form when request link is clicked', () => {
-    const openSupportTicketForm = jest.fn();
-    const wrapper = shallow(<CancellationPanel openSupportTicketForm={openSupportTicketForm} />);
-
-    wrapper.find('UnstyledLink').simulate('click');
-
-    expect(openSupportTicketForm).toHaveBeenCalledWith({ issueId: expect.any(String) });
+  it('renders renew button when pending cancellation ', () => {
```
ewandennis:
```js
it('cancellation button forwards to BB', ...);
it('cancellation button cancels account when BB disabled', ...);
it('renew button renews pending cancellation', ...);
it('refreshes account state after renew', ...);
```

You can mock `Brightback` and either `dive()` or `render()` to make these easier to express.

src/pages/billing/components/PlanSummary.js
```diff
@@ -1,24 +1,25 @@
...
     ? 'free'
     : `$${recurringCharge.toLocaleString()} per ${period || 'month'}`;
   const volume = (planVolumePerPeriod || planVolume).toLocaleString();
-
+  const { effective_date } = pendingCancellation;
   return (
     <React.Fragment>
       <h6 className={styles.Headline}>
-        {volume} emails for {cost}
+        {volume} emails for {cost}{effective_date && <small> to end {formatDate(effective_date)}</small>}
```
ewandennis: Maybe we should be explicit about why this will end:

> ... to end on DATE when your account will be cancelled

src/pages/account/ImmediateCancelAccountPage.js
```diff
@@ -0,0 +1,73 @@
...
+    return cancelAccount()
+      .then(() => fetchAccount() , (error) => {
+        this.setState({ loading: LOAD_STATE.FAILURE, error });
+      })
+      .then(() => {
+        showAlert({
+          message: 'Your plan is set to be cancelled.',
+          type: 'success'
+        });
+        history.push(ACCOUNT_SETTINGS_ROUTE);
```
ewandennis: Lets use `history.replace()` to avoid any surprises when people navigate through their browser history.

PR #1072, opened 2019-06-19
EO-801 Switch over to Alerts v1 API for Alerts List page

kchu1337:
### What Changed
 - Created action/reducer/selector for alerts v1.
 - Added a formatted date and sort key in the selector to modify the list of alerts. 
 - Changed response params from labs/alerts -> v1.  EX:  'alert_metric'->'metric', 'enabled'->'muted'
 - Sorted by last trigger by using a sort key. Also added logic to only show last 4 triggered alerts in the cards panel.

### How To Test
```
    location /api/v1/alerts {
      include cors;
    }
```
 - Navigate to `/alerts-new`.
 - Make sure everything works and that alerts can be toggled and deleted. 

### To Do
- [ ] Possible fix the toggle button bug here.

### DIFF
--- a/src/actions/tests/__snapshots__/alertsV1.test.js.snap
+++ b/src/actions/tests/__snapshots__/alertsV1.test.js.snap
@@ -33,6 +33,7 @@ Array [
       "data": Object {
         "muted": false,
       },
+      "id": "alert-id",
       "method": "PUT",
       "url": "/v1/alerts/alert-id",
     },
--- a/src/actions/tests/__snapshots__/alertsV1.test.js.snap
+++ b/src/actions/tests/__snapshots__/alertsV1.test.js.snap
@@ -1,0 +1,37 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Action Creator: Alerts should dispatch a delete action 1`] = `
+Array [
+  Object {
+    "meta": Object {
+      "id": "alert-id",
+      "method": "DELETE",
+      "url": "/v1/alerts/alert-id",
+    },
+    "type": "DELETE_ALERT_V1",
+  },
+]
+`;
+
+exports[`Action Creator: Alerts should dispatch a list action 1`] = `
+Array [
+  Object {
+    "type": "LIST_ALERTS_V1_SUCCESS",
+  },
+]
+`;
+
+exports[`Action Creator: Alerts should dispatch a set muted status action 1`] = `
+Array [
+  Object {
+    "meta": Object {
+      "data": Object {
+        "muted": false,
+      },
+      "method": "PUT",
+      "url": "/v1/alerts/alert-id",
+    },
+    "type": "SET_ALERT_V1_MUTED_STATUS",
+  },
+]
+`;
--- a/src/actions/tests/__snapshots__/alertsV1.test.js.snap
+++ b/src/actions/tests/__snapshots__/alertsV1.test.js.snap
@@ -16,7 +16,12 @@ Array [
 exports[`Action Creator: Alerts should dispatch a list action 1`] = `
 Array [
   Object {
-    "type": "LIST_ALERTS_V1_SUCCESS",
+    "meta": Object {
+      "method": "GET",
+      "showErrorAlert": false,
+      "url": "/v1/alerts",
+    },
+    "type": "LIST_ALERTS_V1",
   },
 ]
 `;
--- a/src/pages/alerts/ListPageNew.js
+++ b/src/pages/alerts/ListPageNew.js
@@ -2,14 +2,13 @@ import React, { Component } from 'react';
 import { Link } from 'react-router-dom';
 import { Button, Grid, Page, Panel } from '@sparkpost/matchbox';
 import { RemoveRedEye } from '@sparkpost/matchbox-icons';
-import { ApiErrorBanner, DeleteModal, Loading } from 'src/components';
+import { ApiErrorBanner, DeleteModal, Loading, DisplayDate } from 'src/components';
 import { Templates } from 'src/components/images';
 import AlertCollectionNew from './components/AlertCollectionNew';
 import withAlertsList from './containers/ListPage.container';
 import styles from './ListPage.module.scss';
 import { formatDateTime } from 'src/helpers/date';
 import _ from 'lodash';
-import DisplayDate from 'src/components/displayDate/DisplayDate.js';
 
 export class ListPageNew extends Component {
   state = {
--- a/src/pages/alerts/ListPageNew.js
+++ b/src/pages/alerts/ListPageNew.js
@@ -30,14 +30,6 @@ export class ListPageNew extends Component {
     });
   };
 
-  getRecentlyTriggeredAlerts = () => {
-    const { alerts } = this.props;
-    return alerts
-      .filter((alert) => alert.last_triggered !== null) //Remove any alert that has never triggered
-      .sort((a, b) => (a.last_triggered.timestamp > b.last_triggered.timestamp)) //Sorts by last triggered date, descending
-      .slice(0,4);
-  }
-
   handleDelete = () => {
     const { id } = this.state.alertToDelete;
 
@@ -58,38 +50,33 @@ export class ListPageNew extends Component {
 
   renderRecentlyTriggered() {
 
-    const orderedAlerts = this.getRecentlyTriggeredAlerts();
-
-    if (orderedAlerts.length === 0) {
+    const { recentlyTriggeredAlerts } = this.props;
+    if (recentlyTriggeredAlerts.length === 0) {
       return;
     }
 
-    const recentlyTriggered = orderedAlerts.map((alert) => (
-      <Grid.Column
-        xs={12}
-        md={6}
-        lg={3}
-        key = {alert.id}>
-        <Panel
-          accent
-        >
-          <Panel.Section className = {styles.LastTriggeredCard}>
-            <div className = {styles.LastTriggeredTime} >
-              <DisplayDate timestamp={alert.last_triggered_timestamp} formattedDate={alert.last_triggered_formatted} />
-            </div>
-            <h3>{alert.name}</h3>
-          </Panel.Section>
-
-          <Panel.Section className = {styles.Footer}>
-            <Button flat to = {'/alerts-new'/*TODO replace this link*/}><RemoveRedEye className = {styles.Icon}/></Button>
-          </Panel.Section>
-        </Panel>
-      </Grid.Column>));
     return (
       <>
         <h3>Recently Triggered Alerts</h3>
         <Grid>
-          {recentlyTriggered}
+          {recentlyTriggeredAlerts.map((alert) => (
+            <Grid.Column
+              xs={12}
+              md={6}
+              lg={3}
+              key = {alert.id}>
+              <Panel accent>
+                <Panel.Section className = {styles.LastTriggeredCard}>
+                  <div className = {styles.LastTriggeredTime} >
+                    <DisplayDate timestamp={alert.last_triggered_timestamp} formattedDate={alert.last_triggered_formatted} />
+                  </div>
+                  <h3>{alert.name}</h3>
+                </Panel.Section>
+                <Panel.Section className = {styles.Footer}>
+                  <Button flat to = {'/alerts-new'/*TODO replace this link*/}><RemoveRedEye className = {styles.Icon}/></Button>
+                </Panel.Section>
+              </Panel>
+            </Grid.Column>))};
         </Grid>
         </>);
   }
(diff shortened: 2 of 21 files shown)


src/selectors/alertsV1.js
```diff
@@ -0,0 +1,17 @@
+import _ from 'lodash';
+import { createSelector } from 'reselect';
+import { formatDateTime } from '../helpers/date';
+
+const getAlertsList = (state) => state.alertsV1.list;
+
+const appendFormattedDate = (alert) => ({
```
ewandennis: This all looks functional so my feedback here is about the next most important thing: maintainability. In this case, that's about simplicity and naming.

`appendFormattedDate` doesn't fully describe this helper since it also calculates a sort key.

Do we want alerts that have never triggered to appear or last?

The new field names might also be difficult to understand outside of this function. When we display an alert, it would be useful to say "Oh, we're showing last triggered timestamp here" or "We're using last triggered timestamp as the sort key".

I would also pick `camelCase` or `snake_case` for fields in a single object so people can guess at fields when you talk about them.

We could avoid naming the transform at all by folding it into the selector itself, since its small:

```js
export const selectAlertsList = createSelector(
  [getAlertsList],
  (alerts) => alerts.map((alert) => ({
    ...alert,
    sortKey: alert.last_triggered ? alert.last_triggered : '0',
    last_triggered_formatted: alert.last_triggered ? formatDateTime(alert.last_triggered) : null
  }))
);
```

src/selectors/tests/alertsV1.test.js
```diff
@@ -0,0 +1,59 @@
...
+      {
+        id: 'id-3',
+        muted: true,
+        name: 'my alert 3',
+        metric: 'health_score',
+        last_triggered: '2019-06-15T14:48:00.000Z'
+      }
+    ];
+
+    formattedAlerts = [
```
ewandennis: Since `formattedAlerts` is used just for verifying `selectAlertsList`, maybe we should define it there or in the `describe()` for that selector?

src/selectors/tests/alertsV1.test.js
```diff
@@ -0,0 +1,59 @@
+import * as alertsSelectors from '../alertsV1';
+
+
+
+describe('Alerts Selectors: ', () => {
+  let alertsV1; let formattedAlerts;
```
ewandennis: I generally keep to 1 definition per line.

src/selectors/tests/alertsV1.test.js
```diff
@@ -0,0 +1,59 @@
...
+
+  beforeEach(() => {
```
ewandennis: Do we need a `beforeEach()` clause if there is only 1 test?

src/reducers/alertsV1.js
```diff
@@ -0,0 +1,48 @@
...
+      /* DELETE */
+
+    case 'DELETE_ALERT_V1_PENDING':
+      return { ...state, deletePending: true };
+
+    case 'DELETE_ALERT_V1_SUCCESS':
+      return {
+        ...state,
+        deletePending: false,
+        // TODO will need to match subaccount id
```
ewandennis: If this is a requirement for this ticket, I would make sure its in the AC and/or put a TODO on the PR. TODOs in code aren't very visible.

src/actions/tests/__snapshots__/alertsV1.test.js.snap
```diff
@@ -0,0 +1,37 @@
...
+    },
+    "type": "DELETE_ALERT_V1",
+  },
+]
+`;
+
+exports[`Action Creator: Alerts should dispatch a list action 1`] = `
+Array [
+  Object {
+    "type": "LIST_ALERTS_V1_SUCCESS",
```
ewandennis: This doesn't look right. It should be an API call, no?

src/pages/alerts/ListPageNew.js
```diff
@@ -31,10 +30,19 @@ export class ListPageNew extends Component {
     });
   };
 
+  getRecentlyTriggeredAlerts = () => {
+    const { alerts } = this.props;
+    const orderedArray = alerts
+      .filter((alert) => alert.last_triggered !== null) //Remove any alert that has never triggered
+      .sort((a, b) => (a.last_triggered > b.last_triggered) ? -1 : 1) //Sorts by last triggered date, descending
```
ewandennis: I think you can also do `b.last_triggered - a.last_triggered`. Not vital, just a common pattern.

src/pages/alerts/ListPageNew.js
```diff
@@ -31,10 +30,19 @@ export class ListPageNew extends Component {
     });
   };
 
+  getRecentlyTriggeredAlerts = () => {
+    const { alerts } = this.props;
+    const orderedArray = alerts
```
ewandennis: What does `orderedArray` contain? Maybe we could name it for its contents. Something like:

 - `recentAlerts`
 - `recentlyTriggered`
 - `recentlyTriggeredAlerts`

src/pages/alerts/ListPageNew.js
```diff
@@ -51,36 +59,31 @@ export class ListPageNew extends Component {
...
-          <div className = {styles.LastTriggeredTime} ><DisplayDate timestamp={timestamp} formattedDate={lastTriggeredDate} /></div>
-          <h3>{alert.name}</h3>
-        </Panel.Section>
-
-        <Panel.Section className = {styles.Footer}>
-          <Button flat to = {'/alerts-new'}><RemoveRedEye className = {styles.Icon}/></Button>
-        </Panel.Section>
-      </Panel>
-    </Grid.Column>));
+    const orderedAlerts = this.getRecentlyTriggeredAlerts();
```
ewandennis: Since `getRecentlyTriggeredAlerts()` depends only on `state.alerts.list`, maybe it should be a selector.

src/pages/alerts/ListPageNew.js
```diff
@@ -51,36 +59,31 @@ export class ListPageNew extends Component {
...
+
+    if (orderedAlerts.length === 0) {
+      return;
+    }
+
+    const recentlyTriggered = orderedAlerts.map((alert) => (
```
ewandennis: Is it better to put this calculation inline with the containing markup? That way, we can read the markup in the way its emitted:

```jsx
return <>
  <h3>Recently Triggered Alerts</h3>
  <Grid>
    { recentlyTriggered.map(...) }
  </Grid>
</>;
```

src/pages/alerts/components/AlertCollectionNew.js
```diff
@@ -20,34 +18,30 @@ const filterBoxConfig = {
 };
 
 class AlertCollectionNew extends Component {
-  //TODO Add last triggered date and replace link
-  getDetailsLink = ({ id, subaccount_id }) => `/alerts/edit/${id}${setSubaccountQuery(subaccount_id)}`
+  //TODO replace link
```
ewandennis: If this work needs done, lets capture it in Jira and/or in a TODO on this PR.

src/pages/alerts/components/AlertCollectionNew.js
```diff
@@ -20,34 +18,30 @@ const filterBoxConfig = {
...
+  getDetailsLink = ({ id }) => `/alerts/edit/${id}`;
 
   getColumns() {
     const columns = [
       { label: 'Alert Name', sortKey: 'name', width: '40%', className: styles.TabbedCellHeader },
-      { label: 'Metric', sortKey: 'alert_metric' },
-      { label: 'Last Triggered', sortKey: '' },
-      { label: 'Status', sortKey: 'enabled' },
+      { label: 'Metric', sortKey: 'metric' },
+      { label: 'Last Triggered', sortKey: 'sortKey' },
```
ewandennis: If this read something closer to `{ label: 'Last Triggered', sortKey: 'last_triggered_or_null' }`, itd be easier to comprehend.

src/pages/alerts/components/AlertCollectionNew.js
```diff
@@ -20,34 +18,30 @@ const filterBoxConfig = {
...
+    const deleteFn = () => this.props.handleDelete({ id, name });
     return [
       <div className = {styles.TabbedCellBody}>
-        <PageLink to={this.getDetailsLink({ id, subaccount_id })}>{name}</PageLink>
+        <PageLink to={this.getDetailsLink({ id })}>{name}</PageLink>
       </div>,
-      <Tag>{_.get(METRICS, alert_metric, alert_metric)}</Tag>,
-      <DisplayDate timestamp={timestamp} formattedDate={lastTriggeredDate} />,
-      <AlertToggle enabled={enabled} id={id} subaccountId={subaccount_id} />,
+      <Tag>{_.get(METRICS, metric, metric)}</Tag>,
```
ewandennis: Are we expecting `METRICS` not to contain human-readable versions of all `metric` field values? I would spenda moment making sure we cover all the cases here. Using `_.get()` and falling back to a raw field value looks like we aren't confident in the structure of our API records.

src/pages/alerts/components/AlertCollectionNew.js
```diff
@@ -71,7 +65,7 @@ class AlertCollectionNew extends Component {
         getRowData={this.getRowData}
         pagination={true}
         filterBox={filterBoxConfig}
-        defaultSortColumn='name'
+        defaultSortColumn='sortKey'
```
ewandennis: It'd be great to know what we were sorting on 😄

src/pages/alerts/components/AlertToggleNew.js
```diff
@@ -0,0 +1,52 @@
+import React, { Component } from 'react';
```
ewandennis: This looks like an almost direct copy of `AlertToggle`. Is the plan to remove the existing `AlertToggle`, `AlertCollection` and `AlertPage` after the switch over?

If so, will that happen on this ticket?

src/reducers/alertsV1.js
kchu1337: This was in the reducer from the labs endpoint, I will check with Avi/EO to see if this is still needed.

src/actions/tests/__snapshots__/alertsV1.test.js.snap
kchu1337: I think this was from a previous commit. I caught it later on.

src/pages/alerts/components/AlertToggleNew.js
kchu1337: The plan, to my knowledge, is to merge this down under a user feature flag and then create a new ticket to remove the feature flag and change all the new pages to the default ones. I think this was how signalsV2 and alerts was previously done.

src/reducers/alertsV1.js
```diff
@@ -0,0 +1,48 @@
...
+      return { ...state, listError: payload, listPending: false };
+
+    case 'LIST_ALERTS_V1_SUCCESS':
+      return { ...state, list: payload, listPending: false };
+
+    // UPDATE single list row Muted status
+    case 'SET_ALERT_MUTED_V1_STATUS_PENDING':
+      return { ...state, setMutedStatusPending: true };
+
+    case 'SET_ALERT_MUTED_V1_STATUS_SUCCESS':
```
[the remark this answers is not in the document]
kchu1337: Yes and no. It works without updating because the Toggle Component has a local state. There are some minor bugs but I wasn't sure if I should do it in this ticket or make a new bug ticket.

src/reducers/alertsV1.js
kchu1337: Just checked with them. They said this was because previously id was not unique...? But now Id is unique across all accounts and users so we can remove this.

src/reducers/alertsV1.js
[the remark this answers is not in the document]
ewandennis: Making this change _should_ fix the bug where dismissing the delete modal causes all changed `AlertToggle` instances to reset.

src/selectors/alertsV1.js
```diff
@@ -0,0 +1,17 @@
...
+  ...alert,
+  sortKey: (alert.last_triggered) ? alert.last_triggered : '0',
```
ewandennis: Parse `last_triggered` and convert to a timestamp. Use the number `0` for null values.

src/pages/alerts/components/AlertCollectionNew.js
ewandennis: Maybe capture this line of code in comments for that ticket.

src/pages/alerts/ListPageNew.js
```diff
@@ -52,41 +50,33 @@ export class ListPageNew extends Component {
...
+                  <div className = {styles.LastTriggeredTime} >
+                    <DisplayDate timestamp={alert.last_triggered_timestamp} formattedDate={alert.last_triggered_formatted} />
+                  </div>
+                  <h3>{alert.name}</h3>
+                </Panel.Section>
+                <Panel.Section className = {styles.Footer}>
+                  <Button flat to = {'/alerts-new'/*TODO replace this link*/}><RemoveRedEye className = {styles.Icon}/></Button>
+                </Panel.Section>
+              </Panel>
+            </Grid.Column>))};
```
ewandennis: Accidental semicolon there. It appears in the rendered output since it isn't part of any JSX clause.

src/pages/alerts/components/AlertCollectionNew.js
kchu1337: I added some comments in `EO-786` ticket.

PR #1074, opened 2019-06-20
TR-1539 Read only template settings for published template

rajumsys:
Disables settings fields when viewing published template

### What Changed
 - Fields in `Template Settings` tab are disabled when viewing published template
   - Delete button should be active
 - To distinguish (and deep link), modified route to add version
 - List and create pages are updated to take to new routes 

### How To Test
 - From List page, click on draft template -> Template Settings. It should be editable as usual
 - From List page, click on published template -> Template Settings. It should not be editable
 - Create a template. After template creation, it should take you to `/edit/<id>/draft/content`

### Notes:
- Editors are made read only in #1068 
- Navigation between draft and published mode is handled in #1060 

### To Do
- [ ] Address review feedback

### DIFF
--- a/src/config/routes/templates.js
+++ b/src/config/routes/templates.js
@@ -72,7 +72,7 @@ const v2Routes = [
     supportDocSearch: 'template'
   },
   {
-    path: '/templatesv2/edit/:id/:navKey?',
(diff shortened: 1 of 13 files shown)


rajumsys:
I'll merge this into #1068

PR #1075, opened 2019-06-20
EO-836: Updated styling on Single address form

PatrickSison:
### What Changed
 - Updated styling on single address form

### How To Test
 - Navigate to `/recipient-validation`
 - Second tab should say `Single Address`
 - Click on tab
 - Form should be aligned with mock up
 - Check that form works

### To Do
- [ ] Address feedback

### DIFF
--- a/src/pages/recipientValidation/components/SingleAddressForm.js
+++ b/src/pages/recipientValidation/components/SingleAddressForm.js
@@ -6,6 +6,7 @@ import { TextFieldWrapper } from 'src/components';
 import { required, email, maxLength } from 'src/helpers/validation';
 import { singleAddress } from 'src/actions/recipientValidation';
 import SingleResult from './SingleResult';
+import styles from './SingleAddressForm.module.scss';
 
 const formName = 'singleAddressForm';
 export class SingleAddressForm extends Component {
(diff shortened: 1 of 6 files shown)


src/pages/recipientValidation/components/SingleAddressForm.js
```diff
@@ -21,16 +22,21 @@ export class SingleAddressForm extends Component {
...
-              placeholder={'eg. example@mail.com'}
-              validate={[required, email, maxLength(254)]}
-              normalize={(value = '') => value.trim()}
-              connectRight={<Button primary submit disabled={submitDisabled}>{buttonContent}</Button>}
-            />
+            <div className={styles.Header}>Validate a Single Address</div>
+            <p className={styles.Subheader}>Enter the email address below you would like to validate.</p>
+            <div className={styles.Field}>
+              <Field
+                style={{ height: '3.2rem', paddingLeft: '1.5em', fontSize: '.9em' }}
```
rajumsys: can we move it to css?

PatrickSison: Currently styling matches the mockup so I will stick with it. Also tried before moving to css. Hasn't been working without inline css. I can try looking into why styles might be removed.

PR #1077, opened 2019-06-21
AC-830: Updated Email Role to templates

PatrickSison:
### What Changed
 - Renamed role `email` to `Templates`

### How To Test
 - Check that Role is renamed from `email` to `Templates` on manage users page
 - Check that invite user is working correctly
 - Check that modify user role is working correctly
### Notes
 - Should roll out soon after AC-714 so that invites and user modifying isn't broken

### To Do
- [ ] Address any feedback

### DIFF
--- a/src/constants/users.js
+++ b/src/constants/users.js
@@ -1,19 +1,19 @@
 export const ROLES = {
   ADMIN: 'admin',
   DEVELOPER: 'developer',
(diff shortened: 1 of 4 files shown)


ewandennis:
How do we test this today? It looks like `tst` and `stg` both respond with `invalid access_level` when we use the new `templates` role name.

I also see the user list page shows `email` for existing users with the `email` role. Will these be changed at the API level or should we handle that case in the front end?

PR #1083, opened 2019-06-26
Eo 787 - Add alerts Create Page

kchu1337:
### What Changed
 - Created new Create alert Page
 - Created new alert form that separates some Fields into individual components.
     - Filters are their own component
     - The Evaluator is its own component.
     - Subaccounts is its own component.
- Added to validators an inclusive between 2 numbers validator. 
- Implement the Combobox multi-tag typeahead component. 
- Rewrote the function to transform the form field data -> API body.
- Added new constants and helpers to help in showing options and branching 

### How To Test
```
    location /api/v1/alerts {
      include cors;
    }
```
 - Navigate to `/alerts-new/create`.
 - Make sure everything works and that all types of alerts can be created. 

### To Do
 - [x] Address comments from Jon
 - [x] Address comments from Ewan
 - [x] Possibly change email placeholder to 'List of comma delimited emails'. Possibly set subaccounts to [-1] if the field is empty.


src/components/typeahead/ComboBoxTypeahead.js
```diff
@@ -0,0 +1,138 @@
+import React, { useState, useEffect } from 'react';
+import { ComboBox, ComboBoxTextField, ComboBoxMenu } from '@sparkpost/matchbox';
+import Downshift from 'downshift';
+import debounce from 'lodash/debounce';
+import sortMatch from '../../helpers/sortMatch';
+
+export const ComboBoxTypeahead = (props) => {
```
jonambas: This component in general looks good to me. I would add in `propTypes` and `defaultProps` as a form of documentation, and also add comments to the effects

src/components/reduxFormWrappers/ComboBoxTypeaheadWrapper.js
```diff
@@ -0,0 +1,16 @@
...
+// Wrapped matchbox components for use with react-redux Field components
+export default function ComboBoxTypeaheadWrapper({ input, meta, ...rest }) {
+  const { active, error, touched } = meta;
+  return (
+    <ComboBoxTypeahead
+      value={input.value}
+      name={input.name}
+      onChange={input.onChange}
+      error={!active && touched && error ? error : undefined}
+      {...rest}
```
jonambas: This wouldn't really change functionality, but might be a good idea to spread `{...input}` for the `onFocus` and `onBlur` redux-form events

src/pages/alerts/CreatePageNew.js
```diff
@@ -0,0 +1,35 @@
...
+
+export class CreatePageNew extends Component {
+
+  /*
+    Passed as onSubmit to AlertForm. Figures out what updates need to be passed
+    to the createAlert action.
+  */
+  handleCreate = async (values) => {
+    const { createAlert, showAlert, history } = this.props;
+    await createAlert({
```
kchu1337: use promises or await/async. Preferably promises

src/pages/alerts/containers/CreatePageNew.container.js
```diff
@@ -0,0 +1,18 @@
+import { connect } from 'react-redux';
+import { withRouter } from 'react-router-dom';
+import { createAlert, deleteAlert } from 'src/actions/alertsV1';
```
kchu1337: delete delete

src/pages/alerts/containers/CreatePageNew.container.js
```diff
@@ -0,0 +1,18 @@
...
+function withAlertsCreate(WrappedComponent) {
+  const mapDispatchToProps = { createAlert, deleteAlert, showAlert };
+
+  const mapStateToProps = (state, props) => ({
+    error: state.alertsV1.createError,
+    loading: state.alertsV1.createPending,
+    deletePending: state.alertsV1.deletePending
+  });
+
+  return withRouter(connect(mapStateToProps, mapDispatchToProps)(WrappedComponent));
```
kchu1337: Check if withRouter is necessary

src/pages/alerts/components/AlertFormNew.js
```diff
@@ -0,0 +1,150 @@
...
+      change,
+      subaccountsFieldValue,
+      target,
+      source
+    } = this.props;
+
+    const submitText = submitting ? 'Submitting...' : 'Create Alert';
+    const formSpec = getFormSpec(metric);
+
+    const resetFormValues = (value) => {
```
kchu1337: is an event

src/pages/alerts/components/AlertFormNew.js
```diff
@@ -0,0 +1,150 @@
...
+                  component={TextFieldWrapper}
+                  disabled={submitting}
+                  validate={[required, maxLength(24)]}
+                />
+                <div className = {styles.MetricSelector}>
+                  <label>Alert Metric</label>
+                  <Field
+                    name='metric'
+                    component={SelectWrapper}
+                    options={[{ value: '', label: 'Select Metric', disabled: true }, ...getOptions(METRICS)]}
```
kchu1337: add as constant

src/components/typeahead/ComboBoxTypeahead.js
```diff
@@ -0,0 +1,166 @@
...
+    setSelected([]);
+  }, [type]);
+
+  //This is so the list of matches is consistent with the prop results
+  useEffect(() => {
+    setMatches(results);
+  }, [results]);
+
+  //Cancels the debounce on unmount
+  useEffect(() => () => debounceFn.cancel(), [debounceFn]);
```
ewandennis: Its not clear how this behaves given that we share 1 `debounceFn` across _all_ instances of `ComboBoxTypeahead`. I would expect us to create the debounced function on mount and call cancel on unmount.

src/components/typeahead/tests/ComboBoxTypeahead.test.js
```diff
@@ -0,0 +1,54 @@
...
+
+  it('action list does not show when there are no options and the textbox is read-only', () => {
+    const wrapper = subject({ results: []});
+    wrapper.find('ComboBoxTextField').props().onClick({});
+    expect(wrapper.find('ComboBoxTextField').prop('readOnly')).toEqual(true);
+    expect(wrapper.find('ComboBoxMenu').prop('isOpen')).toEqual(false);
+  });
+
+  it('cancels the debounce on unmount', () => {
+    const debounceFn = debounce((fn, input) => fn(input), 0);
```
ewandennis: We can mock or spy on `lodash/debounce` to track how its used in the component.

ewandennis: You could keep a single variable in this test as a mock `debounce.cancel`. Then your module mock above could attach it to the debounced function as `.cancel`.

src/pages/alerts/components/tests/AlertFormNew.test.js
```diff
@@ -0,0 +1,53 @@
...
+    expect(wrapper.instance().props.change).toHaveBeenCalledWith('sending_ip', []);
+    expect(wrapper.instance().props.change).toHaveBeenCalledWith('mailbox_provider', []);
+    expect(wrapper.instance().props.change).toHaveBeenCalledWith('sending_domain', []);
+    expect(wrapper.instance().props.change).toHaveBeenCalledWith('single_filter', { filter_type: 'none', filter_values: []});
+  });
+
+  describe('submit button props', () => {
+
+    it('should disable submit button when pristine or submitting', () => {
+      wrapper.setProps({ pristine: true, submitting: false });
```
ewandennis: These should probably be separate tests, maybe using `jest-in-case`.

src/pages/alerts/components/AlertFormNew.js
```diff
@@ -0,0 +1,114 @@
...
+                  <label>Alert Metric</label>
+                  <Field
+                    name='metric'
+                    component={SelectWrapper}
+                    options={metricsOptions}
+                    helpText={'This assignment is permanent.'}
+                    onChange={resetFormValues}
+                  />
+                </div>
+                {formSpec.hasFilters &&
```
ewandennis: Lets test both sides of this condition.

src/pages/alerts/components/AlertFormNew.js
```diff
@@ -0,0 +1,114 @@
...
+                  <>
+                    <label><h5>Filtered by <span className={styles.OptionalText}>optional</span></h5></label>
+                    <SubaccountField
+                      disabled={submitting}
+                    />
+                    <FilterFields
+                      disabled={submitting}
+                    />
+                </>}
+                {metric !== '' &&
```
ewandennis: Lets test this condition.

src/pages/alerts/components/fields/tests/EvaluatorFields.test.js
```diff
@@ -0,0 +1,65 @@
...
+      source: 'raw'
+    };
+
+    wrapper = shallow(<EvaluatorFields {...props} />);
+  });
+  it('renders correctly', () => {
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('does not render source if it is only 1 option and increases grid size of slider', () => {
```
ewandennis: `and` in a test spec suggests >1 test. Lets split these up.

src/pages/alerts/components/fields/tests/EvaluatorFields.test.js
```diff
@@ -0,0 +1,65 @@
...
+    wrapper.setProps({ metric: 'block_bounce_rate' });
+    expect(wrapper.find({ name: 'source' })).not.toExist();
+    expect(wrapper.find({ id: 'sliderColumn' })).toMatchSnapshot();
```
ewandennis: Lets be explicit instead of using snapshots where ever possible. You could check a prop or attrib value.

src/pages/alerts/components/fields/tests/FilterFields.test.js
```diff
@@ -0,0 +1,53 @@
...
+    wrapper = shallow(<FilterFields {...props}/>);
+    expect(props.listPools).toHaveBeenCalled();
+    expect(props.listSendingDomains).toHaveBeenCalled();
+    expect(props.listSendingIps).toHaveBeenCalled();
+  });
+
+  it('renders single type filter correctly', () => {
+    expect(wrapper.find({ name: 'single_filter.filter_values' })).toExist();
+    expect(wrapper.find({ name: 'sending_ip' })).not.toExist();
+    expect(wrapper).toMatchSnapshot();
```
ewandennis: Lets verify the structure of the rendered output. Its not clear from the snapshot what is "correct".

src/pages/alerts/components/fields/EvaluatorFields.js
```diff
@@ -0,0 +1,104 @@
...
+        <br/>
+        <Field
+          name='value'
+          component={TextFieldWrapper}
+          disabled={disabled}
+          suffix={suffix}
+          validate={numberBetweenInclusive(0, 100)}
+          normalize={normalize}
+          type='number'
+          style={{
```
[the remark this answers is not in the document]
kchu1337: I don't think it is possible for a `TextField` component.

src/pages/alerts/components/fields/tests/FilterFields.test.js
kchu1337: I want to keep the snapshots because I do not have a default `this renders` test as it either renders 1 view or a 2nd view.

src/pages/alerts/components/fields/EvaluatorFields.js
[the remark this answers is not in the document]
jonambas: `3.5.1` includes `align` as a prop, [demo here](https://sparkpost.github.io/matchbox/?selectedKind=Form%7CTextField&selectedStory=aligned%20to%20the%20right&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel)

Also there are other ways to fix vertical alignment.
- Grid is a flex container and accepts a classname (column does not though), could try adding `align-items` to it
- Make a wrapper around each fieldset with `align-self`

ewandennis:
UX notes:
 - The expandable channel `email_addresses` field is required but its error message is hidden until the user expands the field. Its not clear why the form is unsubmittable at that point. Recommend we either surface the field error in the expandable title or subtitle, or as a quick fix just default the expandable to open.
 - Since the slider is used with a text field everywhere, we need to account for the edge cases of filling in the field and keeping the slider in sync. e.g. It is currently possible to input 1000 into a 0-100 limited slider+field combo component.
 - Since the alert name field is non-unique, maybe we should call it "label"
 - Consider providing initial helpful defaults for the trigger values:
    - e.g. monthly sending limit value defaults to 0 - an alert when you hit 0% of your limit isn't too useful (default to 80% ?)
    - e.g. health score defaults to "below" and "0" - an alert when you hit health score below 0 isn't great (default to below 80% ?)
    - e.g. block bounce rate defaults to above 0 - that will presumably trigger a lot (default to above 1%?)
    - e.g. block bounce rate value goes from 0-100 in integer increments. Maybe more useful to allow smaller increments?

src/pages/alerts/components/tests/AlertFormNew.test.js
```diff
@@ -0,0 +1,88 @@
...
+  });
+
+  it('should handle submit', () => {
+    wrapper.find('Form').simulate('submit');
+    expect(props.handleSubmit).toHaveBeenCalled();
+  });
+
+  it('should reset form values when changing metric', () => {
+    wrapper.find({ name: 'metric' }).simulate('change', { target: { value: 'block_bounce_rate' }});
+    expect(wrapper.instance().props.change).toHaveBeenCalledWith('sending_ip', []);
```
ewandennis: Recommend having 1 expect() per test to ensure its clear exactly which part of the contract is being tested.

src/pages/alerts/components/fields/EvaluatorFields.js
```diff
@@ -0,0 +1,104 @@
...
+
+const absoluteNormalize = (value) => Math.abs(value);
+
+export const EvaluatorFields = ({
+  metric,
+  value,
+  source,
+  disabled,
+  change,
+  normalize = absoluteNormalize
```
ewandennis: If we don't set the `normalize` prop, lets not expose it. Instead we can just import `absoluteNormalize` and use it in our implementation.

src/pages/alerts/components/fields/EvaluatorFields.js
```diff
@@ -0,0 +1,104 @@
+import React, { useState, useEffect } from 'react';
+import { connect } from 'react-redux';
+import { Field, formValueSelector, change } from 'redux-form';
+import { SelectWrapper, TextFieldWrapper } from 'src/components/reduxFormWrappers';
+import { getFormSpec, getEvaluatorOptions } from '../../helpers/alertForm';
+import { Grid, Slider, Label } from '@sparkpost/matchbox';
+import { numberBetweenInclusive } from 'src/helpers/validation';
+import { FORM_NAME } from '../../constants/formConstants';
+
+const absoluteNormalize = (value) => Math.abs(value);
```
ewandennis: If we want to test this separately, we can move it to a helper and write unit tests just for it.

PR #1087, opened 2019-06-28
TR-1374 - Edit test data

rajumsys:
Allows user to edit tests data

### What Changed
 - Syncs test data with local storage
 - Uses test data in preview generation

### How To Test
 - Edit Template (v2)
 - Go to `Test Data` tab and add substitution_data (other prop might not be in use now)
 - Go to HTML/Text/AMP and use those subs tag 
 - Preview should render with substituted data
 
 - Separate test data for published and draft template

Notes: 
 - Probably we can enable annotation for test data
 - Currently preview is not re-generated upon changing test data. Should we auto preview as soon as test data is changed? 

### To Do
- [ ] Finish pending tests


bkemper:
Closing, refer to #1206

PR #1093, opened 2019-07-02
EO-871: Updated list results styling

PatrickSison:
### What Changed
 - Updated the styles of list results

### How To Test
 - Go to `/recipient-validation/list`
 - Upload a list
 - Check that the card on the bottom now matches the styles from the mockup

### Notes
 - File Name omitted for now until API update
 - Date moved to left
 - Button kept on right
 - Button text changed to `Download Results`
 - Title changed to `Last Validation`

### To Do
- [ ] Address feedback


src/pages/recipientValidation/components/ListResultsCard.js
```diff
@@ -20,52 +18,62 @@ const ListResultsCard = ({ complete = 'unknown', uploaded, rejectedUrl, status }
...
   const ready = status === 'SUCCESS';
-  const failed = status === 'ERROR';
+  const failed = status === 'ERROR' ;
 
-  let spinner = <div className={styles.Spacer}/>;
   let Icon = InsertDriveFile;
-  let iconClass = null;
   let statusText = null;
 
   if (loading) {
```
kchu1337: Instead of a bunch of `let` and `if` statements, you could make it a function that returns the correct render.
Like
```js
const renderStatus = () => {
if (loading) {
return  <Tag><span className={styles.Loading}><Cached /></span> Processing</Tag>
}
...
```
Could also do a switch-case with `status`

src/pages/recipientValidation/components/ListResultsCard.js
```diff
@@ -20,52 +18,62 @@ const ListResultsCard = ({ complete = 'unknown', uploaded, rejectedUrl, status }
 
   const loading = !complete && status !== 'ERROR';
```
kchu1337: Do you need the `&& status !== 'ERROR';`? If status = `ERROR`, your variable `failed` will be true and `status` and `Icon` get overwritten below.

src/pages/recipientValidation/components/ListResultsCard.js
```diff
@@ -20,52 +18,62 @@ const ListResultsCard = ({ complete = 'unknown', uploaded, rejectedUrl, status }
...
     statusText = 'Completed';
   }
 
   return (
-    <Panel sectioned>
-      <div className={classnames(styles.IconWrapper, iconClass)}>
-        <Icon size={40} />
+    <Panel sectioned accent='gray' title={<div className={styles.PanelHeader}>Last Validation:</div>}>
+      <div className={styles.TableContainer}>
+        <Table>
```
kchu1337: Does it need to be a table vs a grid? Are we going to have multiple entries later?

src/pages/recipientValidation/components/ListResultsCard.js
```diff
@@ -20,52 +18,62 @@ const ListResultsCard = ({ complete = 'unknown', uploaded, rejectedUrl, status }
...
+              <Table.HeaderCell className={styles.TableCell} width='25%'>
+                Status:
+              </Table.HeaderCell>
+            </Table.Row>
+            <Table.Row>
+              <Table.Cell className={styles.TableCell}>
+                {formatDateTime(moment.unix(uploaded))}
+              </Table.Cell>
+              <Table.Cell className={styles.TableCell}>
+                <Tag><Icon /> {statusText}</Tag>
```
kchu1337: If you change the above let-if into a function, then you can just do `{renderStatus()}` here.

src/pages/recipientValidation/components/ListResultsCard.module.scss
```diff
@@ -1,37 +1,32 @@
...
 
-.Spacer {
-  height: rem(37);
+.TableContainer {
+  margin-top: 1rem;
 }
 
-.NoWrap {
-  white-space: nowrap;
+.DownloadButton {
```
kchu1337: The download button vertical align/ y position looks a little low. But that's just imo.

src/pages/recipientValidation/components/tests/__snapshots__/ListResultsCard.test.js.snap
```diff
@@ -2,85 +2,130 @@
 
 exports[`ListResultsCard renders correctly when batch job fails 1`] = `
 <Panel
+  accent="gray"
```
kchu1337: Rather than all snapshot tests, you can just check that <Error> icon exists (if you change to render function, might need to check children text if you keep your let-if). And I guess that the Button does not exist?
Same for the other tests. Probably good to have 1 snapshot test though.

src/pages/recipientValidation/components/ListResultsCard.module.scss
PatrickSison: Was going based off of the mockup where the button was in line with the result row rather than the label row

src/pages/recipientValidation/components/ListResultsCard.js
PatrickSison: It potentially may as it may get a result history later on down the line

src/pages/recipientValidation/components/tests/__snapshots__/ListResultsCard.test.js.snap
PatrickSison: 👍

PR #1095, opened 2019-07-08
AC-838: Updated loading condition

PatrickSison:
[AC-838](https://jira.int.messagesystems.com/browse/AC-838)
### What Changed
 - Updated loading condition which could be causing bug of rendering page before getting billing info

### How To Test
 - On new account, go to `/account/billing`
 - Check that submitting new plan w/ card goes through
 - Check that loading state stays loading until new page

### To Do
- [ ] Address feedback

### Notes
 I think that account should always have created, so then it should be looking at the loading condition regardless.


bkemper:
This can be closed until the bug occurs again.

PR #1096, opened 2019-07-09
AC-746 : Logout invalidates both access and refresh token

ayoung1:

### What Changed
 - Access and Refresh tokens are invalidated when the user clicks the logout button

### How To Test
 - Login on the development webpage
 - Gather access/refresh key from DynamoDB
 - Logout
 - Confirm the keys are removed from Dynamo and can't be used to access api/v1/account

### DIFF
--- a/src/actions/auth.js
+++ b/src/actions/auth.js
@@ -177,7 +177,9 @@ export function logout() {
       return;
     }
     dispatch(invalidateAuthToken(token));
-    dispatch(invalidateAuthToken(refreshToken));
+    if (refreshToken) {
+      dispatch(invalidateAuthToken(refreshToken));
+    }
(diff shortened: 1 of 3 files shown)


src/actions/auth.js
```diff
@@ -156,11 +171,13 @@ export function refresh(token, refreshToken) {
...
   return (dispatch, getState) => {
     const { auth } = getState();
+    const { token, refreshToken } = auth;
     // only log out if currently logged in
     if (!auth.loggedIn) {
       return;
     }
-
+    dispatch(invalidateAuthToken(token));
+    dispatch(invalidateAuthToken(refreshToken));
```
PatrickSison: Should probably make a check for refresh token since it doesn't necessarily exist

PR #1098, opened 2019-07-11
FE-693: Addd GTM event to choose plan form

PatrickSison:
[FE-693](https://jira.int.messagesystems.com/browse/FE-693)

### What Changed
 - Added GTM event 

### How To Test
  - Create an account
  - on onboarding, choose a paid plan
 - Check that the event is being added to window.datalayer 

### To Do
- [ ] Verify that events are being properly tracked

### Notes:
 - I don't know if it's an issue with it being dev server, but I don't see dataLayer changing on local env, but I see it fine on prod. It's getting to the function correctly


bkemper:
Lets close until we have better understanding of the need.

PR #1100, opened 2019-07-11
EO-885 Implement Duplicate Alert logic

kchu1337:
### What Changed
 - Edited the create page to allow for duplication of an alert
 - Created helper to transform api data to form data
 - Created logic for initial values for the Alert form

### How To Test
```
    location /api/v1/alerts {
      include cors;
    }
```
 - Navigate to `/alerts-new`.
 - Click any alert name in the table to open details page.
 - Click on the duplicate button in the top right.
 - Verify that the create page contains all of the duplicated alert's values


src/pages/alerts/DetailsPage.js
```diff
@@ -36,7 +36,7 @@ export class DetailsPage extends Component {
 
     return deleteAlert({ id }).then(() => {
       this.closeDeleteModal();
-      history.push('../alerts-new');
+      history.push('../');
```
bkemper: I recommend using absolute routes.

src/pages/alerts/CreatePageNew.js
```diff
@@ -3,30 +3,55 @@ import { Link } from 'react-router-dom';
...
+    const { loading, getError, getLoading, duplicateId, alert } = this.props;
+
+    if (getLoading) {
+      return <Loading/>;
+    }
+
+    if (getError) {
+      return (
+        <RedirectAndAlert
+          to='/alerts-new'
```
jonambas: I think you would want to redirect back to the details page you were on, no?

src/pages/alerts/DetailsPage.js
```diff
@@ -76,7 +76,7 @@ export class DetailsPage extends Component {
         breadcrumbAction={{ content: 'Back to Alerts', to: '/alerts-new', component: Link }}
         primaryArea={
           <>
-          <Button flat><ContentCopy className={styles.Icon}/>Duplicate</Button>
+          <Button flat component={Link} to={`../create/${id}`}><ContentCopy className={styles.Icon}/>Duplicate</Button>
```
jonambas: Would use an absolute path here

src/pages/alerts/components/AlertFormNew.container.js
```diff
@@ -17,7 +18,7 @@ export default function withAlertForm(WrappedComponent) {
       metric: selector(state, 'metric'),
       single_filter: selector(state, 'single_filter'),
       muted: selector(state, 'muted'),
-      initialValues: DEFAULT_FORM_VALUES
+      initialValues: intialValues ? intialValues : DEFAULT_FORM_VALUES
```
jonambas: Can we append `(Duplicate)` to the alert name?

src/pages/alerts/CreatePageNew.js
```diff
@@ -3,30 +3,55 @@ import { Link } from 'react-router-dom';
...
+          alert={{ type: 'error', message: getError.message }}
+        />
+      );
+    }
+    const initialValues = duplicateId ? formatFromApiToForm(alert) : undefined;
```
jonambas: `formatFromApiToForm` `formatFromFormToApi` naming is a bit confusing

Historically we've put functions like `formatFromApiToForm` inside selectors though

src/pages/alerts/CreatePageNew.js
jonambas: Nm, this is fine

src/pages/alerts/CreatePageNew.js
kchu1337: I could move it into a selector, but I felt that it would be easier to read by putting them in the same helper function. What about using a selector but importing the function from the helper?

src/pages/alerts/helpers/formatFormData.js
```diff
@@ -2,7 +2,7 @@ import _ from 'lodash';
 import { REALTIME_FILTERS } from '../constants/formConstants';
 import { getFormSpec } from './alertForm';
 
-export default function (values) {
+export const formatFormValues = (values) => {
```
bkemper: The file and function have a different name.

src/selectors/alertsV1.js
```diff
@@ -22,4 +25,46 @@ export const selectRecentlyTriggeredAlerts = createSelector(
...
+    default: (() => {})
+  };
+  const formFilters = (getFormFilters[filterType] || getFormFilters.default)();
+
+  const { source, operator, value } = threshold_evaluator;
+
+  const { emails = []} = channels;
+  const email_addresses = emails.join(', ');
+
+  const name = (isDuplicate) ? `${alert.name} (Duplicate)` : alert.name;
```
bkemper: I made a string helper, `tagAsCopy`, for this situation.

src/pages/alerts/CreatePageNew.js
```diff
@@ -3,30 +3,54 @@ import { Link } from 'react-router-dom';
...
+import { formatFormValues } from './helpers/formatFormData';
+import { Loading } from 'src/components';
+import { RedirectAndAlert } from 'src/components/globalAlert';
 
 export class CreatePageNew extends Component {
 
+  componentDidMount() {
+    const { getAlert, duplicateId } = this.props;
+    if (duplicateId) {
+      getAlert({ id: duplicateId });
```
bkemper: Is it the duplicate ID or is it really the ID to duplicate?

src/selectors/alertsV1.js
```diff
@@ -22,4 +25,46 @@ export const selectRecentlyTriggeredAlerts = createSelector(
     .slice(0,4)
 );
 
+export function getInitialValues(alert = {}, isDuplicate) {
```
bkemper: This should be a selector that uses state. 

```js
export const selectAlert => (state) => state.alertsV1.alert;
export const selectAlertFormValues => createSelector([selectAlert], (alert) => {
 // ...
});
```

PR #1103, opened 2019-07-16
AC-888: Dashboard inaccurately reports daily usage for unrestricted accounts

bkemper:
Refer to [AC-888](https://jira.int.messagesystems.com/browse/AC-888)

### What Changed
 - Limits will not be shown unless present
 - Monthly limit will always be displayed when present (before it was hidden if same value as plan volume)

Screen Shot 2019-07-16 at 12 07 16 PM

### How To Test
 - Create a new account
 - `PUT /account/control trust_level=unrestricted`
 - Open the dashboard and confirm limits are not displayed

### DIFF
--- a/src/components/usageReport/UsageReport.js
+++ b/src/components/usageReport/UsageReport.js
@@ -52,10 +52,6 @@ export class UsageReport extends Component {
     const hasDailyLimit = usage.day.limit > 0;
     const hasMonthlyLimit = usage.month.limit > 0;
 
-    const overageMarkup = overage
-      ? <DisplayNumber label='Extra Emails Used' content={overage.toLocaleString()}/>
-      : null;
-
     return (
       <Panel title='Your Usage Report' actions={actions}>
         <Panel.Section>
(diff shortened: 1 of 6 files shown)


src/components/usageReport/UsageReport.js
```diff
@@ -47,53 +47,43 @@ export class UsageReport extends Component {
...
-    const dailyUsage = getPercent(usage.day.used, usage.day.limit);
-
-    const monthlyLimit = usage.month.limit || subscription.plan_volume;
-    const monthlyUsage = getPercent(usage.month.used, monthlyLimit);
-
-    const dailyLimitMarkup = usage.day.limit
-      ? <DisplayNumber label='Daily Limit' content={usage.day.limit.toLocaleString()}/>
-      : null;
-
     const overageMarkup = overage
```
kchu1337: Could move this down and use the ```{overage && ... }``` notation

bkemper:
@kchu1337 This is ready for final review.

PR #1107, opened 2019-07-17
AC-862: Removed X-Sparky header on authenticate

PatrickSison:
[AC-862](https://jira.int.messagesystems.com/browse/AC-862)

### What Changed
 - Removed header on `/authenticate`

### How To Test
 - Run multiple requests by navigating pages. Check that `X-Sparky` header is on any request except `/authenticate`

### Notes
Header is already removed API side, but this just makes sure it's gone client side for `/authenticate`. Another alternative is to add the header as an option, and then make it enabled everywhere but `/authenticate` rather than removing it

### To Do
- [ ] Address feedback

### DIFF
--- a/src/helpers/axiosInstances.js
+++ b/src/helpers/axiosInstances.js
@@ -12,6 +12,11 @@ export const sparkpostAxios = axios.create({
 
 sparkpostAxios.interceptors.response.use(null, sparkpostErrorHandler);
 
+export const sparkpostNoHeaders = axios.create({
+  baseURL: apiBase,
+  withCredentials: true
+});
+
 export const sparkpost = sparkpostAxios;
 
 export const zuora = axios.create({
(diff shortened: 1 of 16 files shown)


src/helpers/http.js
```diff
@@ -29,7 +29,8 @@ function sparkpostLogin(username, password, rememberMe, authHeader = authenticat
     method: 'POST',
     url: '/v1/authenticate',
     data,
-    headers: buildHeaders(authHeader)
+    headers: buildHeaders(authHeader),
+    transformRequest: [(data, headers) => { delete headers['X-Sparky']; return data; }]
```
bkemper: I think we need a new axios instance that doesn't use apiRequestHeaders.

src/helpers/axiosInstances.js
```diff
@@ -9,11 +9,17 @@ export const sparkpostAxios = axios.create({
   headers: apiRequestHeaders,
   withCredentials: true
 });
-
 sparkpostAxios.interceptors.response.use(null, sparkpostErrorHandler);
-
 export const sparkpost = sparkpostAxios;
 
+export const sparkpostPublicAxios = axios.create({
```
bkemper: Do we need to export this?

PR #1110, opened 2019-07-19
AC-848: Added UI Options to hide terminated subaccounts

PatrickSison:
### What Changed
 - UI Option to filter subaccounts

### How To Test
 - Go to `/account/settings`
 - Toggle account option to hide terminated subaccounts
 - Check list page for subaccounts
 - Check typaheads don't show terminated subaccounts if enabled

### To Do
- [ ] Address feedback

### DIFF
--- a/src/pages/account/components/UIOptionsPanel.js
+++ b/src/pages/account/components/UIOptionsPanel.js
@@ -1,0 +1,55 @@
+import React, { Component } from 'react';
+import { Panel, Toggle, Grid } from '@sparkpost/matchbox';
+import LabelledValue from 'src/components/labelledValue/LabelledValue';
+import { isAccountUiOptionSet } from 'src/helpers/conditions/account';
+import { update as updateAccount } from 'src/actions/account';
+import { connect } from 'react-redux';
+
+export class UIOptionsPanel extends Component {
+
+
+  setUIOption = () => {
+    const { updateAccount, hideTermSubEnabled } = this.props;
+    const body = {
+      options: {
+        ui: {
+          hideTerminatedSubaccounts: Boolean(!hideTermSubEnabled)
+        }
+      }
+    };
+    updateAccount(body);
+  };
+
+  render() {
+    const { hideTermSubEnabled, loading } = this.props;
+
+    return (
+      <Panel title='UI Options'>
+        <Panel.Section>
+          <Grid>
+            <Grid.Column xs={11}>
+              <LabelledValue label='Hide Subacounts'>
+                <p>Hide terminated subacounts. Resources associated with terminated subaccounts can still be accessed.</p>
+              </LabelledValue>
+            </Grid.Column>
+            <Grid.Column xs={1}>
+              <Toggle
+                id="hideTerminatedSubaccounts"
+                checked={hideTermSubEnabled}
+                disabled={loading}
+                onChange={this.setUIOption}
+              />
+            </Grid.Column>
+          </Grid>
+        </Panel.Section>
+      </Panel>
+    );
+  }
+}
+
+const mapStateToProps = (state) => ({
+  hideTermSubEnabled: isAccountUiOptionSet('hideTerminatedSubaccounts')(state),
+  loading: state.account.updateLoading
+});
+
+export default connect(mapStateToProps, { updateAccount })(UIOptionsPanel);
--- a/src/pages/account/components/UIOptionsPanel.js
+++ b/src/pages/account/components/UIOptionsPanel.js
@@ -11,7 +11,7 @@ const OPTIONS = [
   {
     key: 'hideTerminatedSubaccounts',
     label: 'Hide Subaccounts',
-    description: 'Hide terminated subacounts. Resources associated with terminated subaccounts can still be accessed.'
+    description: 'Hide terminated subaccounts. Resources associated with terminated subaccounts can still be accessed.'
   }
 ];
 
--- a/src/pages/account/components/UIOptionsPanel.js
+++ b/src/pages/account/components/UIOptionsPanel.js
@@ -2,6 +2,7 @@ import React, { Component } from 'react';
 import { Panel, Toggle, Grid } from '@sparkpost/matchbox';
 import LabelledValue from 'src/components/labelledValue/LabelledValue';
 import { isAccountUiOptionSet } from 'src/helpers/conditions/account';
+import { selectCondition } from 'src/selectors/accessConditionState';
 import { update as updateAccount } from 'src/actions/account';
 import { connect } from 'react-redux';
 
@@ -13,7 +14,7 @@ export class UIOptionsPanel extends Component {
     const body = {
       options: {
         ui: {
-          hideTerminatedSubaccounts: Boolean(!hideTermSubEnabled)
+          hideTerminatedSubaccounts: !hideTermSubEnabled
         }
       }
     };
@@ -24,7 +25,7 @@ export class UIOptionsPanel extends Component {
     const { hideTermSubEnabled, loading } = this.props;
 
     return (
-      <Panel title='UI Options'>
+      <Panel title='Account Options'>
         <Panel.Section>
           <Grid>
             <Grid.Column xs={11}>
@@ -48,7 +49,7 @@ export class UIOptionsPanel extends Component {
 }
 
 const mapStateToProps = (state) => ({
-  hideTermSubEnabled: isAccountUiOptionSet('hideTerminatedSubaccounts')(state),
+  hideTermSubEnabled: selectCondition(isAccountUiOptionSet('hideTerminatedSubaccounts'))(state),
   loading: state.account.updateLoading
 });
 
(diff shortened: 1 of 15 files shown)


bkemper:
How does this work if another resource template, webhook, etc. is assigned to the terminated subaccount?

src/selectors/subaccounts.js
```diff
@@ -33,6 +34,10 @@ export const selectSubaccountFromQuery = createSelector(
   (subaccounts, id) => _.find(subaccounts, { id: Number(id) })
 );
 
+export const selectFilteredSubaccounts = (state) => _.filter(state.subaccounts.list, ({ status }) =>
```
bkemper: Selectors should always be created with `createSelector` unless it is an state or prop assessor (e.g. `getSubaccounts`).

src/pages/account/components/UIOptionsPanel.js
```diff
@@ -0,0 +1,55 @@
...
+      }
+    };
+    updateAccount(body);
+  };
+
+  render() {
+    const { hideTermSubEnabled, loading } = this.props;
+
+    return (
+      <Panel title='UI Options'>
```
bkemper: Is the average user going to know what UI is?

src/selectors/subaccounts.js
bkemper: This `selectFilteredSubaccounts` variable name is misleading.  The list is only filtered if the account option is set.  I think it would be better to build this logic into `selectSubaccounts`.

```
export const selectSubaccounts = createSelector(
  [getSubaccounts, selectCondition(isAccountUiOptionSet('hideTerminatedSubaccounts'))],
  (subaccounts, hideTerminatedSubaccounts) => {
    const visibleSubaccounts = hideTerminatedSubaccounts
      ? subaccounts.filter(({ status }) => status !== 'terminated')
      : subaccounts;

    return visibleSubaccounts.map((subaccount) => formatSubaccount(subaccount));
  }
);
```

If we find cases when even with the flag we need the entire list, we can make a more explicit selector.

bkemper: A benefit to the selector pattern is they are easily testable.  Please make sure this is tested.

src/pages/account/components/UIOptionsPanel.js
```diff
@@ -0,0 +1,55 @@
...
+        <Panel.Section>
+          <Grid>
+            <Grid.Column xs={11}>
+              <LabelledValue label='Hide Subacounts'>
+                <p>Hide terminated subacounts. Resources associated with terminated subaccounts can still be accessed.</p>
```
bkemper: Subaccounts is misspelled.

src/pages/account/components/tests/UIOptionsPanel.test.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import { UIOptionsPanel } from '../UIOptionsPanel';
+
+let props;
+let wrapper;
+
+describe('UIOptionsPanel', () => {
+  beforeEach(() => {
```
bkemper: I really want to get rid of this globals+beforeEach pattern in our tests.  Instead, define a `subject`.

```
const subject = (props = {}) => shallow(
  <UIOptionsPanel 
    hideTermSubEnabled={false} // defines required params for base test case
    updateAccount={() => {}} 
    {...props} // allows overrides for each test case
  />
);
```

src/pages/account/components/tests/UIOptionsPanel.test.js
```diff
@@ -0,0 +1,31 @@
...
+    wrapper = shallow(<UIOptionsPanel {...props} />);
+  });
+
+  it('renders', () => {
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('renders toggle correctly if option enabled', () => {
+    props.hideTermSubEnabled = true;
+    wrapper = shallow(<UIOptionsPanel {...props} />);
```
bkemper: With subject, no mutating and rendering from test.  Instead,  ...

```
const wrapper = subject({ hideTermSubEnabled: true });
expect(wrapper.find('Toggle')).toHaveProp('checked', true);
```

src/pages/account/components/UIOptionsPanel.js
```diff
@@ -0,0 +1,55 @@
...
+            </Grid.Column>
+          </Grid>
+        </Panel.Section>
+      </Panel>
+    );
+  }
+}
+
+const mapStateToProps = (state) => ({
+  hideTermSubEnabled: isAccountUiOptionSet('hideTerminatedSubaccounts')(state),
```
bkemper: Should use the `selectCondition` selector to assure access control is loaded.

src/pages/account/components/UIOptionsPanel.js
```diff
@@ -0,0 +1,55 @@
...
+      <Panel title='UI Options'>
+        <Panel.Section>
+          <Grid>
+            <Grid.Column xs={11}>
+              <LabelledValue label='Hide Subacounts'>
+                <p>Hide terminated subacounts. Resources associated with terminated subaccounts can still be accessed.</p>
+              </LabelledValue>
+            </Grid.Column>
+            <Grid.Column xs={1}>
+              <Toggle
```
bkemper: I could see us adding new options in the near future.  Since options all have the same structure, 'key', 'label', 'description', and a boolean 'value', what do you think about extracting this out into a collection?

```
const options = [
  {
    key: 'hideTerminatedSubaccounts',
    label: 'Hide Subacounts',
    description: 'Hide terminated subacounts. Resources associated with terminated subaccounts can still be accessed.'
  }
  ... 
];
```

src/pages/account/components/UIOptionsPanel.js
```diff
@@ -0,0 +1,55 @@
...
+
+export class UIOptionsPanel extends Component {
+
+
+  setUIOption = () => {
+    const { updateAccount, hideTermSubEnabled } = this.props;
+    const body = {
+      options: {
+        ui: {
+          hideTerminatedSubaccounts: Boolean(!hideTermSubEnabled)
```
bkemper: Do we need `Boolean`?

src/pages/account/components/UIOptionsPanel.js
```diff
@@ -0,0 +1,55 @@
...
+    return (
+      <Panel title='UI Options'>
+        <Panel.Section>
+          <Grid>
+            <Grid.Column xs={11}>
+              <LabelledValue label='Hide Subacounts'>
+                <p>Hide terminated subacounts. Resources associated with terminated subaccounts can still be accessed.</p>
+              </LabelledValue>
+            </Grid.Column>
+            <Grid.Column xs={1}>
```
bkemper: Instead of Grid, we should flexbox it.

Screen Shot 2019-07-26 at 12 40 24 PM

src/pages/account/components/UIOptionsPanel.js
```diff
@@ -0,0 +1,55 @@
...
+  setUIOption = () => {
+    const { updateAccount, hideTermSubEnabled } = this.props;
+    const body = {
+      options: {
+        ui: {
+          hideTerminatedSubaccounts: Boolean(!hideTermSubEnabled)
+        }
+      }
+    };
+    updateAccount(body);
```
bkemper: The update account reducer spreads `body` which will overwrite everything in `account.options`.  I think we need an `updateAccountOption` action.

src/pages/account/components/UIOptionsPanel.js
bkemper:
```

.ToggleRow {
  display: flex;

  :first-child {
    flex-grow: 1;
    padding-right: spacing();
  }
}
```

```
<LabelledValue label='Hide Subacounts'>
  <div className={styles.ToggleRow}>
    <div>
      Hide terminated subacounts. Resources associated with terminated subaccounts can still be accessed.
    </div>
    <div>
      <Toggle
        id="hideTerminatedSubaccounts"
        checked={hideTermSubEnabled}
        disabled={loading}
        onChange={this.setUIOption}
      />
    </div>
  </div>
</LabelledValue>
```

src/pages/account/components/UIOptionsPanel.js
PatrickSison: As updateAccount uses the data from request to modify state, would it be better to keep the same spread + object.assign on the options? Vs making a completely separate action? Not sure if one has advantages over other

PatrickSison: Actually,` ui` is a nested object so it would still get overwritten with object.assign

bkemper: I think it would be safer to make a new action.  Adding more complexity to `updateAccount`, even a deep merge, would leave us in a vulnerable state in the long run.   How about...

```
const setAccountOption = (key, value) => (
  sparkpostApiRequest({
    type: 'SET_ACCOUNT_OPTION',
    meta: {
      method: 'PUT',
      url: '/v1/account',
      data: {
         options: {
           ui: {
             [key]: value
          }
        }
      }
    }
  })
)
```

bkemper:
Before merging, we need to check if accounts with status='active' and compliance_status='terminated' should be hidden.

PR #1112, opened 2019-07-22
EO-893 - Add soft and hard bounce metrics as options

kchu1337:
Added Soft and hard bounce rate as a selectable metric. 

### What Changed
- Added hard and soft bounces to to constants file and added them  

### How To Test
```
    location /api/v1/alerts {
      include cors;
    }
```
 - Create an alert. Verify that soft and hard bounces are options in the alert metric dropdown.
 - Also verify that the evaluator has the correct suffix and options.
 - In the table, the new alert should have the tag with a friendly name. 
 - Click on the alert, the alert should show the correct details.


jonambas:
May want to change the copy on the list page too underneath the title

`Use alerts to be notified when important changes occur in your Health Score, bounce rates, and email usage.`

PR #1113, opened 2019-07-22
AC-885: Frontend login support via access token (issued via the /authenticate API)

PatrickSison:
[AC-885](https://jira.int.messagesystems.com/browse/AC-885)

### What Changed
 - Added a way to login to the web app with the url `/auth?token={token}&username={username}`

### How To Test
 - Open network tools and inspect first `/authenticate` request and grab token
 - Clear cookies/store
 - Go to url `/auth?token={token}&username={username}`
    - If it has a plus symbol in the username, replace with `%2B`
 - Verify that it authenticates and goes through
 - Verify on account where TFA is enabled
 - Verify on account where TFA is required but not enabled on user

### Notes
  - I removed the search params on RedirectAfterLogin as I don't see how it's being used right now and it clears up token/username in url after login. **Please let me know if this isn't the case** or if you think it doesn't need to be removed.

### To Do
- [ ] Address feedback

### DIFF
--- a/src/actions/auth.js
+++ b/src/actions/auth.js
@@ -40,23 +40,28 @@ export function login({ authData = {}, saveCookie = false }) {
  * within redux store
  *
  */
-export function authenticate(username, password, rememberMe = false) {
+export function authenticate(username, password, rememberMe = false, access_token) {
   // return a thunk
   return (dispatch, getState) => {
     const { loggedIn } = getState().auth;
-
+    const isTokenLogin = !!access_token;
     if (loggedIn) {
       return;
     }
 
     dispatch({ type: 'LOGIN_PENDING' });
 
-    return sparkpostLogin(username, password, rememberMe)
+    const maybeLogin = isTokenLogin ? Promise.resolve({ data: { access_token }}) : sparkpostLogin(username, password, rememberMe);
+
+    return maybeLogin
       .then(({ data = {}} = {}) => {
         const authData = { ...data, username };
 
+        //Skips website login if token login
+        if (!isTokenLogin) {
+          dispatch(websiteAuth.authenticate(username, password, rememberMe));
+        }
         // Start website auth token cookie setup process
-        dispatch(websiteAuth.authenticate(username, password, rememberMe));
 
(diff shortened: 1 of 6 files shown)


src/pages/auth/AuthPage.js
```diff
@@ -11,6 +12,15 @@ import RedirectAfterLogin from './components/RedirectAfterLogin';
 import { TFA_ROUTE, ENABLE_TFA_AUTH_ROUTE, SSO_AUTH_ROUTE } from 'src/constants';
 
 export class AuthPage extends React.Component {
+
+  componentDidMount() {
+    const { username, access_token, authenticate } = this.props;
+    // username required as we don't have context for who's token
```
bkemper: Maybe describe why this is needed?

src/actions/auth.js
```diff
@@ -40,23 +40,28 @@ export function login({ authData = {}, saveCookie = false }) {
...
     dispatch({ type: 'LOGIN_PENDING' });
 
-    return sparkpostLogin(username, password, rememberMe)
+    const maybeLogin = isTokenLogin ? Promise.resolve({ data: { access_token }}) : sparkpostLogin(username, password, rememberMe);
+
+    return maybeLogin
       .then(({ data = {}} = {}) => {
         const authData = { ...data, username };
 
+        //Skips website login if token login
```
bkemper: Describe why this is acceptable

src/pages/auth/components/RedirectAfterLogin.js
```diff
@@ -6,9 +6,8 @@ import { DEFAULT_REDIRECT_ROUTE } from 'src/constants';
  * Redirect to the default route.
  * This is meant for use immediately after authentication completes.
  */
-export const RedirectAfterLogin = (props) => {
-  const route = { ...props.location, pathname: DEFAULT_REDIRECT_ROUTE };
-
+export const RedirectAfterLogin = ({ location: { state } = {}}) => {
+  const route = { state, pathname: DEFAULT_REDIRECT_ROUTE };
```
bkemper: 👍

PR #1114, opened 2019-07-22
SA-737 Custom Reports

jonambas:
### What Changed
- Adds ability to save summary report filters and options
- Report options are saved through the users API under `options.ui.customReports`.
- Each report saves `name` and `url` which contains all options.
- `name` is limited to 24 characters
- Saved reports is limited to 10
- Closes `SA-732` as well
- Fixes a bug where the update ui options reducer overwrote existing options in redux

### How To Test
- Add the `feature_reports_enhancements: true` user ui option to your user account
- Visit summary report

### To Do
- [x] Write tests


src/pages/reports/components/CustomReportsList.js
```diff
@@ -0,0 +1,61 @@
...
+  return (
+    <>
+      <Button
+        flat
+        color='orange'
+        size='small'
+        onClick={() => setOpen(!open)}>
+          Saved Reports
+      </Button>
+      <Modal
```
PatrickSison: Do we want a close button for the modal?

src/pages/reports/components/CustomReports.js
```diff
@@ -0,0 +1,108 @@
...
+
+  const isLimited = reports.length >= REPORT_LIMIT;
+  const canOverwrite = _.find(reports, ['name', name.trim()]);
+  const saveDisabled = name === '' || loading;
+
+  return (
+    <Panel.Section>
+      <div className={styles.Wrap}>
+        <div className={styles.NameWrapper}>
+          <TextField
```
PatrickSison: Do we want to disable the field if it's at the limit? If it's enabled, it makes it seem like I'm renaming the report rather than saving it under a different name.

src/pages/reports/components/CustomReportsList.js
PatrickSison: Because of the way the modal renders, there's a certain width on the outside of the white area (to left and right) that can be clicked without triggering onClose

src/pages/reports/components/ReportOptions.js
```diff
@@ -48,9 +49,11 @@ export class ReportOptions extends Component {
   };
 
   render() {
-    const { typeaheadCache, reportOptions, reportLoading, refreshReportOptions, searchOptions, roundToPrecision } = this.props;
+    const { customReportsEnabled, typeaheadCache, reportOptions, reportLoading, refreshReportOptions, searchOptions, roundToPrecision } = this.props;
+
     return (
       <Panel>
+        {customReportsEnabled && <CustomReports searchOptions={searchOptions} />}
```
PatrickSison: Can you add a small test for this?

src/selectors/customReports.js
```diff
@@ -0,0 +1,10 @@
+import { selectCondition } from 'src/selectors/accessConditionState';
+import { isUserUiOptionSet } from 'src/helpers/conditions/user';
+import _ from 'lodash';
+
+export const selectReportsEnhancementsEnabled = (state) => selectCondition(isUserUiOptionSet('feature_reports_enhancements', true))(state);
+
+export const selectCustomReports = (state) => {
+  const reports = _.get(state, 'currentUser.options.ui.customReports') || [];
```
PatrickSison: Is there a difference between doing ` || []` vs using the 3rd param as a default in `_.get()`? I'm not familiar myself

src/actions/customReports.js
```diff
@@ -0,0 +1,33 @@
...
+
+    dispatch(updateUserUIOptions({
+      customReports: deduped
+    }));
+  };
+}
+
+export function deleteReport(name) {
+  return (dispatch, getState) => {
+    const reports = _.get(getState(), 'currentUser.options.ui.customReports') || [];
```
PatrickSison: Same as other. It's pedantic though, so not a necessary change

src/pages/reports/components/CustomReports.js
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    saveReport({
+      name: name.trim(),
+      url: qs.stringify(toSave)
+    });
+  }
+
+  function handleLoad(report) {
+    const { options, filters = []} = parseSearch(report.url);
```
PatrickSison: On the note of parseSearch, could the search potentially bug the same way with a Subaccount with a colon in the name? SA-725

src/selectors/customReports.js
jonambas: There is, but i should make it more clear. When there are no saved tests, the value stored in the api is `false`, for some reason passing it `[]` doesn't update it. Without `|| []` the value `false would be returned`

src/pages/reports/components/CustomReports.js
jonambas: Not sure we want to disable it since the input sort of functions as a header and a field at the same time.

The condition below this `canOverwrite && isLimited` lets the user save over an existing report (same name) with different settings. I could simplify though if its too confusing though, and force users to delete a report before saving a new one

src/pages/reports/components/CustomReportsList.js
jonambas: I ended up just removing the extra container, and using the showCloseButton prop. And made some other css changes for accessibility

src/pages/reports/components/tests/__snapshots__/CustomReportList.test.js.snap
```diff
@@ -0,0 +1,40 @@
...
+  </Button>
+  <Modal
+    onClose={[Function]}
+    open={false}
+    showCloseButton={true}
+  >
+    <Panel
+      title="Saved Reports"
+    >
+      <ReportRow
```
PatrickSison: Haven't done a deep dive into hooks myself. Glad useMemo has a nice snapshot w/ shallow

PatrickSison: I should definitely start adopting hooks a lot more

PR #1118, opened 2019-07-25
SE-118: Upload a file without validating

PatrickSison:
[SE-118](https://jira.int.messagesystems.com/browse/SE-118)

### What Changed
 - Added an alternative upload (no actual action yet)
 - Added UploadedListPage
 - Added UploadedListForm
 - Added routing back.

### How To Test
 - Update with UI option `recipientValidationV2`
 - Check that the upload navigates to the newer page instead of doing a real upload
 - Check that it aligns with the designs

### To Do
- [ ] Address feedback


src/pages/recipientValidation/components/UploadedListForm.module.scss
```diff
@@ -0,0 +1,53 @@
...
+  font-weight: 100;
+  margin-top: .5em;
+  margin-bottom: 2rem;
+
+  .Number {
+    font-weight: 500;
+  }
+}
+
+.RVColumn {
```
kchu1337: Where is this being used?

src/pages/recipientValidation/components/UploadedListForm.js
```diff
@@ -0,0 +1,73 @@
...
+    >
+      <div style={{ float: 'right' }}>
+        <Button onClick={this.handleModal(false)} flat>
+          <Close />
+        </Button>
+      </div>
+      <div className={styles.bodyContainer}>
+        <h3>How was this calculated?</h3>
+        <RecipientValidationPriceTable
+          cellProps={{
```
kchu1337: Can you pass in the style from your scss? Like 
```js
cellProps={{className: styles.Foo}}
```

src/pages/recipientValidation/components/ListForm.js
```diff
@@ -26,15 +28,27 @@ export class ListForm extends Component {
...
-    const { file, handleSubmit, listError, resetUploadError } = this.props;
+    const { file, handleSubmit, listError, resetUploadError, newListUpload } = this.props;
 
     // Redux form validation does not run in the same render cycle after Field's onChange,
     // thus checking props.valid would not work here *shakes fist*
     const valid = !maxFileSize(config.maxRecipVerifUploadSizeBytes)(file) && !fileExtension('csv', 'txt')(file);
 
     if (file && valid && !listError) {
-      handleSubmit(this.handleUpload)();
+      if (newListUpload) {
```
kchu1337: You can use a ternary here like
```js
      const uploadFn = newListUpload ? this.handleNewUpload : this.handleUpload
      handleSubmit(uploadFn)()
```

src/pages/recipientValidation/components/FileUploadWrapper.js
```diff
@@ -28,11 +28,11 @@ export default class FileUploadWrapper extends Component {
 
   render() {
     const { input, meta, uploading } = this.props;
-
     if (uploading) {
       return (
         <fieldset className={styles.Field}>
           <h3 className={styles.Header}>Uploading...</h3>
+          <p>Your list will be ready to validate in just a moment.</p>
```
kchu1337: Will this show up for users without V2?

src/pages/recipientValidation/UploadedListPage.js
```diff
@@ -0,0 +1,82 @@
...
+    // TODO: Use action
+    getJobStatusMock();
+    // TODO: Get usage
+    getUsage();
+
+  }
+
+
+  renderTitle = (date = Date.now()) => ( //TODO: Remove pre-rendered date
+    <div style={{ fontSize: '1rem', fontWeight: 200 }}>
```
kchu1337: Do we have stylistic rules about inline css? If so lets move it to its own scss file.

src/pages/recipientValidation/components/UploadedListForm.js
PatrickSison: I could. That's a good idea

src/pages/recipientValidation/components/ListForm.js
PatrickSison: Only reason it's kept separate is there's a later ticket to remove the old upload later, and i didn't want to mix them too closely

src/pages/recipientValidation/UploadedListPage.js
PatrickSison: 👍

src/actions/recipientValidation.js
```diff
@@ -33,13 +69,25 @@ export function getLatestJob() {
 }
 
 export function getJobStatus(list_id) {
-  return sparkpostApiRequest({
-    type: 'GET_JOB_STATUS',
-    meta: {
-      method: 'GET',
-      url: `v1/recipient-validation/job/${list_id}`
+  //TODO: Unmock action + fix tests
```
PatrickSison: Need to use real based on flag since we can't affect existing flow

kchu1337: Oh wow. Will it be a hard cutover or will people be using V1 and V2? If it's a hard cutover, lets just make the new one a separate action. Like how you had it before but with a payload.

PatrickSison: It works as it is now, with the check on the UI option. You can test between appteam and the account with the flag to double check that the correct action is being taken

PatrickSison: API stubs will be removed/replaced in SE-157, and the cutover will happen in SE-156 on removing the feature flags.

PR #1126, opened 2019-07-31
SE-119: Validate an uploaded file

PatrickSison:
### What Changed
 - User can trigger batch job on uploaded file
 - User can see progress of file after upload

### How To Test
 - Check SE-118 PR for enabling feature flag
 - Upload a file and should automatically go to progress page

### To Do
- [ ] Address feedback

### DIFF
--- a/src/actions/recipientValidation.js
+++ b/src/actions/recipientValidation.js
@@ -46,8 +46,18 @@ export function uploadListNew(data) {
 }
 
 //TODO: wait for endpoint
-export function triggerJob(listId) {
-  return { type: 'TRIGGER_JOB' };
+export function triggerJob(list_id) {
+  return {
+    type: 'TRIGGER_JOB_SUCCESS',
+    payload: {
+      list_id: list_id,
+      batch_status: 'checking_regex',
+      complete: false,
+      original_filename: 'test.csv',
+      address_count: 1234,
+      upload_timestamp: 1565187194
+    }
+  };
   // return sparkpostApiRequest({
   //   type: 'TRIGGER_JOB',
   //   meta: {
@@ -93,7 +103,7 @@ export function getJobStatusMock(list_id) {
     type: 'GET_JOB_STATUS_SUCCESS',
     payload: {
       list_id: list_id,
-      batch_status: 'checking_regex',
+      batch_status: 'queued_for_batch',
       complete: false,
       original_filename: 'test.csv',
       address_count: 1234,
--- a/src/actions/recipientValidation.js
+++ b/src/actions/recipientValidation.js
@@ -93,7 +93,7 @@ export function getJobStatusMock(list_id) {
     type: 'GET_JOB_STATUS_SUCCESS',
     payload: {
       list_id: list_id,
-      batch_status: 'queued_for_batch',
+      batch_status: 'checking_regex',
       complete: false,
       original_filename: 'test.csv',
       address_count: 1234,
--- a/src/actions/recipientValidation.js
+++ b/src/actions/recipientValidation.js
@@ -1,7 +1,7 @@
 import sparkpostApiRequest from 'src/actions/helpers/sparkpostApiRequest';
 
 // If you want to test fail mock response, pass Promise.reject() as another parameter
-const mockThunk = (action, response = Promise.resolve()) => (dispatch) => {
+const mockThunk = (action, response = Promise.resolve(action.payload || {})) => (dispatch) => {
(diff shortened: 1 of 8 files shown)


src/pages/recipientValidation/components/ListProgress.js
```diff
@@ -0,0 +1,57 @@
...
+  'performing_decay_scoring',
+  'performing_role_based_lookup',
+  'performing_disposable_domain',
+  'performing_free_email',
+  'performing_did_you_mean',
+  'uploading_results_to_s3'
+];
+
+const ProgressBar = ({ completed }) => {
+  if (completed > 100) {
```
kchu1337: Theres a function called lerp 😆 in `helpers/math`.

src/pages/recipientValidation/components/ListProgress.js
```diff
@@ -0,0 +1,57 @@
...
+  );
+};
+
+
+
+const ListProgress = ({ job }) => {
+
+  const { status, filename } = job;
+
+  const percentage = 100 * BATCH_STATUS.findIndex((batchStep) => batchStep === status) / BATCH_STATUS.length;
```
kchu1337: Isn't the 2nd part 0-9 and it gets divided by 10? Highest it will go is 90% right?

src/pages/recipientValidation/components/ListProgress.js
```diff
@@ -0,0 +1,57 @@
...
+  return (
+    <div style={{ width: '80%', margin: '20px' }}>
+      <h2>{filename}</h2>
+      <div style={{ marginBottom: 80 }}>
+        <span>Your list is validating. You can track its progress on the recipient validation </span>
+        <UnstyledLink to='/recipient-validation' component={Link}>home page</UnstyledLink>,
+        <span> we'll let you know when validation is complete and your results are ready.</span>
+      </div>
+      <div><strong>Status:</strong> Processing</div>
+      <ProgressBar style={{ marginBottom: '400px' }} completed={percentage}/>
```
kchu1337: Can this be moved to styles.ProgressBar?

src/pages/recipientValidation/components/ListProgress.js
kchu1337: Actually, it's probably better that way. Disregard this comment if its working as intended.

src/pages/recipientValidation/UploadedListPage.js
```diff
@@ -2,6 +2,7 @@ import React, { Component } from 'react';
 import { Page, Panel } from '@sparkpost/matchbox';
 import { Link, withRouter } from 'react-router-dom';
 import UploadedListForm from './components/UploadedListForm';
+import ListProgress from './components/ListProgress';
```
kchu1337: Won't let me add a comment there but on line 35, you can probably just import the actual action and do the same as in SE-118 and not use sparkpostAPIRequest and just mock out a dispatched event

src/pages/recipientValidation/components/ListProgress.js
PatrickSison: What I will probably do is have a check if completed (another field in job) is true, then I'll put the percentage as 100

src/pages/recipientValidation/UploadedListPage.js
```diff
@@ -12,18 +12,66 @@ import _ from 'lodash';
...
+    triggerJob(listId).then(() => {
+      startPolling({
+        key: listId,
+        action: () => this.handlePoll(listId),
+        interval: 5000
+      });
+    });
+  }
+
+  handlePoll = (id) => {
```
kchu1337: Do we want to pass id or just get it from props? Just worried we might be polling for a old id rather than the most recent one if we switch pages? Would at least want to check this case when API comes out

src/actions/recipientValidation.js
```diff
@@ -1,7 +1,7 @@
 import sparkpostApiRequest from 'src/actions/helpers/sparkpostApiRequest';
 
 // If you want to test fail mock response, pass Promise.reject() as another parameter
-const mockThunk = (action, response = Promise.resolve()) => (dispatch) => {
+const mockThunk = (action, response = Promise.resolve(action.payload || {})) => (dispatch) => {
```
kchu1337: Could make it `...response = Promise.resolve ...` and then return `response(action.payload)`. I think it would make it easier when passing a `response` parameter. But I don't see that being used anywhere so it might not be necessary.

PR #1127, opened 2019-08-01
SA-711: Spam Trap Hits by Type are viewable in the UI

bkemper:
Refer to [SA-711](https://jira.int.messagesystems.com/browse/SA-711)

### What Changed

Screen Shot 2019-08-01 at 5 01 46 PM

- Converted bar chart to stacked bar chart
- Added a legend
- Only displayed rate or count in tooltip (instead of both)
- Added types to tooltip

Screen Shot 2019-08-02 at 10 34 45 AM

- Converted preview bart chart to stacked bar chart

### How To Test

_Our staging account does not have spam trap data, so ping me and I'll give you steps to get an account with data._

- Visit /signals/spam-traps/sid/1
- Confirm stacked bar chart is rendering for both "Count" and "Ratio"
- Confirm clicking on a bar changes the recommendations
- Confirm tooltip is providing the correct numbers for both "Count" and "Ratio"


src/pages/signals/constants/spamTrapHitTypes.js
```diff
@@ -0,0 +1,20 @@
+import { toCollection } from 'src/helpers/object';
+
+const spamTrapHitTypes = {
+  trap_hits_parked: {
+    fill: '#50CFDA',
+    label: 'Parked'
```
PatrickSison: Are these industry terms? Or do we need to include some helper/popover text?

bkemper: They are industry terms.  I just added descriptions.

src/pages/signals/SpamTrapPage.js
```diff
@@ -97,19 +115,32 @@ export class SpamTrapPage extends Component {
...
+                        fill
+                      }))
+                      .reverse()
+                  }
+                  yAxisProps={this.getYAxisProps()}
+                  xAxisProps={this.getXAxisProps()}
+                />
+                <Legend
+                  items={spamTrapHitTypesCollection}
+                  tooltipContent={(label) => spamTrapHitTypesByLabel[label].description}
```
PatrickSison: A bit weird to have the data joined this way, but outside the scope of this

src/pages/signals/SpamTrapPage.js
```diff
@@ -97,19 +115,32 @@ export class SpamTrapPage extends Component {
...
-                timeSeries={data}
-                onMouseOver={handleDateHover}
-                onMouseOut={resetDateHover}
-                hovered={hoveredDate}
-                tooltipContent={this.getTooltipContent}
-                yKey={calculation === 'absolute' ? 'trap_hits' : 'relative_trap_hits'}
-                yAxisProps={this.getYAxisProps()}
-                xAxisProps={this.getXAxisProps()}
-              />
+              <div className='LiftTooltip'>
```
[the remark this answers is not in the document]
bkemper: I really don't support this, I just copied it from engagement recency since I was using a similar tooltip.  I think we should always use modules.

src/pages/signals/SpamTrapPage.js
bkemper: I agree.  I would expect an item from items to be provided as the argument, but what can you do.

PR #1131, opened 2019-08-05
AC-630: Ignore free plans in brightback

PatrickSison:
### What Changed
 - Show downgrade, but ignore brightback on free to free plan change

### How To Test
 - On account with deprecated free plan, downgrade to current free plan. Should not enable brightback

### To Do
- [ ] Address any feedback

### DIFF
--- a/src/pages/billing/components/Confirmation.js
+++ b/src/pages/billing/components/Confirmation.js
@@ -68,7 +68,8 @@ export class Confirmation extends React.Component {
     const currentPlanPricing = getPlanPrice(current);
     const selectedPlanPricing = getPlanPrice(selected);
     const isPlanSelected = current.code !== selected.code;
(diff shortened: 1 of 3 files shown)


src/pages/billing/components/Confirmation.js
```diff
@@ -136,7 +137,7 @@ export class Confirmation extends React.Component {
         </Panel.Section>
         <Panel.Section>
           <Brightback
-            condition={Boolean(billingEnabled && isPlanSelected && selected.isFree)}
+            condition={Boolean(billingEnabled && isPlanSelected && selected.isFree && !current.isFree)}
```
bkemper: How can you have `billingEnabled` and `current.isFree`?

PatrickSison: Billing enabled currently is a prop passed down from the form checking if it's self serve billing.

PR #1128, opened 2019-08-01
SA-769 - Inbox Placement Test Details Page (Details Tab)

rajumsys:
Implements details tab of Single Inbox Placement Test tab

### Tests
- Run seedlist api service locally and modify the `GET /inbox-placement/{id}` response to respond something conforming API Spec [1](https://github.com/SparkPost/sparkpost-admin-api-documentation/blob/master/services/inbox_placement_api.md#get-an-inbox-placement-test-get-inbox-placementid) [2](https://github.com/SparkPost/sparkpost-admin-api-documentation/blob/master/services/inbox_placement_api.md#get-inbox-placement-test-results-by-mailbox-provider-get-inbox-placementidmailbox-provider). Reach out to me if you need more info about that. 
- Then visit something like `/inbox-placement/details/101`. ID needs to exist but probably doesn't matter unless you respond based on that. 

### Notes:
- I've put some smaller components in the same module. This saves from all the boilerplate (component and tests) as it's intended for local use (just the other component in the same file). Let me know if you object it. 
- Diff will be reduced on #1130 is merged
- Content tab will be rendered in SA-771

### Todos:
- Use the green/active status component that @21042jim  is working one 
- Get 👍 from Aubrey for design
  - > for hte grid can we offer more space between the from and started and less between started, finished, and name?
- Add any missing tests (waiting for coverage) 
- The provider name in the breakdown needs to be linked to seed details page. but as we don't have the details page yet, i'll add a todo in the respective ticket (which I still need to find).

PR #1137, opened 2019-08-08
SA-771 - Inbox Placement Test Content 

kchu1337:
Implements Content tab of Single Inbox Placement Test

### Tests
- Run seedlist api service locally and modify the `GET /inbox-placement/{id}/content` response to respond something conforming API Spec [1](https://github.com/SparkPost/sparkpost-admin-api-documentation/blob/master/services/inbox_placement_api.md#get-the-content-for-an-inbox-placement-test-get-inbox-placementidcontent). 
- Then navigate to something like `/inbox-placement/details/101` in the URL bar.  DOesn;t need to be `101`. Any number should work depending on how you mock the api response earlier. 
- Navigate to the `content` tab within the page. 

### Todos:
- Use the green/active status component

### DIFF
--- a/src/pages/inboxPlacement/components/TestContent.js
+++ b/src/pages/inboxPlacement/components/TestContent.js
@@ -1,9 +1,8 @@
 import React from 'react';
 import { Panel, CodeBlock, Tabs, Grid } from '@sparkpost/matchbox';
 import useTabs from 'src/hooks/useTabs';
-import styles from './TestContent.module.scss';
-
 import { formatBytes } from 'src/helpers/units';
+import InfoBlock from './InfoBlock';
 
 const TABS = [
   { content: 'Raw Message', key: 'raw_message' },
@@ -20,27 +19,20 @@ const TestContent = ({ content, details }) => {
 
   return (
     <>
-    <Panel title={details.subject} sectioned>
+    <Panel sectioned>
+      <h2>{details.subject}</h2>
       <Grid>
-        <Grid.Column sm={12} md={3}>
-          From:
-          <br/>
-          <span className={styles.FromAddress}>{from_address}</span>
-        </Grid.Column>
-        <Grid.Column xs={12} md={3}>
-          Raw Message Size:
-          <br/>
-          {formatBytes(message_size)}
-        </Grid.Column>
+        <InfoBlock value={from_address} label='From' columnProps={{ sm: 12, md: 4 }}/>
+        <InfoBlock value={formatBytes(message_size)} label='Raw Message Size:' columnProps={{ sm: 12, md: 4 }}/>
       </Grid>
     </Panel>
-      <Panel>
-        <Tabs selected={selectedTabIndex} tabs={tabs} />
-        <Panel.Section>
-          <CodeBlock code={content[selectedTabKey] || ''}/>
-        </Panel.Section>
-      </Panel>
-      </>
+    <Panel>
+      <Tabs selected={selectedTabIndex} tabs={tabs} color='navy' />
+      <Panel.Section>
+        <CodeBlock code={content[selectedTabKey] || ''}/>
+      </Panel.Section>
+    </Panel>
+    </>
   );
 };
 
--- a/src/pages/inboxPlacement/components/TestContent.js
+++ b/src/pages/inboxPlacement/components/TestContent.js
@@ -1,11 +1,48 @@
 import React from 'react';
-import { Panel } from '@sparkpost/matchbox';
+import { Panel, CodeBlock, Tabs, Grid } from '@sparkpost/matchbox';
+import useTabs from 'src/hooks/useTabs';
+import styles from './TestContent.module.scss';
 
-const TestContent = () => (
-  <Panel.Section>
-    <h1>Contents</h1>
-  </Panel.Section>
-);
+import { formatBytes } from 'src/helpers/units';
+
+const TABS = [
+  { content: 'Raw Message', key: 'raw_message' },
+  { content: 'HTML', key: 'html' },
+  { content: 'AMP HTML', key: 'amp' },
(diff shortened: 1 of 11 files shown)


src/pages/inboxPlacement/components/TestContent.js
```diff
@@ -1,11 +1,48 @@
...
+];
+
+const TestContent = ({ content, details }) => {
+  const [selectedTabIndex, tabs] = useTabs(TABS);
+  const selectedTabKey = tabs[selectedTabIndex].key;
+  const { from_address, message_size } = details;
+
+  return (
+    <>
+    <Panel title={details.subject} sectioned>
```
rajumsys: While I like this approach, I guess it adds a border below the title which does not exist in the wireframe. That's why I've added a separate header tag (h3)

src/pages/inboxPlacement/components/TestContent.js
```diff
@@ -1,11 +1,48 @@
...
+      <Grid>
+        <Grid.Column sm={12} md={3}>
+          From:
```
rajumsys: font size is different for label and value. i've a separate component for this which probably can be moved out to own file so that it can be used here https://github.com/SparkPost/2web2ui/pull/1128/files#diff-1ad0685c1de5fcb223b0221006ba9f11R9

however, next one (Valid list unsubscribe) is different. so we can customize that component to pass icon and render accordingly.

src/pages/inboxPlacement/components/TestContent.js
```diff
@@ -1,11 +1,48 @@
...
+          <br/>
+          <span className={styles.FromAddress}>{from_address}</span>
+        </Grid.Column>
+        <Grid.Column xs={12} md={3}>
+          Raw Message Size:
```
rajumsys: checkbox is missing!

src/pages/inboxPlacement/components/TestContent.js
```diff
@@ -1,11 +1,48 @@
...
+          <br/>
+          {formatBytes(message_size)}
+        </Grid.Column>
+      </Grid>
+    </Panel>
+      <Panel>
+        <Tabs selected={selectedTabIndex} tabs={tabs} />
```
rajumsys: may be we'll need some styling for tabs to match design, though IMO they should be implemented inside Tab so they are consistent.

src/pages/inboxPlacement/components/tests/TestContent.test.js
```diff
@@ -0,0 +1,31 @@
+import { shallow } from 'enzyme';
+import React from 'react';
+
+import TestContent from '../TestContent';
+
+
+describe('Inbox Placement Test Content', () => {
```
rajumsys: name it after the component name?

src/pages/inboxPlacement/components/TestContent.js
```diff
@@ -1,11 +1,48 @@
...
+          <span className={styles.FromAddress}>{from_address}</span>
+        </Grid.Column>
+        <Grid.Column xs={12} md={3}>
+          Raw Message Size:
+          <br/>
+          {formatBytes(message_size)}
+        </Grid.Column>
+      </Grid>
+    </Panel>
+      <Panel>
```
rajumsys: indent out

src/pages/inboxPlacement/components/tests/TestContent.test.js
```diff
@@ -0,0 +1,31 @@
...
+    details: mockDetails,
+    content: mockContent
+  };
+
+  it('renders page correctly with defaults', () => {
+    const wrapper = shallow(<TestContent {...props} />);
+    expect(wrapper).toMatchSnapshot();
+  });
+});
+
```
rajumsys: one more test with different active tab will be good addition.

src/pages/inboxPlacement/tests/TestDetailsPage.test.js
```diff
@@ -29,9 +33,10 @@ describe('Page: Single Inbox Placement Test', () => {
     expect(wrapper).toMatchSnapshot();
   });
 
-  it('calls getInboxPlacementTest on load', () => {
+  it('dispatches both action calls on load', () => {
     wrapper = mount(<TestDetailsPage {...props} />);
     expect(mockGetTest).toHaveBeenCalled();
+    expect(mockGetTestContent).toHaveBeenCalled();
```
rajumsys: you've `mockGetTestContent` mocked globally which can (safe here) affect other tests that asserts same (if was called). so it's better to do it inside beforeEach so that they always get cleaned up. if you want to use this way, then please add an `afterEach` and reset the mock.

src/pages/inboxPlacement/components/TestContent.js
kchu1337: Checkbox should be for the valid unsubscribe link which is not part of this ticket.

src/pages/inboxPlacement/tests/TestDetailsPage.test.js
kchu1337: I just used the same format as yours

src/pages/inboxPlacement/components/TestContent.js
```diff
@@ -20,27 +19,20 @@ const TestContent = ({ content, details }) => {
...
     </Panel>
-      <Panel>
-        <Tabs selected={selectedTabIndex} tabs={tabs} />
-        <Panel.Section>
-          <CodeBlock code={content[selectedTabKey] || ''}/>
-        </Panel.Section>
-      </Panel>
-      </>
+    <Panel>
+      <Tabs selected={selectedTabIndex} tabs={tabs} color='navy' />
```
rajumsys: can we take this to css file and use `color(navy)`

PR #1138, opened 2019-08-08
SE-161 Alerts Pre-release polish 

kchu1337:
### What Changed
 - Made all the changes outlined in the AC

### How To Test
 - Confirm that all changes working as intended. 

### To Do
- [x] List | Tooltips should be dark and fix offset
- [x] List | “Active” -> “Mute
- [x] Edit | Breadcrumb and Updating should lead back to Details, not List
- [x] Details: Metric | “Alert Metric” should be “Metric"
- [x] Details: Metric | Monthly Sending Limit:  [Value] above [100]  ->  Above [100%]
- [x] Details: Metric | Bounce Rate: [Value] above [30] -> Above [20%]
- [x] Details: Evaluated | Make above/below bolded as well
- [x] Details: Evaluated | “Evaluated” -> “Condition”
- [x] Details: Filtered By | Delete Colons after items being filtered but make semi-bold. Keep “and”  not bold same with notify by
- [x] Details: Filtered By | Make sure the subaccount section doesn’t show up for accounts without subaccounts 
- [x] Details: Filtered By | Subaccounts [Master and all] shouldn’t appear, there is no filter in this case
- [x] Details: Notify | Remove icons from emails
- [x] Details: Duplicate | Fix: Doesn’t let you press create until the from has been changes
- [x] Details: Active | “Active” -> “Mute"
- [x] Create | Create should push to detail, not List
- [x] Create: Bounces | Remove Selector from Bounce rates 
- [x] Create: Slack | Integrate alerts into your Slack channel -> Integrate this alert with Slack
- [x] Create: Webhooks | “Create a webhook for this alert”

### DIFF
--- a/src/pages/alerts/ListPageNew.js
+++ b/src/pages/alerts/ListPageNew.js
@@ -73,7 +73,7 @@ export class ListPageNew extends Component {
                   <h3>{alert.name}</h3>
                 </Panel.Section>
                 <Panel.Section className = {styles.Footer}>
-                  <Tooltip content='View Details' width='100px'>
+                  <Tooltip dark content='View Details' width='100px' horizontalOffset='-8px'>
                     <Button flat component={Link} to = {`/alerts-new/details/${alert.id}`}><RemoveRedEye className = {styles.Icon}/></Button>
                   </Tooltip>
                 </Panel.Section>
--- a/src/pages/alerts/components/AlertDetails.js
+++ b/src/pages/alerts/components/AlertDetails.js
@@ -5,8 +5,9 @@ import LabelledValue from 'src/components/labelledValue/LabelledValue';
 import { MAILBOX_PROVIDERS } from 'src/constants';
 import styles from './AlertDetails.module.scss';
 import AlertToggle from './AlertToggleNew';
-import { EmailIcon, SlackIcon, WebhookIcon } from 'src/components/icons';
+import { SlackIcon, WebhookIcon } from 'src/components/icons';
 import { Link } from 'react-router-dom';
+import { getEvaluatorOptions } from '../helpers/alertForm';
 
 const extraChannels = [
   {
@@ -21,7 +22,7 @@ const extraChannels = [
   }
 ];
 
-export const AlertDetails = ({ alert, id, subaccountIdToString }) => {
+export const AlertDetails = ({ alert, id, subaccountIdToString, hasSubaccounts }) => {
   const { metric, channels = {}, filters = [], subaccounts = [], threshold_evaluator = {}, any_subaccount, muted } = alert;
   const { source, operator, value } = threshold_evaluator;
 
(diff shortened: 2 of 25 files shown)


src/pages/alerts/ListPageNew.js
```diff
@@ -73,7 +73,7 @@ export class ListPageNew extends Component {
                   <h3>{alert.name}</h3>
                 </Panel.Section>
                 <Panel.Section className = {styles.Footer}>
-                  <Tooltip content='View Details' width='100px'>
+                  <Tooltip dark content='View Details' width='100px' horizontalOffset='-8px'>
```
bkemper: The need for `horizontalOffset` makes me think there is a bug in matchbox or a missing recommendation.

bkemper: It looks like a bug.  Will we always have to set an offset for the tooltip to align correctly?

Screen Shot 2019-08-12 at 3 24 49 PM

src/pages/alerts/components/AlertDetails.js
```diff
@@ -46,27 +47,42 @@ export const AlertDetails = ({ alert, id, subaccountIdToString }) => {
...
-      <span key={'subaccounts'}>
-        Subaccounts: {getSubaccountsTags()}
-      </span>
-    );
-    const filtersTags = filters.map((filter) => (
+    //Shows subaccounts filter if the account has subaccounts and if the subaccount value is not `Master and all`
+    const shouldShowSubaccounts = hasSubaccounts && ((subaccounts.length > 0 && subaccounts[0] !== -1) || any_subaccount);
+    const subaccount = shouldShowSubaccounts
+      ? (
+        <span key={'subaccounts'}>
```
bkemper: Why is this needed?

src/pages/alerts/components/AlertDetails.js
```diff
@@ -78,18 +94,17 @@ export const AlertDetails = ({ alert, id, subaccountIdToString }) => {
     return (
       <>
         {emails && emails.length > 0 && (
-          <div key="email">
-            Email: {emails.map((email) => (
-              <Tag key={email} className={styles.TagsWithIcon}>
-                <EmailIcon className={styles.Icon}/>
+          <div id='email'>
```
bkemper: id or key?

bkemper:
For all the metrics that don't have the option to pick greater than or less than, should we include "Greater Than" as helptext or some where in the label?

Screen Shot 2019-08-14 at 11 02 39 AM

src/pages/alerts/components/tests/__snapshots__/AlertDetails.test.js.snap
```diff
@@ -22,149 +22,157 @@ exports[`Alert Details Component should render the alert details component corre
...
+      label="Condition:"
+    >
+      <Tag
+        className="FirstTag"
+      >
+        Value
+      </Tag>
+      <h6
+        className="BoldInline"
+      />
```
bkemper: This is interesting to have a empty dom element.

PR #1145, opened 2019-08-12
SE-117: UI to show a table of files uploaded in the last 10 days

PatrickSison:
### What Changed
 - Converted list results to list of list-jobs

### How To Test
 - Enable account ui option `recipientValidationV2`
 - Check that the list populates at the bottom

### To Do
- [ ] Address feedback


nicklemmon:
It looks like the UX team was hoping for a visual update on the button here, unless that was tabled in a side convo

### Running Locally
Screen Shot 2019-08-16 at 3 19 00 PM

### InVision
Screen Shot 2019-08-16 at 3 21 20 PM

This may have been discussed elsewhere so forgive me if I'm wrong here.

src/pages/recipientValidation/components/ListResultsCard.module.scss
```diff
@@ -16,14 +16,22 @@
...
+.Ready {
+  color: color(blue);
+}
+
 .PanelHeader {
   font-size: 1.2em;
 }
 
-.TableHeader {
+.TableHeader, .TableRow {
```
nicklemmon: I'm not sure why the bottom border styles were removed here? Removing from the `.TableHeader` makes sense, but the `.TableRow` appears to make the component inconsistent with other tables in the app, and with the UX mock:

Screen Shot 2019-08-16 at 3 26 02 PM

src/pages/recipientValidation/components/ListResultsCard.js
```diff
@@ -1,82 +1,100 @@
...
+        </Table.Cell>
+      </Table.Row>
+    );
   };
 
   return (
-    <Panel sectioned accent='gray' title={<div className={styles.PanelHeader}>Last Validation:</div>}>
+    <Panel sectioned title={<div className={styles.PanelHeader}>Recent Validations</div>}>
       <div className={styles.TableContainer}>
         <Table>
```
nicklemmon: Every `<table>` needs a valid `<caption>` that titles the table. This does not have to be visually available for all users, so you can make use of the Matchbox `visually-hidden()` mixin on that particular element.

* The mixin in question is here: https://github.com/SparkPost/matchbox/blob/master/packages/matchbox/src/styles/config/_accessibility.scss
* More information about using `<caption>`: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/caption

nicklemmon:
> It looks like the UX team was hoping for a visual update on the button here, unless that was tabled in a side convo
> 
> ### Running Locally:
> Screen Shot 2019-08-16 at 3 19 00 PM
> 
> ### InVision:
> Screen Shot 2019-08-16 at 3 21 20 PM
> 
> This may have been discussed elsewhere so forgive me if I'm wrong here.

I spoke with Lynn and she indicates this is OK as-is leveraging the matchbox `<Button/>`.

src/pages/recipientValidation/components/ListResultsCard.module.scss
nicklemmon: Resolved by leveraging `<TableCollection/>` component.

bkemper:
@nicklemmon This is ready for final review.

src/pages/recipientValidation/components/JobsTableCollection.module.scss
```diff
@@ -0,0 +1,22 @@
...
+  color: color(orange);
+}
+
+.LoadingWrapper {
+  padding: rem(100) 0 rem(100);
+  text-align: center;
+}
+
+.Complete {
+  color: color(green);
```
nicklemmon: I believe we are using a slightly different green here than we are in another, similar feature. This is probably why Patrick hardcoded it originally which makes sense. I'm going to try and track it down.

PR #1152, opened 2019-08-16
SE-193 Handle lowercase batch_status in UI

nicklemmon:
Resolves SE-193

### What Changed
* Normalized `batch_status` checking its value in the UI
* Updated relevant tests with more accurate mock data

### How To Test
* Go to `/recipient-validation/list`
* Upload the example .csv file
* Wait for processing to occur and check that the status in the table renders accurately

### To Do
- [x] Figure out why the POST isn't working locally to verify UI update
- [x] Incorporate feedback

### DIFF
--- a/src/reducers/recipientValidation.js
+++ b/src/reducers/recipientValidation.js
@@ -44,7 +44,7 @@ export default (state = initialState, { meta, payload, type }) => {
         jobResults: {
           ...state.jobResults,
           [payload.list_id]: {
-            status: payload.batch_status,
+            status: payload.batch_status.toLowerCase(),
             complete: payload.complete,
             uploaded: payload.upload_timestamp,
             rejectedUrl: payload.rejected_external_url
--- a/src/reducers/recipientValidation.js
+++ b/src/reducers/recipientValidation.js
@@ -44,7 +44,7 @@ export default (state = initialState, { meta, payload, type }) => {
         jobResults: {
           ...state.jobResults,
           [payload.list_id]: {
-            status: payload.batch_status && payload.batch_status.toLowerCase(),
+            status: payload.batch_status ? payload.batch_status.toLowerCase() : null,
             complete: payload.complete,
             uploaded: payload.upload_timestamp,
             rejectedUrl: payload.rejected_external_url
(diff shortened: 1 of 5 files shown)


src/reducers/recipientValidation.js
```diff
@@ -44,7 +44,7 @@ export default (state = initialState, { meta, payload, type }) => {
         jobResults: {
           ...state.jobResults,
           [payload.list_id]: {
-            status: payload.batch_status,
+            status: payload.batch_status.toLowerCase(),
```
bkemper: If you drop select a file for upload, then quickly refresh.

Screen Shot 2019-08-16 at 5 10 55 PM

bkemper: Also, shouldn't we make the same modification for "GET_JOB_STATUS_SUCCESS" action?

bkemper: Also, shouldn't we make the same modification for "GET_JOB_STATUS_SUCCESS" action?

nicklemmon: Nice catch. Think I know what to do.

nicklemmon: Alright, I added a null check and currently can't reproduce this issue afterwards:

*Before:*
`status: payload.batch_status.toLowerCase()`

*After:*
`status: payload.batch_status && payload.batch_status.toLowerCase()`

nicklemmon: Done.

nicklemmon:
OK @bkemper - ready for further review if you see other issues.

Thanks.

src/reducers/recipientValidation.js
```diff
@@ -44,7 +44,7 @@ export default (state = initialState, { meta, payload, type }) => {
         jobResults: {
           ...state.jobResults,
           [payload.list_id]: {
-            status: payload.batch_status,
+            status: payload.batch_status && payload.batch_status.toLowerCase(),
```
bkemper: Wouldn't this set the `status` to `false` when `batch_status` is undefined?

nicklemmon: hmm, for some reason I thought it would be `null` - let me verify.

nicklemmon: I'm seeing undefined actually in a little demo, but let me change it to a ternary just to be safer

PR #1154, opened 2019-08-19
Se 153

shrirupa:
Alert metric recommendations added to slider.

### What Changed
 - Recommended ticks set based on requirements on SE-153.
 - Default value of slider gets set to recommended value when new alert is created, in case of editing and duplicating an alert the value remains same as the old value.
 - matchbox version was updated to get the ticks property of Slider.

### To Do
Address feedback


src/pages/alerts/components/AlertForm.js
```diff
@@ -16,9 +16,7 @@ import withAlertForm from './AlertForm.container';
 
 // Helpers & Validation
 import { emails, ifStringPresent, maxLength, required } from 'src/helpers/validation';
-
 const metricsOptions = [{ value: '', label: 'Select Metric', disabled: true }, ...getOptions(METRICS)];
-
 const notificationChannelData = {
```
kchu1337: If you are getting a max line issue, lets move this into the constants.

src/pages/alerts/components/AlertForm.js
```diff
@@ -1,6 +1,6 @@
+/* eslint-disable max-lines */
```
kchu1337: See my comment below (above)?

src/pages/alerts/components/AlertForm.js
```diff
@@ -47,17 +45,22 @@ const notificationChannelData = {
...
     const formSpec = getFormSpec(event.target.value);
     const { defaultFieldValues } = formSpec;
-
     REALTIME_FILTERS.forEach((filter) => {
       change(filter, []);
     });
     change('single_filter', { filter_type: 'none', filter_values: []});
     defaultFieldValues.forEach(({ fieldName, fieldValue }) => {
-      change(fieldName, fieldValue);
+      if (fieldName === 'value') {
```
kchu1337: Instead of fighting the forEach loop, lets just add a new key `defaultRecommendedValue` or something and then you can do         
```js 
if (isNewAlert && !isDuplicate) {
          change('value`, defaultRecommendedValue);
        }
```

src/pages/alerts/components/AlertForm.js
```diff
@@ -47,17 +45,22 @@ const notificationChannelData = {
 
 export class AlertForm extends Component {
 
-  resetFormValues = (event) => {
+  resetFormValues = (event, isNewAlert, isDuplicate) => {
     const { change } = this.props;
     const formSpec = getFormSpec(event.target.value);
     const { defaultFieldValues } = formSpec;
-
     REALTIME_FILTERS.forEach((filter) => {
```
kchu1337: Sorry, this is unrelated to your ticket, but you can probably remove lines 52-55. This will then save your filters between metric changes. I'll double check but I don't think we need to reset filters anymore.

src/pages/alerts/components/AlertForm.js
```diff
@@ -120,14 +122,17 @@ export class AlertForm extends Component {
                     name='metric'
                     component={SelectWrapper}
                     options={metricsOptions}
-                    onChange={this.resetFormValues}
+                    onChange={(event) => this.resetFormValues(event, isNewAlert, isDuplicate)}
```
kchu1337: Generally, we want to avoid anon functions especially in forms due to how redux-form handles validation and stuff. If you need the 2 other props you can destructure them from `this.props` in the function itself.

src/pages/alerts/components/AlertForm.js
```diff
@@ -120,14 +122,17 @@ export class AlertForm extends Component {
...
                     validate={required}
                   />
                 </div>
                 {metric !== '' &&
                 <div className={styles.Evaluator}>
                   <EvaluatorFields
+                    key={isNewAlert && !isDuplicate ? metric : 'evaluator-field'}
```
kchu1337: Can it work with just ```key={metric}```? If not, this is fine.

src/pages/alerts/components/AlertForm.js
```diff
@@ -120,14 +122,17 @@ export class AlertForm extends Component {
...
                     disabled={submitting}
+                    isNewAlert={isNewAlert}
```
kchu1337: If you are always using these two in an AND conjunction, you can simplify the prop to pass down by something like 
```js
shouldUpdateRecommendation = {isNewAlert && !isDuplicate}
```

src/pages/alerts/components/fields/EvaluatorFields.js
```diff
@@ -5,42 +5,65 @@ import { SelectWrapper, TextFieldWrapper } from 'src/components/reduxFormWrapper
...
+  change,
+  isNewAlert,
+  isDuplicate
 }) => {
 
   const [sliderValue, setSliderValue] = useState(value);
 
+
   const changeValueField = (val) => {
+    setSliderValue(val);
```
kchu1337: Lets move this into a separate function. Something like `changeSliderAndValue`. You can call `changeValueField` and `changeSlider` from it. This works fine as is, but I feel like it's better to separate functionality. 
This is called when the slider changes. If we set the slider value after slider changes, I am worried it might make a loop. Like  slider onChange->changeValueField changes slider-> slider onChange ... ect

src/pages/alerts/components/fields/EvaluatorFields.js
```diff
@@ -5,42 +5,65 @@ import { SelectWrapper, TextFieldWrapper } from 'src/components/reduxFormWrapper
...
+    }
+  };
+
+  const setValueOnOperatorChange = (event) => {
+    if (isNewAlert && !isDuplicate) {
+      changeValueField(RECOMMENDED_METRIC_VALUE[metric][source][event.target.value]);
     }
   };
 
+  const getTicksRecommendedValue = (sourceOptions, operatorOptions) => sourceOptions.length > 1
```
kchu1337: See my comment on the constants file. But you might want to do ```_.get(RECOMMENDED_METRIC_VALUE, [metric,source,operator], 0)```  to be safe.

src/pages/alerts/components/fields/EvaluatorFields.js
```diff
@@ -51,7 +74,7 @@ export const EvaluatorFields = ({
             component={SelectWrapper}
             disabled={disabled}
             options={sourceOptions}
-            onChange={setOperatorOnSourceChange}
+            onChange={(event) => setOperatorOnSourceChange(event, metric)}
```
kchu1337: Same here about anon functions. You can destructure metrics from props.

src/pages/alerts/components/fields/EvaluatorFields.js
```diff
@@ -5,42 +5,65 @@ import { SelectWrapper, TextFieldWrapper } from 'src/components/reduxFormWrapper
...
+    setSliderValue(val);
     change(FORM_NAME, 'value', val);
   };
 
   const changeSlider = (event) => {
     setSliderValue(event.target.value);
   };
 
-  const setOperatorOnSourceChange = (event) => {
+  const setOperatorOnSourceChange = (event,metric) => {
```
kchu1337: Since it's also setting recommended value, let's rename to ```handleSourceChange```

src/pages/alerts/constants/formConstants.js
```diff
@@ -16,6 +16,21 @@ export const FILTERS_FRIENDLY_NAMES = {
   sending_ip: 'Sending IP'
 };
 
+export const RECOMMENDED_METRIC_VALUE = {
+  monthly_sending_limit: 80,
```
kchu1337: All of these metrics also have source and operator values. So if you make this 
```js
monthly_sending_limit: {
    raw: {
      gt: 80
    }
```
You won't have to all of the if statements.

package.json
```diff
@@ -56,8 +56,10 @@
   },
   "dependencies": {
     "@sparkpost/boomerang": "^1.0.6",
-    "@sparkpost/matchbox": "^3.5.3",
+    "@sparkpost/matchbox": "^3.6.0",
     "@sparkpost/matchbox-icons": "^1.1.12",
+    "acorn-jsx": "^5.0.1",
```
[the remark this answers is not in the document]
shrirupa: The others were the part of npm Warnings when I updated matchbox. I can uninstall them.

src/pages/alerts/components/AlertForm.js
shrirupa: Good point

src/pages/alerts/components/AlertForm.js
shrirupa: resolved as below

src/pages/alerts/components/AlertForm.js
shrirupa: point taken.

src/pages/alerts/components/AlertForm.js
shrirupa: I will keep this in mind.

src/pages/alerts/components/AlertForm.js
shrirupa: ok

src/pages/alerts/components/AlertForm.js
shrirupa: good point

src/pages/alerts/components/fields/EvaluatorFields.js
shrirupa: done

src/pages/alerts/components/fields/EvaluatorFields.js
shrirupa: done

src/pages/alerts/components/fields/EvaluatorFields.js
shrirupa: I have changed this for clarity of code but I don't think earlier code causes infinite loop, I am not sure why.

src/pages/alerts/components/fields/EvaluatorFields.js
shrirupa: done

src/pages/alerts/constants/formConstants.js
shrirupa: ok will change it, my thinking making it this way as I wanted it to resemble actual structure of different metrics.

src/pages/alerts/components/AlertForm.js
shrirupa: done

src/pages/alerts/components/fields/EvaluatorFields.js
```diff
@@ -87,7 +97,7 @@ export const EvaluatorFields = ({
           normalize={Math.abs}
           type='number'
           align='right'
-          onChange={changeSlider}
+          onChange={changeValueField}
```
kchu1337: You don't need this since the field is being controlled by redux-form.

src/pages/alerts/components/fields/tests/EvaluatorFields.test.js
```diff
@@ -64,13 +65,37 @@ describe('Evaluator Fields Component', () => {
   });
 
   it('changes slider value when value field value changes', () => {
-    expect(wrapper.find('Slider').prop('value')).toEqual(50);
-    wrapper.find({ name: 'value' }).simulate('change', { target: { value: 60 }});
-    expect(wrapper.find('Slider').prop('value')).toEqual(60);
+    wrapper.find({ name: 'value' }).simulate('change', 60);
+    expect(props.change).toHaveBeenCalledWith(FORM_NAME, 'value', 60);
```
kchu1337: Oh yea, this is why I used a hook, so that I could test that the slider changes on textfield change. Without the hook the only way to test is to actually change the value prop itself. So like 
```js
    expect(wrapper.find('Slider').prop('value')).toEqual(50);
    wrapper.setProps({value:60})
    expect(wrapper.find('Slider').prop('value')).toEqual(60);
```
This is because we export the component without connecting it to redux.

PR #1166, opened 2019-08-30
Sa 873 - Fix Test Name rendering in list and details page

kchu1337:
### What Changed
 - Don't render test name and the divider if the name is not present.
 - Make the details page actually render the test name rather than the string. Or `None` if not exist.

### How To Test
 - Go to/inbox-placement. Check that the table does not show the divider. 
 - Click on any of the tests. Check that the test name is `None` in the details page.

### DIFF
--- a/src/pages/inboxPlacement/TestListPage.js
+++ b/src/pages/inboxPlacement/TestListPage.js
@@ -61,11 +61,11 @@ const FilterSortCollectionRow = ({ id, status, subject, test_name, from_address,
             </div>
             <div className = {styles.TestName}>
               {test_name &&
-                <>
+                <span id={'testName'}>
                   <strong>{test_name}</strong>
                   <strong className = {styles.Divider}>{'|'}</strong>
-                </>}
-              <span>{from_address}</span>
+                </span>}
+              <span id={'fromAddress'}>{from_address}</span>
             </div>
             <div className = {styles.TestSchedule}>
               <span ><Schedule className={styles.ScheduleIcon}/></span>
--- a/src/pages/inboxPlacement/TestListPage.js
+++ b/src/pages/inboxPlacement/TestListPage.js
@@ -34,14 +34,14 @@ const filterBoxConfig = {
     </div>)
 };
 
-const FilterSortCollectionRow = ({ id, status, subject, test_name, from_address, start_time, end_time, placement }, iterator) => (
+const FilterSortCollectionRow = ({ id, status, subject, test_name, from_address, start_time, end_time, placement }) => (
   [
     <Table.Row
-      key={iterator}
+      key={id}
       className={styles.TableRow}
       rowData={[
-        <div className = {styles.TabbedCellBody}>
-          <div className = {styles.SubjectContainer}>
+        <div className={styles.TabbedCellBody}>
+          <div className={styles.SubjectContainer}>
             <div className={styles.TooltipWrapper}>
               <Tooltip
                 disabled={status === STATUS.COMPLETED}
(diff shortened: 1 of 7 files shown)


src/pages/inboxPlacement/TestListPage.js
```diff
@@ -60,9 +60,12 @@ const FilterSortCollectionRow = ({ id, status, subject, test_name, from_address,
               </PageLink>
             </div>
             <div className = {styles.TestName}>
-              <strong>{test_name}</strong>
-              <strong className = {styles.Divider}>{'|'}</strong>
-              <span>{from_address}</span>
+              {test_name &&
+                <span id={'testName'}>
```
bkemper: Why are we setting id's or `<span />`s?

src/pages/inboxPlacement/TestListPage.js
```diff
@@ -60,9 +60,12 @@ const FilterSortCollectionRow = ({ id, status, subject, test_name, from_address,
               </PageLink>
             </div>
             <div className = {styles.TestName}>
-              <strong>{test_name}</strong>
-              <strong className = {styles.Divider}>{'|'}</strong>
-              <span>{from_address}</span>
+              {test_name &&
```
bkemper: Shouldn't we hide the `<div />` container as well?

src/pages/inboxPlacement/components/tests/TestDetails.test.js
```diff
@@ -21,6 +22,16 @@ describe('Component: TestDetails', () => {
...
   });
 
+  it('renders test name as None when it does not exist', () => {
+    const detailsWithoutName = {
+      start_time: '2019-07-08T15:49:56.954Z',
+      end_time: null,
+      subject: 'Fooo'
+    };
+    const wrapper = subject({ details: detailsWithoutName });
+    expect(wrapper.find('InfoBlock').at(3).prop('value')).toEqual('None');
```
bkemper: `toHaveProp('value', 'None')`

src/pages/inboxPlacement/TestListPage.js
kchu1337: The style is also always applied to the `from address`

src/pages/inboxPlacement/TestListPage.js
kchu1337: I did it as a shortcut to find the tag for testing purposes.

src/pages/inboxPlacement/TestListPage.js
```diff
@@ -34,14 +34,14 @@ const filterBoxConfig = {
     </div>)
 };
 
-const FilterSortCollectionRow = ({ id, status, subject, test_name, from_address, start_time, end_time, placement }, iterator) => (
+const FilterSortCollectionRow = ({ id, status, subject, test_name, from_address, start_time, end_time, placement }) => (
   [
     <Table.Row
-      key={iterator}
+      key={id}
```
bkemper: @kchu1337 I don't know if it is safe to do this, but I think it is better to make this a component.

src/pages/inboxPlacement/tests/TestListPage.test.js
```diff
@@ -45,26 +45,22 @@ describe('Page: Test List', () => {
...
 
   describe('Test Name:', () => {
-
-    const renderFirstColumn = (rowProps) => {
+    const renderFirstColumn = (props) => {
       const wrapper = subject();
-      const rowFn = wrapper.find('FilterSortCollection').prop('rowComponent');
-      const row = rowFn(rowProps)[0];
-      return shallow(shallow(row).prop('children')[0]);
+      const Rows = wrapper.find('FilterSortCollection').prop('rowComponent');
```
bkemper: This would definitely be easier to test if this component lived in a separate file.

PR #1167, opened 2019-09-03
AC 984 -  Feature comparison modal

shrirupa:
#1147 # What Changed
     > Created a FeatureComparisonModal to show feature comparison between plans.
   
### To Do
- [ ] Need to put in the logic to open the modal, depends on AC-983


src/pages/billing/components/FeatureComparisonModal.js
```diff
@@ -0,0 +1,62 @@
...
+    </Table.HeaderCell>
+  </Table.Row>;
+}
+function Icon({ value }) {
+  return value ? <Check/> : <Close/>;
+}
+export function renderCell(cellValue) {
+  if (typeof cellValue === 'boolean') { return <Icon value={cellValue}/> ; }
+  if (typeof cellValue === 'string') {
+    if (cellValue.indexOf('\n') !== -1) {
```
kchu1337: You can do 
```js
    return cellValue.split('\n').map((item, key) => <div key={key} style={{ fontSize: (key > 0) ? '12px' : '' }}>{item}</div>);
```
Less if statements.
If you make the style a class, then you should be able to do something like `classNames={key > 0 && styles.foo}`

src/pages/billing/components/FeatureComparisonModal.js
```diff
@@ -0,0 +1,62 @@
...
+    <Table.HeaderCell colSpan={colSpan} className={styles.GroupHeading}>
+      {groupName}
+    </Table.HeaderCell>
+  </Table.Row>;
+}
+function Icon({ value }) {
+  return value ? <Check/> : <Close/>;
+}
+export function renderCell(cellValue) {
+  if (typeof cellValue === 'boolean') { return <Icon value={cellValue}/> ; }
```
kchu1337: Up to you, but I feel like you can just return the icon directly instead of calling another function to return the icon.

src/pages/billing/components/FeatureComparisonModal.js
```diff
@@ -0,0 +1,62 @@
...
+  return cellValue;
+}
+function ComparisonModal({ open, handleClose }) {
+  return (<Modal open={open} showCloseButton={true} onClose={handleClose} >
+    <Panel>
+      <div className={styles.FeatureComparisonTable}>
+        <Table>
+          <HeaderRow plans={PLANS}/>
+        </Table>
+        {_.map(FEATURE_COMPARISON, (featureArray, groupName) => <>
```
kchu1337: I don't think you need the extra `<>`

src/pages/billing/components/FeatureComparisonModal.js
```diff
@@ -0,0 +1,62 @@
...
+          <Table>
```
kchu1337: Whenever you use a map function, I think you need to add a key prop or else React will yet at you in the console.

src/pages/billing/components/FeatureComparisonModal.js
```diff
@@ -0,0 +1,62 @@
...
+    <Table.Cell className={styles.FeatureCell}> {featureName} </Table.Cell>
+    <Table.Cell className={classNames(styles.FeatureComaprisonCell, styles.PlanOne)}> {renderCell(testAccount)} </Table.Cell>
+    <Table.Cell className={classNames(styles.FeatureComaprisonCell, styles.PlanTwo)}> {renderCell(starterPlans)} </Table.Cell>
+    <Table.Cell className={classNames(styles.FeatureComaprisonCell, styles.PlanThree)}> {renderCell(premierPlans)} </Table.Cell>
+  </Table.Row>;
+}
+export function HeaderRow({ plans }) {
+  return <Table.Row>
+    <Table.Cell className={styles.FeatureCell}>  </Table.Cell>
+    <Table.Cell className={classNames(styles.FeatureComaprisonCell, styles.PlanOne, styles.TableCaption)}> {plans[0]} </Table.Cell>
```
kchu1337: Might be able to use a `map` function here and above in the `Row` component.

src/pages/billing/components/FeatureComparisonModal.module.scss
```diff
@@ -0,0 +1,42 @@
...
+    border-top: 1px solid rgb(205, 205, 209);
+    border-bottom: 1px solid rgb(205, 205, 209);
+}
+.FeatureCell {
+    border-left: 1px solid rgb(205, 205, 209);
+    width: 40%;
+    padding-left: 2em;
+    max-height: 40px;
+}
+.FeatureComaprisonCell {
```
kchu1337: Misspelled `Comparison`

src/pages/billing/components/FeatureComparisonModal.module.scss
```diff
@@ -0,0 +1,42 @@
...
+    border-left: 1px solid rgb(205, 205, 209);
+    width: 20%;
+    text-align: center;
+    max-height: 40px;
```
kchu1337: I would add  `word-break: break-word;` here. If you do, you can remove it from the other class below.

src/pages/billing/constants.js
```diff
@@ -0,0 +1,42 @@
+export const FEATURE_COMPARISON = {
+  'Standard Features': [
+    ['Signals Predictive Analytics', true, true,true],
```
kchu1337: You can go 1 or 2 levels deeper and make the arrays into objects. So like 
```js
'Standard Features': {
    'Signals Predictive Analytics': [true, true, true]
}
```
You can make that array into an object as well but I don't think that's necessary. Just add a comment so that we know which array index corresponds to which plan.

src/pages/billing/components/tests/FeatureComparisonModal.test.js
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import { renderCell, HeaderRow, GroupHeading, Row } from '../FeatureComparisonModal';
+import { PLANS } from '../../constants';
+describe('FeatureComparisonModal: ', () => {
+  const props = {
+    open: true,
+    plans: PLANS,
+    handleClose: jest.fn((x) => true),
+    onClose: jest.fn((x) => true),
```
kchu1337: You don't need the other handleClose and you can just do jest.fn() since we don't care about the return value of the onclose function.

src/pages/billing/components/tests/FeatureComparisonModal.test.js
```diff
@@ -0,0 +1,48 @@
...
+    groupName: 'Standard Features',
+    colSpan: 3,
+    featureName: 'Signals Predictive Analytics',
+    testAccount: true,
+    starterPlans: true,
+    premierPlans: true
+  };
+  describe('Row: ', () => {
+    it('should render correctly', () => {
+      const wrapper = shallow(<Row {...props} />);
```
kchu1337: You are passing in a lot of unnecessary props. I would pass in only the props that the Component is expecting. In this case it looks like it doesn't matter but it could in another test case.

src/pages/billing/components/BillingSummary.js
```diff
@@ -15,15 +15,15 @@ import { formatFullNumber } from 'src/helpers/units';
...
 const PAYMENT_MODAL = 'payment';
 const CONTACT_MODAL = 'contact';
 const IP_MODAL = 'ip';
 const RV_MODAL = 'recipient_validation';
 
 export default class BillingSummary extends Component {
   state = {
-    show: false
+    show: false,
+    openComparisonModal: true
```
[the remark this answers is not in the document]
shrirupa: yes, missed removing this.

src/pages/billing/components/FeatureComparisonModal.js
shrirupa: missed it! thanks

src/pages/billing/components/FeatureComparisonModal.js
shrirupa: you are right!. There were other console errors too, removed them.

src/pages/billing/components/FeatureComparisonModal.js
shrirupa: done

src/pages/billing/components/FeatureComparisonModal.js
shrirupa: ok. `classNames={key > 0 && styles.foo}` doing this throws console warning, replaced it with 
`index > 0 ? styles.SmallerFont : ''`

src/pages/billing/components/FeatureComparisonModal.module.scss
shrirupa: word break is only needed for PlanNames, and not for all places where FeatureComparisonCell is used. So I guess its better to keep it in PlanNames.

src/pages/billing/constants.js
shrirupa: done

src/pages/billing/components/FeatureComparisonModal.js
shrirupa: done

src/pages/billing/components/tests/FeatureComparisonModal.test.js
shrirupa: done

src/pages/billing/components/tests/FeatureComparisonModal.test.js
shrirupa: right! done.

src/pages/billing/components/FeatureComparisonModal.js
```diff
@@ -5,21 +5,22 @@ import _ from 'lodash';
...
 import { Check, Close } from '@sparkpost/matchbox-icons';
-export function Row({ featureName, ...value }) {
+export function Row({ featureName, ...featureValues }) {
   return <Table.Row>
     <Table.Cell className={styles.FeatureCell}> {featureName} </Table.Cell>
-    <Table.Cell className={classNames(styles.FeatureComparisonCell, styles.PlanOne)}> {renderCell(value[PLANS[0]])} </Table.Cell>
-    <Table.Cell className={classNames(styles.FeatureComparisonCell, styles.PlanTwo)}> {renderCell(value[PLANS[1]])} </Table.Cell>
-    <Table.Cell className={classNames(styles.FeatureComparisonCell, styles.PlanThree)}> {renderCell(value[PLANS[2]])} </Table.Cell>
+    {_.map(featureValues, (val, key) => <Table.Cell key={key} className={classNames(styles.FeatureComparisonCell,
+      key === 'Test Account' ? styles.PlanOne : '',
```
PatrickSison: could be a bit cleaner as `key === 'Test Account' && styles.PlanOne`

src/pages/billing/components/FeatureComparisonModal.js
```diff
@@ -5,21 +5,22 @@ import _ from 'lodash';
...
 }
 export function HeaderRow({ plans }) {
   return <Table.Row>
     <Table.Cell className={styles.FeatureCell}>  </Table.Cell>
-    {plans.map((plan,index) => <Table.Cell key={plan} className={classNames(styles.FeatureComparisonCell,
-      index + 1 === 1 ? styles.PlanOne : '',
-      index + 1 === 2 ? styles.PlanTwo : '',
-      index + 1 === 3 ? styles.PlanThree : '',
+    {plans.map((plan) => <Table.Cell key={plan} className={classNames(styles.FeatureComparisonCell,
+      plan === 'Test Account' ? styles.PlanOne : '',
```
PatrickSison: Same as other comment about style

src/pages/billing/components/FeatureComparisonModal.module.scss
```diff
@@ -0,0 +1,46 @@
+@import '~@sparkpost/matchbox/src/styles/config.scss';
+
+.GroupHeading {
+    background: #f5f5f8;
+    border-top: 1px solid rgb(205, 205, 209);
```
PatrickSison: I would avoid using rgb() and use color(gray, {lightValue}) to keep consistency with other colors. Same with the background color. The accent colors are ok as they are individual instances

src/pages/billing/components/FeatureComparisonModal.js
kchu1337: Even better would be to define an object like 
```js
const planStyles = {
  'Test Account': styles.PlanOne,
  'Starter Plans': styles.PlanTwo,
  'Premier Plans': styles.PlanThree
};
```
and then you can just do `classNames(styles.FeatureComparisonCell, planStyles[planName])`
Also, let's rename `key` to `planName` or `planType` and and `val` to `planValues`. So it's more clear what they mean.

shrirupa: ok done. with classNames library being used `key === 'Test Account' && styles.PlanOne` doesn't show a error as in the case of just using `className={index > 0 && styles.SmallerFont}`.

src/pages/billing/components/FeatureComparisonModal.js
shrirupa: done.

src/pages/billing/components/FeatureComparisonModal.js
```diff
@@ -28,17 +31,12 @@ export function GroupHeading({ groupName, colSpan }) {
...
 export function renderCell(cellValue) {
-  if (typeof cellValue === 'boolean') { return <Icon value={cellValue}/> ; }
+  if (typeof cellValue === 'boolean') { return cellValue ? <Check/> : <Close/>; }
   if (typeof cellValue === 'string') {
-    if (cellValue.indexOf('\n') !== -1) {
-      return <div> {cellValue.substring(0,cellValue.indexOf('\n'))} <br/>
-        <span style={{ fontSize: '12px' }}> {cellValue.substring(cellValue.indexOf('\n'),cellValue.length)}</span>
-      </div>;
-    }
+    return <div>{cellValue.split('\n').map((item, index) => <div key={index} className={index > 0 ? styles.SmallerFont : ''}>{item}</div>)}</div>
```
kchu1337: I don't think you need the first `<div>` tag.

src/pages/billing/components/tests/FeatureComparisonModal.test.js
```diff
@@ -1,35 +1,48 @@
 import React from 'react';
 import { shallow } from 'enzyme';
-import { renderCell, HeaderRow, GroupHeading, Row } from '../FeatureComparisonModal';
+import ComparisonModal, { renderCell, HeaderRow, GroupHeading, Row } from '../FeatureComparisonModal';
 import { PLANS } from '../../constants';
```
kchu1337: Let's mock the constants file so the snapshots are smaller/easier to read and so that a modification to the plans doesn't need snapshot updates.

src/pages/billing/components/tests/FeatureComparisonModal.test.js
```diff
@@ -1,35 +1,48 @@
...
+    it('should render correctly', () => {
+      const wrapper = shallow(<ComparisonModal {...props}/>);
+      expect(wrapper).toMatchSnapshot();
+    });
+
+  });
   describe('Row: ', () => {
+    const props = {
+      featureName: 'Signals Predictive Analytics',
+      featuerValues: {
```
kchu1337: Doesn't matter, but you misspelled feature

src/pages/billing/components/FeatureComparisonModal.js
shrirupa: This is needed for tests. When I test renderCell for the case of strings and don't pass the enclosing <div> tag it complains saying `TypeError: ShallowWrapper can only wrap valid elements`

src/pages/billing/components/tests/FeatureComparisonModal.test.js
shrirupa: corrected it . lol, i should install a spell check.

src/pages/billing/components/FeatureComparisonModal.module.scss
shrirupa: done

src/pages/billing/components/FeatureComparisonModal.js
```diff
@@ -0,0 +1,72 @@
+import React from 'react';
+import { Modal, Table, Panel } from '@sparkpost/matchbox';
+import { FEATURE_COMPARISON, PLANS } from './../constants';
+import _ from 'lodash';
+import styles from './FeatureComparisonModal.module.scss';
+import classNames from 'classnames';
+import { Check, Close } from '@sparkpost/matchbox-icons';
+export function Row({ featureName, ...featureValues }) {
+  return <Table.Row>
+    <Table.Cell className={styles.FeatureCell}> {featureName} </Table.Cell>
```
PatrickSison: Would it be worth reducing some of the width on this column? Could possibly help with some of the word wrapping

src/pages/billing/components/tests/FeatureComparisonModal.test.js
shrirupa: done!

src/pages/billing/components/FeatureComparisonModal.js
shrirupa: ok

src/pages/billing/components/PlanSelect.js
```diff
@@ -1,15 +1,31 @@
...
+};
+
 
-export const SelectedPlan = ({ plan, onChange }) => (
-  <Panel title='Your New Plan'>
+export const SelectedPlan = ({ plan, onChange }) => {
+  const { isShowing, toggle } = useModal(false);
+  return <>
+  <FeatureComparisonModal open={isShowing} handleClose={toggle}/>
+  <Panel title={<PanelTitle title={'Your New Plan'} toggleModal={toggle}/>}>
```
PatrickSison: On Panel, there's an `actions` prop where the button can be added rather than through the title

src/pages/billing/components/PlanSelect.js
```diff
@@ -59,9 +90,12 @@ const PlanSelectSection = ({ plans, currentPlan, onSelect }) => {
   ));
 
   return (
-    <Panel title='Select a Plan'>
+    <>
+    <FeatureComparisonModal open={isShowing} handleClose={toggle}/>
+    <Panel title={<PanelTitle title={'Select a Plan'} toggleModal={toggle}/>}>
```
PatrickSison: Same as above

src/pages/billing/components/PlanSelect.module.scss
```diff
@@ -34,3 +34,7 @@
   position: absolute;
   transform: translate(-20px, 2px);
 }
+
```
PatrickSison: Should be able to remove if modal open button is put into `actions` prop

src/pages/billing/components/FeatureComparisonModal.js
```diff
@@ -0,0 +1,72 @@
...
+export function GroupHeading({ groupName, colSpan }) {
+  return <Table.Row>
+    <Table.HeaderCell colSpan={colSpan} className={styles.GroupHeading}>
+      {groupName}
+    </Table.HeaderCell>
+  </Table.Row>;
+}
+
+export function renderCell(cellValue) {
+  if (typeof cellValue === 'boolean') { return cellValue ? <Check/> : <Close/>; }
```
PatrickSison: It looks like in the mock ups, the `X`(Close) should have a gray color instead of the column color

src/pages/billing/components/FeatureComparisonModal.module.scss
```diff
@@ -0,0 +1,45 @@
+@import '~@sparkpost/matchbox/src/styles/config.scss';
+
+.GroupHeading {
+    background: color(gray,9);
+    border-top: 1px solid color(gray,5);
+    border-bottom: 1px solid color(gray,5);
+}
+.FeatureCell {
+    border-left: 1px solid color(gray,5);
```
PatrickSison: I think this can be removed since there doesn't need to be a left border

src/pages/billing/components/FeatureComparisonModal.js
shrirupa: nice catch!

src/pages/billing/components/FeatureComparisonModal.module.scss
```diff
@@ -0,0 +1,47 @@
...
+    color: rgb(60, 146, 240);
+}
+.PlanThree {
+    color: rgb(250, 100, 35);
+}
+.PlanName{
+    height: 80px;
+    padding: 31px;
+    font-size: 24px;
+    font-weight: bold;
```
PatrickSison: Would probably change to `font-weight: 600;` to get closer to mock up

src/pages/billing/components/FeatureComparisonModal.module.scss
```diff
@@ -0,0 +1,47 @@
...
+    background: color(gray,9);
+    border-top: 1px solid color(gray,5);
+    border-bottom: 1px solid color(gray,5);
+}
+.FeatureCell {
+    width: 31%;
+    padding-left: 2em;
+    max-height: 40px;
+}
+.FeatureComparisonCell {
```
PatrickSison: Would add maybe `font-size: 120%`

PR #1168, opened 2019-09-05
AC-983: New Billing Plan Select

PatrickSison:
### What Changed
 - Add new select for plans

### How To Test
 - Add account ui option 'account_feature_limits'
 - Go to `/account/billing/plan`
 - Verify the look compared to snapshot

### To Do
- [ ] ~Mock new bundle and plan endpoint~
- [ ] ~Map new bundles and plans to new plan select~
- [x] Let select functionality work in form
- [x] Add testing for new form

### DIFF
--- a/src/pages/billing/components/CurrentPlanSection.js
+++ b/src/pages/billing/components/CurrentPlanSection.js
@@ -1,0 +1,18 @@
+import React from 'react';
+import { Panel } from '@sparkpost/matchbox';
+import PlanPrice from 'src/components/billing/PlanPrice';
+import styles from './CurrentPlanSection.module.scss';
+import { PLAN_TIERS } from 'src/constants';
+
+const CurrentPlanSection = ({ currentPlan }) => (
+  <Panel title='Current Plan'>
+    <Panel.Section className={styles.currentPlan}>
+      <div className={styles.Title}>{PLAN_TIERS[currentPlan.tier]}</div>
+      <div className={styles.Plan}>
+        <PlanPrice showOverage showIp showCsm plan={currentPlan} />
+      </div>
+    </Panel.Section>
+  </Panel>
+);
+
+export default CurrentPlanSection;
--- a/src/pages/billing/components/PlanSelect.js
+++ b/src/pages/billing/components/PlanSelect.js
@@ -1,0 +1,51 @@
+import React from 'react';
+import { Panel, Button } from '@sparkpost/matchbox';
+import { PLAN_TIERS } from 'src/constants';
+import PlanPrice from 'src/components/billing/PlanPrice';
+
+import styles from './PlanSelect.module.scss';
+
+const PlanSelectSection = ({ plans, currentPlan, onSelect }) => {
+  const TIERS = [
+    { key: 'default' },
+    { key: 'test', label: PLAN_TIERS.test },
+    { key: 'starter', label: PLAN_TIERS.starter },
+    { key: 'premier', label: PLAN_TIERS.premier }
+  ];
+
+  const planList = TIERS.map((tier) => {
+    const tierPlans = plans[tier.key];
+    if (!tierPlans) {
+      return;
+    }
+
+    return (
+      <Panel.Section key={`tier_section_${tier.key}`}>
+        <div className={styles.tierLabel}>{tier.label}</div>
+        <div className={styles.tierPlans}>
+          {
+            plans[tier.key].map((bundle) => (
+              <div className={styles.PlanRow} key={`plan_row_${bundle.code}`}>
+                <div>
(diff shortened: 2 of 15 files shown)


src/pages/billing/forms/NewChangePlanForm.js
```diff
@@ -0,0 +1,100 @@
...
+    </form>
+  );
+};
+
+const mapStateToProps = (state, props) => {
+  const { code: planCode, promo: promoCode } = qs.parse(props.location.search);
+
+  return {
+    plans: selectTieredVisiblePlans(state),
+    initialValues: changePlanInitialValues(state, { planCode, promoCode }),
```
shrirupa: looks like keys planCode and promoCode have been renamed to code and promo in line 73. I think this should be 
`initialValues: changePlanInitialValues(state, { code, promo })`

src/pages/billing/components/CurrentPlanSection.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { Panel } from '@sparkpost/matchbox';
+import PlanPrice from 'src/components/billing/PlanPrice';
+import styles from './CurrentPlanSection.module.scss';
+import { PLAN_TIERS } from 'src/constants';
+
+const CurrentPlanSection = ({ currentPlan }) => (
+  <Panel title='Current Plan'>
+    <Panel.Section className={styles.currentPlan}>
```
shrirupa: `<Panel sectioned title='Current Plan'>` also works instead of having a <Panel.Section> again.

src/pages/billing/components/PlanSelect.js
```diff
@@ -0,0 +1,81 @@
...
+import { Check } from '@sparkpost/matchbox-icons';
+import { PLAN_TIERS } from 'src/constants';
+import PlanPrice from 'src/components/billing/PlanPrice';
+import cx from 'classnames';
+
+import styles from './PlanSelect.module.scss';
+
+export const SelectedPlan = ({ plan, onChange }) => (
+  <Panel title='Your New Plan'>
+    <Panel.Section>
```
shrirupa: Same as above.

src/pages/billing/components/PlanSelect.js
```diff
@@ -0,0 +1,81 @@
...
+          </div>
+        </div>
+      </div>
+    </Panel.Section>
+  </Panel>
+);
+
+const PlanSelectSection = ({ plans, currentPlan, onSelect }) => {
+  const TIERS = [
+    { key: 'default' },
```
shrirupa: not sure why this is needed

src/pages/billing/components/PlanSelect.js
```diff
@@ -0,0 +1,81 @@
...
+    { key: 'test', label: PLAN_TIERS.test },
+    { key: 'starter', label: PLAN_TIERS.starter },
+    { key: 'premier', label: PLAN_TIERS.premier }
+  ];
+
+  const planList = TIERS.map((tier) => {
+    const tierPlans = plans[tier.key];
+    if (!tierPlans) {
```
shrirupa: same as above.

src/pages/billing/components/PlanSelect.js
```diff
@@ -0,0 +1,81 @@
...
+
+const PlanSelectSection = ({ plans, currentPlan, onSelect }) => {
+  const TIERS = [
+    { key: 'default' },
+    { key: 'test', label: PLAN_TIERS.test },
+    { key: 'starter', label: PLAN_TIERS.starter },
+    { key: 'premier', label: PLAN_TIERS.premier }
+  ];
+
+  const planList = TIERS.map((tier) => {
```
shrirupa: instead of using TIERS here, can't we use PLAN_TIERS here instead.

src/pages/billing/components/CurrentPlanSection.js
PatrickSison: This provides a standard padding rather than me providing it

src/pages/billing/components/PlanSelect.js
PatrickSison: You're right. It's an artifact of the old plan picker as hidden plans would fall under the default tier (which would not have a tier), but since there's a second "page", it can be removed now.

src/pages/billing/components/PlanSelect.js
PatrickSison: Not sure what you mean

src/pages/billing/components/PlanSelect.js
PatrickSison: The map of the plans should have gone through that, but it doesn't. The logic is also an artifact of displaying the default plans when that shouldn't be the case anymore. Will remove

PR #1169, opened 2019-09-05
TR-1512 Duplicate Template

nicklemmon:
[TR-1512](https://jira.int.messagesystems.com/browse/TR-1512)

### What Changed
- Updated `<DraftModeActions/>` component, adding new `<DuplicateTemplate/>`
- Added new `<DuplicateTemplateModal/>` component that is outside of the existing `<Popover/>` to help address toggling behavior, keeping it independent from the Popover itself
- Extracted new `<SaveAndPublishConfirmationModal/>` component to address similar behavior with the `<SaveAndPublish/>` component
- Updated `<SaveAndPublish/>` component to avoid having nested `<button>` elements within `<a>` elements
- Updated some styling for the `<EditNavigation/>` component to mirror design specs more closely
- Updated relevant tests, wrote new tests
- Updated Matchbox to at least version `3.7.0` to allow use of the `<ScreenReaderOnly>` component
- Added `<ButtonWrapper>` generic component for consistent positioning of button elements within the UI - includes template, styles, and documentation
- Added screen reader only content to the `<FullPage/>` component to ensure screen reader users properly interpret the back button

### How To Test
1. Navigate to http://localhost:3100/templatesV2
1. Select a template that is in the draft status
1. Click on the popover icon to the right of the 'Save and Publish' button:

Screen Shot 2019-09-05 at 10 42 01 AM

1. Click on 'Duplicate Template'
1. A modal should appear with extra copy appended to the relevant text fields (`(COPY)` for the name field or `-copy` for the ID field)
1. After confirming, the user should be re-routed to `/templatesV2`
1. The duplicated template should now appear on the list page

### To Do
- [x] Incorporate feedback
- [ ] __QUESTION__ - if I duplicate a template that has been published previously, is the new duplicated template supposed to render in the table with a status of 'Published'?


src/pages/templatesV2/components/EditNavigation.js
```diff
@@ -18,6 +18,8 @@ const EditNavigation = ({ primaryArea }) => {
             })}
             key={key}
             onClick={() => { setNavigation(key); }}
+            to="javascript:void(0)"
```
bkemper: Also, by setting the href, the default anchor selector style is applied which sets the color to blue when `:active`.
Screen Shot 2019-09-16 at 7 03 05 AM

src/pages/templatesV2/components/EditNavigation.module.scss
```diff
@@ -11,21 +11,32 @@
...
 .NavigationLink {
+  font-size: rem(18);
   color: color(gray, 4); // to match default text
   display: inline-block;
-  padding: spacing(large) spacing();
-
-  &:first-child {
-    padding-left: spacing(large);
-  }
+  padding: spacing(large) spacing(large);
```
bkemper: This could be reduced to `padding: spacing(large)`.

src/pages/templatesV2/components/EditNavigation.module.scss
```diff
@@ -11,21 +11,32 @@
...
+
+  &:after {
+    content: "";
+    display: block;
+    width: 75%;
+    position: absolute;
+    left: 50%;
+    right: 50%;
+    transform: translateX(-50%);
+    bottom: -1px;
```
bkemper: I think this is a unique situation with the bottom border set to `2px`, but the bottom border size could be variable.  Does this need to lay on top of the border?  Or could it always safely sit at `bottom: 0px`?

Screen Shot 2019-09-16 at 7 14 56 AM

src/pages/templatesV2/components/editorActions/DraftModeActions.js
```diff
@@ -1,40 +1,96 @@
...
+import SaveAndPublishConfirmationModal from './SaveAndPublishConfirmationModal';
 import ViewPublished from './ViewPublished';
 import SaveDraft from './SaveDraft';
 import styles from './Actions.module.scss';
 import useEditorContext from '../../hooks/useEditorContext';
+import DuplicateTemplate from './DuplicateTemplate';
+import DuplicateTemplateModal from './DuplicateTemplateModal';
 
-export default () => {
+const DraftModeActions = () => {
```
bkemper: 👍

src/pages/templatesV2/components/editorActions/EditDraft.js
```diff
@@ -8,12 +8,20 @@ import { setSubaccountQuery } from 'src/helpers/subaccounts';
 export default ({ className, children }) => {
   const { draft, history } = useEditorContext();
 
-  const onClick = () => {
+  // TODO: Why break this out in to a function when it is effectively just a link with params?
```
bkemper: I agree.  This should be a `PageLink`.

src/pages/templatesV2/components/editorActions/tests/SaveAndPublishConfirmationModal.test.js
```diff
@@ -0,0 +1,68 @@
...
+    await wrapper.find('ConfirmationModal').prop('onConfirm')(); //invoke attached func
+    expect(publishDraft).toHaveBeenCalled();
+    expect(push).toHaveBeenCalledWith('/templatesv2/edit/foo/published/content');
+  });
+
+  it('redirects to published path with subaccount upon publishing', async () => {
+    const push = jest.fn();
+    const publishDraft = jest.fn(() => Promise.resolve());
+    const wrapper = subject({ history: { push }, publishDraft, draft: { id: 'foo', subaccount_id: 101 }});
+    await wrapper.find('ConfirmationModal').prop('onConfirm')(); //invoke attached func
```
bkemper: I'm interested to know you opinion on this...

Props that begin with `on` can be invoked with `simulate`.  For example, `wrapper.find('ConfirmationModal').simulate('confirm')`.  

However, `simulate` won't return a promise, so you can't reliably `await` it, but the `await` is still needed.

src/pages/templatesV2/components/EditNavigation.js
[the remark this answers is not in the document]
nicklemmon: Definitely! One of those things we could add to make it easier to build things accessibly without actually having to know what's going on under the hood. I'll open an issue with Matchbox.

nicklemmon: Interesting! I'll look in to overriding those styles.

src/pages/templatesV2/components/EditNavigation.module.scss
nicklemmon: Nice catch, I'll take a look.

src/pages/templatesV2/components/editorActions/DuplicateTemplate.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import { UnstyledLink } from '@sparkpost/matchbox';
+import { ContentCopy } from '@sparkpost/matchbox-icons';
+
+const DuplicateTemplate = (props) => {
+  const { className, onClick } = props;
```
bkemper: These can be deconstructed where `prop` argument is declared.

src/pages/templatesV2/components/editorActions/tests/DraftModeActions.test.js
```diff
@@ -22,4 +22,22 @@ describe('DraftModeActions', () => {
...
     expect(subject({ hasPublished: false }).find('ViewPublished')).not.toExist();
   });
+
+  it('Sets the `DuplicateTemplateModal` open prop to `true` when clicking on `DuplicateTemplate`', () => {
+    const wrapper = subject();
+    const duplicateButton = wrapper.find('DuplicateTemplate');
+
+    duplicateButton.simulate('click');
+
+    expect(wrapper.find('DuplicateTemplateModal').props().open).toEqual(true);
```
bkemper: jest-enzyme provides some nice test matches such as, [`toHaveProp`](https://www.npmjs.com/package/jest-enzyme#tohaveprop).

src/pages/templatesV2/components/editorActions/tests/DraftModeActions.test.js
```diff
@@ -22,4 +22,22 @@ describe('DraftModeActions', () => {
...
+  });
+
+  it('Sets the `SaveAndPublishConfirmationModal` oen prop to `true` when clicking on `SaveAndPublish`', () => {
```
bkemper: This is a literal description of what is happening to our code.  Instead, we should describe tests from a users perspective.  For example, "opens confirmation modal when save and publish button is clicked".

src/pages/templatesV2/components/editorActions/DraftModeActions.js
```diff
@@ -1,40 +1,96 @@
...
+            }
+          </div>
+        </Popover>
+
+        <DuplicateTemplateModal
+          open={isDuplicateModalOpen}
+          onClose={handleModalClose}
+        />
+
+        <SaveAndPublishConfirmationModal
```
bkemper: Should these action modals be rendered in the same component as the button, `SaveAndPublish`?  If they are colocated, then the state can live them and this actions component shouldn't continue to grow.

src/pages/templatesV2/components/editorActions/DuplicateTemplate.js
nicklemmon: If that's the preference I can do that - seems like that's generally how things are in the project now that I look at it. I usually prefer not to destructure up there as I feel like it's less clear where that data is coming from but I'm flexible!

src/pages/templatesV2/components/editorActions/tests/DraftModeActions.test.js
nicklemmon: hmm, I'll have to take a look.

src/pages/templatesV2/components/editorActions/DraftModeActions.js
nicklemmon: So I did have them with the button originally, but because the button is within a popover, it means I can't close the popover when the user opens the modal as the modal can't be rendered unless the popover is also rendered. Does that make sense?

src/pages/templatesV2/components/EditNavigation.js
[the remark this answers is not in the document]
nicklemmon: Opened issue this issue in the Matchbox repo:
https://github.com/SparkPost/matchbox/issues/247

src/pages/templatesV2/components/EditNavigation.module.scss
nicklemmon: Setting it to `0` gets it above the line - easier to maintain in the long run as well.

src/pages/templatesV2/components/editorActions/DraftModeActions.js
nicklemmon: I'm going to leave this as-is for now, but let me know if you have another idea that might resolve this particular problem.

src/pages/templatesV2/components/editorActions/tests/DraftModeActions.test.js
nicklemmon: Hmm, I see what you're saying - unfortunately, all we can truly assert with this test is that the prop is passing, not that the modal is actually visible/open! If we were not doing shallow rendering then we could make that assertion.

If the `DuplicateTemplateModal` were incorrectly handling the `open` prop, then that test description wouldn't really be accurate if that makes sense.

Maybe it would be worth refactoring this to use `mount()` instead of `shallow()`?

src/pages/templatesV2/components/editorActions/tests/SaveAndPublishConfirmationModal.test.js
nicklemmon: Truthfully I just migrated these tests from another file after a component extraction, so I can't speak much to the details of these tests - I was hoping to preserve them to ensure I wasn't breaking anything despite refactoring a few things.

After doing a little light digging, this looks like a really promising way to handle it:
https://github.com/airbnb/enzyme/issues/346#issuecomment-304535773

src/pages/templatesV2/components/editorActions/tests/EditDraft.test.js
```diff
@@ -21,8 +21,10 @@ describe('EditDraft', () => {
   });
 
   it('renders edit draft action with provided children', () => {
+    /* eslint-disable no-useless-escape */
     expect(subject(null,
-      <span>Click here</span>).find('UnstyledLink').html()).toMatch('<a><span>Click here</span></a>');
+      <span>Click here</span>).find('UnstyledLink').html()).toMatch('<a href=\"javascript:void(0);\" target=\"\" rel=\"\" role=\"button\"><span>Click here</span></a>');
```
bkemper: This would be testing Matchbox.  I think we could really benefit from a custom matcher.

```
it('renders edit draft action with provided children', () => {
    const children = <span>Click here</span>;
    const wrapper = subject(null, children);

    expect(wrapper).toHaveTextContent('Click Here');
  });
```

src/pages/templatesV2/components/editorActions/DuplicateTemplate.js
bkemper: I'm indifferent.  Let it if you think it is better.

src/pages/templatesV2/components/editorActions/DraftModeActions.js
bkemper: Ahh, good point.  Leave it.

src/pages/templatesV2/components/editorActions/tests/DraftModeActions.test.js
bkemper: Oh no, don't change the test.  We want to leave it a unit test.

src/pages/templatesV2/components/editorActions/tests/EditDraft.test.js
nicklemmon: Yeah previously this component had a custom method that was invoking `history.push()` so this is outdated actually - I deleted this test in the latest push to reflect that.

Thanks!

src/pages/templatesV2/components/editorActions/tests/SaveAndPublishConfirmationModal.test.js
bkemper: Hmm... I think I understand the magic.

```
  it('redirects to published path with subaccount upon publishing', () => {
    const push = jest.fn();
    const promise = Promise.resolve();
    const publishDraft = jest.fn(() => promise);
    const wrapper = subject({ history: { push }, publishDraft, draft: { id: 'foo', subaccount_id: 101 }});

    wrapper.find('ConfirmationModal').simulate('confirm');

    return promise.then(() => {
      expect(publishDraft).toHaveBeenCalled();
      expect(push).toHaveBeenCalledWith('/templatesv2/edit/foo/published/content?subaccount=101');
    });
  });
```

bkemper: Very clever.

src/pages/templatesV2/components/editorActions/DuplicateTemplateModal.js
```diff
@@ -0,0 +1,103 @@
...
+        <Panel
+          title="Duplicate Template"
+          sectioned
+        >
+          <form onSubmit={(e) => handleSubmit(e)}>
+            <TextField
+              id="template-name"
+              name="templateName"
+              label="Template Name"
+              required // not working as I would expect...
```
bkemper: This just decorates the TextField with an asterisk.

src/pages/templatesV2/components/editorActions/DuplicateTemplateModal.js
```diff
@@ -0,0 +1,103 @@
...
+              value={draftName ? draftName : null}
```
bkemper: I think we should try to avoid `null` as much as possible.  It will force you to do extra work (e.g. `!draftName || draftName.length === 0`).  

Screen Shot 2019-09-16 at 12 37 43 PM

src/pages/templatesV2/components/editorActions/DuplicateTemplateModal.js
```diff
@@ -0,0 +1,103 @@
...
+  const [draftId, setDraftId] = useState(initialDraftId);
+  const [hasNameError, setNameError] = useState(false);
+  const [hasDraftError, setDraftError] = useState(false);
+  const [hasSuccessRedirect, setSuccessRedirect] = useState(false);
+
+  const handleSubmit = (e) => {
+    e.preventDefault();
+
+    if (!draftName || draftName.length === 0) {
+      setNameError(true);
```
bkemper: Should we move this to `onChange` handler?

src/pages/templatesV2/components/editorActions/DuplicateTemplateModal.js
```diff
@@ -0,0 +1,103 @@
...
+    if (!draftId || draftId.length === 0) {
+      setDraftError(true);
+    }
+
+    if (draftName && draftId) {
+      setNameError(false);
+      setDraftError(false);
+
+      createTemplate({
+        ...draft,
```
bkemper: I think there are only three values we need from draft, content, options, and shared_with_subaccounts, and we should explicitly set them to avoid spreading in the published flag and other metadata.

```
createTemplate({
  content: draft.content,
  name: draftName,
  id: draftId,
  options: draft.options,
  shared_with_subaccounts: draft.shared_with_subaccounts
}).then(() => setSuccessRedirect(true));
```

src/pages/templatesV2/components/editorActions/DuplicateTemplateModal.js
nicklemmon: That makes sense to me.

PR #1171, opened 2019-09-08
TR-1742: Prevent reporting-only users from accessing Recipient Validation

bkemper:
Refer to [TR-1742](https://jira.int.messagesystems.com/browse/TR-1742);

### What Changed
 - Hide the "Recipient Validation" nav item for developer, reporting, and templates users

### How To Test 
- [ ] As an admin, confirm nav item is displayed
- [ ] As a developer, confirm nav item is hidden
- [ ] As a reporting, confirm nav item is hidden
- [ ] As a templates, confirm nav item is hidden


src/config/navItems/index.js
```diff
@@ -98,6 +106,9 @@ export default [
   },
   {
     label: 'Recipient Validation',
+    // this route is a redirect, its condition is not correctly inherited, so this condition will
```
PatrickSison: Maybe something that can resolve the issue is changing the way the redirect works. I put it initially in the routes file, but maybe it can happen in the RecipientValidation page. When navigating to `/recipient-validation` without `recipient_validation` feature flag, it is redirected to `/recipient-validation/list` still but shows `RVDisabledPage`

PatrickSison: Maybe we can handle redirect in the page itself (so it can preserve route if rendering RVDisabledPage, but not if it renders actual page)

PatrickSison: Maybe more difficult than it's worth since would have to be more checks on when to redirect and when to not

PR #1174, opened 2019-09-09
Sa 603 - Add day precision to Summary report datepicker

kchu1337:
### What Changed
 - Updated getMomentPrecision to include `day` as a precision option. 
 - The datepicker should disable the time text field for date choices > 2 days. Should also default to 12:00 am on the from date to 11:59 on the to date.
 - Refactored some code so it's easier to understand IMO.
 - Updated tests. 

### How To Test
 - Test that the datepicker on the report pages correctly show date ranges of 12:00 - 11:59 for day precision. 
 - Test that hour and minute precision works as before.
 - Test that nothing is breaking on the signals page. The signals page uses the same precision functions and datepicker. 

### To Do
- [ ] Undo some refactoring if it makes less sense than before.

### DIFF
--- a/src/components/datePicker/ManualEntryForm.js
+++ b/src/components/datePicker/ManualEntryForm.js
@@ -4,7 +4,7 @@ import _ from 'lodash';
 import { Grid, TextField } from '@sparkpost/matchbox';
 import { ArrowForward } from '@sparkpost/matchbox-icons';
 import { formatInputDate, formatInputTime, parseDatetime } from 'src/helpers/date';
-import { getValidDateRange, getPrecision } from 'src/helpers/metrics';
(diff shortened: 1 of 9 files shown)


nicklemmon:
Overall this looks good - my main feedback is about the view more than the implementation, and isn't actually the result of changes made as a part of this PR.

src/components/datePicker/ManualEntryForm.js
```diff
@@ -83,6 +83,7 @@ export default class ManualEntryForm extends Component {
 
     let precisionLabel = null;
     let precisionLabelValue;
+    let momentPrecision = '';
```
nicklemmon: Does this need to be set to an empty string in this case? I would think:

```javascript
let momentPrecision;
``` 

Would work just fine.

PR #1178, opened 2019-09-13
SA-603 Datepicker Day Precision fix

kchu1337:
Fix datepicker from https://github.com/SparkPost/2web2ui/pull/1174. 

Prevent to date from being set into the future by not rounding the to if it's already set to now.

Also disable rounding in suppressions and events.

### DIFF
--- a/src/helpers/signals.js
+++ b/src/helpers/signals.js
@@ -76,7 +76,7 @@ export const getDates = ({ from, relativeRange, to, now = new Date() } = {}) => 
 
   if (relativeRange !== 'custom') {
     options = {
-      ...getRelativeDates(relativeRange, { now: moment(now).subtract(1, 'day') }),
+      ...getRelativeDates(relativeRange, { now: moment(now).subtract(1, 'day').toDate() }),
       relativeRange: relativeRange
     };
   }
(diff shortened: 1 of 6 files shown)


src/helpers/signals.js
```diff
@@ -76,7 +76,7 @@ export const getDates = ({ from, relativeRange, to, now = new Date() } = {}) =>
 
   if (relativeRange !== 'custom') {
     options = {
-      ...getRelativeDates(relativeRange, { now: moment(now).subtract(1, 'day') }),
+      ...getRelativeDates(relativeRange, { now: moment(now).subtract(1, 'day').toDate() }),
```
bkemper: 👍

src/helpers/tests/metrics.test.js
```diff
@@ -200,6 +200,14 @@ describe('metrics helpers', () => {
       expect(resFrom.toISOString()).toEqual(expectedValue.from);
       expect(resTo.toISOString()).toEqual(expectedValue.to);
     }, allCases);
+
+    it('should not round To if it is same as now', () => {
+      const now = '2018-12-18T10:02';
+      const date = moment(new Date(now));
+      Date.now = jest.fn(() => date);
```
bkemper: Don't need this, just pass string dates to `roundBoundaries`.

src/helpers/tests/metrics.test.js
```diff
@@ -200,6 +200,14 @@ describe('metrics helpers', () => {
       expect(resFrom.toISOString()).toEqual(expectedValue.from);
       expect(resTo.toISOString()).toEqual(expectedValue.to);
     }, allCases);
+
+    it('should not round To if it is same as now', () => {
+      const now = '2018-12-18T10:02';
+      const date = moment(new Date(now));
```
bkemper: We should try to avoid using `new Date` in tests because we cannot control the timezone.

PR #1179, opened 2019-09-13
SE-224 Polish Alerts Create/Edit form

jmoyse:
* Removed the IP section when theres no IPs on the account
* Updated sliders to have a precision of 0 instead of 2. 
* Various tiny copy changes

### DIFF
--- a/src/pages/alerts/components/fields/EvaluatorFields.js
+++ b/src/pages/alerts/components/fields/EvaluatorFields.js
@@ -50,7 +50,7 @@ export const EvaluatorFields = ({
     <Grid className={styles.Grid}>
       {sourceOptions.length > 1 && (
         <Grid.Column sm={12} md={3}>
-          <Label>Score</Label>
+          <Label>Evaluated</Label>
           <Field
             name='source'
             component={SelectWrapper}
--- a/src/pages/alerts/components/fields/EvaluatorFields.js
+++ b/src/pages/alerts/components/fields/EvaluatorFields.js
@@ -50,7 +50,7 @@ export const EvaluatorFields = ({
     <Grid className={styles.Grid}>
       {sourceOptions.length > 1 && (
         <Grid.Column sm={12} md={3}>
-          <Label>Evaluated</Label>
+          <Label>Score</Label>
           <Field
             name='source'
             component={SelectWrapper}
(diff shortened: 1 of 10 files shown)


src/pages/alerts/components/fields/EvaluatorFields.js
```diff
@@ -50,7 +50,7 @@ export const EvaluatorFields = ({
     <Grid className={styles.Grid}>
       {sourceOptions.length > 1 && (
         <Grid.Column sm={12} md={3}>
-          <Label>Evaluated</Label>
+          <Label>Score</Label>
```
PatrickSison: Is this the one that needs to change as well? Or just the conversion of `raw` to `Score` in `formConstants.js`

src/pages/alerts/components/fields/FilterFields.js
```diff
@@ -104,18 +104,31 @@ export class FilterFields extends Component {
...
-          name={value}
-          component={ComboBoxTypeaheadWrapper}
-          results={filterTypeaheadResults[value]}
-          label={label}
-          defaultSelected={this.props[value]}
-          {...extraProps[value]}
-        />)
-      ));
+      formSpec.filterOptions
+        .filter((option) => shouldRenderFilter(option))
```
PatrickSison: I like the filter happening before-hand 👍. One thing we can change now though is removing the condition  on the field itself (line 76 and 77) since it checks the same condition but it won't matter since the field wouldn't be rendered

src/pages/alerts/components/fields/EvaluatorFields.js
jmoyse: Good catch. I'll get that fixed.

src/pages/alerts/helpers/alertForm.js
```diff
@@ -65,18 +65,23 @@ export const getEvaluatorOptions = (metric, source) => {
...
+      return { sliderLabel: 'Percent Change', sliderPrecision: 0 };
+    } else if (source !== 'raw') {
       return { sliderLabel: 'Percent Change', sliderPrecision: 2 };
     } else {
       switch (metric) {
         case 'health_score':
           return { sliderLabel: 'Score', sliderPrecision: 1 };
         case 'monthly_sending_limit':
           return { sliderLabel: 'Percent Used', sliderPrecision: 0 };
+        case 'week_over_week':
```
PatrickSison: I don't think these 2 lines are needed since they would be coming from source and would be checked already in the first condition

PR #1180, opened 2019-09-16
TR-1345 - Send test email

nicklemmon:
[TR-1345](https://jira.int.messagesystems.com/browse/TR-1345) - Sending a Test Email from Templates V2 UI.

### What Changed
- Adds new `tohaveTextContent()` Jest matcher (full credit to @bkemper)
- Addresses small issue with templates selector that was returning template `text` as `undefined`
- Adds new `className` prop to the `<Loading/>` component
- Adds new `<MultiEmailField/>` re-usable component with:
    * `useMultiEmailField` hook
    * documentation
    * supporting tests
- Exposes the `sendPreview` and `showAlert` action for use on `<EditAndPreviewPage/>`
- Adds new screen reader only content to `<PreviewControlBar/>` buttons
- Adds relevant HTML attributes to `<PreviewControlBar/>` buttons to make them keyboard focusable
- Adds new styles to `<PreviewControlBar/>` to ensure the active state is not communicated through color alone
- Adds new `<SendTestEmalButton/>` component that renders the "Send a Test" button and relevant Modal with form elements to send a preview email

### How To Test
- Navigate to `http://localhost:3100/templatesV2`
- Click on an email template in the list
- Click on 'Send a Test'
- Send a test email using the UI
- Things to check:
    * The `<MultiEmailField/>` component has some fancy interaction! Spacebar and blur adds the text entry to the email list, errors are thrown if entry is not a valid email
    * Submitting the form throws an error if no entry is present
    * Opening the modal saves the state of the draft
    * Sending a preview closes the modal automatically
    * Check the unit tests for additional test cases that can be checked manually

### To Do
- [ ] Incorporate feedback
- [X] Ensure the draft saves when the user opens the preview modal
- [X] Add supporting tests


config/jest/setup.js
```diff
@@ -22,11 +22,11 @@ console.error = (message) => {
 // mock out a file that uses require.context under the hood
 jest.mock('src/components/notifications/staticMarkdownNotifications', () => [
   {
-    component: function() {},
+    component: function () {},
```
nicklemmon: Just made this change to address a linter error - I guess the linter doesn't fail a commit for these files?

config/jest/setup.js
```diff
@@ -55,5 +55,5 @@ Object.defineProperty(global.window.location, 'assign', { value: jest.fn(), conf
 // Show a stack track for unhandled rejections to help
 // track them down.
 process.on('unhandledRejection', (reason) => {
-	console.log(reason)
+  console.log(reason);
```
nicklemmon: Again, just fixing a linter error here.

src/components/loading/Loading.js
```diff
@@ -5,8 +5,8 @@ import styles from './Loading.module.scss';
 /**
  * A centered loading animation
  */
-export const Loading = function Loading() {
-  return <LoadingSVG className={styles.Center}/>;
+export const Loading = function Loading({ className }) {
```
nicklemmon: If there is a reason `className` wasn't included as a prop before I'm all ears. Seems useful though.

src/components/loading/Loading.js
```diff
@@ -1,14 +1,20 @@
...
  */
-export const Loading = function Loading() {
-  return <LoadingSVG className={styles.Center}/>;
+export const Loading = function Loading(props) {
+  const { className } = props;
+
+  return <LoadingSVG className={classnames(styles.Center, className)}/>;
 };
 
+Loading.propTypes = {
```
nicklemmon: Is there a reason this wasn't exposed previously? Seems useful, so kind of surprised it wasn't available before.

src/pages/templatesV2/components/SendTestEmailButton.js
```diff
@@ -0,0 +1,224 @@
+/* eslint-disable max-lines */
```
nicklemmon: This is a long component! This *could* be broken up to to more components, i.e., the form and the modal, but the form and the modal have a lot of interplaying state so I left them all here. If you have an approach that would work to break this up, I'm all ears. Might just be a lack of knowledge on my part.

src/pages/templatesV2/components/SendTestEmailButton.module.scss
```diff
@@ -0,0 +1,14 @@
+@import '~@sparkpost/matchbox/src/styles/config.scss';
+
+.LoadingWrapper {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
+
+.Loading {
+  // resetting positioning to allow the loader to render inside the modal
```
nicklemmon: I would think this CSS should exist somewhere else already? Not sure where though...should maybe even become a prop on the modal component like `isLoading` to do the rendering for us.

nicklemmon:
@bkemper Ready when you have time - this is another fairly large one

config/jest/matchers/toHaveTextContent.js
```diff
@@ -0,0 +1,10 @@
+const toHaveTextContent = (received, expectedContent) => {
+  const text = received.debug && received.debug();
+
+  return {
+    message: () => `expected component to have content "${expectedContent}"`,
+    pass: RegExp(expectedContent).test(text)
+  };
+};
+
+export default toHaveTextContent;
```
bkemper: 🙏

src/components/loading/Loading.js
bkemper: I think it was to avoid custom scss.\

src/pages/templatesV2/components/SendTestEmailButton.js
```diff
@@ -0,0 +1,175 @@
...
+    clearForm();
+    setModalOpen(false);
+    setMultiEmailError('');
+  };
+
+  const handleSubmit = (e) => {
+    e.preventDefault();
+
+    if (multiEmailList.length === 0) {
+      setMultiEmailError('Please enter a valid email address');
```
bkemper: We should return from a guard clause.

src/pages/templatesV2/components/SendTestEmailButton.js
```diff
@@ -0,0 +1,175 @@
...
+                  onKeyDownAndBlur={(e) => handleMultiEmailKeyDownAndBlur(e)}
+                  onRemoveEmail={handleMultiEmailRemove}
+                  error={multiEmailError}
+                  value={multiEmailValue}
+                  emailList={multiEmailList}
+                />
+
+                <TextField
+                  id="text-field-test-email-from"
+                  label="From:"
```
bkemper: Colons are not used on the other fields, so we should probably removed them until our style changes.

src/pages/templatesV2/components/SendTestEmailButton.js
```diff
@@ -0,0 +1,175 @@
...
+    setMultiEmailList([]);
+  };
+
+  const handleModalOpen = () => {
+    setModalOpen(true);
+    setModalLoading(true);
+
+    // Save the template, then allow the user to send a preview
+    // The preview can be send whether the current draft was successfully saved or not.
+    updateDraft({ id: templateId, content }, subaccountId)
```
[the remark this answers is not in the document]
nicklemmon: To surface the decision to save? Hmm, that seems like lots of prompting in a row I would think.

src/pages/templatesV2/components/SendTestEmailButton.js
```diff
@@ -0,0 +1,175 @@
...
+  const handleSubmit = (e) => {
+    e.preventDefault();
+
+    if (multiEmailList.length === 0) {
+      setMultiEmailError('Please enter a valid email address');
+    } else {
+      setModalLoading(true);
+    }
+
+    if (multiEmailList.length && templateId) {
```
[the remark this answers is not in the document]
nicklemmon: Good point! The component will never render without one.

src/pages/templatesV2/components/SendTestEmailButton.js
nicklemmon: I'm fine to do that, but would you mind explaining why in this case?

src/components/multiEmailField/tests/MultiEmailField.test.js
```diff
@@ -0,0 +1,165 @@
...
+  it('renders with default props for `selectedItems`, `value`, and `error`', () => {
+    const wrapper = subject();
+
+    expect(wrapper).toHaveProp('selectedItems', []);
+    expect(wrapper).toHaveProp('value', '');
+    expect(wrapper).toHaveProp('error', '');
+  });
+});
+
+describe('MultiEmailField with useMultiEmailField', () => {
```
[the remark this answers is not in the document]
nicklemmon: So basically I didn't want to test `useMultiEmailField` because it's an implementation detail of the component, and won't ever be used independently of the component itself. So testing it with the component makes sense, but not in isolation:
https://kentcdodds.com/blog/react-hooks-whats-going-to-happen-to-my-tests

src/components/multiEmailField/useMultiEmailField.js
```diff
@@ -0,0 +1,75 @@
...
+    if (e.keyCode === 13) {
+      e.preventDefault();
+    }
+
+    // Remove the last email from the list when the user deletes
+    // and no in progress value is present in the field
+    if (e.keyCode === 8 && !multiEmailValue) {
+      setMultiEmailList(multiEmailList.filter((email, index) => {
+        if (index + 1 !== multiEmailList.length) {
+          return email;
```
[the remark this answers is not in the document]
nicklemmon: Duh. Good catch.

src/components/multiEmailField/useMultiEmailField.js
```diff
@@ -0,0 +1,75 @@
...
+
+  const handleMultiEmailChange = (e) => {
+    setMultiEmailValue(e.target.value);
+    setMultiEmailError('');
+  };
+
+  const handleMultiEmailRemove = (target) => {
+    setMultiEmailList(multiEmailList.filter((item) => {
+      if (target !== item) {
+        return item;
```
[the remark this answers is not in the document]
nicklemmon: Duh again.

src/components/multiEmailField/tests/MultiEmailField.test.js
```diff
@@ -0,0 +1,165 @@
...
+
+    wrapper.simulate('change', { target: { value: 'hello@me.com' }});
+    wrapper.simulate('keyDownAndBlur', { type: 'keyDown', keyCode: 32, preventDefault: jest.fn() });
+
+    // Invoking `wrapper.props().onRemove()` wasn't triggering the deletion - works when tested manually
+    wrapper
+      .dive()
+      .dive()
+      .find('Tag')
+      .simulate('remove');
```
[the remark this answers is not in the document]
nicklemmon: I'll invoke the prop instead.

src/pages/templatesV2/components/SendTestEmailButton.js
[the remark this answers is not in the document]
nicklemmon: I did address one of these two - maybe we can talk with Lynn about the other - no need to save the template when in published mode since no editing will have taken place.

src/components/multiEmailField/tests/MultiEmailField.test.js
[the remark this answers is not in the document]
bkemper: I completely agree that we don't want to test implementation details.  Your explanation makes sense for this specific case.  In the future, if we are making reusable hooks that are not tightly coupled to a specific component, I think it will be necessary to exclusively test a hook with a light wrapper like [useTabs.test.js](https://github.com/SparkPost/2web2ui/blob/master/src/hooks/useTabs/useTabs.test.js).

src/components/multiEmailField/useMultiEmailField.js
```diff
@@ -0,0 +1,75 @@
...
+
+  const handleMultiEmailKeyDownAndBlur = (e) => {
+    // prevent form submission when using the enter key
+    if (e.keyCode === 13) {
+      e.preventDefault();
+    }
+
+    // Remove the last email from the list when the user deletes
+    // and no in progress value is present in the field
+    if (e.keyCode === 8 && !multiEmailValue) {
```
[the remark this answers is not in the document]
bkemper: I just took another look.  I don't think we need this code.

src/pages/templatesV2/components/SendTestEmailButton.js
bkemper: The benefit of a guard clause is to improve readability by less nesting and avoid unnecessary conditions.  Here is an example, [Replace Nested Conditional with Guard Clauses](https://refactoring.guru/replace-nested-conditional-with-guard-clauses).  When applied to your code...

```
// with guard clause
const handleSubmit = (e) => {
  e.preventDefault();

  if (multiEmailList.length === 0) {
    setMultiEmailError('Please enter a valid email address');
    return;
  }

  setModalLoading(true);
  sendPreview();
};
```

nicklemmon: Of course. This is a really great article too.

src/components/multiEmailField/useMultiEmailField.js
```diff
@@ -0,0 +1,65 @@
...
+  const handleMultiEmailKeyDownAndBlur = (e) => {
+    // Remove the last email from the list when the user deletes
+    // and no in progress value is present in the field
+    if (e.keyCode === 8 && !multiEmailValue) {
+      setMultiEmailList(multiEmailList.filter((email, index) => index + 1 !== multiEmailList.length));
+    }
+
+    if (e.type === 'blur' || e.keyCode === 32 || e.keyCode === 13) {
+      if (e.type === 'keydown') {
+        e.preventDefault(); // Prevents spaces from being written to the field and prevents the enter key from submitting the form
```
bkemper: Adding single comment

src/components/multiEmailField/useMultiEmailField.js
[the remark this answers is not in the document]
nicklemmon: Filed this issue:
https://github.com/SparkPost/matchbox/issues/258

PR #1182, opened 2019-09-17
SA-888 - Use Health Score endpoint to get injections data

kchu1337:
### What Changed
 - Used `total_injection_count` from the health score endpoint in the selector to get injections.
 - Removed the call to `/signals/injection` in the health score page and all related code.
 - Allow the injection chart to be shown for mailbox provider facet.

### How To Test
 - Best to point to production and use the 108 account but there is still one data point in staging from June 10th.
 - Go to the health score overview page `signals/health-score`. Check that the number of injections came from the health score endpoint by checking the UI value and the value returned by the API call.
 - Do the same thing for the details view. You can access the details view by clicking one of hte facet links in table in the health score overview page.
 - Check that the injections chart appears for mailbox-provider facet. Do this by filtering by mailbox provider in the overview page and then clicking on the mailbox provider facet.

### DIFF
--- a/src/selectors/signals.js
+++ b/src/selectors/signals.js
@@ -294,7 +294,7 @@ export const selectHealthScoreDetails = createSelector(
     });
 
     // Merge in rankings
-    const mergedHistory = _.map(filledHistory, (healthData) => ({ ranking: rankHealthScore(roundToPlaces(healthData.health_score * 100, 1)), ...healthData }));
+    const mergedHistory = filledHistory.map((healthData) => ({ ranking: rankHealthScore(roundToPlaces(healthData.health_score * 100, 1)), ...healthData }));
 
     const isEmpty = mergedHistory.every((values) => values.health_score === null);
 
--- a/src/selectors/signals.js
+++ b/src/selectors/signals.js
@@ -293,7 +293,7 @@ export const selectHealthScoreDetails = createSelector(
       from, to
     });
 
-    // Merge in injections and rankings
+    // Merge in rankings
     const mergedHistory = _.map(filledHistory, (healthData) => ({ ranking: rankHealthScore(roundToPlaces(healthData.health_score * 100, 1)), ...healthData }));
 
     const isEmpty = mergedHistory.every((values) => values.health_score === null);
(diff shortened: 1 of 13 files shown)


src/selectors/signals.js
```diff
@@ -268,9 +267,10 @@ export const selectHealthScoreDetails = createSelector(
     const match = data.find((item) => String(item[facet]) === facetId) || {};
 
     const history = _.get(match, 'history', []);
-    const normalizedHistory = history.map(({ dt: date, weights, ...values }) => ({
+    const normalizedHistory = history.map(({ dt: date, weights, total_injection_count: injections, ...values }) => ({
```
nicklemmon: While we're here, can this be refactored to just use `Array.prototype.map` instead of Lodash?

src/selectors/signals.js
```diff
@@ -287,16 +287,14 @@ export const selectHealthScoreDetails = createSelector(
...
       from, to
     });
 
-    // Merge in injections and rankings
-    const mergedHistory = _.map(filledHistory, (healthData) => {
-      const spamData = _.find(spamDetails.data, ['date', healthData.date]);
-      return { injections: spamData.injections, ranking: rankHealthScore(roundToPlaces(healthData.health_score * 100, 1)), ...healthData };
-    });
+    // Merge in rankings
+    const mergedHistory = _.map(filledHistory, (healthData) => ({ ranking: rankHealthScore(roundToPlaces(healthData.health_score * 100, 1)), ...healthData }));
```
nicklemmon: See comment above - is there a reason to use Lodash here?

src/selectors/signals.js
```diff
@@ -426,21 +424,19 @@ export const selectHealthScoreOverviewData = createSelector(
...
 export const selectCurrentHealthScoreDashboard = createSelector(
-  [getCurrentHealthScoreData, getOptions, getInjections],
-  ({ data, loading, error }, { from, to }, injections) => {
+  [getCurrentHealthScoreData, getOptions],
+  ({ data, loading, error }, { from, to }) => {
     const accountData = _.find(data, ['sid', -1]) || {};
     const history = accountData.history || [];
 
-    const normalizedHistory = history.map(({ dt: date, health_score, ...values }) => {
+    const normalizedHistory = history.map(({ dt: date, health_score, total_injection_count: injections, ...values }) => {
```
nicklemmon: A nice, vanilla `.map` right here

src/selectors/tests/__snapshots__/signals.test.js.snap
```diff
@@ -1155,7 +1155,7 @@ Object {
       Object {
         "date": "2018-01-02",
         "health_score": 0.8,
-        "injections": null,
+        "injections": 12345,
```
nicklemmon: It would probably be worth having some `null` values in here just to make sure the selector is handing that in a predictable way

src/selectors/signals.js
kchu1337: Do you mean the `_.get`?

PR #1183, opened 2019-09-17
SA-893 - Use Total Injection instead of total recipients

kchu1337:
### What Changed
 - Changed table header for engagement recency from `Total Recipients` to `Total Injections`

### How To Test
 - Go to `/signals/engagement`. Check that the rightmost table header reads `Total Injections`.


PatrickSison:
Ticket withdrawn for SA-892

PR #1184, opened 2019-09-17
SA-892: Sort by injection counts

PatrickSison:
### What Changed
 - Added sortable column for injectionand engagement recency
 - Added `order_by` mapping for keys

### How To Test
 - Go to `/signals/engagement`
 - Verify that the injection column is sortable

### To Do
- [x] Check if spam traps will have filter

### Notes
 - Spam traps endpoints `order_by` doesn't currently seem to have a way to order by injections. Total injections are also currently calculated on the front end

### DIFF
--- a/src/pages/signals/components/EngagementRecencyOverview.js
+++ b/src/pages/signals/components/EngagementRecencyOverview.js
@@ -194,6 +194,7 @@ class EngagementRecencyOverview extends React.Component {
             align="right"
             dataKey="total_engagement"
             label="Total Recipients"
+            sortable
             width="15%"
             component={({ total_engagement }) => <NumericDataCell value={total_engagement} />}
           />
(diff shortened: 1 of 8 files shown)


src/pages/signals/components/EngagementRecencyOverview.js
```diff
@@ -192,10 +192,11 @@ class EngagementRecencyOverview extends React.Component {
           />
           <Column
             align="right"
-            dataKey="total_engagement"
```
bkemper: Can't this key be completely removed, `EngagementRecencyOverview.test.js` and `selectEngagementRecencyOverviewData` selector?

PatrickSison: 👍

PR #1185, opened 2019-09-17
SE-260: Alert create page freezes for certain enterprise customers

bkemper:
Refer to [SE-260](https://jira.int.messagesystems.com/browse/SE-260)

### What Changed
 - Combo box typeahead should only render a max of 100 options
 - Combo box menu should remain open after a selection

### How To Test
 - Open /alerts/create page using an account with thousands of subaccounts (e.g. brian.kemper+se-260@sparkpost.com)
 - Select "Health Score" from "Alert Metric" dropdown

_This is the point when your browser would freeze because all the subaccounts were being written to DOM for the combo box menu.  Now, it should not freeze and allow you to continue with creation of an alert._

### DIFF
--- a/src/components/typeahead/ComboBoxTypeahead.js
+++ b/src/components/typeahead/ComboBoxTypeahead.js
@@ -1,131 +1,93 @@
 import React, { useState, useEffect } from 'react';
 import { ComboBox, ComboBoxTextField, ComboBoxMenu } from '@sparkpost/matchbox';
 import Downshift from 'downshift';
-import debounce from 'lodash/debounce';
-import sortMatch from 'src/helpers/sortMatch';
 import PropTypes from 'prop-types';
+import { useDebouncedCallback } from 'use-debounce';
+import sortMatch from 'src/helpers/sortMatch';
 
-export const ComboBoxTypeahead = (props) => {
-  const {
-    error,
-    results,
-    itemToString,
-    maxNumberOfResults,
-    onChange,
-    disabled,
-    label,
-    name,
-    selectedMap,
-    placeholder,
-    readOnly,
-    defaultSelected,
-    debounceFn //Needed for testing
-  } = props;
+export const ComboBoxTypeahead = ({
+  defaultSelected,
+  disabled,
+  error,
+  itemToString,
+  label,
+  maxNumberOfResults,
+  name,
+  onChange,
+  placeholder,
+  readOnly,
+  results,
+  selectedMap
+}) => {
+  const [inputValue, setInputValue] = useState('');
   const [selected, setSelected] = useState(defaultSelected);
-  const [matches, setMatches] = useState(results);
-  const [updateMatches, setUpdateMatches] = useState(() => undefined);
+  const [menuItems, setMenuItems] = useState([]);
+  const [updateMenuItems] = useDebouncedCallback((value) => {
+    const items = value ? sortMatch(results, value, itemToString) : results;
+    setMenuItems(items.slice(0, maxNumberOfResults));
+  }, 400);
 
-  //Updates the value of this component
   useEffect(() => {
     onChange(selected.map(selectedMap));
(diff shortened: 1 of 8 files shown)


src/components/typeahead/ComboBoxTypeahead.js
```diff
@@ -1,147 +1,116 @@
...
   const [selected, setSelected] = useState(defaultSelected);
-  const [matches, setMatches] = useState(results);
-  const [updateMatches, setUpdateMatches] = useState(() => undefined);
+  const [menuItems, setMenuItems] = useState([]);
+  const [updateMenuItems] = useDebouncedCallback((value) => {
+    const items = value ? sortMatch(results, value, itemToString) : results;
+    setMenuItems(items.slice(0, maxNumberOfResults));
+  }, 400);
 
-  //Updates the value of this component
```
kchu1337: Let's put this back. It's not super obvious at first glance what the effect does.

src/components/typeahead/ComboBoxTypeahead.js
```diff
@@ -1,147 +1,116 @@
...
-
-    const rootProps = getRootProps({
-      refKey: 'rootRef'
-    });
 
-    const items = matches
+    return changes;
+  };
+
+  const typeaheadfn = ({
```
kchu1337: I think you can just add ``` setInputValue(inputValue);``` somewhere in this function. Its a render function so it will re-render/re-run whenever its prop(inputValue) changes right?

src/components/typeahead/ComboBoxTypeahead.js
```diff
@@ -1,147 +1,116 @@
...
     );
-  }
+  };
 
   return (
-    <Downshift itemToString={itemToString} stateReducer={stateReducer} >
+    <Downshift
+      defaultHighlightedIndex={0}
+      itemToString={itemToString}
+
```
kchu1337: extra space

src/components/typeahead/ComboBoxTypeahead.js
bkemper: I think that would read nicer, but putting a set state inside a render function scares me.

bkemper:
@kchu1337 This is ready for final review.

src/components/typeahead/ComboBoxTypeahead.js
```diff
@@ -26,7 +26,7 @@ export const ComboBoxTypeahead = ({
     const items = value ? sortMatch(results, value, itemToString) : results;
     const nextMenuItems = items.slice(0, maxNumberOfResults);
     setMenuItems(nextMenuItems);
-  }, 400);
+  }, 300);
```
kchu1337: I was messing around with this yesterday and I think we are missing something like 
```js
  // Updates changes in initial values
  useEffect(() => {
    setSelectedItems(defaultSelected);
  }, [defaultSelected]);
```
With his, the subaccounts will no longer show id:(###). But hte change might be jarring so maybe we need loading state?

src/components/typeahead/ComboBoxTypeahead.js
```diff
@@ -71,26 +80,29 @@ export const ComboBoxTypeahead = ({
     isOpen,
     openMenu
   }) => {
+    const hasSelectedItems = Boolean(selectedItems.length);
+    const isSelectedItemExclusive = isExclusiveItem(selectedItems[0]);
     const items = menuItems
-      .filter((item) => ( // remove selected items from menu
-        !selectedItems.some((selectedItem) => selectedMap(selectedItem) === selectedMap(item))
+      .filter((item) => (
```
kchu1337: We could also do 
```js
        //removes all ptions if an exclusive item is already selected
        if (isSelectedItemExclusive) {
          return false;
        }
```
Then we don't need to change the isOpen logic.

bkemper: Nice recommendation!

PR #1186, opened 2019-09-18
AC-987: Feature comparison section

PatrickSison:
### What Changed
 - Added feature change section which lists features that would be removed or need to change before plan change

### How To Test
 - Go to new plan select
- Choose free or lower plan tier
 - Check that feature changes populate and should be acknowledged

### To Do
- [ ] Address feedback
- [ ] Verify copy

### DIFF
--- a/src/pages/billing/components/FeatureChangeSection.js
+++ b/src/pages/billing/components/FeatureChangeSection.js
@@ -1,0 +1,60 @@
+import React from 'react';
+import { Panel } from '@sparkpost/matchbox';
+import { CheckCircle, Warning } from '@sparkpost/matchbox-icons';
+
+import { useFeatureChangeContext } from '../context/FeatureChangeContext';
+
+import cx from 'classnames';
+import styles from './FeatureChangeSection.module.scss';
+
+const Feature = ({ key, value, label, description, action }) => (
+  <Panel.Section key={`confirm_${key}`}>
+    <div className={styles.Label}>{label}</div>
+    <div className={styles.Feature}>
+      <div>
+        {description}
+      </div>
+      <div>
+        {
+          value ? (
+            <CheckCircle className={styles.FeatureCheckIcon}/>
+          ) : action(key)
+        }
+      </div>
+    </div>
+  </Panel.Section>
+);
+
+const FeatureChangeSection = () => {
+  const { isReady, features } = useFeatureChangeContext();
+
+  const renderCTA = () => {
+    const Icon = isReady ? CheckCircle : Warning;
+    const description = isReady
+      ? <div>
+        <strong>Your features have been updated</strong>
+        <span>, please continue with your plan change.</span>
+      </div>
+      : <div>
+        <span>Your new plan has additional limits on features you currently use. See the list below to </span>
+        <strong>make the necessary changes before you can change plans.</strong>
+      </div>;
+    return (
+      <Panel.Section>
+        <div className={styles.FeatureListStatus}>
+          <Icon className={cx(styles.FeatureListIcon, isReady ? styles.success : styles.danger)}/>
+          {description}
+        </div>
+      </Panel.Section>
+    );
+  };
+
+  return (
+    <Panel accent={isReady ? 'green' : 'red'} title='Changes to Features'>
+      {renderCTA()}
+      {features.map(Feature)}
+    </Panel>
+  );
+};
+
+export default FeatureChangeSection;
--- a/src/pages/billing/components/FeatureChangeSection.js
+++ b/src/pages/billing/components/FeatureChangeSection.js
@@ -2,6 +2,7 @@ import React from 'react';
 import { Panel } from '@sparkpost/matchbox';
 import { CheckCircle, Warning } from '@sparkpost/matchbox-icons';
 
+import { Loading } from 'src/components/loading/Loading';
 import { useFeatureChangeContext } from '../context/FeatureChangeContext';
 
 import cx from 'classnames';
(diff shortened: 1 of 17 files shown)


src/pages/billing/forms/NewChangePlanForm.js
```diff
@@ -6,66 +6,83 @@ import { reduxForm } from 'redux-form';
...
+    if (promo && selectedBundle) {
       applyPromoCode(promo);
     }
-  },[applyPromoCode, promo, selectedPlan, verifyPromoCode]);
+  },[applyPromoCode, promo, selectedBundle, verifyPromoCode]);
+
+  //clears out requestParams when user changes plan
   useEffect(() => {
-    if (!selectedPlan) { //clears out requestParams when user changes plan
+    if (!selectedBundle) {
```
PatrickSison: Bugged effect right now since plans were loaded globally, but now should be dependent here

src/pages/billing/components/FeatureChangeSection.js
```diff
@@ -0,0 +1,71 @@
...
+  if (loading) {
+    return (
+      <Panel sectioned style={{ minHeight: '200px' }}>
+        <Loading />
+      </Panel>
+    );
+  }
+
+  const renderCTA = () => {
+    const Icon = isReady ? CheckCircle : Warning;
```
shrirupa: Since this condition is being checked thrice, May be it is better to group it in to if-else. then to make this function more concise we can move this    ` <Panel.Section name='feature-change-status'> <div className={styles.FeatureListStatus}> . ` inside `FeatureChangeSection` return block.

src/pages/billing/context/FeatureChangeContext.js
```diff
@@ -0,0 +1,165 @@
...
+  return (
+    <FeatureChangeContext.Provider value={value}>
+      {children}
+    </FeatureChangeContext.Provider>
+  );
+};
+
+const mapStateToProps = (state) => ({
+  plans: selectPlansByKey(state),
+  subscription: state.billing.subscription,
```
shrirupa: May be we can have a selector format the `subscription` to have its structure same as `selectedBundle`. This would make 'diff'-ing it simpler.

src/pages/billing/context/FeatureChangeContext.js
```diff
@@ -0,0 +1,165 @@
...
+const FeatureChangeContext = createContext({});
+
+export const FeatureChangeProvider = ({
+  children,
+  plans,
+  subscription,
+  selectedBundle,
+  loading,
+  getSubscription
+}) => {
```
shrirupa: May be we can do away with this. I have put down some ideas/changes which might get us there - 

- ```const FeatureChangeSection = ({ selectedBundle, subscription, loading }) => { ```
```
  subscription: newSelector(state.billing.subscription),
  loading: state.billing.loading
});
export default connect(mapStateToProps, { getSubscription })(FeatureChangeSection);
```
 ->we don't need access to plans if we make a selector that already formats out subscription in the same way selectedBundle is formatted.

- reducer can be like this - 
```
function reducer(state, { type, updates }) {
  switch (type) {
    case 'WINDOW FOCUSED':
      getSubscription();
      break;
    case 'UPDATE FEATURES':
      return getNewState(state, updates);//we can have the function do the same thing calculateDifferences does  here
  }
}
```
- calculateDiffernces may be pulled out into a separate component.

src/pages/billing/components/FeatureChangeSection.js
```diff
@@ -0,0 +1,71 @@
...
+  <Panel.Section key={`confirm_${key}`}>
+    <div className={styles.Label}>{label}</div>
+    <div className={styles.Feature}>
+      <div className={styles.description}>
+        {description}
+      </div>
+      <div>
+        {
+          value ? (
+            <CheckCircle className={styles.FeatureCheckIcon}/>
```
shrirupa: We won't have to do this check if `calculateDifferences` assigns` <CheckCircle>` to `actions` key.

src/pages/billing/context/FeatureChangeContext.js
```diff
@@ -0,0 +1,165 @@
...
+          if (actions.dedicated_ip || !comparedPlan) {
+            resObject[product] = {
+              label: 'Dedicated IPs',
+              description: (
+                <div>
+                  <span>
+                    Your new plan doesn't include dedicated IPs.
+                    Your current IPs will now be billed at $20/month each, or you can
+                  </span>
+                  <SupportTicketLink issueId="general_issue">
```
shrirupa: needs a space between "you can" and "submit".

shrirupa:
Question regarding UI - If the old plan had more limit for subaccounts and the selected plan has less limit but the old plan had lesser quantity than new limit. Should the subaccounts show the <CheckCircle> or should it show "Got it" Button.
Screen Shot 2019-10-04 at 5 39 13 PM

src/pages/billing/context/FeatureChangeContext.js
PatrickSison: > suggested changes for NewChangePlanForm.js

Not a big fan of constantly passing a large amount of props down. Part of the idea is that the props can exist outside of the component, and then the props can easily be accessed in any part outside the provider. What I might do instead is pull the provider higher up, so that the ready state can be used by the form to lock submission and display a the submission button (for later on)

PatrickSison:
> Question regarding UI - If the old plan had more limit for subaccounts and the selected plan has less limit but the old plan had lesser quantity than new limit. Should the subaccounts show the `<CheckCircle>` or should it show "Got it" Button.

In this case, we would not show the section at all. It only renders sections that need to be addressed (like if quantity needs to be fixed) or plan changes that happen automatically (like removal of TFA/SSO)

src/pages/billing/context/FeatureChangeContext.js
PatrickSison: I think I'll move the context provider up in AC-988 as the scope of this ticket is starting to get out of hand

PR #1187, opened 2019-09-18
AC-985: Promo Code field

shrirupa:
Promo Code Field Changed to look and function like - [this](https://sparkpost.invisionapp.com/share/DGTCT4JW6JN#/screens/380587103)

### What Changed
 - Promo code is now applied when the apply button is clicked
 - Promo code can be removed by clicking on remove button 

### How To Test
 - Check if this field works correctly on `/account/billing/plan`(requires feature flag - `account_feature_limits` ) and `/onboarding/plan`.

### To Do
- [ ] Address feedback.


src/components/billing/PromoCodeNew.js
```diff
@@ -0,0 +1,42 @@
...
+    }
+    return <Button onClick={handleClick} >Apply</Button>;
+  };
+  const renderSuffix = (condition) => {
+    if (condition) {
+      return <LoadingSVG size="XSmall" />;
+    }
+    return <></>;
+  };
+  const isDisabled = () => promoPending || typeof selectedPromo.promoCode === 'string';
```
PatrickSison: Not sure I understand the check on promo code for it as a `string`. Couldn't we just check for the existence of an applied promo?

src/components/billing/PromoCodeNew.js
```diff
@@ -0,0 +1,42 @@
...
+  const handleClick = () => {
+    applyPromoCode(promocode);
+  };
+  const renderConnectRight = (condition) => {
+    if (condition) {
+      return <Button onClick={clearPromoCode} >Remove</Button>;
+    }
+    return <Button onClick={handleClick} >Apply</Button>;
+  };
+  const renderSuffix = (condition) => {
```
PatrickSison: Could this be reduced to something simple such as `(condition) => (condition && <LoadingSVG />)` or `(condition) => condition ? <LoadingSVG /> : null`?

PatrickSison: Can we also rename this to something like `renderLoading`?

src/components/billing/PromoCodeNew.js
```diff
@@ -0,0 +1,42 @@
...
+  const { applyPromoCode, clearPromoCode } = handlePromoCode;
+  const { promoError, promoPending, selectedPromo } = promoCodeObj;
+  const [promocode, setPromocode] = useState(selectedPromo.promoCode || '');
+  const handleChange = (event) => {
+    setPromocode(event.target.value);
+  };
+  const handleClick = () => {
+    applyPromoCode(promocode);
+  };
+  const renderConnectRight = (condition) => {
```
PatrickSison: Can we name this to `renderApplyButton` or `renderActionButton` or something along those lines?

src/pages/billing/components/PlanSelect.js
```diff
@@ -8,6 +8,7 @@ import cx from 'classnames';
 import _ from 'lodash';
 import styles from './PlanSelect.module.scss';
 import { useState } from 'react';
+import PromoCodeNew from '../../../components/billing/PromoCodeNew';
```
PatrickSison: The import can be changed to be from `'src/components/billing/PromoCodeNew'`

src/pages/billing/forms/NewChangePlanForm.js
```diff
@@ -25,21 +23,26 @@ export const ChangePlanForm = ({
...
+  billing,
+  clearPromoCode,
   // initialValues: {
   //   promoCode
   // },
   currentPlan
 }) => {
   const [selectedPlan, selectPlan] = useState(null);
-
+  const { promoPending, promoError, selectedPromo = {}} = billing;
```
PatrickSison: One issue we run into is that on changing a plan, the promo code/promo error persist, which won't work for the future as the promo code may not necessarily apply to the newly selected plan and the persisting error would not make to show in this case. We might be able to make a hook that has the promo code and the selected plan as dependencies and verify whenever those changes and control the error/applied promo

PatrickSison: This would accomplish what some of the work that redux-form would do, but give us more control over it than redux-form does

src/pages/billing/forms/NewChangePlanForm.js
```diff
@@ -75,22 +90,23 @@ const mapStateToProps = (state, props) => {
   return {
     plans: selectTieredVisiblePlans(state),
     initialValues: changePlanInitialValues(state, { planCode, promoCode }),
-    currentPlan: currentPlanSelector(state)
+    currentPlan: currentPlanSelector(state),
+    billing: state.billing
```
PatrickSison: Rather than pass the entire billing state, we can pass what we need by making a selector that can look for promo information. Or if we choose to go with the hook route, we may be able to remove this entirely

src/components/billing/PromoCodeNew.js
shrirupa: so earlier i had something like promoPending || selected.promoCode but in case of selected.promocode = 'something', the disabled from becomes a string and the console complains received a string value instead of boolean, so I swapped it out to this condition -> `typeof selectedPromo.promoCode === 'string'` . but you are right may be i could simplify it to _.has(selectedPromo, promocode).

src/components/billing/PromoCodeNew.js
```diff
@@ -0,0 +1,42 @@
+import React, { useState } from 'react';
```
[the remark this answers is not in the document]
shrirupa: As discussed, since it(PromoCode) is still being used in Confirmation page which will change soon, we can swap it later.

src/pages/billing/forms/NewChangePlanForm.js
shrirupa: Ok, instead of creating a new hook for this, instead I added a useEffect for selectedPlan and called clearPromoCode.

src/pages/billing/forms/NewChangePlanForm.js
shrirupa: ok, added a selector.

src/pages/onboarding/ChoosePlan.js
```diff
@@ -121,12 +79,21 @@ export class OnboardingPlanPage extends Component {
     }
   }
   render() {
-    const { loading, plans, submitting, selectedPlan = {}, billing } = this.props;
-
+    const { loading, plans, submitting, selectedPlan = {}, billing, clearPromoCode } = this.props;
+    const { selectedPromo = {}, promoError, promoPending } = billing;
+    const promoCodeObj = {
```
PatrickSison: I'm not a big fan of generating this object, but seems to generally be a systemic issue with a lack of a general standard for handling loading/error state in our app. Not sure best solution without some larger redux work

src/pages/billing/forms/NewChangePlanForm.js
```diff
@@ -25,21 +23,30 @@ export const ChangePlanForm = ({
...
+  useEffect(() => {
+    if (!selectedPlan) {
+      clearPromoCode();
+    }
+  },[clearPromoCode, selectedPlan]);
   //TODO: Implement in AC-986
   // useEffect(() => { console.log(selectedPlan, promoCode)}, [verifyPromoCode, promoCode, selectedPlan]);
 
+  const applyPromoCode = (promoCode) => {
+    verifyPromoCode({ promoCode , billingId: selectedPlan.billingId, meta: { promoCode }});
```
PatrickSison: add to the meta object a flag as `showErrorAlert: false`. That prevents the pop up alert from showing as we show the alert already in the field itself

src/components/billing/PromoCodeNew.js
```diff
@@ -0,0 +1,38 @@
...
+  const isDisabled = () => promoPending || _.has(selectedPromo, 'promoCode');
+  const displayErrorMessage = () => promoError ? promoError.message : '';
+  return <TextField
+    name="promo"
+    label="Promo Code"
+    disabled={isDisabled()}
+    connectRight={renderActionButton(selectedPromo.promoCode)}
+    onChange={handleChange}
+    suffix={renderLoading(promoPending)}
+    errorInLabel={true}
```
PatrickSison: Up to you, but should we remove this? Only reason is other fields in the ChoosePlan.js show the error at the bottom of the field, and this field is not consistent with it

src/pages/billing/forms/NewChangePlanForm.js
shrirupa: Thanks!

src/components/billing/PromoCodeNew.js
shrirupa: I made it like this to keep it consistent with the earlier PromoCode field, but you are right it should be the same for all the fields in the form.

src/components/billing/PromoCodeNew.js
```diff
@@ -15,9 +15,9 @@ const PromoCodeNew = ({ promoCodeObj, handlePromoCode }) => {
   };
   const renderActionButton = (condition) => {
     if (condition) {
-      return <Button onClick={clearPromoCode} >Remove</Button>;
+      return <Button onClick={clearPromoCode}>Remove</Button>;
     }
-    return <Button onClick={handleClick} >Apply</Button>;
+    return <Button onClick={handleClick} disabled={promoPending}>Apply</Button>;
```
PatrickSison: 👍

src/pages/onboarding/components/CreditCardSection.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import { Panel } from '@sparkpost/matchbox';
+import PaymentForm from 'src/pages/billing/forms/fields/PaymentForm';
+import BillingAddressForm from 'src/pages/billing/forms/fields/BillingAddressForm';
+import { FORMS } from 'src/constants';
+
+const CreditCardSection = ({ billing, submitting, selectedPlan = {}}) => {
+  if (selectedPlan.isFree) {
```
PatrickSison: At this point, the selectedPlan is only being used to check if it's free. The same condition is checked in the same rendering where it is coming from. Could this condition be consolidated?

src/pages/onboarding/ChoosePlan.js
```diff
@@ -136,10 +103,17 @@ export class OnboardingPlanPage extends Component {
         <Grid>
           <Grid.Column>
             <Panel>
-              <PlanPicker selectedPromo={billing.selectedPromo} disabled={disableSubmit} plans={plans} onChange={this.onPlanSelect}/>
+              <PlanPicker selectedPromo={selectedPromo} disabled={disableSubmit} plans={plans} onChange={this.onPlanSelect}/>
               <AccessControl condition={not(isAws)}>
-                {!selectedPlan.isFree && this.renderPromoCodeField()}
-                {this.renderCCSection()}
+                {!selectedPlan.isFree &&
```
PatrickSison: This condition could probably be consolidated with the one within the CreditCardSection

src/components/billing/PromoCodeNew.js
```diff
@@ -0,0 +1,37 @@
+import React, { useState } from 'react';
+import { TextField, Button } from '@sparkpost/matchbox';
+import { LoadingSVG } from 'src/components';
+import _ from 'lodash';
+
+const PromoCodeNew = ({ promoCodeObj, handlePromoCode }) => {
+  const { applyPromoCode, clearPromoCode } = handlePromoCode;
+  const { promoError, promoPending, selectedPromo } = promoCodeObj;
+  const [promocode, setPromocode] = useState(selectedPromo.promoCode || '');
```
PatrickSison: Sorry, super pedantic and not a necessary change. Can we camel case to `promoCode` and `setPromoCode`?

src/pages/onboarding/ChoosePlan.js
PatrickSison: Also can we have a test that checks if both components are rendered?

src/pages/onboarding/components/tests/CreditCardSection.test.js
```diff
@@ -0,0 +1,38 @@
...
+  it('should render correctly when selectedPlan is Free', () => {
+    props.isPlanFree = true;
+    wrapper.setProps(props);
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('should render correctly when form is being submitted', () => {
+    props.submitting = true;
+    wrapper.setProps(props);
+    expect(wrapper).toMatchSnapshot();
```
PatrickSison: We should try to reduce our reliance on snapshots as it makes the test more difficult to maintain with us checking the snapshot. In this case we can be explicit and find the exact element and check its props:
```javascript
expect(wrapper.find('BillingAddressForm').prop('disabled')).toEqual(true);
```

src/components/billing/tests/PromoCodeNew.test.js
```diff
@@ -0,0 +1,61 @@
...
+
+
+  it('should render with loading spinner', () => {
+    props.promoCodeObj = {
+      promoError: false,
+      promoPending: true,
+      selectedPromo: {}
+    };
+    wrapper.setProps(props);
+    expect(wrapper).toMatchSnapshot();
```
PatrickSison: This test can be explicit

src/components/billing/tests/PromoCodeNew.test.js
```diff
@@ -0,0 +1,61 @@
...
+  });
+
+  it('should render with error', () => {
+    props.promoCodeObj = {
+      promoError: { message: 'Oh no!' },
+      promoPending: false,
+      selectedPromo: {}
+    };
+    wrapper.setProps(props);
+    expect(wrapper).toMatchSnapshot();
```
PatrickSison: This test can also be explicit

src/components/billing/tests/PromoCodeNew.test.js
```diff
@@ -0,0 +1,61 @@
...
+  it('text field should disable and button should change to remove when a promocode is applied',() => {
+    props.promoCodeObj = {
+      promoError: false,
+      promoPending: false,
+      selectedPromo: {
+        promoCode: 'TESTCODE'
+      }
+    };
+    wrapper.setProps(props);
+    expect(wrapper).toMatchSnapshot();
```
PatrickSison: This test can also be explicit

PR #1189, opened 2019-09-20
TR-1781 - Fix creating new template in V2

nicklemmon:
[TR-1781](https://jira.int.messagesystems.com/browse/TR-1781)

### What Changed
- Text content passed to create method will travel with `POST` as the API does not support a null value for the text field
- This was introduced as the result of a prior bug fix for the original templates implementation

### How To Test
- Go to `/templatesV2` and click 'Create New'
- Complete the form and click 'Next'

### To Do
- [ ] Incorporate feedback
- [X] Add supporting test to check that request is passing along an empty string via `content.text` < -- Nevermind - updating the existing test accomplishes this as the `CreatePage.test.js` is already making this check

### DIFF
--- a/src/actions/helpers/templates.js
+++ b/src/actions/helpers/templates.js
@@ -18,8 +18,8 @@ export const getTestDataKey = ({ id, username, mode }) => ([ 'tpldata', username
 export const shapeContent = (content = {}) => {
   const shapedContent = {
     ...content,
+    text: content.text,
     reply_to: _.isEmpty(content.reply_to) ? undefined : content.reply_to,
-    text: _.isEmpty(content.text) ? undefined : content.text,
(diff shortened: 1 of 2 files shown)


src/actions/helpers/templates.js
```diff
@@ -18,8 +18,8 @@ export const getTestDataKey = ({ id, username, mode }) => ([ 'tpldata', username
 export const shapeContent = (content = {}) => {
   const shapedContent = {
     ...content,
+    text: content.text,
```
bkemper: Maybe add a comment for why we allow an empty text part?

nicklemmon: That definitely makes sense

PR #1190, opened 2019-09-23
SA-890 -Selected and Hovered dates are highlighted

kchu1337:
### What Changed
 - Selected and hovered dates have opacity 1 while unselected have opacity ~.7~ .5 for spam trap and engagement recency. All charts on the health score detail page remain the same. 
 - Changed tests to assertion tests. 

### How To Test
 - Log into account with signals data. Go to engagement recency details page or spam trap details page. Verify that hovering and selecting the bars shows a highlighted state.

### To Do
- [x] Per UX request, selecting an already selected bar results in default state.

### DIFF
--- a/src/pages/signals/EngagementRecencyPage.js
+++ b/src/pages/signals/EngagementRecencyPage.js
@@ -33,7 +33,7 @@ export class EngagementRecencyPage extends Component {
     tickFormatter: (tick) => `${roundToPlaces(tick * 100, 0)}%`,
     domain: [0, 1],
     ticks: [0, 0.25, 0.5, 0.75, 1.0]
-  })
+  });
 
   getXAxisProps = () => {
     const { xTicks } = this.props;
@@ -41,7 +41,7 @@ export class EngagementRecencyPage extends Component {
       ticks: xTicks,
       tickFormatter: (tick) => moment(tick).format('M/D')
     };
-  }
+  };
 
   getTooltipContent = ({ payload = {}}) => (
     <Fragment>
@@ -55,10 +55,25 @@ export class EngagementRecencyPage extends Component {
         />
       ))}
     </Fragment>
-  )
+  );
 
   renderContent = () => {
-    const { data = [], empty, error, facet, facetId, gap, handleDateSelect, loading, selectedDate, subaccountId } = this.props;
+    const {
+      data = [],
+      empty,
+      error,
+      facet,
+      facetId,
+      gap,
+      handleDateSelect,
+      loading,
+      selectedDate,
+      subaccountId,
+      hoveredDate,
+      handleDateHover,
+      resetDateHover
+    } = this.props;
+
     const selectedCohorts = _.find(data, ['date', selectedDate]) || {};
     let chartPanel;
 
@@ -87,8 +102,11 @@ export class EngagementRecencyPage extends Component {
               <div className='LiftTooltip'>
                 <BarChart
                   gap={gap}
+                  onMouseOver={handleDateHover}
+                  onMouseOut={resetDateHover}
                   onClick={handleDateSelect}
                   selected={selectedDate}
+                  hovered={hoveredDate}
                   timeSeries={data}
                   tooltipContent={this.getTooltipContent}
                   tooltipWidth='250px'
@@ -111,7 +129,7 @@ export class EngagementRecencyPage extends Component {
         </Grid.Column>
       </Grid>
     );
-  }
+  };
 
   render() {
     const { facet, facetId, subaccountId } = this.props;
--- a/src/pages/signals/components/charts/barchart/BarChart.js
+++ b/src/pages/signals/components/charts/barchart/BarChart.js
@@ -30,7 +30,7 @@ class BarChart extends Component {
       }
       return isActiveDate ? {opacity : 1} : {opacity : .5}
     },
-    (key,isActiveDate,selectedFill) =>`${key}${isActiveDate}${selectedFill}`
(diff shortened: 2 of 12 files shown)


src/pages/signals/EngagementRecencyPage.js
```diff
@@ -55,10 +55,25 @@ export class EngagementRecencyPage extends Component {
         />
       ))}
     </Fragment>
-  )
+  );
 
   renderContent = () => {
-    const { data = [], empty, error, facet, facetId, gap, handleDateSelect, loading, selectedDate, subaccountId } = this.props;
+    const {
```
nicklemmon: I'm a big fan of breaking up destructuring like this - it helps with merge conflicts as well.

nicklemmon:
Screen Shot 2019-09-30 at 3 10 26 PM

I'm getting some rendering problems at the `/signals/health-score` route

src/pages/signals/components/charts/barchart/BarChart.js
```diff
@@ -94,6 +95,7 @@ class BarChart extends Component {
             barCategoryGap={gap}
             data={timeSeries}
             margin={margin}
+            cursor={(disableHover) ? 'pointer' : undefined}
```
nicklemmon: This variable name is confusing - the way this line reads is, 'If hover is disabled, make the CSS cursor a pointer, otherwise leave it undefined' - but that doesn't make much sense. Maybe `disableHover` means disable showing a tooltip on hover or something?

Either way, it's still confusing to me.

If it does the latter, something like:
```
disableTooltipOnHover
```

*or* rather than using negative language, use positive language (which is usually easier to understand):

```
hasTooltipOnHover
```

src/pages/signals/components/charts/barchart/BarChart.js
```diff
@@ -22,6 +22,17 @@ import healthScoreThresholds from '../../../constants/healthScoreThresholds';
...
 class BarChart extends Component {
+
+  getSelectedStyle = _.memoize((key, isActiveDate, selectedFill) =>
+    {
+      if (key === 'health_score' || key === 'injections' || key === 'weight_value')  {
+       return isActiveDate ? {fill: selectedFill} : {};
+      }
+      return isActiveDate ? {opacity : 1} : {opacity : .5}
+    },
+    (key,isActiveDate,selectedFill) =>`${key}${isActiveDate}${selectedFill}`
```
nicklemmon: Not sure why the linter doesn't catch this, but spaces between params is usually easier to read:

```
(key, isActiveDate, selectedFill) =>
```

src/pages/signals/components/charts/barchart/tests/BarChart.test.js
```diff
@@ -35,27 +37,65 @@ describe('BarChart Component', () => {
   });
 
   it('renders a normal bar chart with correct fill for selected/hovered threshold events', () => {
-    wrapper.setProps({ selected: '2011-01-01', yKey: 'health_score', timeSeries: [{ health_score: 75, ranking: 'warning', date: '2011-01-01' }]});
-    const payload = { fill: '#fill', health_score: 75, ranking: 'warning', date: '2011-01-01' };
-    expect(wrapper.find('Bar').at(0).props().shape(payload).props).toMatchSnapshot();
+    const key = 'health_score';
+    wrapper.setProps({ selected: '2011-01-01', yKey: key, timeSeries: [{ [key]: 75, ranking: 'warning', date: '2011-01-01' }]});
```
nicklemmon: Nice work bailing on the snapshots 👍

src/pages/signals/components/charts/barchart/tests/BarChart.test.js
```diff
@@ -35,27 +37,65 @@ describe('BarChart Component', () => {
...
+  });
+
+  it('renders a normal bar chart with correct fill for selected/hovered stacked bar events', () => {
+    const key = 'c_new';
+    wrapper.setProps({ selected: '2011-01-01', yKey: key, timeSeries: [{ [key]: 75, date: '2011-01-01' }]});
+    const payload = { fill: '#fill', [key]: 75, date: '2011-01-01' };
+    expect(wrapper.find({ dataKey: key }).at(0).props().shape(payload).props.opacity).toEqual(1);
+  });
+
+  it('renders a pointer cursor on the entire chart if disablehover is true', () => {
```
nicklemmon: Per my earlier comment, ideally this test could have helped to communicate what this prop is usd for, but I'm admittedly pretty confused about what it's for still.

src/pages/signals/containers/withDateSelection.js
```diff
@@ -25,17 +25,28 @@ export class WithDateSelection extends Component {
     const { selectedDate } = this.state;
 
     const dataSetChanged = prevProps.data !== data;
-    let selectedDataByDay = _.find(data, ['date', selectedDate]);
+    const selectedDataByDay = _.find(data, ['date', selectedDate]);
```
nicklemmon: Probably and opportunity to just use `Array.find` here (even though I can see all you did was change the variable from `let` to `const`)

nicklemmon:
Again, I would think the linter would catxh this, but there are some unused variables on line 50 of `withDateSelection.js`:

Screen Shot 2019-09-30 at 3 40 10 PM

src/pages/signals/components/charts/barchart/BarChart.js
kchu1337: In the preview, the cursor will still change into the pointer when hovering over the bars but no tooltip shows. Since they are links I decided to change the whole chart to use a pointer for the cursor. I will add a `isLink` prop which will accomplish the same but without using `disableHover`

PR #1199, opened 2019-09-26
SD-942 - Incorporate Shared Pool in the select dropdown, add supporti…

nicklemmon:
[SD-941 - Provide Shared Pool Option in AutoIP Warmup UI](https://jira.int.messagesystems.com/browse/SD-942)

### What Changed
- Incorporated check for IPs with `auto_warmup_enabled` that renders the 'Shared Pool' `<option>` when present
- Added supporting tests

### How To Test
1. Navigate to `/account/ip-pools/edit/carnold_all_cold`
1. The 'Overflow Pool' `<select>` should render with 'Shared Pool' selected
1. Switch to the `master` branch
1. The 'Overflow Pool' `<select>` should render with 'None' selected

*NOTE FOR REVIEWERS:* I believe I implemented this correctly, but double checking the business requirements would be a good idea - I'm new to this feature so want to make sure I'm not introducing anything unusual!

### To Do
- [x] Incorporate feedback

### DIFF
--- a/src/pages/ipPools/components/PoolForm.js
+++ b/src/pages/ipPools/components/PoolForm.js
@@ -30,7 +30,24 @@ export class PoolForm extends Component {
       };
     }));
 
+    // If the pool has available IPs with auto warmup enabled,
+    // render the 'Shared Pool' option in the `<select/>`
+    const getPoolsWithAutoWarmup = () => pools.filter((currentPool) => {
+      if (currentPool.ips) {
+        const ipsWithAutoWarmup = currentPool.ips.filter((ip) => ip.auto_warmup_enabled);
+
+        if (ipsWithAutoWarmup.length > 0) {
+          return currentPool;
+        }
+      }
+    });
+
+    if (getPoolsWithAutoWarmup().length) {
+      overflowPools.unshift({ label: 'Shared Pool', value: 'shared pool' });
+    }
+
     overflowPools.unshift({ label: 'None', value: '' });
+
     return overflowPools;
   }
 
--- a/src/pages/ipPools/components/PoolForm.js
+++ b/src/pages/ipPools/components/PoolForm.js
@@ -1,5 +1,4 @@
 import React, { Component } from 'react';
-import _ from 'lodash';
 import { connect } from 'react-redux';
 import { Field, reduxForm } from 'redux-form';
 import { withRouter } from 'react-router-dom';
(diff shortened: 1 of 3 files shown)


src/pages/ipPools/components/PoolForm.js
```diff
@@ -30,7 +30,24 @@ export class PoolForm extends Component {
       };
     }));
 
+    // If the pool has available IPs with auto warmup enabled,
+    // render the 'Shared Pool' option in the `<select/>`
+    const getPoolsWithAutoWarmup = () => pools.filter((currentPool) => {
```
kchu1337: Can just do 
```js
pools.some(({ ips }) => ips.some((ip) => ip.auto_warmup_enabled))
``` 
to get a boolean value on whether any ips have auto warmup.

src/pages/ipPools/components/PoolForm.js
```diff
@@ -30,7 +30,24 @@ export class PoolForm extends Component {
       };
     }));
```
kchu1337: I think we can use `.filter(Boolean)` instead of `_.compact` above. But that's also not part of your changes so it'll be up to you if you want to change and test it. Just gets rid of lodash.

src/pages/ipPools/components/tests/PoolForm.test.js
```diff
@@ -94,8 +94,44 @@ describe('PoolForm tests', () => {
...
+        pools: [{
+          name: 'Fake Pool',
+          id: 'fake-pool',
+          ips: [{
+            external_ip: '1.1.1.1'
+          }]
+        }]
+      });
+
+      expect(wrapper.find(component).prop('options').includes({ label: 'Shared Pool', value: 'shared pool' })).toBe(false);
```
kchu1337: I think includes checks by reference as well.
For example if you type 
```js
console.log([{label:'shared pool'}].includes({label:'shared pool'}));
```
the result is false.

src/pages/ipPools/components/PoolForm.js
nicklemmon: Nice! I pretty much never use `.some` but this is obviously a great case for it. I should use it more often.

src/pages/ipPools/components/PoolForm.js
nicklemmon: Good call - I don't mind doing that.

src/pages/ipPools/components/tests/PoolForm.test.js
nicklemmon: oh weird. Good catch. I checked that a different label name caused it to fail, but not this scenario.

nicklemmon: Yeah totally silly - objects with the exact same values aren't technically the same object, so this test won't work. I refactored this guy in a manner that works now. Using `.some()` as well!

nicklemmon:
@kchu1337 Ready for re-review - no need to deploy this today or anything, but we can get it queued up at least.

PR #1206, opened 2019-10-01
TR-1374 - Templates Editor V2 Test Data

nicklemmon:
[TR-1374](https://jira.int.messagesystems.com/browse/TR-1374)

### What Changed
- Added new templates actions that update test data using `window.localStorage` instead of `localforage` (`createV2`, `updateV2`, `publishV2`, `sendPreviewV2`)
- Added new actions to update, retrieve, and delete test data from `localStorage` (`setTestDataV2`, `deleteTestDataV2`, `getTestDataV2`)
- Added new `EditTestDataSection` component to render the test data tab
- Updated the editor theme to `tomorrow_night_eighties` because it looks super sweet
- Updated background color of editor wrapper to match the new syntax theme
- Updated `PreviewSection` component to conditionally render the preview depending on what kind of data has been provided when the user has selected the `test_data` tab
- Adds new `useEditorTestData` hook that returns `testData`, `parsedTestData`, and the `setTestData` function
- Updated AMP linting script from Google
- Added defaults for template options in the event no options are received from the accounts endpoint

### How To Test
1. Navigation to `/templatesV2`
2. Check that `Test Data` tab renders on the lefthand side and renders the test data editor when active
3. Check that test data renders in the preview by adding this HTML:
```html
<h1>Hello {{world}}</h1>
```

and this test data:
```json
{
  "substitution_data": {
    "world": "hello"
  },
  "metadata": {},
  "options": {}
}
```

And the preview should render as:

> Hello hello

4. When the template is a "DRAFT", edit the test data, click "Save Draft" in the top right menu and verify that data is preserved when that particular template mounts and unmounts.
5. When in "DRAFT" click "Save and Publish" and submit the modal - verify that the new published version of the template has the same dest data as the draft version that produced it
6. Verify that deleting a template removes the test data from local storage by verifying in the console with `window.localStorage`
7. Verify that duplicating a template within the list page view also duplicates the test data when the copy is created
8. Verify that duplicating a template within the template view using the top right menu also duplicates the test data when the copy is created

### To Do
- [x] Incorporate feedback
- [x] Add tests for updated actions
- [x] Get assistance testing `useEditorTestData`


bkemper:
lemmon-flavor 😆

package-lock.json
```diff
@@ -6755,7 +6755,7 @@
     },
     "ansi-escapes": {
       "version": "3.1.0",
-      "resolved": "https://registry.npmjs.org/ansi-escapes/-/ansi-escapes-3.1.0.tgz",
+      "resolved": "http://registry.npmjs.org/ansi-escapes/-/ansi-escapes-3.1.0.tgz",
```
bkemper: Ugh.

src/actions/templates.js
```diff
@@ -126,6 +126,34 @@ export function update(data, subaccountId, params = {}) {
...
+
+    try {
+      await dispatch(update(data, subaccountId));
+
+      await dispatch(sparkpostApiRequest({
+        type: 'PUBLISH_TEMPLATE',
+        meta: {
+          method: 'PUT',
+          url: `/v1/templates/${id}`,
+          data: { ...data, published: true },
```
bkemper: Do we need to spread data here if we are awaiting update?

src/pages/templatesV2/CreatePage.js
```diff
@@ -14,18 +14,50 @@ export default class CreatePage extends Component {
...
+      ...values,
+      content: {
+        ...values.content,
+        text: '' // add some content to avoid api validation error
+      }
+    };
+    const templateId = values.id;
+    const createPromise = create(formData);
+    // This is used as a place to store meta and substitution data temporarily until test data can be stored with a draft
+    const createRecipientListPromise = createRecipientList({
```
bkemper: As discussed, we should maintain the current experience using localstorage.

src/pages/templatesV2/components/Editor.js
```diff
@@ -3,7 +3,7 @@ import AceEditor from 'react-ace';
 import useEditor from '../hooks/useEditor';
 import 'brace/ext/searchbox';
 import 'brace/mode/text';
-import 'brace/theme/chaos';
+import 'brace/theme/tomorrow_night_eighties';
```
bkemper: Did you run this by Lynn?

nicklemmon:
> Do we want to only save test data when the template is saved? I would recommend separating them.
> The test data needs to be used when sending a test email.

The test data is sent with the preview so it is used - it just isn't saved unless done so explicitly.

You're not sure about that?

src/actions/templates.js
```diff
@@ -101,6 +101,41 @@ export function create(data) {
...
 }
 
+export function createV2(data) {
+  const {
+    id,
+    assignTo,
+    subaccount,
+    content,
+    testData,
+    ...formData
```
bkemper: :)

src/pages/templatesV2/components/Editor.js
nicklemmon: I did not, but I can!

src/actions/templates.js
```diff
@@ -140,7 +233,7 @@ export function publish(data, subaccountId) {
         meta: {
           method: 'PUT',
           url: `/v1/templates/${id}`,
-          data: { published: true },
+          data: { ...data, published: true },
```
bkemper: Do we need this?

src/actions/templates.js
```diff
@@ -126,6 +161,64 @@ export function update(data, subaccountId, params = {}) {
...
+    try {
+      await dispatch(update(data, subaccountId));
+      dispatch(setTestDataV2({ id, mode: 'published', data: testData }));
+
+      await dispatch(sparkpostApiRequest({
+        type: 'PUBLISH_TEMPLATE',
+        meta: {
+          method: 'PUT',
+          url: `/v1/templates/${id}`,
+          data: { ...data, published: true },
```
bkemper: I think spreading data is dangerous.  It is going to include id, testData, and other data that is not needed.

src/actions/templates.js
```diff
@@ -163,14 +256,66 @@ export function deleteTemplate(id, subaccountId) {
...
   return (dispatch, getState) => {
     const username = getState().currentUser.username;
     const testData = typeof data === 'object' ? JSON.stringify(data) : data;
+
     return localforage.setItem(getTestDataKey({ id, username, mode }), testData).then(() => dispatch({ type: 'SET_TEMPLATE_TEST_DATA' }));
   };
 }
 
+export function getTestDataV2({ id, mode }) {
+  return (dispatch, getState) => {
```
bkemper: This and all the other v2 test data action thunks aren't dispatching an action.  This to me is an indicator that we should move this out of redux.

src/actions/tests/templates.test.js
```diff
@@ -19,9 +19,11 @@ describe('Action Creator: Templates', () => {
     }
   };
 
-  beforeEach(async() => {
+  beforeEach(async () => {
     localforage.setItem = jest.fn((a) => Promise.resolve(a));
     localforage.getItem = jest.fn(() => Promise.resolve(null));
+    window.localStorage.setItem = jest.fn();
+    window.localStorage.getItem = jest.fn();
```
bkemper: I just saw the spies below, so I'm assuming these didn't work and can be removed.

src/actions/tests/templates.test.js
```diff
@@ -152,4 +154,116 @@ describe('Action Creator: Templates', () => {
...
+    it('updates local storage with passed in data by invoking localStorage.setItem', async () => {
+      const data = {
+        options: {},
+        substitution_data: {
+          foo: 'bar'
+        },
+        metadata: {}
+      };
+      // See: https://github.com/facebook/jest/issues/6798#issuecomment-440988627
+      const spy = jest.spyOn(window.localStorage.__proto__, 'setItem');
```
bkemper: This is a good find, but what about a more global approach of mocking localStorage?

```
// add to config/jest/setup.js
const mockLocalStorage = {
  clear: jest.fn(),
  getItem: jest.fn(),
  removeItem: jest.fn(),
  setItem: jest.fn()
};

Object.defineProperty(global.window, 'localStorage', { value: mockLocalStorage });
```

```
// replace
const spy = jest.spyOn(window.localStorage.__proto__, 'setItem');
expect(spy).toHaveBeenCalledWith('key', JSON.stringify(data));

// with
expect(window.localStorage.setItem).toHaveBeenCalledWith('key', JSON.stringify(data));
```

src/pages/templatesV2/CreatePage.js
```diff
@@ -14,18 +14,44 @@ export default class CreatePage extends Component {
...
+    };
+    const templateId = values.id;
+    const testDataBase = {
+      options: {},
+      substitution_data: {},
+      metadata: {}
+    };
+    createTemplateV2({
+      ...formData,
+      testData: testDataBase
```
bkemper: hmm... we will probably need a follow-up ticket to set this based on account defaults.

src/pages/templatesV2/ListPage.js
```diff
@@ -42,12 +48,13 @@ export default class ListPage extends Component {
...
 
   toggleDuplicateModal = (template) => {
-    const { getDraft } = this.props;
+    const { getDraft, getTestDataV2 } = this.props;
 
     getDraft(template.id)
       .then((res) => {
         this.setState({
           templateToDuplicate: res,
+          testDataToDuplicate: getTestDataV2({ id: res.id, mode: res.published ? 'published' : 'draft' }),
```
bkemper: Like I was saying before, typically you don't want to rely on the response from the action creator.  Instead, you get the value from the redux store, but then again maybe we don't need to make it an action creator.

src/pages/templatesV2/components/EditTestDataSection.js
```diff
@@ -0,0 +1,24 @@
...
+    isPublishedMode
+  } = useEditorContext();
+
+  return (
+    <Editor
+      mode="json"
+      name="test-data-editor"
+      onChange={(value) => setTestData(value)}
+      value={testData}
+      readOnly={isPublishedMode}
```
bkemper: This should be editable, so users can preview and send tests of the published template.

src/pages/templatesV2/components/Editor.js
```diff
@@ -26,10 +26,10 @@ const Editor = ({
           ...editorProps,
           $blockScrolling: Infinity
         }}
-        fontSize={12}
+        fontSize={14}
         height="100%" // must set height on wrapper
         highlightActiveLine
-        // note, must global import modes
+        // note, must global import modes from https://www.npmjs.com/package/brace
```
bkemper: ❤️

src/pages/templatesV2/components/PreviewSection.js
```diff
@@ -9,25 +9,53 @@ import PreviewContainer from './PreviewContainer';
...
+          return `<p style="white-space: pre-wrap">${preview.text}</p>`;
+        } else {
+          return '';
+        }
+
+      case 'text':
+        // Must wrap text content in <p> to apply style and must be a string for injecting into iframe
+        return `<p style="white-space: pre-wrap">${preview.text}</p>`;
+
+      default:
```
bkemper: Is this a case?

src/pages/templatesV2/components/PreviewSection.js
```diff
@@ -9,25 +9,53 @@ import PreviewContainer from './PreviewContainer';
...
-  const content = currentTabKey === 'text'
-    ? `<p style="white-space: pre-wrap">${preview[currentTabKey]}</p>`
-    : preview[currentTabKey];
+  const getPreviewContent = (tabKey) => {
+    switch (tabKey) {
+      case 'html':
+      case 'amp_html':
+        return preview[tabKey];
+
+      case 'test_data':
```
bkemper: This function and case statement, specifically this case took me a minute to understand what was happening.  What do you think about simplifying this logic into two conditions that isolate the exceptions (when my current tab key isn't a content part use the first available content, and wrap text part with <p>).


```
  const currentContentKey = !(currentTabKey in preview)
    ? ['html', 'amp_html', 'text'].find((key) => (preview[key] || '').trim())
    : currentTabKey;

  const content = currentContentKey === 'text'
    ? `<p style="white-space: pre-wrap">${preview[currentContentKey]}</p>`
    : preview[currentContentKey];
```

src/pages/templatesV2/components/editorActions/PublishedModeActions.js
```diff
@@ -93,8 +95,10 @@ const PublishedModeActions = () => {
           onClose={handleModalClose}
           template={draft}
           contentToDuplicate={content}
-          createTemplate={createTemplate}
+          testDataToDuplicate={testData}
+          createTemplate={createTemplateV2}
```
bkemper: Hmm... maybe the modal component should just use the context hook.

src/actions/templates.js
nicklemmon: This might be a large-ish refactor - what do you think about moving forward with this implementation and cutting a ticket to refactor?

src/actions/templates.js
nicklemmon: Let me double check why I did that.

src/pages/templatesV2/components/EditTestDataSection.js
nicklemmon: It should always be editable? As the UX exists currently, if the template is published none of it is editable - not just test data, but the HTML, the text, and the AMP tabs are all read only.

It was that way before I started working, but could be an inaccurate implementation of requirements?

src/pages/templatesV2/components/editorActions/PublishedModeActions.js
nicklemmon: Well, the editor context exists on the edit view, but there is no such context for the list page, so I chose the prop route. We could add some context to the ListPage to address?

src/pages/templatesV2/ListPage.js
nicklemmon: In general, it feels like Redux should basically be removed from this feature, but I'm not sure we have time to do that before the end of the month...

src/actions/tests/templates.test.js
nicklemmon: I actually tried that and it didn't work for some reason - this was the only way I could get it to work.

src/actions/templates.js
nicklemmon: You're right, not needed. I read the docs incorrectly.

src/actions/tests/templates.test.js
nicklemmon: hmm, never mind - I tried what you posted and that works just fine. Not sure what I was missing earlier!

src/pages/templatesV2/components/PreviewSection.js
nicklemmon: Thinking it over, I think this actually messes with the logic a bit - in my mind, the preview should render blank if you are on "Text" and have no content there - I was mainly trying to make sure that when on test data, the first available. What about something like this?

```js
const formatTextContent = (textContent) => `<p style="white-space: pre-wrap">${textContent}</p>`;
  const getPreviewContent = (tabKey) => {
    if (tabKey === 'text') {
      return formatTextContent(preview.text);
    }

    if (tabKey === 'test_data') {
      const keyWithContent = ['html', 'amp_html', 'text'].find((key) => preview[key]);

      if (keyWithContent === 'text') {
        return formatTextContent(preview.text);
      }

      return preview[keyWithContent];
    }

    return preview[tabKey];
  };
```

nicklemmon:
@bkemper Ready for another pass when you have time. Thanks again 👍

nicklemmon:
@bkemper Just a heads up, I went ahead and added some Cypress tests as well to sort of to an end-to-end run so that refactoring will be less risky with future branches.

src/pages/templatesV2/components/editorActions/PublishedModeActions.js
bkemper: Ahh, gotcha.

src/pages/templatesV2/components/Editor.js
nicklemmon: She was cool with the change.

src/selectors/account.js
```diff
@@ -23,9 +23,9 @@ export const selectDefaultTemplateOptions = createSelector(
   [getAccount],
   ({
     options: {
-      click_tracking,
-      rest_tracking_default,
-      transactional_default
+      click_tracking = false,
+      rest_tracking_default = false,
```
bkemper: I'm sorry, these defaults need to match the API defaults.

```
click_tracking = true
rest_tracking_default = true
transactional_default = false
```

nicklemmon: Done, and added a supporting test to boot.

src/pages/templatesV2/components/editorActions/SaveAndPublishConfirmationModal.js
```diff
@@ -1,28 +1,35 @@
...
-  const handleConfirm = useCallback(() => publishDraft(draft, draft.subaccount_id)
-    .then(() => {
-      setHasSaved(true);
-      setSuccessRedirect(true);
-    }), [draft, publishDraft, setHasSaved]);
+  const handleConfirm = () => {
+    publishDraftV2({
+      ...draft,
+      content,
+      options: parsedTestData.options, // NOTE FOR REVIEWERS: Not sure if this is the right thing to pass here - could be wrong
```
bkemper: This needs to be removed.  Test data should only be used with preview and sending a test email.

src/pages/templatesV2/hooks/useEditorTestData.js
```diff
@@ -0,0 +1,49 @@
...
+
+const useEditorTestData = (props) => {
+  const {
+    draft,
+    templateTestData,
+    getTestDataV2,
+    isPublishedMode
+  } = props;
+  const [testData, setTestData] = useState(templateTestData);
+  const [parsedTestData, setParsedTestData] = useState(undefined);
```
bkemper: Should we default this to an empty object the way we do when a parsing error is caught?

src/pages/templatesV2/components/editorActions/SaveAndPublishConfirmationModal.js
nicklemmon: Thanks for catching that.

src/pages/templatesV2/hooks/useEditorTestData.js
nicklemmon: Sure! Extra safe that way. I'll double check it doesn't cause anything weird.

src/pages/templatesV2/hooks/useEditorTestData.js
```diff
@@ -0,0 +1,49 @@
...
+          id: draft.id,
+          mode: isPublishedMode ? 'published' : 'draft'
+        })
+      }, null, 2));
+    }
+  }, [draft, getTestDataV2, isPublishedMode]);
+
+  useEffect(() => {
+    try {
+      setParsedTestData(JSON.parse(testData));
```
bkemper: Does this need to be a different side effect?  I think there is really only one side effect, when draft changes set test data and parse it.

bkemper: Please ignore.  I see now.

src/actions/templates.js
```diff
@@ -239,3 +384,34 @@ export function sendPreview({ id, mode, emails, from, subaccountId }) {
...
   };
 }
+
+export function sendPreviewV2({ id, mode, emails, from, subaccountId }) {
+  const recipients = emails.map((email) => ({
+    address: { email }
+  }));
+
+  return async (dispatch) => {
+    const { payload: testData = {}} = await dispatch(getTestDataV2({ id, mode }));
```
bkemper: The return value is the `testData`, so this deconstruction is not needed.

src/actions/templates.js
```diff
@@ -163,14 +256,66 @@ export function deleteTemplate(id, subaccountId) {
...
   };
 }
 
+export function getTestDataV2({ id, mode }) {
+  return (dispatch, getState) => {
+    const username = getState().currentUser.username;
+    const rawData = window.localStorage.getItem(getTestDataKey({ id, username, mode }));
+
+    try {
+      return JSON.parse(rawData);
```
bkemper: To prevent that null case, maybe something like `JSON.parse(rawData) || {};`.

src/pages/templatesV2/hooks/useEditorTestData.js
```diff
@@ -0,0 +1,49 @@
...
+    } catch {
```
bkemper: Catch without the error!!! 👍

src/pages/templatesV2/components/editorActions/SaveAndPublishConfirmationModal.js
```diff
@@ -1,28 +1,35 @@
...
+      testData
```
bkemper: This should pass `parsedTestData` to avoid saving anything the user types in.

src/pages/templatesV2/components/editorActions/SaveDraft.js
```diff
@@ -8,17 +8,22 @@ const SaveDraft = (props) => {
...
   const handleClick = useCallback(() => {
     if (onClick) {
       onClick();
     }
 
-    updateDraft({ id: draft.id, content }, draft.subaccount_id)
+    updateDraftV2({
+      id: draft.id,
+      content,
+      testData
```
bkemper: This should pass `parsedTestData` to avoid saving anything the user types in.

src/pages/templatesV2/hooks/tests/useEditorTestData.test.js
```diff
@@ -0,0 +1,68 @@
...
+    Attempted to log "Error: Warning: An update to %s inside a test was not wrapped in act(...).
+
+    When testing, code that causes React state updates should be wrapped into act(...):
+
+    act(() => {
+      // fire events that update state
+    });
+    // assert on the output
+  */
+  // it('updates test data via `getRecipientList` when the draft state changes', async () => {
```
bkemper: I can help tomorrow.

src/pages/templatesV2/components/editorActions/SaveAndPublishConfirmationModal.js
nicklemmon: I can update this for all of the methods if that's your concern - though technically a user could put whatever they wanted in to local storage and send it along with a request.

src/pages/templatesV2/hooks/tests/useEditorTestData.test.js
nicklemmon: @bkemper All other feedback wrapped up except this one. Let me know if you have time to take a look.

src/actions/templates.js
```diff
@@ -239,3 +384,34 @@ export function sendPreview({ id, mode, emails, from, subaccountId }) {
...
   };
 }
+
+export function sendPreviewV2({ id, mode, emails, from, subaccountId }) {
+  const recipients = emails.map((email) => ({
+    address: { email }
+  }));
+
+  return async (dispatch) => {
+    const { testData = {}} = await dispatch(getTestDataV2({ id, mode }));
```
bkemper: Remove destructuring.

nicklemmon:
@bkemper Alright, added those last tests. Let me know what other feedback you have.

Thanks!

src/pages/templatesV2/hooks/tests/useEditorTestData.test.js
```diff
@@ -0,0 +1,89 @@
...
+      metadata: { meta: 'data' },
+      options: { foo: 'bar' }
+    };
+    const mockGetTestData = () => preExistingTestData;
+    const wrapper = useTestWrapper({
+      draft,
+      getTestDataV2: mockGetTestData
+    });
+    const { parsedTestData } = useHook(wrapper);
+    expect(parsedTestData).toEqual(preExistingTestData);
```
bkemper: 👏

src/pages/templatesV2/hooks/useEditorTestData.js
```diff
@@ -0,0 +1,54 @@
...
+          id: draft.id,
+          mode: isPublishedMode ? 'published' : 'draft'
+        })
+      }, null, 2));
+    }
+  }, [draft, getTestDataV2, isPublishedMode]);
+
+  useEffect(() => {
+    try {
+      setParsedTestData(JSON.parse(testData));
```
bkemper: To avoid getting in a bad state, what do you think about taking a very defensive approach...

```
const results = JSON.parse(testData);
const nextParsedTestData = _.pick(results, ['metadata', 'options', 'substitution_data']);

setParsedTestData(nextParsedTestData);
```

cypress/tests/templates/templates.js
```diff
@@ -0,0 +1,192 @@
...
+    cy.server();
+
+    cy.route('POST', '/api/v1/utils/content-previewer').as('contentPreviewer');
+    cy.route('POST', '/api/v1/templates/**').as('templatesPost');
+    cy.route('PUT', '/api/v1/templates/**').as('templatesPut');
+    cy.route('GET', '/api/v1/templates/**').as('templatesGet');
+    cy.route('DELETE', '/api/v1/templates/**').as('templatesDelete');
+  });
+
+  describe('runs through the creation, duplication, and deletion flow end-to-end', () => {
```
bkemper: Very cool, I'm super interested in learning more!

src/pages/templatesV2/hooks/useEditorTestData.js
nicklemmon: Screen Shot 2019-10-24 at 8 47 36 AM

Hmm, still need to have the `JSON.parse()` in a `try/catch` to make this work.

nicklemmon:
```js
let results;

    try {
      results = JSON.parse(testData);
    } catch (err) {
      // See: https://2ality.com/2017/08/optional-catch-binding.html#use-case%3A-json.parse()
      if (err instanceof SyntaxError) {
        results = {};
      } else {
        throw err;
      }
    }
    const nextParsedTestData = _.pick(results, ['metadata', 'options', 'substitution_data']);

    setParsedTestData(nextParsedTestData);
```

This does the trick though. Thoughts?

bkemper: 👏   This is perfect.

PR #1211, opened 2019-10-03
SA-897 - Health Score V3 visualization. 

kchu1337:
### What Changed
 - Added the Health Score chart to `signals/health-scoreV3`.

### How To Test
 - Enable the `allow_health_score_V3` flag. 
 - Go to any health score details page. Change the url manually from `signals/health-score` to `signals/health-scoreV3`. Keep the other route params.
 - Check that the graph is working. 

### Notes
Most of the stuff was copied from #1016


src/selectors/signals.js
```diff
@@ -312,6 +312,76 @@ export const selectHealthScoreDetails = createSelector(
...
+      },
+      from, to
+    });
+
+    // Merge in rankings
+    const mergedHistory = filledHistory.map((healthData) => {
+      const ranking = rankHealthScore(roundToPlaces(healthData.health_score * 100, 1));
+      return ({
+        ranking: ranking,
+        fill: fillHealthScoreColor(ranking),
```
kchu1337: This should be the only difference between this selector and `selectHealthScoreDetails`

src/selectors/signals.js
```diff
@@ -18,13 +18,13 @@ export const getOptions = (state, { now = moment().subtract(1, 'day'), ...option
 });
 
 // Redux store
-export const getSpamHitsData = (state, props) => _.get(state, 'signals.spamHits', {});
-export const getEngagementRecencyData = (state, props) => _.get(state, 'signals.engagementRecency', {});
-export const getEngagementRateByCohortData = (state, props) => _.get(state, 'signals.engagementRateByCohort', {});
-export const getUnsubscribeRateByCohortData = (state, props) => _.get(state, 'signals.unsubscribeRateByCohort', {});
-export const getComplaintsByCohortData = (state, props) => _.get(state, 'signals.complaintsByCohort', {});
-export const getHealthScoreData = (state, props) => _.get(state, 'signals.healthScore', {});
-export const getCurrentHealthScoreData = (state, props) => _.get(state, 'signals.currentHealthScore', {});
```
kchu1337: Eslint yelling at me 😓

src/selectors/signals.js
```diff
@@ -387,7 +457,7 @@ export const selectEngagementRecencyOverview = createSelector(
 
 export const selectHealthScoreOverviewData = createSelector(
   getHealthScoreData, getOptions,
-  ({ data }, { from, to }) => data.map(({ current_health_score, WoW, ...rowOfData }) => {
```
kchu1337: Eslint yelling at me 😓

src/pages/signals/components/HealthScoreLegend.module.scss
```diff
@@ -0,0 +1,45 @@
...
+  background: #4194ED;
+}
+
+.Desc {
+  position: relative;
+  top: 13px;
+
+  font-size: font-size(200);
+  line-height: 1em;
+  color: color(gray, 4);
```
nicklemmon: This element does not meet required color contrast ratios:
Screen Shot 2019-10-07 at 1 03 02 PM

src/pages/signals/HealthScorePageV3.js
```diff
@@ -22,13 +110,31 @@ export function HealthScorePageV3(props) {
         >
           Create Alert
         </Button>}>
+      <Panel>
+        <Panel.Section>
+          <Grid>
```
nicklemmon: The chart is pretty much inaccessible for screen reader only users, and as a result, needs to be hidden from those users. In addition, the data needs to be exposed for screen reader only users in a different format.

- So for the `<Grid>` component here you can add the `aria-hidden="true"`
- To present the data for screen reader only users, add a `<table></table>` that is wrapped in the Matchbox `<ScreenReaderOnly>` only component - the table just needs to render the data (in this case, probably all of the available health scores. It would also be worth adding a column that indicates the "status" of that particular health score that maps to the legend that you built for this feature.

Let me know if you could use some help implementing this - we haven't really done this yet.

src/selectors/signals.js
nicklemmon: If this didn't break anything, then this is a good update 👍

src/selectors/signals.js
```diff
@@ -312,6 +312,38 @@ export const selectHealthScoreDetails = createSelector(
   }
 );
 
+const fillHealthScoreColor = _.memoize((rank) => {
+  switch (rank) {
+    case ('danger'):
+      return '#D1E4F4';
+    case ('warning'):
+      return '#91C5FD';
+    case ('good'):
```
nicklemmon: I don't think the parens are necessary here?

```js
switch (rank) {
  case 'danger':
    return ...
  case 'warning':
    return ...
}
```

src/pages/signals/components/HealthScoreLegend.module.scss
```diff
@@ -0,0 +1,45 @@
...
+}
+
+
+.Label {
+  flex: 0 0 0;
```
nicklemmon: Shouldn't this work?

```css
flex: 0;
```

src/pages/signals/HealthScorePageV3.js
```diff
@@ -1,12 +1,100 @@
-import React from 'react';
+import React, { useCallback } from 'react';
```
nicklemmon: I'm seeing this when navigating between `/signals/health-score` and `/signals/health-scoreV3/<param>/<param>`:
Screen Shot 2019-10-07 at 11 15 45 AM

This might not actually be a legit issue once the old implementation is phased out.

src/pages/signals/HealthScorePageV3.js
```diff
@@ -1,12 +1,100 @@
...
+import moment from 'moment';
+import HealthScoreMetric from './components/HealthScoreMetric';
+import DateFilter from './components/filters/DateFilter';
+import HealthScoreLegend from './components/HealthScoreLegend';
+import Callout from 'src/components/callout/Callout';
+import { Loading } from 'src/components';
 
 export function HealthScorePageV3(props) {
-  const { facet, facetId, subaccountId } = props;
+  const {
```
nicklemmon: The page `<title></title>` is not updating to reflect the current page state. It should probably be:

`Master Account | Health Score | Signals | SparkPost`

Can we use React Helmet to achieve this?

src/pages/signals/HealthScorePageV3.js
```diff
@@ -1,12 +1,100 @@
...
+        xAxisRefLines={newModelLine}
+        yKey='health_score'
+        yAxisProps={{
+          ticks: [0, 0.55, 0.8, 1],
+          tickFormatter: (tick) => parseInt(tick * 100)
+        }}
+        xAxisProps={getXAxisProps()}
+      />);
+  };
+
```
nicklemmon: Just putting this comment in a random place haha! So are these two features in the mock out of scope for now?

Screen Shot 2019-10-07 at 1 28 53 PM
Screen Shot 2019-10-07 at 1 28 51 PM

src/pages/signals/HealthScorePageV3.js
```diff
@@ -1,12 +1,100 @@
...
+    return (
+      <BarChart
+        margin={newModelMarginsHealthScore}
+        gap={gap}
+        onClick={handleDateSelect}
+        onMouseOver={handleDateHover}
+        onMouseOut={resetDateHover}
+        disableHover={false}
+        selected={selectedDate}
+        hovered={hoveredDate}
```
nicklemmon: Is there any way to make these interactions occur on keyboard focus? This is technically a big a11y problem otherwise...if this may require a significant component update then it would probably be worth getting a new ticket written to do so.

src/pages/signals/components/tests/HealthScoreMetric.test.js
```diff
@@ -0,0 +1,22 @@
...
+      injections: 100
+    };
+    wrapper = shallow(<HealthScoreMetric {...props}/>);
+  });
+
+  it('renders correctly', () => {
+    expect(wrapper).toMatchSnapshot();
+  });
+
+});
```
nicklemmon: Not sure this test has a lot of value either. Since it isn't going to be re-used anywhere else, it would only get in the way of future refactoring. Ideally, this could be tested a level above this component to make said refactoring possible.

src/pages/signals/HealthScorePageV3.js
kchu1337: I am using a wrapper for the Page component in matchbox. From your Cyrpess PR it looks like it might be an issue with the matchbox component?

src/pages/signals/HealthScorePageV3.js
kchu1337: Yes, the additional charts + recommendations will be in another ticket

src/pages/signals/HealthScorePageV3.js
nicklemmon: It looks like the page titles are controlled in the routes config:
https://github.com/SparkPost/2web2ui/blob/SA-897/src/config/routes/index.js#L242-L331

src/pages/signals/HealthScorePageV3.js
kchu1337: We use an external library Recharts and it looks like it only supports keyboard events.

src/config/routes/index.js
```diff
@@ -279,7 +279,7 @@ const routes = [
     component: signals.HealthScorePageV3,
     condition: isAccountUiOptionSet('allow_health_score_v3'),
     layout: App,
-    title: 'Signals',
+    title: 'Signals | Health Score',
```
nicklemmon: Other way around actually:

`Health Score | Signals`

It renders from most specific to least specific:

`Health Score | Signals | SparkPost`

src/pages/signals/HealthScorePageV3.js
nicklemmon: Will resolve this as part of a future ticket related to chart a11y

src/pages/signals/HealthScorePageV3.js
nicklemmon: @kchu1337 were you able to render the screen reader only `<table>`? All I see is the aria attribute on the Panel.

nicklemmon: Screen Shot 2019-10-16 at 10 21 32 AM

So adding `aria-hidden` here isn't the right approach to take until the screen reader only `<table>` is present - if that isn't being tackled now, I would remove it.

src/pages/signals/components/HealthScoreLegend.module.scss
```diff
@@ -0,0 +1,33 @@
...
+}
+
+.danger, .warning, .good {
+  position: relative;
+  height: 10px;
+}
+
+.Desc {
+  position: relative;
+  top: 13px;
```
nicklemmon: Is there a reason standard spacing units aren't used here, i.e. `spacing(small)`?

PR #1212, opened 2019-10-04
AC-1013 - New help contact address for TFA or SSO users who are locked out

shrirupa:
AC-1013 - New help contact address for TFA or SSO users who are locked out

### What Changed
- All errors associated with TFA logging in append "Please contact login.issues@sparkpost.com for assistance" in the UI.
- All errors associated with SSO logging in append "Please contact login.issues@sparkpost.com for assistance" in the UI.

### How To Test
 - Try out TFA login and check the error message.
 - Try out SSO login and check the error message.

### To Do
- [ ] Address Feedback

### DIFF
--- a/src/pages/auth/components/SsoLoginForm.js
+++ b/src/pages/auth/components/SsoLoginForm.js
@@ -13,7 +13,7 @@ export class SsoLoginForm extends React.Component {
 
     return (
       <React.Fragment>
-        {loginError && <Error error={`${loginError} Please contact login.issues@sparkpost.com for assistance`} />}
+        {loginError && <Error error={`${loginError}. Please contact login.issues@sparkpost.com for assistance.`} />}
         <form onSubmit={handleSubmit}>
           <Field
             autoFocus
--- a/src/pages/auth/components/SsoLoginForm.js
+++ b/src/pages/auth/components/SsoLoginForm.js
@@ -13,7 +13,7 @@ export class SsoLoginForm extends React.Component {
 
     return (
       <React.Fragment>
-        {loginError && <Error error={loginError} />}
+        {loginError && <Error error={`${loginError} Please contact login.issues@sparkpost.com for assistance`} />}
         <form onSubmit={handleSubmit}>
           <Field
             autoFocus
--- a/src/pages/auth/components/TfaForm.js
+++ b/src/pages/auth/components/TfaForm.js
@@ -5,7 +5,7 @@ import { required } from 'src/helpers/validation';
 import { TextFieldWrapper } from 'src/components';
 import { verifyAndLogin } from 'src/actions/tfa';
 
-import { Button, UnstyledLink } from '@sparkpost/matchbox';
+import { Button, Error, UnstyledLink } from '@sparkpost/matchbox';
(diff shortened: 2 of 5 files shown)


src/pages/auth/components/TfaForm.js
```diff
@@ -11,18 +11,18 @@ import { LINKS } from 'src/constants';
 export class TfaForm extends Component {
 
   handleSubmit = ({ code }) => {
-    const { enabled, ...authData } = this.props.tfa;
+    const { _enabled, ...authData } = this.props.tfa;
```
PatrickSison: The reducer sets the variable as enabled so it does not remove it from spread anymore. It might be worth changing to `{ enabled: _enabled, ...authData }` so that we keep it marked as being unused but removes it from the spread

src/pages/auth/components/SsoLoginForm.js
```diff
@@ -13,7 +13,7 @@ export class SsoLoginForm extends React.Component {
 
     return (
       <React.Fragment>
-        {loginError && <Error error={loginError} />}
+        {loginError && <Error error={`${loginError}. Please contact login.issues@sparkpost.com for assistance.`} />}
```
PatrickSison: Might be thinking too edge case, but if the login error (which comes externally) has a period, there will be duplicate periods

src/pages/auth/components/TfaForm.js
```diff
@@ -11,18 +11,18 @@ import { LINKS } from 'src/constants';
...
 
   handleSubmit = ({ code }) => {
-    const { enabled, ...authData } = this.props.tfa;
+    const { enabled: _enabled, ...authData } = this.props.tfa;
 
     return this.props.verifyAndLogin({ authData, code }).catch((err) => {
       if (err.response.status === 400) {
         throw new SubmissionError({
-          code: 'The code is invalid'
+          code: 'The code is invalid. Please contact login.issues@sparkpost.com for assistance.'
```
PatrickSison: Is there potentially other errors that could occur during this time that we should be catching here? Any 5XX errors?

shrirupa: I don't know, ticket specifies the issue of users being locked out, 5XX errors are server related, should we be handling them by throwing this error message ? I

shrirupa: I will add one for 403

src/pages/auth/components/TfaForm.js
```diff
@@ -40,7 +40,7 @@ export class TfaForm extends Component {
             component={TextFieldWrapper}
             validate={required}
           />
-
+          {error && <Error wrapper='div' error={error}/>}
```
PatrickSison: The error needs to be repositioned. 
Screen Shot 2019-10-10 at 3 55 55 PM

You can probably remove this line, and in <Field .../>, stick a prop as `error={error`

shrirupa: Field doesn't take in `error` prop. I will try to modify the css of `Field`.

PatrickSison: It seemed to work for me. I was able to pass in the error that was created

PR #1213, opened 2019-10-04
SA-954 - Improve Selected and Hovered states

kchu1337:
### What Changed
 - Added a flag `shouldHighlightSelected` to Barchart. This is set to false in the default state which shows the latest recommendations but does not highlight the bar. 
 - Added the flag as props to signals charts. 

### How To Test
 - Go to health score/spam trap/engagement recency pages and hover/select the bars in each graph.

### DIFF
--- a/src/pages/signals/components/charts/barchart/BarChart.js
+++ b/src/pages/signals/components/charts/barchart/BarChart.js
@@ -23,17 +23,18 @@ import healthScoreThresholds from '../../../constants/healthScoreThresholds';
  */
 class BarChart extends Component {
 
-  getSelectedStyle = _.memoize((key, isActiveDate, selectedFill) =>
+  getSelectedStyle = _.memoize((key, isActiveDate, ranking, activeFill) =>
     {
+      const selectedFill = (key === 'health_score' && ranking) ? healthScoreThresholds[ranking].barColor : activeFill;
       if (key === 'health_score' || key === 'injections' || key === 'weight_value')  {
        return isActiveDate ? {fill: selectedFill} : {};
       }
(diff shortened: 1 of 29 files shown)


src/pages/signals/components/charts/barchart/BarChart.js
```diff
@@ -197,6 +200,7 @@ BarChart.defaultProps = {
   gap: 1,
   height: 250,
   margin: { top: 12, left: 18, right: 0, bottom: 5 },
+  shouldHighlightSelected: true,
```
shrirupa: since propTypes are used on this page, should we have `shouldHighlightSelected` added in that `shouldHighlightSelected: PropTypes.bool`

src/pages/signals/components/charts/barchart/tests/BarChart.test.js
```diff
@@ -50,6 +50,13 @@ describe('BarChart Component', () => {
     expect(wrapper.find({ dataKey: key }).at(0).props().shape(payload).props.fill).toEqual('#fill');
```
shrirupa: should we have a test to for selecting and unselecting action on bar chart rather than having it on withDateSelection, as that is checking the implementation detail there.

kchu1337: I don't think it is necessary since the `Bar` component comes from a library and it has a built in prop `onCLick` that takes a function. There's docs here: http://recharts.org/en-US/api/Bar.
IMO, we would be basically testing that the `onClick` prop works properly since we would mock the click action in jest for a unit tests..
It might make sense to do so for a functional test but not for a unit test.

shrirupa:
Looks good to me.

PR #1215, opened 2019-10-08
AC-987: Fixed promo code state bug

PatrickSison:
### What Changed
 - Fix bug with query params being cleared before loading bundles or async initialize state

### How To Test
 - Navigate to http://app.sparkpost.test/account/billing/plan?code=100K-premier-0519&promo=THXFISH2 on staging
 - Check that behavior is correct

### To Do
- [ ] Address feedback


PatrickSison:
One thing I'd like to look into later is how we use hooks in an async loading component. Otherwise, we're gonna end up with a lot of janky checks on if one effect was run before another.

PR #1219, opened 2019-10-10
SE-269 Updating look and feel of MessageEventsPage

jmoyse:
This is updating the events search components to latest mocks. This removes the view details button and replaces it with clicking directly on the row. The selected row should now highlight on a mouse over as well. Subject is now included in the table in addition to the ordering of the rows being switched around. 

The view button component and tests have been removed since they are no longer needed.


src/pages/reports/messageEvents/components/tests/MessageEventRow.test.js
```diff
@@ -0,0 +1,23 @@
...
+      message_id: '123abc',
+      event_id: '456xyz',
+      history: {
+        push: jest.fn()
+      }
+    }
+  };
+
+  it('renders correctly', () => {
+    expect(shallow(<MessageEventRow {...props} />)).toMatchSnapshot();
```
bkemper: Shouldn't we test the click handler?

src/pages/reports/messageEvents/components/MessageEventRow.module.scss
```diff
@@ -0,0 +1,19 @@
...
+    cursor: pointer;
+    background: lighten(color(gray, 9), 1);
+  }
+}
+
+.Cell {
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  color: color(gray, 4);
```
bkemper: Do we need to change the font color to play nicely with the hover style?

src/pages/reports/messageEvents/components/MessageEventRow.module.scss
```diff
@@ -0,0 +1,19 @@
...
+  &:not(.selected):hover {
+    cursor: pointer;
+    background: lighten(color(gray, 9), 1);
+  }
+}
+
+.Cell {
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
```
bkemper: Setting the ellipsis on every cell will cause the columns not to wrap and in this example, at about 1200px, I start to lose the timestamp column.  Personally, I think it is easier to read an email address when it doesn't wrap, but we should pass to UX for feedback.

Screen Shot 2019-10-10 at 4 16 49 PM

src/pages/reports/messageEvents/components/MessageEventRow.js
```diff
@@ -0,0 +1,30 @@
...
+
+export class MessageEventRow extends Component {
+  onClick = () => {
+    const url = getDetailsPath(this.props.event.message_id, this.props.event.event_id);
+    this.props.history.push(url);
+  };
+
+  render() {
+    return (
+      <Table.Row className={styles.Row} onClick={this.onClick}>
```
nicklemmon: So adding an `onClick` handlers to a `<tr>` isn't going to work from an a11y perspective - basically screen readers won't know how to handle this scenario, and keyboard-only users won't have a way to use the button.

I would advise working with UX to re-work the design a bit. Having the clickable element inside a table cell would be the better direction to take.

jmoyse:
From the discussions we're going to backlog this ticket until an UX redesign can occur. We can potentially use the work started here in the new version.

PR #1224, opened 2019-10-15
AC-988: Plan change submission checks conditions

PatrickSison:
### What Changed
 - Submit button shows up if conditions are met

### How To Test
 - Go to `/account/billing/plan`
 - Select a plan, and then acknowledge (for subaccounts, just reenter tab several times)
 - Check that submit button comes up once changes are fixed

### To Do
- [ ] Add tests
- [ ] Move submit button to new component
- [ ] Address feedback


src/pages/billing/components/SubmitSection.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
```
nicklemmon: Don't know where to put this so just putting it at the top!

Screen Shot 2019-10-17 at 3 52 50 PM

This was after selecting the first plan on `/account/billing/plan`

src/pages/billing/forms/NewChangePlanForm.js
```diff
@@ -73,8 +81,15 @@ export const ChangePlanForm = ({
     }
   },[selectedBundle, code, allBundles, updateRoute]);
 
+  const onSubmit = () => {
+  };
+
+  if (loading) {
+    return (<Loading />);
```
nicklemmon: No need for the parens here I don't think

src/pages/billing/forms/NewChangePlanForm.js
```diff
@@ -73,8 +81,15 @@ export const ChangePlanForm = ({
     }
   },[selectedBundle, code, allBundles, updateRoute]);
 
+  const onSubmit = () => {
```
nicklemmon: Empty function here looks like?

src/pages/billing/forms/NewChangePlanForm.js
```diff
@@ -73,8 +81,15 @@ export const ChangePlanForm = ({
...
+  };
+
+  if (loading) {
+    return (<Loading />);
+  }
+
   return (
-    <form >
+    <form onSubmit={handleSubmit(onSubmit)}>
```
nicklemmon: Looks like passed in prop is using the aforementioned empty function as an argument for some reason. Not sure why!

src/pages/billing/forms/NewChangePlanForm.js
```diff
@@ -118,7 +134,8 @@ const mapStateToProps = (state, props) => {
     bundles: selectTieredVisibleBundles(state),
     initialValues: changePlanInitialValues(state, { planCode, promoCode }),
     currentPlan: currentPlanSelector(state),
-    promoCodeObj: getPromoCodeObject(state)
+    promoCodeObj: getPromoCodeObject(state),
+    billing: state.billing
```
nicklemmon: Is this being referenced somewhere? Not clear why this was added - though I'm probably missing something.

src/pages/billing/components/SubmitSection.js
```diff
@@ -0,0 +1,35 @@
...
+import Brightback from 'src/components/brightback/Brightback';
+import config from 'src/config';
+import { Button } from '@sparkpost/matchbox';
+import { useFeatureChangeContext } from '../context/FeatureChangeContext';
+
+export const SubmitButton = ({ loading, condition }) => (
```
nicklemmon: I think if you brought this in to the `SubmitSection` return statement instead of breaking it out, then testing would be simpler (and no export required I don't think)

src/pages/billing/forms/NewChangePlanForm.js
PatrickSison: It's a placeholder, but it's easy enough to re-add back in. Will remove for now

src/pages/billing/forms/NewChangePlanForm.js
PatrickSison: The wonderful world of redux-form 😭

src/pages/billing/forms/NewChangePlanForm.js
PatrickSison: Placeholder. Will remove for now

src/pages/billing/components/SubmitSection.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
```
nicklemmon: I'm seeing that repeated Favicon request thing in this branch too?? What is that about.  But it only happens after I sit on `account/billing/plan` for about a minute.

I checked the `/dashboard` route to see if I could reproduce there, but no dice.

*BUT* I can reproduce the issue on `master` so it clearly isn't a problem with this branch.

src/pages/billing/components/SubmitSection.js
```diff
@@ -0,0 +1,33 @@
...
+import { useFeatureChangeContext } from '../context/FeatureChangeContext';
+
+const SubmitSection = ({ brightbackCondition, loading }) => {
+  const { isReady } = useFeatureChangeContext();
+  if (!isReady) {
+    return null;
+  }
+
+  return (
+    (
```
nicklemmon: What's with the extra parens here? Surprised ESLint wasn't worried about that one.

src/pages/billing/components/SubmitSection.js
PatrickSison: Fixed the issue. Removed the effect that was causing it. The effect only needed to happen upon an action, so put it there, rather than a derivative of it.

nicklemmon: Nice!

src/pages/billing/forms/NewChangePlanForm.js
```diff
@@ -64,17 +81,12 @@ export const ChangePlanForm = ({
...
-  },[selectedBundle, code, allBundles, updateRoute]);
+  if (loading) {
+    return <Loading />;
+  }
 
   return (
-    <form >
+    <form>
       <Grid>
         <Grid.Column xs={8}>
```
nicklemmon: Can we address this? I know this is technically beyond the scope of the current work, but would be an easy change to introduce:
Screen Shot 2019-10-22 at 3 28 11 PM

src/pages/billing/forms/NewChangePlanForm.js
```diff
@@ -15,47 +18,61 @@ import { FeatureChangeContextProvider } from '../context/FeatureChangeContext';
...
   promoCodeObj,
   clearPromoCode,
   getBundles
+
+  //redux-form
+  //handleSubmit
+
 }) => {
+
+  useEffect(() => { getBillingCountries(); }, [getBillingCountries]);
```
nicklemmon: hmm, so these effects run when `getBillingCountries` is available as a prop? Does this not work with an empty dependency array or is there a point in time in which there is no value for this prop?

In fact, I think you can combine them?

```js
useEffect(() => {
  getBillingCountries()
  getBillingInfo();
}, [])
```

src/pages/billing/components/tests/SubmitSection.test.js
```diff
@@ -0,0 +1,47 @@
...
+    it('should render the button', () => {
+      const wrapper = subject();
+      const button = getButton(wrapper, { enabled: true, to: 'url' });
+      expect(button).toMatchSnapshot();
+    });
+
+    it('should not use url if not enabled', () => {
+      const wrapper = subject();
+      const button = getButton(wrapper, { enabled: false, to: 'url' });
+      expect(button).toMatchSnapshot();
```
nicklemmon: I get why you're using a snapshot here, but I think an explicit check would be more self-documenting.

PatrickSison: I'll try this one again as explicit. Had trouble testing this explicitly since I couldn't shallow render it correctly.

nicklemmon: I'm not opposed to a full mount or even using `.dive()`!

src/pages/billing/forms/NewChangePlanForm.module.scss
```diff
@@ -0,0 +1,33 @@
+@import '~@sparkpost/matchbox/src/styles/config.scss';
+
```
nicklemmon: Nice.

Would you mind cleaning up spacing a bit in this file? A lot of blank empty lines and things. Do we have a Sass linter set up? It looks like maybe no. Maybe we should get an FE ticket for that.

nicklemmon:
@PatrickSison Had one small bit of feedback and looks like you probably have a snapshot failure as well.

We can roll this out tomorrow AM first thing.

PR #1225, opened 2019-10-15
SA-770 - Display results by seed addresses for mailbox provider

kchu1337:
### What Changed
 - Update StopTest component to be more independent. Moved the dispatched calls into the component rather than being passed as a prop.
 - Updated the InfoBox component to have custom css styling. Also removed the break tag as it didn't seem to affect the current implementations. 
 - Added new routes and action/reducers for the all messages page /endpoint. 
 - Created the all messages page. Currently only supports mailbox-providers but code for future sending ip support is included. 
 - Added link to the all messages page in the details page in the breakdown by mailbox provider section

### How To Test
 - Go to /inbox-placement. Click on any (real) test. On the details page, scroll down and click on one of the mailbox-providers. Verify that the new page looks like the mocks. 

### To Do
- [ ] UX feedback

### DIFF
--- a/src/actions/inboxPlacement.js
+++ b/src/actions/inboxPlacement.js
@@ -67,7 +67,14 @@ export function getAllInboxPlacementMessages(id, filters) {
     meta: {
       method: 'GET',
       url: `/v1/inbox-placement/${id}/messages`,
-      params: filters
+      params: filters,
+      showErrorAlert: false
     }
+  });
+}
+
+export function resetState() {
+  return (dispatch) => dispatch({
+    type: 'RESET_STATE'
   });
 }
--- a/src/actions/inboxPlacement.js
+++ b/src/actions/inboxPlacement.js
@@ -60,3 +60,14 @@ export function getInboxPlacementTestContent(id) {
     }
   });
 }
+
+export function getAllInboxPlacementMessages(id, filters) {
+  return sparkpostApiRequest({
+    type: 'GET_ALL_INBOX_PLACEMENT_MESSAGES',
+    meta: {
+      method: 'GET',
+      url: `/v1/inbox-placement/${id}/messages`,
+      params: filters
+    }
+  });
+}
--- a/src/pages/inboxPlacement/AllMessagesPage.js
+++ b/src/pages/inboxPlacement/AllMessagesPage.js
@@ -11,23 +11,22 @@ import InfoBlock from './components/InfoBlock';
 import styles from './AllMessagesPage.module.scss';
 import { formatPercent } from 'src/helpers/units';
 
-export const AllMessagesPage = (props) => {
-  const {
-    history,
-    id,
-    filterType,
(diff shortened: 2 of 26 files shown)


shrirupa:
Screen Shot 2019-10-17 at 11 05 33 AM
I think these columns should have equal width like in mock.May be we can add 
`width: 20% in .HeaderCell`

src/pages/inboxPlacement/AllMessagesPage.js
```diff
@@ -0,0 +1,118 @@
...
+    loading,
+    status,
+    sent,
+    placement,
+    authentication,
+    error,
+    getInboxPlacementByProviders,
+    getAllInboxPlacementMessages,
+    StopTestComponent = StopTest //This is for unit testing purposes
+  } = props;
```
shrirupa: Not necessary, but since we are not using `props` anywhere why not pass it destructured.
```
 export const AllMessagesPage = ({
    history,
    id,
    filterType,
    filterName,
```

shrirupa:
`Sent, Inbox, Spam, Missing, SPF, DKIM, DMARC` look smaller to me in mocks, may be font-size: 0.8rem

src/pages/inboxPlacement/components/AllMessagesCollection.js
```diff
@@ -0,0 +1,52 @@
...
+    case 'fail':
+      return <span className={styles.Fail}>Fail</span>;
+    default:
+      return '---';
+  }
+};
+const HeaderComponent = () => (<thead>
+  <Table.Row className={styles.HeaderRow}>
+    <Table.HeaderCell className={styles.HeaderCell}></Table.HeaderCell>
+    <Table.HeaderCell className={styles.HeaderCell}>SPF</Table.HeaderCell>
```
shrirupa: font-size and font-weight these are same as that of Sent, Inbox, Spam, Missing labels in mocks. 
```
font-size: 0.8rem;
font-weight: 500;
```

shrirupa:
> `Sent, Inbox, Spam, Missing, SPF, DKIM, DMARC` look smaller to me in mocks, may be font-size: 0.8rem

Other than this, Looks good to me.

src/actions/inboxPlacement.js
```diff
@@ -67,7 +67,14 @@ export function getAllInboxPlacementMessages(id, filters) {
...
       method: 'GET',
       url: `/v1/inbox-placement/${id}/messages`,
-      params: filters
+      params: filters,
+      showErrorAlert: false
     }
   });
 }
+
+export function resetState() {
```
shrirupa: Can't we have this done in 
```
    case 'GET_ALL_INBOX_PLACEMENT_MESSAGES_FAIL':
      return { state: initialState, getAllMessagesPending: false, getAllMessagesError: payload };
```

PR #1228, opened 2019-10-20
TR-1750: AMP enabled Test Emails Sent in-app always send with open tracking enabled

bkemper:
Refer to [TR-1750](https://jira.int.messagesystems.com/browse/TR-1750)

### What Changed
 - If template options are not define for a template, they should be initialized with default account options for both new and old editors 

### How To Test

Must test all template actions in both old and new template editors.  The expectation is if the template options are not set, they will use the account option defaults.

- Create
- Duplicate
- Edit
- Publish
- Preview
- Send Test

### DIFF
--- a/src/pages/templates/CreatePage.container.js
+++ b/src/pages/templates/CreatePage.container.js
@@ -7,7 +7,7 @@ import _ from 'lodash';
 import { create, getDraft } from 'src/actions/templates';
 import { showAlert } from 'src/actions/globalAlert';
 
-// Selectors
+import { selectDefaultTemplateOptions } from 'src/selectors/accessConditionState';
 import { selectAndCloneDraftById, selectDefaultTestData } from 'src/selectors/templates';
 
 import CreatePage from './CreatePage';
@@ -24,6 +24,7 @@ const mapStateToProps = (state, props) => ({
   formName: FORM_NAME,
   initialValues: {
     assignTo: 'master',
+    options: selectDefaultTemplateOptions(state),
     testData: selectDefaultTestData(),
     ...selectAndCloneDraftById(state, props.match.params.id)
   }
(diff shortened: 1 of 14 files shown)


src/selectors/accessConditionState.js
```diff
@@ -11,6 +11,21 @@ export const getCurrentAccountPlan = createSelector(
   (account, plans) => plans.find((plan) => plan.code === account.subscription.code) || {}
 );
 
+export const selectDefaultTemplateOptions = createSelector(
```
nicklemmon: I think it might be worth adding a comment just describing why this selector is in this file - it made sense when you explained it, but might be future to write down.

bkemper:
I ended up adding a comment and moving the selector with the other account selectors.  This is ready for final review.

src/selectors/accessConditionState.js
```diff
@@ -11,21 +11,6 @@ export const getCurrentAccountPlan = createSelector(
   (account, plans) => plans.find((plan) => plan.code === account.subscription.code) || {}
 );
 
-export const selectDefaultTemplateOptions = createSelector(
```
nicklemmon: Screen Shot 2019-10-22 at 9 03 48 AM

Tried to make another pass, but looks like some imports are a bit messed up!

bkemper: 😊

src/pages/templates/CreatePage.container.js
```diff
@@ -7,8 +7,8 @@ import _ from 'lodash';
 import { create, getDraft } from 'src/actions/templates';
```
nicklemmon: Screen Shot 2019-10-22 at 11 00 26 AM

The options don't appear to be making it through to the send preview request - that being said, I'm not sure that actually matters! The API docs don't reference options as a possibility, so this part of the request is probably doing nothing?

bkemper: The options are set on the template and our preview request is made with the template_id.  So, the preview will be inherit the options from the template.

PR #1232, opened 2019-10-23
AC-1033 - Credit Card in Form

shrirupa:
AC-1033 - Credit Card in Form

### What Changed
- In the new change plan form, the credit card form appears
- Credit card section loads once content in FeatureComparisonSection is complete
- Credit card section loads existing credit card information if available

### How To Test
 - Navigate to account/billing/plan
 - Check that credit card form loads if no card is present after the FeatureChangeSection is complete.
 - Check that credit card section loads with saved card information after the FeatureChangeSection is complete.
 -Credit card section doesn't show up when selected plan is free.

### To Do
- [ ] Address Feedback

### DIFF
--- a/src/pages/billing/components/CardSection.js
+++ b/src/pages/billing/components/CardSection.js
@@ -3,6 +3,7 @@ import { Panel } from '@sparkpost/matchbox';
 import CardSummary from './CardSummary';
 import PaymentForm from '../forms/fields/PaymentForm';
 import BillingAddressForm from '../forms/fields/BillingAddressForm';
+import { useFeatureChangeContext } from '../context/FeatureChangeContext';
 
 const FORMNAME = 'changePlan';
 
@@ -13,10 +14,14 @@ const CardSection = ({
   submitting,
   canUpdateBillingInfo
 }) => {
+  const { isReady } = useFeatureChangeContext();
   const [useSavedCC, setUseSavedCC] = useState(true);
 
   const handleCardToggle = () => setUseSavedCC(!useSavedCC);
 
+  if (!isReady) {
+    return null;
+  }
   if (selectedPlan.isFree) {
     return null; // CC not required on free plans
   }
--- a/src/pages/billing/components/CardSection.js
+++ b/src/pages/billing/components/CardSection.js
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React from 'react';
 import { Panel } from '@sparkpost/matchbox';
 import CardSummary from './CardSummary';
 import PaymentForm from '../forms/fields/PaymentForm';
@@ -13,12 +13,12 @@ const CardSection = ({
   account,
   submitting,
   canUpdateBillingInfo,
+  useSavedCC,
+  handleCardToggle,
   isNewChangePlanForm //TODO: remove this when removing the OldChangePlanForm
 }) => {
   const { isReady, loading } = useFeatureChangeContext();
-  const [useSavedCC, setUseSavedCC] = useState(true);
 
-  const handleCardToggle = () => setUseSavedCC(!useSavedCC);
 
   if ((!isReady || loading) && isNewChangePlanForm) {
     return null;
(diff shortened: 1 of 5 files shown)


src/pages/billing/components/CardSection.js
```diff
@@ -1,21 +1,28 @@
...
   account,
-  useSavedCC,
   submitting,
-  handleCardToggle,
-  canUpdateBillingInfo
+  canUpdateBillingInfo,
+  isNewChangePlanForm //TODO: remove this when removing the OldChangePlanForm
 }) => {
+  const { isReady, loading } = useFeatureChangeContext();
+  const [useSavedCC, setUseSavedCC] = useState(true);
```
PatrickSison: This will probably need to be brought up to the NewChangePlanForm level at some point so submission knows to use the form vs using the preexisting credit card

src/pages/billing/components/CardSection.js
```diff
@@ -1,21 +1,28 @@
...
 
 const FORMNAME = 'changePlan';
 
 const CardSection = ({
   countries,
   selectedPlan,
   account,
-  useSavedCC,
   submitting,
-  handleCardToggle,
```
PatrickSison: I don't think we can get rid of this yet because of the usage on the current form right now

src/pages/billing/components/tests/CardSection.test.js
```diff
@@ -1,42 +1,38 @@
...
-  expect(wrapper.find('Connect(PaymentForm)')).toExist();
-});
+  it('should render payment form if no billing info', () => {
+    const wrapper = subject({ account: {}});
+    expect(wrapper.find('CardSummary')).not.toExist();
+    expect(wrapper.find('Connect(PaymentForm)')).toExist();
+  });
 
-
-it('should handle toggle', () => {
```
PatrickSison: We will need this test still because we need the handle toggle for the current form. We may also need it for the newer form as the submission has to be aware of if user is using SavedCC

src/pages/billing/components/CardSection.js
```diff
@@ -1,21 +1,28 @@
...
-  canUpdateBillingInfo
+  canUpdateBillingInfo,
+  isNewChangePlanForm //TODO: remove this when removing the OldChangePlanForm
 }) => {
+  const { isReady, loading } = useFeatureChangeContext();
+  const [useSavedCC, setUseSavedCC] = useState(true);
 
+  const handleCardToggle = () => setUseSavedCC(!useSavedCC);
+
+  if ((!isReady || loading) && isNewChangePlanForm) {
```
PatrickSison: This condition is a bit weird in terms of UX. Same for the submit section, but in the reverse way. 

Just checking isReady make the button render at first (when it's checking conditions) and disappears once it loads.

This condition on the otherhand doesn't do a first render, but then disappears each time returning to the tab. We should probably consolidate the conditions so it's consistent, but not sure which is a better experience

src/pages/billing/components/CardSection.js
shrirupa: I just the made handling the toggling internal to the component, it doesn't effect the working of Current ChangePlanForm. rechecked it.

src/pages/billing/components/tests/CardSection.test.js
shrirupa: checked it. this works.

src/pages/billing/components/CardSection.js
shrirupa: yes, I had the same doubt. I find it better for card section and button to not be there if feature section is loading. How do you think we should proceed ?

src/pages/billing/components/CardSection.js
PatrickSison: It still toggles, but the variable for what state it's in is checked by ChangePlanForm:
https://github.com/SparkPost/2web2ui/blob/master/src/pages/billing/forms/ChangePlanForm.js#L89

src/pages/billing/components/CardSection.js
PatrickSison: Makes sense. I agree with using the loading condition on the card form

src/pages/billing/components/CardSection.js
shrirupa: Missed it! Thanks.

PR #1234, opened 2019-10-23
TR-1850 - Replace use of array sort with Lodash orderBy for consisten…

nicklemmon:
[TR-1850](https://jira.int.messagesystems.com/browse/TR-1850)

### What Changed
- Replaced use of `Array.prototype.sort` with `_.orderBy` for consistency with TableCollection

### How To Test
- Verify that recent activity panels show the same first four results as the table collection

### To Do
- [x] Incorporate feedback

### DIFF
--- a/src/pages/templatesV2/components/RecentActivity.js
+++ b/src/pages/templatesV2/components/RecentActivity.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import _ from 'lodash';
 import { Panel } from '@sparkpost/matchbox';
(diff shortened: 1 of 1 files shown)


nicklemmon:
@PatrickSison This appears to address the sorting issue. To verify, I grabbed the templates list JSON and hardcoded it in to my component locally. The `Array.prototype.sort` reproduced the bug, using `_.orderBy` resolved it.

Let me know if you have questions!

PR #1245, opened 2019-10-29
TR-1899 - Add  style to editor tabs that are in read only mode

nicklemmon:
[TR-1899](https://jira.int.messagesystems.com/browse/TR-1899)

### What Changed
- Add `<Tag/>` component instance for editor tabs when in published mode:

Screen Shot 2019-10-31 at 1 28 29 PM

- Swap vertical kebab with horizontal kebab after chat with UX
- Added conditional rendering to the popover
- Updated editor styles when in read only mode

### How To Test
- Go to `/templatesV2` and click on a template in draft mode
- Click 'Save and Publish' in the top right
- Verify that the Popover is no longer available on the editor tabs
- Verify that the `<Tag/>` renders for the HTML, AMP HTML, and Text tabs but *not* for the Test Data tab

### To Do
- [x] Add supporting tests
- [ ] Incorporate feedback

### DIFF
--- a/src/pages/templatesV2/components/EditSection.js
+++ b/src/pages/templatesV2/components/EditSection.js
@@ -5,19 +5,24 @@ import {
   Button,
   Popover,
   ActionList,
-  ScreenReaderOnly
+  ScreenReaderOnly,
+  Tag
 } from '@sparkpost/matchbox';
-import { MoreVert } from '@sparkpost/matchbox-icons';
+import { MoreHoriz } from '@sparkpost/matchbox-icons';
 import tabs from '../constants/editTabs';
 import InsertSnippetModal from './InsertSnippetModal';
 import useEditorContext from '../hooks/useEditorContext';
 import styles from './EditSection.module.scss';
 
 const EditSection = () => {
-  const { currentTabIndex, setTab } = useEditorContext();
+  const { currentTabIndex, setTab, isPublishedMode } = useEditorContext();
   const [isPopoverOpen, setPopoverOpen] = useState(false);
   const [isSnippetModalOpen, setSnippetModalOpen] = useState(false);
-  const TabComponent = tabs[currentTabIndex].render;
+
+  const currentTab = tabs[currentTabIndex];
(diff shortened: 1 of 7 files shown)


src/pages/templatesV2/components/EditSection.js
```diff
@@ -5,19 +5,24 @@ import {
...
 const EditSection = () => {
-  const { currentTabIndex, setTab } = useEditorContext();
+  const { currentTabIndex, setTab, isPublishedMode } = useEditorContext();
   const [isPopoverOpen, setPopoverOpen] = useState(false);
   const [isSnippetModalOpen, setSnippetModalOpen] = useState(false);
-  const TabComponent = tabs[currentTabIndex].render;
+
+  const currentTab = tabs[currentTabIndex];
+  const TabComponent = currentTab.render;
+  const hasReadOnlyTag = isPublishedMode && (currentTab.key === 'html' || currentTab.key === 'amp_html' || currentTab.key === 'text');
```
bkemper: Is it better to use `key` from the tabs or `currentTabKey` from context?

bkemper: Also, this is the second time we have used this list of tab keys to do something.  They are also used in `PreviewSection`.  Should we pull this list/logic into context, so it can be shared by both?

src/pages/templatesV2/components/EditSection.js
```diff
@@ -36,33 +41,43 @@ const EditSection = () => {
...
-              <MoreVert/>
+        {hasReadOnlyTag &&
+          <div className={styles.TagWrapper}>
+            <Tag color="yellow">
+              <span>Read Only</span>
+            </Tag>
+          </div>
+        }
+
+        {hasPopover &&
```
bkemper: Won't this show for test data tab?

nicklemmon: I wouldn't think so since the test data can be edited in published mode or draft mode

PR #1247, opened 2019-10-30
AC-990: Remove mock endpoints

PatrickSison:
### What Changed
- Remove mock endpoints
- Seperate bundles endpoint from plans endpoint

### How To Test
- Check that new form submission works (with UI feature flag account_feature_limits)
- Check that the old form submission works

### To Do
- [ ] Address feedback
- [ ] Fix testing


src/actions/billing.js
```diff
@@ -5,36 +5,24 @@ import { list as getSendingIps } from './sendingIps';
...
-    }
-  });
-}
 
 /**
  * Updates plan
  * @param {string} code
  */
-export function updateSubscription({ code, promoCode, meta = {}}) {
+export function updateSubscription({ code, bundle = code, promoCode, meta = {}}) {
```
shrirupa: Not sure I understand, why is bundle = code ?

src/pages/billing/context/ChangePlanContext.js
```diff
@@ -0,0 +1,43 @@
...
+  rawPlans: plans,
+  ...value
+}) => {
+  useEffect(() => { getBillingCountries(); }, [getBillingCountries]);
+  useEffect(() => { getBillingInfo(); }, [getBillingInfo]);
+  useEffect(() => { getBundles(); }, [getBundles]);
+  useEffect(() => { getPlans(); }, [getPlans]);
+
+  const plansByKey = useMemo(() => _.keyBy(plans, 'plan'), [plans]);
+  // const availableBundles = useMemo(() => {
```
shrirupa: is this commented code for future use, or can be removed ?

src/actions/billing.js
PatrickSison: This was used since the new endpoint expects bundle, rather than code. Rather than updating tests and the form for now, old form just gets mapped to bundle for now, then it will be removed later on

src/pages/billing/context/ChangePlanContext.js
PatrickSison: It can be removed. Added the memoization somewhere else

PR #1252, opened 2019-11-01
TR-1897 - Add loading state while DuplicateTemplateModal is fetching a template

nicklemmon:
[TR-1897](https://jira.int.messagesystems.com/browse/TR-1897)

### What Changed
- Remove closing modal as part of a Promise chain
- use Redux store to control loading state
- Updated minimum height of loading state to more closely match height of the modal content

### How To Test
- On the `/templatesV2` list page, click on any of the duplicate buttons and verify that a loading state is now present

### To Do
- [ ] Incorporate feedback

### DIFF
--- a/src/components/panelLoading/PanelLoading.js
+++ b/src/components/panelLoading/PanelLoading.js
@@ -3,14 +3,23 @@ import styles from './PanelLoading.module.scss';
 import { Panel } from '@sparkpost/matchbox';
 import { Loading } from 'src/components/loading/Loading';
 
-const PanelLoading = ({ minHeight }) => (
-  <Panel className={styles.Loading} style={{ minHeight }}>
-    <Loading />
-  </Panel>
-);
+const PanelLoading = (props) => {
+  const { minHeight, accent } = props;
+
+  return (
+    <Panel
+      className={styles.Loading}
+      style={{ minHeight }}
+      accent={accent}
+    >
+      <Loading />
+    </Panel>
+  );
+};
 
 PanelLoading.defaultProps = {
-  minHeight: '400px'
+  minHeight: '400px',
+  accent: false
 };
 
 export default PanelLoading;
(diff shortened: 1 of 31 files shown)


src/pages/templatesV2/components/editorActions/DuplicateTemplateModal.js
```diff
@@ -107,7 +102,7 @@ const DuplicateTemplateModal = (props) => {
   if (isLoading) {
     return (
       <ModalWrapper {...modalProps}>
-        <PanelLoading minHeight={'300px'}/>
+        <PanelLoading minHeight={'325px'}/>
```
PatrickSison: If we make the height of the loading panel close the existing duplicate panel, should we add the accent as well?

nicklemmon: That's a great idea.

nicklemmon: Addressed it. Also added the `accent` prop to the `InsertSnippetModal` for consistency.

src/components/panelLoading/PanelLoading.js
```diff
@@ -3,14 +3,23 @@ import styles from './PanelLoading.module.scss';
 import { Panel } from '@sparkpost/matchbox';
 import { Loading } from 'src/components/loading/Loading';
 
-const PanelLoading = ({ minHeight }) => (
-  <Panel className={styles.Loading} style={{ minHeight }}>
-    <Loading />
-  </Panel>
-);
+const PanelLoading = (props) => {
+  const { minHeight, accent } = props;
```
PatrickSison: I think we need to default `accent` to false. The loading panel is used in other places, and I think without the prop, it assumes it's true

nicklemmon: I did default it to `false`, actually! In the `PanelLoading.defaultProps` object.

PatrickSison: Whoops! I was looking at my modification to code without pulling your changes 🤦‍♂

PR #1254, opened 2019-11-01
TR-1901 - When updating the draft from the template settings view, the updated state of the template content should be saved

shrirupa:
TR-1901 - When updating the draft from the template settings view, the updated state of the template content should be saved

### What Changed
 - updated state of template is now being saved when switching tabs and when settings are updated.

### How To Test
 - Make changes to the HTML tab of the editor
 - Swap to "Template Settings"
 - Save those settings
 - Saved Indicator changes to SAVED
 - Return to the HTML tab and verify that the content stays the same

### To Do
- [ ] Address feedback

### DIFF
--- a/src/pages/templatesV2/components/settings/SettingsForm.js
+++ b/src/pages/templatesV2/components/settings/SettingsForm.js
@@ -18,8 +18,11 @@ export default class SettingsForm extends React.Component {
       updateDraftV2,
       parsedTestData,
       subaccountId,
-      showAlert
+      showAlert,
+      content
     } = this.props;
+
+    values.content = content;
 
     return updateDraftV2({
       id: draft.id,
--- a/src/pages/templatesV2/components/settings/SettingsForm.js
+++ b/src/pages/templatesV2/components/settings/SettingsForm.js
@@ -22,12 +22,11 @@ export default class SettingsForm extends React.Component {
       content
     } = this.props;
 
-    values.content = content;
-
     return updateDraftV2({
       id: draft.id,
       parsedTestData,
-      ...values
+      ...values,
+      content
     }, subaccountId)
       .then(() => {
         showAlert({ type: 'success', message: 'Template settings updated.' });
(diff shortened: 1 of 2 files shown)


src/pages/templatesV2/components/settings/SettingsForm.js
```diff
@@ -18,13 +18,15 @@ export default class SettingsForm extends React.Component {
       updateDraftV2,
       parsedTestData,
       subaccountId,
-      showAlert
+      showAlert,
+      content
```
nicklemmon: @shrirupa This does work, but we need to also update the saved indicator - maybe this should be bundled in with `updateDraftV2`...

In the meantime, you can grab the `setHasSaved` function from props and set that to `true` in addition to showing the alert.

Thanks again for taking this on!

shrirupa: I didn't think about this, I will add it.

src/pages/templatesV2/components/settings/SettingsForm.js
```diff
@@ -12,21 +12,28 @@ import styles from './SettingsForm.module.scss';
...
       subaccountId,
-      showAlert
+      showAlert,
+      content,
+      setHasSaved
     } = this.props;
 
+    const { text , html } = content;
+
+    values.content = { ...values.content , text, html };
```
nicklemmon: It took me awhile to figure out what was going on here! But now I get it.

Would it be simpler to do `values.content = { ...content }`? There are actually more fields in that content object than just the text and HTML so I want to capture those too.

Thanks!

src/pages/templatesV2/components/settings/SettingsForm.js
```diff
@@ -12,21 +12,28 @@ import styles from './SettingsForm.module.scss';
...
+
     return updateDraftV2({
       id: draft.id,
       parsedTestData,
       ...values
     }, subaccountId)
       .then(() => {
+        setHasSaved(true);
```
nicklemmon: 👍 Nice!

src/pages/templatesV2/components/settings/SettingsForm.js
nicklemmon: OK we talked this over in person, and sounds like this won't work, but there has to be some way to grab the current state of the content rather than the outdated content when the `SettingsForm` mounts? Let's think about this more before doing it this way - this just seems like it would produce a bug in the medium/long term.

PR #1256, opened 2019-11-04
TR-1946 - Resolve issue with RecentActivity component returning nothing when fewer than 3 templates are present

nicklemmon:
[TR-1946](https://jira.int.messagesystems.com/browse/TR-1946)

### What Changed
- When fewer than three templates are present, `null` is returned instead of nothing
- Updated tests to fail by using a full `mount`, then verified that the fix addressed the issue

### How To Test
- With a new account (or any account with fewer than 3 templates), go to `/templatesV2` and verify that no Recent Activity component renders *and* the app does not render the error boundary UI

### To Do
- [x] Incorporate feedback


nicklemmon:
> LGTM. Looking forward to getting rid of these shallow renders in the future 👍

Yessir! Will take some doing but looking forward to that.

PR #1257, opened 2019-11-04
SA-966 Display test results grouped by region

gfriedmansp:
### What Changed
 - Added Select filter in "Placement Breakdown" section with 2 options: Provider and Region. (/inbox-placement/details/73/details).
 - Upon selection of "Region" option, table data below will change to be grouped by region instead of provider.

### How To Test
 - Open inbox placement page and ensure the list has data with the "Provider" option being the default. 
 - Change the selection to "Region" and ensure the data in the table lists the names of each region in the test in start case form. (i.e. North America as opposed to north america).
 - Change back to "Provider" option and make sure the data changes back.


src/pages/inboxPlacement/components/TestDetails.js
```diff
@@ -9,9 +9,33 @@ import styles from './TestDetails.module.scss';
 import TimeToReceiveSection from './TimeToReceiveSection';
 import AuthenticationResults from './AuthenticationResults';
 
-const TestDetails = ({ details, placementsByProvider }) => {
+export const PLACEMENT_FILTER_TYPES = Object.freeze({
```
kchu1337: We have a constants folder under `src/pages/inboxPlacement/constants`, so lets make a new file and put this constant in there.

src/pages/inboxPlacement/components/TestDetails.js
```diff
@@ -9,9 +9,33 @@ import styles from './TestDetails.module.scss';
...
+  'PROVIDER': 'Provider',
```
kchu1337: Not a big deal but why not use `mailbox-provider` and `region` so we can use these 2 terms everywhere? Also, the mocks have it as `Mailbox Provider` and not `Provider`.

src/pages/inboxPlacement/components/TestDetails.js
```diff
@@ -9,9 +9,33 @@ import styles from './TestDetails.module.scss';
...
+  'REGION': 'Region'
+});
+const PLACEMENTS_BY_OPTIONS = Object.values(PLACEMENT_FILTER_TYPES);
```
kchu1337: The options in the Select can take options in the form of `{label: 'foo', value: 'bar'}`. Lets make an array of this type of object. Then for state, instead of the index, we can directly use the keys of `PLACEMENT_FILTER_TYPES`.
Something like 
```js
const PLACEMENTS_BY_OPTIONS = Object.keys(PLACEMENT_FILTER_TYPES).map((key) => ({ label: PLACEMENT_FILTER_TYPES[key], value: key }));

```

src/pages/inboxPlacement/components/TestDetails.js
```diff
@@ -49,7 +70,10 @@ const TestDetails = ({ details, placementsByProvider }) => {
         </Grid>
       </Panel>
       <Panel title={panelTitle}>
-        <ProvidersBreakdown data={placementsByProvider}/>
+        <div className={styles.PlacementFilter}>
+          <Select options={PLACEMENTS_TYPE_OPTIONS} onChange={onFilterChange} />
```
kchu1337: We should also add a test to simulate changing the selector.

src/selectors/inboxPlacement.js
```diff
@@ -5,3 +5,13 @@ export const getSeeds = (state) => state.inboxPlacement.seeds;
 
 export const selectReferenceSeed = createSelector(
   [getSeeds], (seeds) => seeds.find((seed) => endsWith(seed, '@seed.sparkpost.com')));
+
+const getInboxPlacement = (state) => state.inboxPlacement;
+
+export const selectTestDetailsPageError = createSelector(
```
kchu1337: We also have tests @ `src/selectors/tests/inboxPlacement.test.js`

PR #1258, opened 2019-11-05
TR-1948 - Resolve bug with updating an existing template from the draft state to the published state

nicklemmon:
[TR-1948](https://jira.int.messagesystems.com/browse/TR-1948)

### What Changed
- Updated Matchbox version to get a11y updates
- Updated `publishV2` to use `updateV2` instead of `update`
- Renamed some variables for clarity's sake
- Updated markup for rendering status on the list page
- Removed unnecessary spreading of data on the `SaveAndPublishConfirmationModal`
- Updated templates selector to accurately reflect the template's current status

### How To Test
1. Go to `/templatesV2`
1. Create a new template
1. Go back to the list page and verify that the template is marked as a 'Draft'
1. Open that template and publish it
1. Go back to the list page and verify that the template is marked as 'Published'
1. Open the template and click 'Edit Draft'
1. Make a change to the draft and click 'Save and Publish'
1. Verify that the changes that you made were preserved
1. Click 'Edit Draft' and make any change to the template
1. Using the menu in the top right, click 'Save Draft'
1. Return to the list page
1. Verify that there are now two templates in the list, one with 'Draft' and one with 'Published' - the published draft should indicate that there are unpublished changes in the draft.
1. Take a break because that was a lot of work 😩

### To Do
- [ ] Incorporate feedback

### DIFF
--- a/src/pages/templatesV2/EditAndPreviewPage.container.js
+++ b/src/pages/templatesV2/EditAndPreviewPage.container.js
@@ -37,15 +37,16 @@ const mapStateToProps = (state, props) => {
   const draft = selectDraftTemplateById(state, id);
   const published = selectPublishedTemplateById(state, id);
   const isPublishedMode = props.match.params.version === 'published';
-  const draftOrPublished = draft || published;
-  const hasDraft = draftOrPublished && draftOrPublished.has_draft;
-  const hasPublished = draftOrPublished && draftOrPublished.has_published;
+  const template = draft || published;
+  const hasDraft = template && template.has_draft;
+  const hasPublished = template && template.has_published;
   const snippets = state.snippets.items;
 
   return {
     draft,
-    snippets,
     published,
+    template,
+    snippets,
     isPublishedMode,
     hasDraft,
     hasPublished,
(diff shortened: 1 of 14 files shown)


src/selectors/templates.js
```diff
@@ -103,14 +103,24 @@ export const selectTemplatesForListTable = createSelector(
   [selectTemplates], (templates) => {
     const templatesForListing = [];
     templates.forEach((template) => {
+      const isPublished = template.published;
```
PatrickSison: Will draft templates have the property `published` as true if it's the same as the published? trying to digest how the API works

src/pages/templatesV2/EditAndPreviewPage.container.js
```diff
@@ -37,15 +37,16 @@ const mapStateToProps = (state, props) => {
...
+  const template = draft || published;
+  const hasDraft = template && template.has_draft;
+  const hasPublished = template && template.has_published;
   const snippets = state.snippets.items;
 
   return {
     draft,
-    snippets,
     published,
+    template,
```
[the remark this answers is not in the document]
nicklemmon: yeeeah this is a weird thing for me as well - it made developing this feature a lot more confusing.

src/selectors/templates.js
nicklemmon: it looks like, no - if you are dealing with a draft `published` is `false`. This is where the weirdness of the API design causes problems - ideally these fields like `has_published` and `has_draft` would be totally unnecessary if the the draft/published transition happened in a more logical manner.

https://developers.sparkpost.com/api/templates/#header-template-object

PR #1262, opened 2019-11-06
SA-1001: Filter Inbox Placement Tests by Text Field

gfriedmansp:
### What Changed
 - Added a search bar to the top of the inbox placement details page
 - You can now filter the list by Mailbox Providers by case insensitive text matching of region name or mailbox provider name
 - You can now filter the list by Region by case insensitive text matching of region name

### How To Test
 - Use /inbox-placement/details/74/details on 108 account
 - Search "america" on the mailbox provider's list and you should see at&t and comcast
 - Switch to filtering by region and you should see North America

### DIFF
--- a/src/pages/inboxPlacement/components/TestDetails.js
+++ b/src/pages/inboxPlacement/components/TestDetails.js
@@ -18,9 +18,9 @@ const TestDetails = ({ details, placementsByProvider, placementsByRegion }) => {
   const [breakdownType, setBreakdownType] = useState(PLACEMENTS_TYPE_OPTIONS[0].value);
   const [searchPlacements, setSearchPlacements] = useState('');
 
-  const onSearchChange = useCallback((e) => {
-    setSearchPlacements(e.target.value);
-  }, []);
+  const debouncedSetSearchPlacements = useCallback(_.debounce((placements) => setSearchPlacements(placements), 300), []);
+
+  const onSearchChange = useCallback((e) => debouncedSetSearchPlacements(e.target.value), [debouncedSetSearchPlacements]);
 
   const onFilterChange = useCallback((e) => {
     setBreakdownType(e.target.value);
--- a/src/pages/inboxPlacement/components/TestDetails.js
+++ b/src/pages/inboxPlacement/components/TestDetails.js
@@ -35,11 +35,11 @@ const TestDetails = ({ details, placementsByProvider, placementsByRegion }) => {
   switch (breakdownType) {
     case PLACEMENT_FILTER_TYPES.REGION:
       breakdownData = placementsByRegion;
-      textFieldPlaceholder = 'Region';
+      textFieldPlaceholder = PLACEMENT_FILTER_LABELS[breakdownType];
       break;
     default:
       breakdownData = placementsByProvider;
-      textFieldPlaceholder = 'Mailbox Provider or Region';
+      textFieldPlaceholder = `${PLACEMENT_FILTER_LABELS[PLACEMENT_FILTER_TYPES.MAILBOX_PROVIDER]} or ${PLACEMENT_FILTER_LABELS[PLACEMENT_FILTER_TYPES.REGION]}`;
       break;
   }
 
--- a/src/pages/inboxPlacement/components/TestDetails.js
+++ b/src/pages/inboxPlacement/components/TestDetails.js
@@ -1,6 +1,6 @@
 import React, { useState, useCallback } from 'react';
 import { format } from 'date-fns';
-import { Grid, Panel, Select, TextField } from '@sparkpost/matchbox';
+import { Grid, Panel, Select, TextField, Label, ScreenReaderOnly } from '@sparkpost/matchbox';
 import _ from 'lodash';
 import FolderPlacementBarChart from './FolderPlacementBarChart';
 import PlacementBreakdown from './PlacementBreakdown';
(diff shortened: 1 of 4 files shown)


nicklemmon:
Hey, @gfriedmansp could you provide the route to the particular placement test you're referring to in your PR?

src/pages/inboxPlacement/components/TestDetails.js
```diff
@@ -68,8 +90,13 @@ const TestDetails = ({ details, placementsByProvider, placementsByRegion }) => {
...
       </Panel>
       <Panel title={panelTitle}>
-        <div className={styles.PlacementFilter}>
-          <Select options={PLACEMENTS_TYPE_OPTIONS} onChange={onFilterChange} />
+        <div className={styles.PlacementFilterContainer}>
+          <div className={styles.PlacementFilterSelect}>
+            <Select options={PLACEMENTS_TYPE_OPTIONS} onChange={onFilterChange} />
+          </div>
+          <div className={styles.PlacementFilterTextField}>
+            <TextField suffix={<Search />} onChange={onSearchChange} placeholder={textFieldPlaceholder} />
```
nicklemmon: So every form control needs a label in order for it to be valid (and screen reader accessible). In this case, you can import the `<Label>` from Matchbox and then make sure that both the `<Select/>` and `<TextField/>` have matching `id` props passed in. This associates the two via the HTML `for` attribute.

Now the mocks probably don't want these labels to be visible, which is totally fine. In that case, wrap each `<Label/>` in the Matchbox `<ScreenReaderOnly/>` component and you'll be good to go.

src/pages/inboxPlacement/components/TestDetails.module.scss
```diff
@@ -22,7 +22,16 @@
...
+  display: flex;
+}
+
+.PlacementFilterSelect {
   max-width: 200px;
 }
+
+.PlacementFilterTextField {
+  flex: 1;
+  margin-left: spacing(base);
```
nicklemmon: Screen Shot 2019-11-07 at 9 48 05 AM

It would probably be good to make both of these full width on smaller screen sizes. You could either use the Matchbox `<Grid/>` component to accomplish this or you could use CSS grid if you would like. Either way.

The table is obviously super janky, but looks like that has been an issue before this ticket so no need to address now.

src/pages/inboxPlacement/components/tests/TestDetails.test.js
```diff
@@ -81,4 +82,98 @@ describe('Component: TestDetails', () => {
...
+    expect(wrapper.find('PlacementBreakdown').prop('data')).toEqual([dogmail, gmail]);
+
+    // No results
+    wrapper.find('TextField').simulate('change', { target: { value: 'this wont match anything' }});
+    expect(wrapper.find('PlacementBreakdown').prop('data')).toEqual([]);
+
+    // Search by case insensitive region match
+    wrapper.find('TextField').simulate('change', { target: { value: 'america' }});
+    expect(wrapper.find('PlacementBreakdown').prop('data')).toEqual([dogmail]);
+  });
```
nicklemmon: We have just started to use `data-id` attributes to make things easier to select in tests (and less brittle) as well as for use with Pendo for analytics stuff.

Would you mind adding one to the `TextField` and `Select` and targeting those attributes instead of the elements themselves? Should make your tests a little more future proof.

Your tests are very easy to read. Well done! 👍

gfriedmansp:
> Hey, @gfriedmansp could you provide the route to the particular placement test you're referring to in your PR?

/inbox-placement/details/74/details

src/pages/inboxPlacement/components/TestDetails.js
```diff
@@ -22,15 +29,30 @@ const TestDetails = ({ details, placementsByProvider, placementsByRegion }) => {
...
+    const searchRegex = new RegExp(_.escapeRegExp(searchPlacements), 'i');
+    switch (breakdownType) {
+      case PLACEMENT_FILTER_TYPES.REGION:
+        breakdownData = breakdownData.filter((item) => item.region.match(searchRegex));
+        break;
+      default:
+        breakdownData = breakdownData.filter((item) => item.region.match(searchRegex) || item.mailbox_provider.match(searchRegex));
+        break;
+    }
+  }
```
nicklemmon: I'm a little confused by the use of `switch` in these two cases when there are only two conditions?

I'm assuming you used a `switch` since another one was already present for the filtering, but just wanted to see what you thought. This is obviously a subjective point since this approach should be the same either way functionally.

gfriedmansp: > I'm a little confused by the use of `switch` in these two cases when there are only two conditions?
> 
> I'm assuming you used a `switch` since another one was already present for the filtering, but just wanted to see what you thought. This is obviously a subjective point since this approach should be the same either way functionally.

This was simply because I _know_ filtering by IP address is coming soon and it felt like it made  sense for the person who adds that to not need to refactor

nicklemmon: And I did not know that! So makes sense.

nicklemmon:
> > Hey, @gfriedmansp could you provide the route to the particular placement test you're referring to in your PR?
> 
> /inbox-placement/details/74/details

Hmm, says 'Resource could not be found'

gfriedmansp:
> > > Hey, @gfriedmansp could you provide the route to the particular placement test you're referring to in your PR?
> > 
> > 
> > /inbox-placement/details/74/details
> 
> Hmm, says 'Resource could not be found'

Oh sorry, I updated the description above, but not my post: It's on the 108 prod account

src/pages/inboxPlacement/components/TestDetails.js
```diff
@@ -68,10 +90,21 @@ const TestDetails = ({ details, placementsByProvider, placementsByRegion }) => {
         </Grid>
       </Panel>
       <Panel title={panelTitle}>
-        <div className={styles.PlacementFilter}>
-          <Select options={PLACEMENTS_TYPE_OPTIONS} onChange={onFilterChange} />
+        <div className={styles.PlacementFilterContainer}>
+          <div className={styles.PlacementFilterSelect}>
+            <ScreenReaderOnly>
+              <Label id="select-placement-type">Placement Type</Label>
```
nicklemmon: Nice! So this isn't 100% working because there is a bug with the Matchbox `<Select/>` component. This will be rolling out with the next Matchbox update and then this will work.

PR #1267, opened 2019-11-07
FE-737: Mobile Nav Closes on Click of NavItem

gfriedmansp:
### What Changed
 - The mobile navigation will now properly close when you click on a nav item

### How To Test
 - Shrink window down to mobile size
 - Open the navigation
 - Toggle a NavGroup
   - Ensure this does not close the navigation
 - Click on an individual NavItem
   - Ensure this properly closes the navigation


src/components/navigation/Navigation.js
```diff
@@ -18,17 +18,17 @@ export class Navigation extends Component {
...
     return navItems.map((item, key) => {
-      const props = { ...item, toggleMobileNav: this.toggleMobileNav, location, key: key };
+      const props = { ...item, toggleMobileNav: this.toggleMobileNav, location, key: key, mobile };
       return item.children ? <NavGroup {...props} /> : <NavItem {...props} />;
     });
   }
 
   toggleMobileNav = () => {
-    this.setState({ open: !this.state.open });
+    this.setState(({ open: isOpen }) => ({ open: !isOpen }));
```
gfriedmansp: We should always use the functional setState when the next state depends on the current state.

src/components/navigation/Navigation.js
```diff
@@ -18,17 +18,17 @@ export class Navigation extends Component {
     open: false
   };
 
-  renderItems() {
+  renderItems({ mobile }) {
     const { navItems, location } = this.props;
 
     return navItems.map((item, key) => {
-      const props = { ...item, toggleMobileNav: this.toggleMobileNav, location, key: key };
+      const props = { ...item, toggleMobileNav: this.toggleMobileNav, location, key: key, mobile };
```
nicklemmon: Haven't dug in to this code before - I'm not a big fan of using JS to replace CSS media queries, which I would _think_ perform better? Maybe with the redesign we could address this.

src/components/navigation/Navigation.js
nicklemmon: Nice!

nicklemmon: This may need to be handled elsewhere...when clicking on the accounts dropdown and navigating there as well as while clicking on the "SparkPost" logo, the mobile nav stays open. Screenshots for what I'm referring to:

Screen Shot 2019-11-08 at 9 54 24 AM
Screen Shot 2019-11-08 at 9 54 27 AM

src/components/navigation/tests/__snapshots__/Navigation.test.js.snap
```diff
@@ -21,16 +21,19 @@ exports[`Navigation should render on desktop 1`] = `
           <NavItem
             key="0"
             location="/foo/bar"
+            mobile={false}
```
nicklemmon: Is there another way we can write a test for this?

src/components/navigation/Navigation.js
gfriedmansp: Yeah, it feels pretty bad when you reduce the browser window that the nav "slides" away. The redesign is probably the best place to handle/refactor that.

src/components/navigation/Navigation.js
gfriedmansp: Instead of handling it elsewhere, it seems like we should include the top bar as the "overlay", which would otherwise close the left nav. Doesn't make sense that some places would be interactive and others aren't with the nav open.

Want me to handle this here or in a new ticket?

src/components/navigation/tests/__snapshots__/Navigation.test.js.snap
gfriedmansp: I briefly looked into writing up a test, but it felt like something that would be better handled as a browser test as opposed to anything snapshott-y. This was just an updated snap for another test.

nicklemmon: Nice! So I would say go ahead and try to write a Cypress test for this. It isn't hooked up to the pipeline, but it will be shortly.

src/components/navigation/Navigation.js
nicklemmon: Hmm, I'm not sure I agree! Opening the account menu shouldn't necessarily close the nav?

Would there be a way to use some middleware that on a route change the navigation state is set to `false`? That would handle any scenario instead of specific events.

cypress/tests/navigation/navigation.spec.js
```diff
@@ -0,0 +1,57 @@
...
+  });
+
+  it('navigation closes after opening admin menu', () => {
+    openNavigation();
+    openAccountMenu();
+    assertNavigationIsNotVisible();
+    assertAccountMenuIsVisible();
+  });
+
+  it('admin menu closes after opening navigation', () => {
```
nicklemmon: I think the assertion verbiage is a little confusing here - this is kind of nitpicky admittedly - probably just removing the word `'navigation'` would do it?

```
'Mobile navigation closes after clicking child menu item'
```

instead of:

```
Mobile navigation navigation stays open after clicking menu item with child
```

cypress/tests/navigation/navigation.spec.js
```diff
@@ -0,0 +1,57 @@
+/// <reference types="Cypress" />
+
+describe('Mobile Navigation', () => {
+  beforeEach(() => {
+    cy.viewport(500, 1000);
+  });
+
+  const accountDropdownSelector = '[data-id="account-dropdown-link"]';
+  const navigationButtonSelector = '[data-id="top-navigation-link"]';
```
nicklemmon: I think `-button` instead of `-link` might be a better name here - even if it's code as an `<a>` it should be coded as a `<button>` and should be in the future

PR #1269, opened 2019-11-08
FE-754 Do Not Trigger Keyboard on iOS for Select a Plan Dropdown

gfriedmansp:
### What Changed
 - The keyboard in iOS does not open when you use the "custom select" when selecting a plan on the /onboarding/plan page.

### How To Test
 - Use the device simulator from xcode and navigate to the /onboarding/plan page. (Turns out you can go directly there without making a new account)
 - Press the "Select a Plan" dropdown and see that the keyboard does not trigger itself open (Note: You need to have the simulator set up to not use the hardware keyboard for the on-screen keyboard to pop up to begin with).
 - Ensure that choosing an option besides the default indeed changes the value


gfriedmansp:
If anyone has an idea of how to write a test for this (iOS keyboard triggering), let me know 👍

PR #1270, opened 2019-11-08
FE-726 - Add Content to Icon Buttons and Make Said Buttons Keyboard Accessible

nicklemmon:
[FE-726](https://jira.int.messagesystems.com/browse/FE-726)

### What Changed
- Added screen reader only content to many icon buttons in the navigation
- Updated markup to use `<button>` elements instead of anchor tags
- Added new `button-reset` Sass mixin to reset button styles
- Updating spacing between elements for consistency
- Updated failing tests
- Added `data-id` attributes to interactive items in the top navigation
- Added new `<IconButton/>` component with documentation and supporting unit tests
- Upgraded Storybook to version `5.2.6`
- Added documentation for many 2web2ui components:
  * `ActionPopover`
  * `Callout`
  * `LabelledValue`
  * `Loading`
  * `Note`
  * `SparkPostLogo`
  * `ButtonWrapper`
  * `IconButton`
  * `CopyField`
  * `DatePicker`
  * `DateSelector`
  * `MultiEmailField`
  * `DownloadLink`
  * `ExternalLink`
  * `PageLink`
  * `Empty`
  * `Loading`
  * `FocusContainer`
  * ...and an introduction page
- Added new build NPM script to eventually allow deployment

### How To Test
- Using a screen reader (or not!), verify that all buttons in the navigation have content
   - With a screen reader, you can use Voiceover, focus on the elements to verify -> https://www.apple.com/voiceover/info/guide/_1124.html
   - Without a screen reader, inspect the DOM to verify that all buttons have content even if it isn't visible
- Verify that all buttons within the navigation are interactive using the keyboard only - **hint:** use tab and shift+tab
- Run `npm install && npm run storybook-dev` to view the updated components and updated StoryBook documentation site

### To Do
- [ ] Incorporate feedback


nicklemmon:
> Instead of custom style with a scss mixin and using `<ScreenReaderOnly />` for every button, can this be consolidated in a reusable component that requires onClick, data-id, and screen reader text as props and resets button style?

That's a great idea! I do think the button reset is worth having, *or* we could just do a global `<button>` CSS reset so that people are encouraged to use the right markup without additional work. Global changes are a tad scary of course...

src/components/iconButton/IconButton.js
```diff
@@ -0,0 +1,41 @@
...
+    </button>
+  );
+};
+
+IconButton.propTypes = {
+  onClick: PropTypes.func.isRequired,
+  screenReaderLabel: PropTypes.string.isRequired,
+  className: PropTypes.string,
+  title: PropTypes.string,
+  ['data-id']: PropTypes.string
```
bkemper: Are the brackets needed?

nicklemmon: Looks like no! 👍

src/components/navigation/components/AccountDropdown.js
```diff
@@ -20,9 +20,11 @@ export class AccountDropdown extends Component {
   renderActivator = () => (
     <WindowSizeContext.Consumer>
       {({ mobile }) => (
-        <UnstyledLink className={styles.Email} onClick={this.toggleDropdown}>
-          {(mobile || !this.props.email) ? <Person size={24}/> : <Fragment>{this.props.email} <ArrowDropDown/></Fragment>}
-        </UnstyledLink>
+        <button className={styles.Email} onClick={this.toggleDropdown} data-id="nav-button-accounts">
```
bkemper: Can we use the new `<IconButton />` here?

src/components/navigation/components/AccountDropdown.module.scss
```diff
@@ -1,17 +1,24 @@
 @import '~@sparkpost/matchbox/src/styles/config.scss';
+@import 'src/styles/resets';
 
 .Email {
+  @include button-reset();
```
bkemper: If so, I don't think we will need much of this scss.

src/components/navigation/components/Top.js
```diff
@@ -14,10 +14,21 @@ import { DEFAULT_REDIRECT_ROUTE } from 'src/constants';
 export class Top extends Component {
   renderMobile = () => (
     <div className={styles.Top}>
-      <UnstyledLink onClick={this.props.toggleMobileNav} className={styles.Menu}>
+      <button
```
bkemper: Can we use the new `<IconButton />` component here?

src/components/navigation/components/AccountDropdown.js
nicklemmon: Hey, we can't use it here because this button has visible content inside of it (the user's email address). If it were just an icon we could use it, but in this case this is a one-off component used only in the navigation.

nicklemmon:
@bkemper Ready for another pass.

src/components/navigation/components/AccountDropdown.js
bkemper: Then do we need the ScreenReaderOnly?

PR #1273, opened 2019-11-12
SA-1002: Add inbox placement messages filtered by region

kchu1337:
### What Changed
 - Added all messages filtered by region

### How To Test
 - To to /inbox-placement and click on any tests to access the details page
 - Switch dropdown to region. 
 - Click on any of the regions.
 - The all messages page should load. Verify that filtering is correct by checking that the number sent matches the number of messages in the table.

### DIFF
--- a/src/pages/inboxPlacement/AllMessagesPage.js
+++ b/src/pages/inboxPlacement/AllMessagesPage.js
@@ -1,6 +1,7 @@
 import React, { useEffect, useCallback } from 'react';
 import { connect } from 'react-redux';
 import { Grid, Page, Panel } from '@sparkpost/matchbox';
+import _ from 'lodash';
 
 import { Loading } from 'src/components';
 import { getInboxPlacementByProviders, getInboxPlacementByRegions, getAllInboxPlacementMessages, resetState } from 'src/actions/inboxPlacement';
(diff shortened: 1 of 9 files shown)


src/pages/inboxPlacement/AllMessagesPage.js
```diff
@@ -106,10 +124,17 @@ export const AllMessagesPage = ({
 
 function mapStateToProps(state, props) {
   const { id, filterType, filterName } = props.match.params;
-  const result = ((filterType === 'mailbox-provider')
-    ? state.inboxPlacement.placementsByProvider.find(({ mailbox_provider }) => mailbox_provider === filterName)
-    : undefined) || //TODO add sending ip
-    {};
+  //Runs an inline function that returns a value from a switch statement
```
gfriedmansp: Think this would be best as a selector so we don't have to define a function and re-calculate that result every time our state/props change

src/pages/inboxPlacement/components/PlacementBreakdown.js
```diff
@@ -34,9 +34,15 @@ const WrapperComponent = ({ children }) => (<div>
 const getPlacementNameByType = (type, { id, mailbox_provider, region }) => {
   switch (type) {
     case PLACEMENT_FILTER_TYPES.REGION:
-      return <strong>{_.startCase(region)}</strong>;
+      return (
+        <PageLink to={`/inbox-placement/details/${id}/${PLACEMENT_FILTER_TYPES.REGION}/${region}`}>
+          <strong>{_.startCase(region)}</strong>
```
gfriedmansp: This is probably my fault, but it might be a good idea to pull this formatting functionality into a util of some sort. Maybe a `formatRegionForDisplay` or something - similar to how we have the formatter for provider names.

PR #1274, opened 2019-11-12
SE-269: UX improvements to message events table

PatrickSison:
### What Changed
 - New Column (subject), truncated with ellipses
- Order of columns rearranged, Event, subject, Recipient, from, timestamp, view details
- Timestamp shows full date after 60 seconds

### How To Test
 - Go to `/reports/message-events`
 -Check that order of columns is correct
- Check that older timestamps show the full timestamp
 - Check that subject truncates if too long.

### To Do
- [ ] Address feedback

### Notes
 - The timestamp doesn't transition to full timestamp after 1 minute since TimeAgo watches for new time, but we currently don't


src/pages/reports/messageEvents/MessageEventsPage.module.scss
```diff
@@ -13,3 +13,10 @@
   margin-left: 5px;
   margin-bottom: 3px;
 }
+
+.MessageSubject {
+  text-overflow: ellipsis;
+  overflow: hidden;
+  white-space: nowrap;
+  max-width: 200px;
```
kchu1337: The column width seems to stay the same size after resizing the window width to be lower. I found `width: 15vw` to work better.

PR #1275, opened 2019-11-13
AC-1080 - Sending module guide set

shrirupa:
AC-1080 - Sending module guide set

### What Changed
- Includes the options for "Show Me SparkPost" guide set and the "Let's Code" guide set
- reduced the margin for CardContent.
- BreadCrumbs are not the part of this ticket.

### How To Test
- Navigate to /dashboard.
- check if the second step of guide is similar to [mocks](https://sparkpost.invisionapp.com/share/DHULLYL45GB#/screens/390961094)

### To Do
- [ ] Address Feedback

### DIFF
--- a/src/components/card/Card.module.scss
+++ b/src/components/card/Card.module.scss
@@ -1,32 +1,32 @@
 @import '~@sparkpost/matchbox/src/styles/config.scss';
 
-
 .CardContainer{
-    border-radius: 5px;
-    border: 1px solid color(gray,5);
-    padding: 15px;
+  border-radius: 5px;
+  border: 1px solid color(gray,5);
+  padding: 15px;
+  margin-top: spacing('small')
 }
 .CardTitle{
-    display: block;
-    color: color(gray,3);
-    font-weight: 600;
-    font-size: font-size(500);
+  display: block;
+  color: color(gray,3);
+  font-weight: 600;
+  font-size: font-size(500);
 }
 .CardContent{
-    display: inline-block;
-    margin-top: spacing('small')*2;
-    margin-bottom: spacing('small');
-    font-size: font-size(500);
+  display: inline-block;
+  margin-top: spacing('small')*2;
+  margin-bottom: spacing('small');
+  font-size: font-size(500);
 }
 
 .right {
-    text-align: right;
+  text-align: right;
 }
 
 .center {
-    text-align: center;
+  text-align: center;
 }
 
 .left {
-    text-align: left;
+  text-align: left;
 }
--- a/src/components/card/Card.module.scss
+++ b/src/components/card/Card.module.scss
@@ -15,7 +15,7 @@
 .CardContent{
     display: inline-block;
     margin-top: spacing('small')*2;
-    margin-bottom: spacing('larger');
+    margin-bottom: spacing('small');
     font-size: font-size(500);
 }
 
--- a/src/components/card/Card.module.scss
+++ b/src/components/card/Card.module.scss
@@ -1,7 +1,6 @@
 @import '~@sparkpost/matchbox/src/styles/config.scss';
 
 
-
 .CardContainer{
     border-radius: 5px;
     border: 1px solid color(gray,5);
--- a/src/pages/dashboard/components/GettingStartedGuide.js
+++ b/src/pages/dashboard/components/GettingStartedGuide.js
@@ -16,7 +16,6 @@ export const GettingStartedGuide = ({ isGuideAtBottom, moveGuideAtBottom }) => {
   const renderStep = () => {
     switch (stepName) {
       case 'Features':
-      default:
         return <Grid>
           <Grid.Column xs={12}>
             <Card>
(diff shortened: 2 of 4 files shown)


src/pages/dashboard/components/GettingStartedGuide.js
```diff
@@ -30,6 +29,36 @@ export const GettingStartedGuide = ({ isGuideAtBottom, moveGuideAtBottom }) => {
             </Card>
           </Grid.Column>
         </Grid>;
+      case 'Sending':
+        return <>
+        <h2 className={styles.SendingStepHeading}>Where Would You Like to Begin?</h2>
```
nicklemmon: So technically "Features" should probably be a heading, and looking at the page hierarchy, features should be an `<h3>` and 'Where Would You Like to Begin?' should be an `<h4>`:

Screen Shot 2019-11-14 at 10 29 32 AM

You can always use `role="heading"` and `aria-level` if you don't want the styling that comes with the `<h>` tags

src/components/card/Card.module.scss
```diff
@@ -1,7 +1,6 @@
 @import '~@sparkpost/matchbox/src/styles/config.scss';
 
 
-
 .CardContainer{
```
nicklemmon: The spacing in this file looks weird? Is your editor set up to use two spaces instead of 4 for tabs?

src/pages/dashboard/components/GettingStartedGuide.js
```diff
@@ -16,7 +16,6 @@ export const GettingStartedGuide = ({ isGuideAtBottom, moveGuideAtBottom }) => {
   const renderStep = () => {
```
nicklemmon: When "Start Sending" is clicked, keyboard focus needs to be manually moved to the container element:

Screen Shot 2019-11-14 at 10 37 17 AM

You can accomplish this with a combination of adding `tabIndex="-1"` to that `<div>` and using refs to focus the element. With hooks, this can be a little tricky, but I believe the right one to use is `useLayoutEffect`?

This article looks pretty good:
https://dev.to/westbrookc16/managing-focus-in-react-3n13

src/components/card/Card.module.scss
```diff
@@ -16,7 +15,7 @@
 .CardContent{
     display: inline-block;
     margin-top: spacing('small')*2;
-    margin-bottom: spacing('larger');
+    margin-bottom: spacing('small');
```
nicklemmon: 👍 NICE!

src/pages/dashboard/components/tests/GettingStartedGuide.test.js
```diff
@@ -3,12 +3,26 @@ import { shallow } from 'enzyme';
...
+    expect(subject({ isGuideAtBottom: false }).find('Panel').prop('actions')).toEqual(expect.arrayContaining([{ 'color': 'blue', 'content': <span> Move to Bottom <ArrowDownward size="20" /> </span>, 'onClick': defaultProps.moveGuideAtBottom }]));
   });
-});
-
 
+  it('should render Sending Step when Start Sending Button is clicked', () => {
+    const instance = subject();
+    instance.find('Button').simulate('click');
+    expect(instance.find('Card')).toHaveLength(2);
+    instance.find('Button').forEach((button) => {
```
nicklemmon: If you want to simplify this, you could use a new matchers we added:

```js
expect(instance.toHaveTextContent('Show Me SparkPost')).toBeTrue();
expect(instance.toHaveTextContent('Let\'s Code')).toBeTrue();
```

src/pages/dashboard/components/GettingStartedGuide.js
```diff
@@ -30,6 +29,36 @@ export const GettingStartedGuide = ({ isGuideAtBottom, moveGuideAtBottom }) => {
             </Card>
           </Grid.Column>
         </Grid>;
+      case 'Sending':
+        return <>
+        <h2 className={styles.SendingStepHeading}>Where Would You Like to Begin?</h2>
+        <Grid>
```
nicklemmon: Screen Shot 2019-11-14 at 10 57 16 AM

My guess is there should be a gutter between these two?

src/components/card/Card.module.scss
shrirupa: my editor is setup for 4 spaces for tab. not sure how it looks weird.

nicklemmon: Oh, I meant that it should be set for 2 instead of 4! All of the other files look like they use 2 spaces.

src/pages/dashboard/components/GettingStartedGuide.js
shrirupa: I am not sure that if this correct implementation in response to your feedback, but what I changed is when you now click on Start Sending then the focus moves to 'Show Me SparkPost' card.

nicklemmon: 👍 Let me check it out!

nicklemmon: It doesn't appear to be working...

You know what, it looks like you could leverage this component to do the work for you:
https://github.com/SparkPost/2web2ui/blob/ded78fdd7868e55616c2e6d4496123cac0de9390/src/components/focusContainer/FocusContainer.js

src/pages/dashboard/components/tests/GettingStartedGuide.test.js
shrirupa: That's nice! Thanks.

PR #1276, opened 2019-11-18
SE-379: List incidents on alert details page

PatrickSison:
### What Changed
 - Added alert incidents table
 - Added alert incidents action/reducer

### How To Test
 - Using appteam account (prod works nicely), go to `/alerts`
 - Go to one of the alerts page
 - Check that table is at bottom
    - Check empty state
    - Check that table properly renders active tag

### To Do
- [ ] Address feedback
- [x] Confirm AC with`last_fired + 45 minutes`
- [ ] Error state with not being able to access alert details with too many filters from incidents API

### DIFF
--- a/src/pages/alerts/components/AlertIncidents.js
+++ b/src/pages/alerts/components/AlertIncidents.js
@@ -1,0 +1,95 @@
+import React from 'react';
+import { Panel, Table, Tag } from '@sparkpost/matchbox';
+import { TableCollection, Empty } from 'src/components';
+import { METRICS, FILTERS_FRIENDLY_NAMES } from '../constants/formConstants';
+import { getEvaluatorOptions } from '../helpers/alertForm';
+import { roundToPlaces } from 'src/helpers/units';
+import moment from 'moment';
+import _ from 'lodash';
+import styles from './AlertIncidents.module.scss';
+
+const formatDateTime = (timestamp) => moment(timestamp).format('MMM DD, YYYY [at] h:mma');
+
+const AlertIncidents = ({ incidents = [], alert, subaccountIdToString }) => {
+
+  const { metric, source } = alert;
+  const { suffix } = getEvaluatorOptions(metric, source);
+
+  const columns = [
+    { label: 'Triggered', sortKey: 'first_fired' },
+    { label: 'Resolved', sortKey: 'last_fired' },
+    { label: 'For' },
+    { label: METRICS[metric], sortKey: 'triggered_value', width: '20%' }
+  ];
+
+  const getRowData = (props) => {
+    const { first_fired, last_fired, status, triggered_value, filters } = props;
+
+
+    const renderTags = () => {
+      const { subaccount_id, ...rest } = filters;
+
+      const subaccountTag = subaccount_id &&
+      <span key='subaccounts'>
+        <h6 className={styles.BoldInline}>Subaccounts</h6>
+        &nbsp;
+        <Tag>{subaccountIdToString(subaccount_id)}</Tag>
+      </span>;
+
+      const otherTags = _.map(rest, (value, key) => {
+        const finalValue = value === 'NULL' ? null : value;
+        return (
+          finalValue &&
+          <span key={key}>
+            <h6 className={styles.BoldInline}>{FILTERS_FRIENDLY_NAMES[key]}</h6>
+            &nbsp;
+            <Tag>{finalValue}</Tag>
+          </span>
+        );
+      });
+
+      return (
+        [
+          subaccountTag,
+          otherTags.length > 0 ? ' and ' : null,
+          ...otherTags
+        ]
+      );
+    };
+
+    return [
+      <div>{formatDateTime(first_fired)}</div>,
+      <div>{status === 'Active' ? <Tag color='yellow'>Active</Tag> : formatDateTime(last_fired)}</div>,
+      <div>{renderTags()}</div>,
+      <div className={styles.paddedCell}>{roundToPlaces(triggered_value, 3)}{suffix}</div>
+    ];
+  };
+
+  const TableWrapper = (props) => (
+    <>
+      <div>
+        <Table>{props.children}</Table>
+      </div>
+    </>
+  );
+
+  return (
+    <Panel title='Alert Incidents'>
+      {incidents.length <= 0
+        ? <Empty message='No incidents'/>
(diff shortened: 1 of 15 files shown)


src/pages/alerts/components/AlertIncidents.js
```diff
@@ -0,0 +1,103 @@
...
+import { TableCollection, Empty } from 'src/components';
+import { METRICS, FILTERS_FRIENDLY_NAMES } from '../constants/formConstants';
+import { MAILBOX_PROVIDERS } from 'src/constants';
+import { getEvaluatorOptions } from '../helpers/alertForm';
+import { roundToPlaces } from 'src/helpers/units';
+import moment from 'moment';
+import _ from 'lodash';
+import styles from './AlertIncidents.module.scss';
+
+const formatDateTime = (timestamp) => moment(timestamp).format('MMM DD, YYYY [at] h:mma');
```
shrirupa: May be we should use `formatDateTime` or `formatDate` in `src/helpers/date.js `and also may be add this format `'MMM DD, YYYY [at] h:mma'` to src/config/default.js as a new key if it going to be used in other places too.

src/pages/alerts/components/AlertIncidents.js
```diff
@@ -0,0 +1,103 @@
...
+      <div>{formatDateTime(first_fired)}</div>,
+      <div>{status === 'Active' ? <Tag color='yellow'>Active</Tag> : formatDateTime(last_fired)}</div>,
+      <div className={styles.IncidentFilters}>{renderTags()}</div>,
+      <div className={styles.paddedCell}>{roundToPlaces(triggered_value, 3)}{suffix}</div>
+    ];
+  };
+
+  const TableWrapper = (props) => (
+    <>
+      <div>
```
shrirupa: <> and <div> are not required, right ?

src/pages/alerts/components/AlertIncidents.js
```diff
@@ -0,0 +1,103 @@
...
+        <Table>{props.children}</Table>
+      </div>
+    </>
+  );
+
+  return (
+    <Panel title='Alert Incidents'>
+      {incidents.length <= 0
+        ? <Empty message='No incidents'/>
```
shrirupa: Is loading state required for this table?

src/pages/alerts/components/AlertIncidents.module.scss
```diff
@@ -0,0 +1,14 @@
+@import '~@sparkpost/matchbox/src/styles/config.scss';
+
+.BoldInline{
+  display: inline-block;
+}
+
+.paddedCell {
+  padding: spacing(small) spacing();
```
shrirupa: Right padding needs be slightly more may be  ` padding: spacing(small) spacing(larger)*2; `

src/pages/alerts/components/AlertIncidents.js
```diff
@@ -0,0 +1,103 @@
...
+        if (!acc) {
+          return [tag];
+        }
+        if (!tag) {
+          return acc;
+        }
+        return [...acc, <span key={'join_and'}> and </span>, tag];
+      }, null);
+
+      return joinTags;
```
shrirupa: Nicely done 👍

src/pages/alerts/components/AlertIncidents.js
```diff
@@ -0,0 +1,103 @@
...
+
+      const joinTags = filterTags.reduce((acc, tag) => {
+        if (!acc) {
+          return [tag];
+        }
+        if (!tag) {
+          return acc;
+        }
+        return [...acc, <span key={'join_and'}> and </span>, tag];
+      }, null);
```
shrirupa: not necessary but i think instead of null, we can have filterTags[0] and also can remove `if (!acc) { return [tag];}`

src/pages/alerts/components/AlertIncidents.js
shrirupa: not necessary but i think instead of null (line 63), we can have filterTags[0] and also can remove `if (!acc) { return [tag];}`

src/pages/alerts/components/tests/AlertIncidents.test.js
```diff
@@ -0,0 +1,60 @@
...
+import React from 'react';
+import AlertIncidents from '../AlertIncidents';
+
+describe('Alert Details Component', () => {
+
+  const alert = {
+    name: 'My Alert Name',
+    metric: 'health_score',
+    channels: {
+      emails: ['Myemail@email.com', 'email@ddress.com'],
```
shrirupa: I think we need one more test for renderTags functionality ?  may be a snapshot ?

src/pages/alerts/tests/DetailsPage.test.js
```diff
@@ -32,6 +34,11 @@ describe('Page: Alert Details', () => {
     expect(props.listSubaccounts).toHaveBeenCalled();
   });
 
+  it('should attempt to load incidents', () => {
+    wrapper - shallow(<DetailsPage {...props} />);
```
shrirupa: - is = ?

src/pages/alerts/components/AlertIncidents.js
PatrickSison: Wouldn't it end up adding the first tag twice then since it's added at the start of the reducer? Although I could do an index check, but preference is towards the current implementation. What do you think?

src/pages/alerts/components/tests/AlertIncidents.test.js
PatrickSison: Yeah, I'll look into adding more testing for it 👍

src/pages/alerts/tests/DetailsPage.test.js
PatrickSison: 🤦‍♂ Yup

src/pages/alerts/components/AlertIncidents.module.scss
PatrickSison: Padding was defined by cell. I realized other tables actually right align the column header. I put that change in, but let me know what you think

src/pages/alerts/components/AlertIncidents.js
PatrickSison: Actually, probably not. The page itself currently has a loading state, and the loading state for alert and incidents are shared

PR #1278, opened 2019-11-18
FE-761 Create utility hooks for page filter state

gfriedmansp:
### What Changed
 - Created hook called `useEffectAfterMounting` which will function as a `useEffect`, but will not run on component mount.
- Created hook called `usePageFilters` that will:
  - Maintain a state of filters
  - Hydrate the filter state from the URL
  - Update the route when the filters change
### How To Test
 - There's technically no way to test this visually at the moment, but ensure the test suite runs and that the tests are testing _well_


src/hooks/usePageFilters/usePageFilters.js
```diff
@@ -0,0 +1,140 @@
...
+ *   page: {
+ *     validate: val => !isNaN(val) && val > 0 && val < 10,
+ *     normalize: val => val * 1, // Convert from string to number
+ *     defaultValue: 0,
+ *     excludeFromRoute: false,
+ *   }
+ * });
+ */
+const usePageFilters = whitelist => {
+  const { requestParams = {}, updateRoute } = useRouter();
```
bkemper: `requestParams` should always be an object, so we can remove the default.

src/hooks/usePageFilters/usePageFilters.js
```diff
@@ -0,0 +1,140 @@
...
+  const defaultFilters = useRef(
+    normalizeFilterState(omitFiltersExcludedFromRoute(requestParams, whitelist), whitelist),
+  );
+
+  const cleanReducer = useCallback(
```
bkemper: This `useCallback` is not necessary.

PR #1283, opened 2019-11-20
AC-1081 - Breadcrumbs for the modules

shrirupa:
AC-1081 - Breadcrumbs for the modules

### What Changed
 - Added a shared components to display BreadCrumbs.
 - Added breadCrumbs to the GettingStartedGuide

### How To Test
 - Navigate to /dashboard.
 - check if breadcrumbs are similar to [mocks](https://sparkpost.invisionapp.com/share/DHULLYL45GB#/screens/390961094)

### To Do
- [x] Address Feedback.

### DIFF
--- a/src/components/breadcrumbs/BreadCrumbs.js
+++ b/src/components/breadcrumbs/BreadCrumbs.js
@@ -1,0 +1,17 @@
+import React from 'react';
+import styles from './BreadCrumbs.module.scss';
+import classNames from 'classnames';
+export const BreadCrumbs = ({ children }) => (
+  <div className={styles.BreadCrumbContainer}>
+    {React.Children.map(children, (child, index) => (
+  <>
+  &nbsp; {child} &nbsp; {index + 1 !== React.Children.count(children) && '>'}
+  </>
+    ))}
+  </div>
+);
+
+
+export const BreadCrumbsItem = ({ children, active, onClick }) => (
+  <span onClick={onClick} className={classNames(!active && styles.InActiveItem, styles.BreadCrumbItem)}>{children} </span>
+);
--- a/src/components/breadcrumbs/BreadCrumbs.js
+++ b/src/components/breadcrumbs/BreadCrumbs.js
@@ -5,7 +5,7 @@ export const BreadCrumbs = ({ children }) => (
   <div className={styles.BreadCrumbContainer}>
     {React.Children.map(children, (child, index) => (
   <>
-  &nbsp; {child} &nbsp; {index + 1 !== React.Children.count(children) && <div className={styles.Arrow}>></div>}
+  {child} {index + 1 !== React.Children.count(children) && <div className={styles.Arrow}>&nbsp;>&nbsp;</div>}
   </>
     ))}
   </div>
--- a/src/components/breadcrumbs/BreadCrumbs.js
+++ b/src/components/breadcrumbs/BreadCrumbs.js
@@ -5,7 +5,7 @@ export const BreadCrumbs = ({ children }) => (
   <div className={styles.BreadCrumbContainer}>
     {React.Children.map(children, (child, index) => (
   <>
-  &nbsp; {child} &nbsp; {index + 1 !== React.Children.count(children) && '>'}
+  &nbsp; {child} &nbsp; {index + 1 !== React.Children.count(children) && <div className={styles.Arrow}>></div>}
   </>
     ))}
   </div>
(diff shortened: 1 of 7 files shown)


src/pages/dashboard/components/GettingStartedGuide.js
```diff
@@ -67,7 +74,14 @@ export const GettingStartedGuide = ({ isGuideAtBottom, moveGuideAtBottom }) => {
   };
   return <>
         <Panel title='Getting Started' actions={actions} sectioned >
-          <p><strong> Features</strong> </p>
+          <BreadCrumbs>
+            {breadCrumbsItems[stepName].map((item) => (
```
kchu1337: Instead of creating the individual items as children, why not pass in the array as a prop and let `<BreadCrumbs>` figure out the render logic

src/components/breadcrumbs/BreadCrumbs.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import styles from './BreadCrumbs.module.scss';
+import classNames from 'classnames';
+export const BreadCrumbs = ({ children }) => (
+  <div className={styles.BreadCrumbContainer}>
+    {React.Children.map(children, (child, index) => (
+  <>
+  &nbsp; {child} &nbsp; {index + 1 !== React.Children.count(children) && <div className={styles.Arrow}>></div>}
```
kchu1337: Adding to my pervious comment, Instead of child, you could just do `<BreadCrumbItem {...props}/>`.
Also, this is a UX thing but maybe instead of having the space at the beginning, you could have the space after the carat `>`

src/pages/dashboard/components/GettingStartedGuide.js
```diff
@@ -4,8 +4,15 @@ import { ArrowDownward, Send } from '@sparkpost/matchbox-icons';
 import { Card, CardTitle, CardContent, CardActions } from 'src/components';
 import ButtonWrapper from 'src/components/buttonWrapper';
 import styles from './GettingStartedGuide.module.scss';
+import { BreadCrumbs, BreadCrumbsItem } from '../../../components';
```
kchu1337: Use `src/components`.

src/pages/dashboard/components/tests/GettingStartedGuide.test.js
```diff
@@ -24,4 +24,23 @@ describe('GettingStartedGuide', () => {
...
+    const instance = subject();
+    instance.find('Button').simulate('click');
+    instance.find({ children: 'Show Me SparkPost' }).simulate('click');
+    instance.find('BreadCrumbsItem').at(1).simulate('click');
+    expect(instance).toHaveTextContent('Show Me SparkPost');
+    expect(instance).toHaveTextContent('Let&#39;s Code');
+
+  });
+
+  it('should render the BreadCrumbItem as active corresponding to the Step', () => {
```
kchu1337: Isn't this more of a unit test for the component itself?

src/pages/dashboard/components/GettingStartedGuide.js
shrirupa: If I pass prop like may be - 
`<BreadCrumbs items={ ['Features', 'Sending', 'Show Me SparkPost']} activeStep={'Show Me SparkPost'}/>
` 
I find the approach of less accommodating for following use cases -
1.if the items are JSX then passing props will not be readable.
2. finding the activeStep step logic will get complicated if items that I pass in the array have JSX. We will end up accommodating lot of scenarios for this.
3. If we have a use case in future we want to disable individual steps then we will end up passing one more prop here.(too many props will make it harder to read and understand)

What do you think ?

src/pages/dashboard/components/tests/GettingStartedGuide.test.js
shrirupa: In the current approach the step that is active is controlled by GettingStartedGuide that's the reason to have this test here.

src/pages/dashboard/components/GettingStartedGuide.js
kchu1337: Ok that's fine, I didn't think of future/other use cases.

src/pages/dashboard/components/tests/GettingStartedGuide.test.js
kchu1337: OK, I see.

PR #1284, opened 2019-11-20
FE-419 Use Prettier + Stylelint for code formatting. Update ESLint.

gfriedmansp:
### What Changed
 - Added the prettier code formatter
 - Added the stylelint style formatter
 - Update ESLint
 - Add plugins to turn off ESLint's code formatting rules
 - Add plugins to turn off Prettier's style formatting rules

### How To Test
 - Turn on format on save in VSCode (or other IDE, might have to google how to do so, but ensure that "Prettier" is in charge of it.
 - Save a js/jsx file that you know has formatting issues / change a file to have a bad format (add a string with double quotes or something). Save it, ensure the file is properly formatted.
 - Do the same with a scss file.
 - Turn off format on save, repeat the process and make a dummy branch (git checkout -b dummy-branch), commit the unformatted code, and make sure it gets formatted.


.stylelintrc
```diff
@@ -0,0 +1,4 @@
+{
+  "plugins": ["stylelint-prettier"],
+  "extends": ["stylelint-prettier/recommended"]
+}
```
nicklemmon: Should have an empty line at the end for git reasons. I think there is a way to make VSCode do this by default?

nicklemmon:
@gfriedmansp everything works as expected!

One question, shouldn't we commit editor settings files to the repo so onboarding is easier? For some reason the `.vscode` folder is ignored in the repo as it stands currently. 

Something like this?
https://github.com/nicklemmon/citrus-react/blob/master/.vscode/settings.json

PR #1286, opened 2019-11-22
AC-1082 : "Show Me SparkPost" list for onboarding

shrirupa:
AC-1082 : "Show Me SparkPost" list for onboarding

### What Changed
- Added the GuideListItem.
- List shows up when Show Me SparkPost button is clicked.
- Breadcrumb includes "Show Me SparkPost".

### How To Test
 - Navigate to /dashboard.
 - Click on "Start Sending" and then on "Show Me SparkPost".
 - check if list is similar to [mocks](https://sparkpost.invisionapp.com/share/DHULLYL45GB#/screens/390961095)
- adding itemCompleted={true} to one of the <GuideListItem> and check if it is similar to mocks. 
- List persists as the active list through session until switching off of guide set explicitly.

### To Do
- [ ] Address feedback.
- [x]  Write tests

### DIFF
--- a/src/pages/dashboard/components/GettingStartedGuide.js
+++ b/src/pages/dashboard/components/GettingStartedGuide.js
@@ -28,59 +28,84 @@ export const GettingStartedGuide = ({ isGuideAtBottom, moveGuideAtBottom }) => {
   const renderStep = () => {
     switch (stepName) {
       case 'Features':
-        return <Grid>
-          <Grid.Column xs={12} >
-            <Card >
-              <CardTitle><Send size='20' className={styles.SendIcon}/>   &nbsp;Sending with Sparkpost</CardTitle>
-              <CardContent><p className={styles.FeaturesCardContent}>Learn how to send emails, integrate our API into your code, and make the most of our powerful analytics.</p></CardContent>
-              <CardActions>
-                <ButtonWrapper>
-                  <Button color='orange' onClick={() => setStepName('Sending')}>Start Sending</Button>
-                </ButtonWrapper>
-              </CardActions>
-            </Card>
-          </Grid.Column>
-        </Grid>;
+        return <Panel.Section>
+          <Grid>
+            <Grid.Column xs={12} >
+              <Card >
+                <CardTitle><Send size='20' className={styles.SendIcon}/>   &nbsp;Sending with Sparkpost</CardTitle>
+                <CardContent><p className={styles.FeaturesCardContent}>Learn how to send emails, integrate our API into your code, and make the most of our powerful analytics.</p></CardContent>
+                <CardActions>
+                  <ButtonWrapper>
+                    <Button color='orange' onClick={() => setStepName('Sending')}>Start Sending</Button>
+                  </ButtonWrapper>
+                </CardActions>
+              </Card>
+            </Grid.Column>
+          </Grid>
+        </Panel.Section>
+        ;
       case 'Sending':
-        return <>
-        <p className={styles.SendingStepHeading} role="heading" aria-level="4" ref={guideHeadingRef} tabIndex={-1}>Where Would You Like to Begin?</p>
-        <Grid>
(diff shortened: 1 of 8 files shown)


src/pages/dashboard/components/GuideListItem.js
```diff
@@ -0,0 +1,33 @@
...
+      <div className={styles.CheckBoxContainer}>
+        {itemCompleted ? <CheckCircleOutline size={36} className={styles.CheckCircleIcon}/> : <RadioButtonUnchecked size={36}/>}
+      </div>
+      <div className={styles.ListItemContainer}>
+        {children}
+      </div>
+    </Grid.Column>
+    <Grid.Column md={3} xs={12}>
+      <div className={styles.ListActionContainer}>
+        <Button onClick={onClick} color={(!itemCompleted && 'orange') || null}> {name} </Button>
```
kchu1337: UX comment: Is the button disabled? Also the button looks more like the one with an outline prop.

src/pages/dashboard/components/GuideListItem.module.scss
```diff
@@ -0,0 +1,37 @@
...
+
+.ListActionContainer{
+  float: right;
+}
+
+.CheckCircleIcon{
+  color: status-color(success);
+}
+
+.GuideListItem{
```
kchu1337: UX comment: Looks like the alignment is different in the mocks. A few pixels down and to the right.

src/pages/dashboard/components/tests/GettingStartedGuide.test.js
```diff
@@ -7,7 +7,8 @@ import { ArrowDownward } from '@sparkpost/matchbox-icons';
 describe('GettingStartedGuide', () => {
   const defaultProps = {
     isGuideAtBottom: false,
-    moveGuideAtBottom: jest.fn()
+    moveGuideAtBottom: jest.fn(),
+    storeStepName: jest.fn()
```
kchu1337: Could check that this function is called when the button is clicked.

src/pages/dashboard/components/tests/GuideListItem.test.js
```diff
@@ -0,0 +1,27 @@
...
+    children: null,
+    action: {
+      name: 'Button',
+      onClick: jest.fn()
+    }
+  };
+
+  const subject = (props) => (shallow(<GuideListItem {...defaultProps} {...props} />));
+
+  it('should render a orange button only when list item is not complete', () => {
```
kchu1337: Could split each into 2 separate tests.

src/pages/dashboard/components/GettingStartedGuide.js
```diff
@@ -19,69 +20,109 @@ export const GettingStartedGuide = ({ isGuideAtBottom, moveGuideAtBottom }) => {
...
+  const [stepName, setStepName] = useState(storedStepName || 'Features');
   const guideHeadingRef = useRef(null);
   useEffect(() => {
     if (guideHeadingRef.current) { guideHeadingRef.current.focus(); }
   }, [stepName]);
+  const renderBreadCrumbs = () => (<BreadCrumbs>
+    {breadCrumbsItems[stepName].map((item) => (
+      <BreadCrumbsItem
+        key={item}
+        onClick={() => setStepName(item)}
```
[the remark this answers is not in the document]
shrirupa: Good catch! missed this.

src/pages/dashboard/components/GuideListItem.js
shrirupa: Button is not disabled, user should be able to revisit it. But you are right it looks more like the outline prop one.

src/pages/dashboard/components/tests/GuideListItem.test.js
shrirupa: Since its a if-else like condition, I think we should have it together.

PR #1287, opened 2019-11-23
FE-765 - Adds React Testing Library

nicklemmon:
[FE-765](https://jira.int.messagesystems.com/browse/FE-765)

### What Changed
- Installs `@testing-library/react` ➡️ https://github.com/testing-library/react-testing-library
- Installs supporting matcher library `@testing-library/jest-dom` ➡️ https://github.com/testing-library/jest-dom
- Installs supporting user event simulation library `@testing-library/user-event` for typing and click simulation ➡️ https://github.com/testing-library/user-event
- Installs supporting ESLint plugin to encourage best practices `eslint-plugin-testing-library` ➡️ https://github.com/Belco90/eslint-plugin-testing-library
- Added `data-id` attribute to `<PanelLoading/>` component
- Configured default test ID attribute to `data-id` instead of `data-testid`
- Updated Jest setup script to catch warnings with exception for bug with React 16.8.0 - see comments for more information
- Filed new issue with Matchbox regarding use of `data-test` instead of `data-id` in that library
- Updated `<SendTestEmailButton/>` component tests using React testing library both as an example and to prove that the new setup functions as expected

### How To Test
- Checkout this branch, run `npm install`
- Run all tests and verify that they continue to pass
- Write a new test anywhere in the app using React Testing Library

### To Do
- [ ] Incorporate feedback
- [ ] Update Wiki documentation to discourage use of Enzyme...?

### DIFF
--- a/config/jest/setup.js
+++ b/config/jest/setup.js
@@ -3,6 +3,7 @@
 import raf from './tempPolyfills';
 import Enzyme from 'enzyme';
 import Adapter from 'enzyme-adapter-react-16';
+import '@testing-library/jest-dom/extend-expect';
 import * as matchers from './matchers';
 import setupPortals from 'src/__testHelpers__/setupPortals';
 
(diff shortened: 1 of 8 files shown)


config/jest/setup.js
```diff
@@ -14,11 +16,32 @@ expect.extend(matchers); // register custom matchers
...
+// this is just a little hack to silence a warning that we'll get until we
+// upgrade to 16.9: https://github.com/facebook/react/pull/14853
+const originalError = console.error = (message) => {
+  // Fail tests on any warning
   throw new Error(message);
 };
 
+beforeAll(() => {
+  console.error = (...args) => {
+    if (/Warning.*not wrapped in act/.test(args[0])) {
```
gfriedmansp: This looks scary - is it related to the hack to silence other warnings? I've seen this error message before, but it seemed relevant when I had it /shrug.

nicklemmon: Take a look here and see what you think:
https://github.com/testing-library/react-testing-library#suppressing-unnecessary-warnings-on-react-dom-168

Basically, it will show in every test using React Testing Library without it...and then fail the test. Any other ways around it?

gfriedmansp: 👍 All good, wanted to make sure it was related and can be removed once we upgrade

PR #1293, opened 2019-11-26
SE-222 Alerts email validation and duplicate removal

jmoyse:
Additional validation was added to make sure the count of email list is less than 10. This is also being checked server side, but having only some validated client side seemed out of place to me.

### What Changed
 - Emails inserted into the customer alerts edit and create page now have additional sanitation to remove duplicates.
 - Client side check to make sure the list doesn't go above 10 (already enforced server side)

### How To Test
 - Create/Edit an alert in the UI. https://app.sparkpost.com/alerts/create
 - Set the alert to go to email and add the same email multiple times. 
 - Try to add more than 10 emails.


src/pages/alerts/constants/formConstants.js
```diff
@@ -65,12 +65,14 @@ const slackSubtitle = <Fragment>
...
 
+const countEmails = (emails) => emails.length <= 10;
+
 export const NOTIFICATION_CHANNEL_DATA = {
   emails: {
     icon: <EmailIcon />,
     subtitle: 'Receive notifications through email. One address per line. Up to 10 addresses.',
     fieldProps: {
-      validate: ifStringPresent(emails),
+      validate: [ifStringPresent(emails), countEmails],
```
PatrickSison: This validator doesn't run correctly since it comes back as a string. Would probably need to be structured similarly to `ifStringPresent`

I would recommend moving this validator into `src/helpers/validation.js`, change the validator to something like
```javascript
const countEmails = (str) => multilineStringToArray(str).length <= 10 ? undefined : 'Must contain less than 10 emails';
```

then change the validator for the `formConstants` to contain
```javascript
validate: [ifStringPresent(emails), ifStringPresent(countEmails)]
```

src/pages/alerts/helpers/validateEmailList.js
```diff
@@ -1,6 +1,7 @@
 import _ from 'lodash';
```
PatrickSison: Is this file being imported anywhere? This might be able to take the place of the previous suggestion.

src/pages/alerts/helpers/formatFormValues.js
```diff
@@ -55,7 +55,7 @@ export default function formatFormValues(values) {
   const webhook = (values.webhook || '').trim();
 
   if (emails) {
-    channels.emails = multilineStringToArray(emails);
+    channels.emails = [...new Set(multilineStringToArray(emails))];
```
PatrickSison: What is this doing?

gfriedmansp: de-duplicating

jmoyse: Just recreates the array as a Set (which won't allow duplicates), then returning it back. There are probably more performant wants of doing it, but I feel like this is a nice balance of performance and readability. I'm open to other suggestions if you think it's not clear enough.

PatrickSison: Ahhhh, ok. No, that's a good way 👍

src/pages/alerts/constants/formConstants.js
jmoyse: Good catch. I fixed the validation issue and moved the whole check to the the `emails` function. Seems better to have it all in one place.

src/helpers/validation.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable max-lines */
```
jmoyse: This file was basically at the 150 lines tipping point before adding the few extra lines for the validator. The `eslint-disable` probably isn't the best choice, but it looks like its being used in a few other places as well. I can refactor some stuff around to get under 150 lines. Your call.

gfriedmansp: I don't think 2web2ui has that rule anymore (as of recent). Let me know if it doesn't let you commit it without the disable line so I can fix it!

src/helpers/validation.js
```diff
@@ -24,10 +25,13 @@ export function email(value) {
...
 export function emails(str) {
   const values = multilineStringToArray(str);
-
-  return values.length && values.every(isEmailAddress)
-    ? undefined
-    : 'Must be a comma separated list of valid Email Addresses';
+  let message = undefined;
+  if (values.length <= 0 || !values.every(isEmailAddress)) {
+    message = 'Must be a comma separated list of valid Email Addresses';
+  } else if (values.length >= 10) {
```
bkemper: I see that this helper is only used on the alerts form, but it would be beneficial to break this logic into another helper, `maxItems(10)`.  Something like this...

```
const maxItems = (max) => (str) => {
  const values = multilineStringToArray(str);

  if (values.length > max) {
    return `Must contain no more than ${max} items`;
  }
}
```

bkemper: Also, is the max 9 emails or 10 emails?

src/pages/alerts/helpers/formatFormValues.js
bkemper: Is this too late in the process to deduplicate?  Or should we be doing it as a normalize step?

Screen Shot 2019-12-06 at 4 00 06 PM

src/pages/alerts/helpers/tests/validateEmailList.test.js
```diff
@@ -12,6 +12,11 @@ test('invalid, none valid', () => {
   expect(validateEmailList('  emailaddress1.com')).toEqual('Must enter at least one valid Email Addresses');
 });
 
+test('that no more than 10 email addresses have been entered', () => {
+  const tooManyEmails = [...Array(11).keys()].map((x) => `jane.doe${x}@gmail.com`).join(', ');
```
bkemper: Interesting... I have always used `Array.from(Array(11))` to achieve the same thing as `[...Array(11).keys()]`

src/pages/alerts/helpers/validateEmailList.js
bkemper: I don't see this being used either.  It might be time to retire it.

src/helpers/validation.js
jmoyse: From what I can see in the customer-alerts it looks like 10. I pulled the same error message from the alerts API.
https://github.com/SparkPost/customer-alerts/blob/master/packages/api/src/lib/validators/create.ts#L44

src/pages/alerts/helpers/formatFormValues.js
jmoyse: Good catch. I went changed the order of the de-duplication and length check. It will remove the duplicate and get the count after now.

PR #1302, opened 2019-12-03
AC-1083 - "Show Me SparkPost" checklist item for sending test email

shrirupa:
AC-1083 - "Show Me SparkPost" checklist item for sending test email

### What Changed
 - "Show Me SparkPost" Button directs to the Templates page / starts the guide in AC-1107
 - Condition is stored in UI option on clicking the button and navigating to the Templates page. (The ListItem will have a green tick and button is no longer orange)

### How To Test
 -  Click on 'Start Sending' then 'Show Me SparkPost' then 'Send a Test Email'.
 -  Check if click on Send a Test Email button redirects to /templates page, check after redirection 
      the list item is completed(green tick and button is no longer orange. )

### To Do
- [ ] Address Feedback

### DIFF
--- a/src/pages/dashboard/components/GettingStartedGuide.js
+++ b/src/pages/dashboard/components/GettingStartedGuide.js
@@ -6,24 +6,20 @@ import ButtonWrapper from 'src/components/buttonWrapper';
 import styles from './GettingStartedGuide.module.scss';
 import { BreadCrumbs, BreadCrumbsItem } from 'src/components';
 import { GuideListItem, GuideListItemTitle, GuideListItemDescription } from './GuideListItem';
+import { GUIDE_IDS } from '../constants';
 
-export const GettingStartedGuide = ({
-  onboarding = {},
-  // isGuideAtBottom,
-  onboardingActiveStep,
-  history,
-  setAccountOption,
-  ...rest
-}) => {
+export const GettingStartedGuide = ({ onboarding = {}, history, setAccountOption }) => {
   const breadCrumbsItems = {
     Features: ['Features'],
     Sending: ['Features', 'Sending'],
     'Show Me SparkPost': ['Features', 'Sending', 'Show Me SparkPost'],
     "Let's Code": ['Features', 'Sending', "Let's Code"],
   };
-  const { isGuideAtBottom = false } = onboarding;
+  const { isGuideAtBottom = false, active_step, send_test_email_completed } = onboarding;
 
-  const { sendTestEmail } = rest;
+  const setOnboardingAccountOption = (obj = {}) => {
+    setAccountOption('onboarding', { ...onboarding, ...obj });
+  };
 
   const actions = isGuideAtBottom
     ? null
@@ -36,12 +32,11 @@ export const GettingStartedGuide = ({
             </span>
           ),
           color: 'blue',
-          onClick: () =>
-            setAccountOption('onboarding', { ...onboarding, ...{ isGuideAtBottom: true } }),
+          onClick: () => setOnboardingAccountOption({ isGuideAtBottom: true }),
         },
       ];
   //stepName could be Features,Sending,Show Me Sparkpost, Let's Code
-  const [stepName, setStepName] = useState(onboardingActiveStep || 'Features');
+  const [stepName, setStepName] = useState(active_step || 'Features');
   const guideHeadingRef = useRef(null);
   useEffect(() => {
     if (guideHeadingRef.current) {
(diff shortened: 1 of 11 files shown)


src/pages/dashboard/components/GettingStartedGuide.js
```diff
@@ -7,122 +7,211 @@ import styles from './GettingStartedGuide.module.scss';
...
+          color: 'blue',
+          onClick: () => setAccountOption('isGuideAtBottom', true),
+        },
+      ];
   //stepName could be Features,Sending,Show Me Sparkpost, Let's Code
   const [stepName, setStepName] = useState(storedStepName || 'Features');
   const guideHeadingRef = useRef(null);
   useEffect(() => {
-    if (guideHeadingRef.current) { guideHeadingRef.current.focus(); }
+    if (guideHeadingRef.current) {
```
PatrickSison: Out of curiosity, is this an accessibility thing? I'm not familiar with it. It always ends up highlighting the Sending step

src/pages/dashboard/components/GettingStartedGuide.js
```diff
@@ -7,122 +7,211 @@ import styles from './GettingStartedGuide.module.scss';
...
+          {item}
+        </BreadCrumbsItem>
+      ))}
+    </BreadCrumbs>
+  );
 
-  const setAndStoreStepName = (stepName) => {
-    storeStepName(stepName);
+  const setAndStoreStepName = stepName => {
+    setAccountOption('stepName', stepName);
```
PatrickSison: Not sure if it's too late to do this, but what do you think about renaming the option that this is stored under? I'm wondering if `stepName` is too vague and if we should name it something more specific?

src/pages/dashboard/components/GettingStartedGuide.js
```diff
@@ -7,122 +7,211 @@ import styles from './GettingStartedGuide.module.scss';
...
     setStepName(stepName);
   };
+  const handleAction = action => {
+    switch (action) {
+      case 'Send Test Email':
+        setAccountOption('sendTestEmail', true);
```
PatrickSison: A similar idea with this. Should we preface this with something that indicates that these account options fall under `onboarding_sending_blahblah` or nesting another object? Not a huge deal, but maybe to avoid collision with names in the future with actual account UI options.

src/pages/dashboard/components/GettingStartedGuide.js
```diff
@@ -7,122 +7,211 @@ import styles from './GettingStartedGuide.module.scss';
...
-                    <Button color='orange' onClick={() => setAndStoreStepName('Sending')}>Start Sending</Button>
-                  </ButtonWrapper>
-                </CardActions>
-              </Card>
-            </Grid.Column>
-          </Grid>
-        </Panel.Section>;
-
+        return (
+          <Panel.Section>
```
PatrickSison: Should the panel section and breadcrumbs be extracted as well? I guess the circumstance is if something errors and we can't find a proper step and there's no clean empty state.

src/pages/dashboard/components/GettingStartedGuide.js
```diff
@@ -7,122 +7,211 @@ import styles from './GettingStartedGuide.module.scss';
...
-    storeStepName(stepName);
+  const setAndStoreStepName = stepName => {
+    setAccountOption('stepName', stepName);
     setStepName(stepName);
   };
+  const handleAction = action => {
+    switch (action) {
+      case 'Send Test Email':
+        setAccountOption('sendTestEmail', true);
+        history.push('/templates?pendo=6RgwDLtUU5Ynp20auFvU9Qjbpqg');
```
PatrickSison: Should we store the pendo guide IDs in a constant file?

src/pages/dashboard/components/GettingStartedGuide.js
shrirupa: yeah, its for screenreaders.

src/pages/dashboard/components/GettingStartedGuide.js
shrirupa: Its not late at all, what do you think this should be renamed to `current_guide_step` ?

src/pages/dashboard/components/GettingStartedGuide.js
shrirupa: yes you are right, I am going to put all these keys under `onboarding` object.

src/pages/dashboard/components/GettingStartedGuide.js
shrirupa: I did not understand what errors are we talking about here ? and clean empty state ?

src/pages/dashboard/components/GettingStartedGuide.js
```diff
@@ -6,123 +6,210 @@ import ButtonWrapper from 'src/components/buttonWrapper';
...
   };
-  const actions = isGuideAtBottom ? null : [{
-    content: <span> Move to Bottom <ArrowDownward size='20'/> </span>,
-    color: 'blue',
-    onClick: moveGuideAtBottom
-  }];
+  const { isGuideAtBottom = false, active_step, send_test_email_completed } = onboarding;
+
+  const setOnboardingAccountOption = (obj = {}) => {
+    setAccountOption('onboarding', { ...onboarding, ...obj });
```
PatrickSison: Do we need to spread the entire object onboarding object?

shrirupa: yes, planning to have state of all the step in this object, eg , Analytics_step_completed, send_test_email_completed.

PatrickSison: Not sure if it's a control endpoint difference, but the keying seems to work without spread. For example, I can make a request with
```javascript
"foo": {
  "bar": true
}
```

then

```javascript
"foo": {
  "bar2": true
}
```

and the final object would still come back as

```javascript
"foo": {
  "bar": true,
  "bar2": true
}
```

PR #1306, opened 2019-12-05
FE-766 - Remove in-app notifications

shrirupa:
FE-766 - Remove in-app notifications

### What Changed
 - Removed dropdown from top navigation
 - Removed all markdown files and code to import them including webpack configuration

### How To Test
 - verify there is no in app notification
 - verify the storybook still works
 - verify that all components that were used only by in-app notification are removed.

### To Do
- [ ] Address feedback.


shrirupa:
> Looks like you missed `src/selectors/helpers/transformNotifications.js` & `src/reducers/notifications.js`

Good catch!

PR #1307, opened 2019-12-05
FE-768 R&D - Swapping between Hibana styles and existing styles

jonambas:

### What Changed
     > Outline changes to user flows, any surprising architectural decisions 
     > Alignment between PR commits and this section helps your reviewers.

### How To Test
 - Describe any special dev environment setup required.
     > Anything outside basic hamurai-managed stuff counts as "special".

### To Do
- [ ] Tag new routes in Pendo


src/components/hibanaWrappers.js
```diff
@@ -0,0 +1,45 @@
...
+  'marginY',
+  'm',
+];
+
+// Demonstrates omitting new component props
+export function TextField(props) {
+  const isHibanaEnabled = useHibana();
+
+  if (isHibanaEnabled) {
+    return <NewTextField {...props} />;
```
jonambas: TextField doesn't make use of system props yet, this is just a code example

src/components/hibanaWrappers.js
```diff
@@ -0,0 +1,45 @@
...
+  'mr',
+  'marginBottom',
+  'mb',
+  'marginLeft',
+  'ml',
+  'marginX',
+  'mx ',
+  'marginY',
+  'm',
+];
```
jonambas: These props could also be handled in 3.x matchbox instead of here

src/components/hibanaWrappers.js
jonambas: Also, we could potentially loop through components instead of manually creating a wrapper for each one

src/context/Hibana.js
```diff
@@ -0,0 +1,46 @@
...
+
+// A context provider to manage switching between two Matchbox versions 
+// Provides a button toggle
+// Any account feature flag check would be present in this component
+ 
+export function Hibana(props) {
+  const [enabled, setEnabled] = React.useState(defaultContext);
+
+  return (
+    <HibanaContext.Provider value={enabled}>
```
jonambas: Can also pass through `setEnabled` here if we need to programatically toggle anywhere else

nicklemmon:
Successfully implemented in the app. Closing.

PR #1312, opened 2019-12-10
SA-1040 Add Inbox Placement Filtering

gfriedmansp:
### What Changed
 - Added a TypeAhead to Inbox Placement Trends page
 - Added a DateFilter to Inbox Placement Trends page
 - Managed state using the new `usePageFilters` hook

### How To Test
 - `/inbox-placement` page will add some default filters to URL (from 30 days ago, to now)
 - The chart will populate with those default filters
 - When changing the date in the date picker, the url will change with the new date + update chart
 - When using the typeahead, you can choose filters and the chart will update + the filters will appear below
 - Refreshing the page at any time will use the previously chosen filters as defaults (deep link) (the text in the typeahead will not persist and that's okay)
 - Choosing a 30 day range, waiting 2 minutes so it's technically not 30 days from now will change the datepicker selection to custom on refresh. This again, is for deep linking.

### To Do
- [x] Update tests


src/pages/inboxPlacement/components/TrendsFilters.js
```diff
@@ -0,0 +1,137 @@
...
+                updateFilters({
+                  dateRange: {
+                    from: formatApiTimestamp(from),
+                    to: formatApiTimestamp(to),
+                  },
+                  range: 'custom',
+                });
+              }
+            }}
+            roundToPrecision={true}
```
PatrickSison: Not sure if it is relevant, but the date picker shows a precision of being to the hour, but I don't think the data supplied is broken down by hour

src/components/typeahead/Typeahead.js
```diff
@@ -19,13 +19,13 @@ export const TypeaheadItem = ({ id, label }) => (
 export class Typeahead extends Component {
```
PatrickSison: Prettier

gfriedmansp:
> Screen Shot 2019-12-17 at 12 04 41 PM
> 
> MEGA BAR!

Resolve by adding 7 day minimum range to date picker

src/pages/inboxPlacement/components/TrendsFilters.js
gfriedmansp: So I think this was necessary because of the whole UTC vs. local time thing. The data itself is rolled up per UTC day, so if the user doesn't supply a time, it could actually be a different day depending on their time zone. Not sure the best way to handle it in this case.

src/reducers/inboxPlacement.js
```diff
@@ -38,11 +38,25 @@ export default (state = initialState, { type, payload, meta }) => {
...
+    case 'GET_INBOX_PLACEMENT_TRENDS_FILTER_VALUES_SUCCESS': {
+      return {
+        ...state,
+        getTrendsFilterValuesPending: false,
+        trendsFilterValues: payload,
+        getTrendsFilterValuesError: null,
+      };
+    }
+    case 'GET_INBOX_PLACEMENT_TRENDS_FILTER_VALUES_FAIL':
+      return { ...state, getTrendsFilterValuesPending: false, getTrendsFilterValuesError: payload };
```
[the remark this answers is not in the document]
gfriedmansp: I believe its used in the typeahead with `disabled={trendsFilterValuesloading || trendsFilterValuesError}`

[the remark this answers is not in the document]
PatrickSison: Ahh, missed that. Cool 👍

PR #1323, opened 2019-12-17
SE-463: Alert incidents polish

PatrickSison:
### What Changed
 - Alert edit form disabled changing metric/evaluator
 - Health score (raw) now removes percent correctly in alert incidents
 - Health score label will now show appropriate evaluator as column header in incidents table
 - Added keys to table
 - Changed precision of values

### How To Test
 - Go to `/alerts` page and click on a health score alert (or create one)
    - Create form should be able to change the metric/evaluator
    - Edit form should disable metric field and evaluator (if health score metric alert)
 - Verify alert incident table changes are made

### DIFF
--- a/src/pages/alerts/components/AlertIncidents.js
+++ b/src/pages/alerts/components/AlertIncidents.js
@@ -3,7 +3,7 @@ import { Panel, Table, Tag } from '@sparkpost/matchbox';
 import { TableCollection, Empty } from 'src/components';
 import { FORMATS } from 'src/constants';
 import { formatDateTime } from 'src/helpers/date';
-import { METRICS, FILTERS_FRIENDLY_NAMES } from '../constants/formConstants';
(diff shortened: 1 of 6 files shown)


src/pages/alerts/components/AlertIncidents.js
```diff
@@ -12,24 +12,27 @@ import _ from 'lodash';
 import styles from './AlertIncidents.module.scss';
 
 const AlertIncidents = ({ incidents = [], alert, subaccountIdToString }) => {
-
-  const { metric, source } = alert;
+  const { metric, threshold_evaluator = {} } = alert;
+  const { source } = threshold_evaluator;
```
shrirupa: Not important, skip it if you want - `const { metric, threshold_evaluator: { source } = {} } = alert;`

PR #1326, opened 2019-12-19
AC-1168: Added link to external docs

PatrickSison:
### What Changed
 - Added link to external docs in onboarding checklist
 - Checklist item after clicking on link to the docs

### How To Test
 - Go to the dashboard with UI feature flag `messaging_onboarding` enabled
 - Verify under "Let's Code" checklist that copy is correct (see [AC-1168](https://jira.int.messagesystems.com/browse/AC-1168))
 - Verify link works correctly
 - Verify that check appears after clicking link and persists


src/pages/dashboard/components/GettingStartedGuide.js
```diff
@@ -188,17 +192,28 @@ export const GettingStartedGuide = ({
...
+            <Panel.Section>
+              <CheckListItem
+                {...LETS_CODE_LIST['View Developer Docs']}
+                itemCompleted={view_developer_docs_completed}
+                action={{
+                  to: 'https://developers.sparkpost.com/api',
+                  external: true,
+                }}
+              />
+            </Panel.Section>
```
nicklemmon: Screen Shot 2019-12-20 at 10 19 16 AM

Styles for smaller screen sizes could be improved - icons are a little all over the place. The button position looks fine, but it could use some space around it. Things look fine on larger screen sizes.

src/pages/dashboard/components/GuideListItem.js
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import { RadioButtonUnchecked, CheckCircleOutline } from '@sparkpost/matchbox-icons';
 import { Button, Grid } from '@sparkpost/matchbox';
 import styles from './GuideListItem.module.scss';
-export const GuideListItem = ({ itemCompleted, children, action: { name, onClick } }) => (
+export const GuideListItem = ({ itemCompleted, children, action: { name, onClick, ...rest } }) => (
```
nicklemmon: I think we are trying to avoid prop spreading like this - in future versions of React (in our case, versions that are already out) arbitrary props will make their way to the DOM if children don't control for them.

src/pages/dashboard/components/tests/GettingStartedGuide.test.js
```diff
@@ -101,7 +101,26 @@ describe('GettingStartedGuide', () => {
...
+      .onClick();
+
+    expect(defaultProps.setAccountOption).toHaveBeenCalledWith('onboarding', {
+      view_developer_docs_completed: true,
+    });
+
+    expect(instance.find('Button[data-id="View Developer Docs"]')).toHaveProp(
+      'to',
+      'https://developers.sparkpost.com/api',
+    );
```
nicklemmon: Have you tried using React Testing Library for these tests? Since Redux Form isn't involved, they would be ideal. I'm worried that these tests will prevent refactoring since they rely so heavily on component names rather than the actual DOM they output.

If there is a big rush to get this out the door then we could save this for another time, but if not, I think this would be a great learning opportunity, and it would really improve the quality of these tests.

src/pages/dashboard/constants.js
```diff
@@ -13,12 +13,12 @@ export const SENDING_STEP_LIST = {
...
     name: 'Show Me SparkPost',
     content:
-      'Send your first email in one click and dive right into what SparkPost can do for your email strategy',
+      'Send your first email in one click and dive right into what SparkPost can do for your email strategy.',
   },
   "Let's Code": {
     name: "Let's Code",
     content:
-      "Ready to integrate via SMTP or API? We'll get you set up ASAP so you can start building with SparkPost",
+      "Ready to integrate via SMTP or API? We'll get you set up ASAP so you can start building with SparkPost.",
```
nicklemmon: For these abbreviations, try using the new `<Abbreviation/>` component. You can see docs for it right next to the component in `src/components`

src/pages/dashboard/components/tests/GettingStartedGuide.test.js
PatrickSison: I think these are simple enough to be able to convert to react testing library 👍

src/pages/dashboard/constants.js
PatrickSison: Cool, I'll add it in

src/pages/dashboard/components/GettingStartedGuide.js
nicklemmon: This is totally fixed. Nice.

I did run in to one more issue with responsive-ness (not a word I don't think...) but it is outside the scope of this ticket. Just wanted to make note of it. There should be gap between these two `<Card/>` components. This could be refactored to use CSS grid, in which case the `grid-gap` value would be applied whether it is between columns or rows. Very handy.

Screen Shot 2020-01-06 at 10 45 22 AM

src/pages/dashboard/components/tests/GettingStartedGuide.test.js
```diff
@@ -45,94 +47,76 @@ describe('GettingStartedGuide', () => {
...
 
   it('should navigate to templates page when Send a Test Email button is clicked', () => {
-    const instance = subject({ onboarding: { active_step: 'Show Me SparkPost' } }, mount);
-    instance
-      .find('GuideListItem')
-      .at(0)
-      .prop('action')
-      .onClick();
+    const { queryByText } = subject({ onboarding: { active_step: 'Show Me SparkPost' } });
+    userEvent.click(queryByText('Send Test Email'));
```
nicklemmon: Nice! Much better.

PR #1327, opened 2019-12-19
SA-1005: Add to Watchlist page

gfriedmansp:
### What Changed
 - Created a page at `/blacklist/watchlist/add`
 - Page handles adding a single resource (sending domain or ip address) to the watchlist
 - Page offers ability to "save" and redirect back to `blacklist/watchlist` or to "save and add another" which does the same minus the redirect

### How To Test
 - Add something to the watchlist using save
   - Ensure it redirects 
   - Ensure it displays a success message
 - Add something to the watchlist using save and add another
   - Ensure it does not redirect
   - Ensure it clears input
   - Ensure it displays a success message
 - Try to re-add an existing resource
  - Ensure it displays the error message below the resource
  - Ensure the error message goes away when you change the input


src/actions/blacklist.js
```diff
@@ -11,6 +11,24 @@ export function listMonitors() {
   });
 }
 
+export function watchlistAdd(resource) {
+  try {
```
shrirupa: I think try-catch block can be removed, since sparkpostApiRequest has error handling in it.

src/pages/blacklist/WatchlistAddPage.js
```diff
@@ -0,0 +1,108 @@
...
+  const handleSubmit = useCallback(
+    e => {
+      e.preventDefault();
+      watchlistAdd(resource)
+        .then(({ resource }) => {
+          showAlert({ type: 'success', message: `Added ${resource} to Watchlist` });
+          history.push(`/blacklist/watchlist`);
+        })
+        .catch(() => {
+          // Don't do anything since `submitError` prop will be passed
```
shrirupa: Can we remove this if this is not being used ?

src/pages/blacklist/WatchlistAddPage.js
```diff
@@ -0,0 +1,108 @@
...
+  watchlistAdd,
+  showAlert,
+  history,
+  submitPending,
+  submitError,
+}) => {
+  const [resource, setResource] = useState('');
+  const [hideSubmitError, setHideSubmitError] = useState(false);
+
+  const onChange = useCallback(
```
shrirupa: Sorry, but I did not understand the reason to wrap `onChange` `handleSubmit` and `handleSaveAndNew` in useCallback.

src/actions/blacklist.js
gfriedmansp: We need the try catch to manually reject the Promise, since sparkpostApiRequest just throws an exception when it gets a 400 error.

src/pages/blacklist/WatchlistAddPage.js
gfriedmansp: Nope, it is being used, we're just not doing anything when it catches. Otherwise we'd have an uncaught promise rejection

src/pages/blacklist/WatchlistAddPage.js
gfriedmansp: Without the useCallback, every time this component renders (every keypress in the input), it would create a new function for both of these functions. It's just a performance gain (and probably a good habit)

shrirupa: I am ok with you having it like this. but I wonder how much of a performance gain is it in this case because these functions are not really doing a lot of computing and component rerenders only when its props changes 🤷‍♀ 
Came accross this -> https://kentcdodds.com/blog/usememo-and-usecallback

PR #1328, opened 2019-12-19
TR-2047 - Lower max file size for RV from 200MB to 20MB

nicklemmon:
[TR-2047](https://jira.int.messagesystems.com/browse/TR-2047)

### What Changed
- Reformatted two files with Prettier
- Lowered max RV file size upload from 200MB to 20MB (NOTE: the config defines the limit in bytes, which was derived from https://www.gbmb.org/bytes-to-mb)

### How To Test
- Updated your local `ngnix.conf` (mine was in `/usr/local/etc/openresty`) file to accept larger file uploads, then restart openresty:
  - Directions here: https://stackoverflow.com/questions/26717013/how-to-edit-nginx-conf-to-increase-file-size-upload
- Go to `/recipient-validation/list` and upload a file that is above 20MB and another that is below 20 MB - **For whoever picks up the review, I have some uploaded CSV files in a Google Drive folder that I will share with you**
- Verify that the file size under 20 MB uploads successfully
- Verify that the file size over 20 MB renders an error

### To Do
- [ ] Incorporate feedback

### DIFF
--- a/src/config/default.js
+++ b/src/config/default.js
@@ -56,7 +56,7 @@ const config = identifier => ({
     cookieName: 'heroku-nav-data',
   },
   maxUploadSizeBytes: 20000000,
-  maxRecipVerifUploadSizeBytes: 20971520,
+  maxRecipVerifUploadSizeBytes: 20971520, // NGNIX configures max upload size in megabytes, translating to this value in bytes
   metricsPrecisionMap: [
     { time: 60, value: '1min', format: 'ha' },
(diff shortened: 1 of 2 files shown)


nicklemmon:
> I think, I am doing something wrong setting this up.
> I did get an error when I uploaded a file over 20MB but api gives a 500 when I am trying to upload the 12 MB file that you shared. But again I was able to upload that demo file on RV. 🤔

@shrirupa Hmm, did you make the Ngnix config changes?

PR #1341, opened 2020-01-02
TR-2038 - Error message for large files when uploading recipient lists larger than 10 MB

nicklemmon:
[TR-2038](https://jira.int.messagesystems.com/browse/TR-2038)

### What Changed
- Updates configuration to recipient list validation limit of 12 MB (in bytes)
- Adds `asyncValidate` function 
- Adds focus styles for file upload fields
- Prettier reformatting on the recipient list creation form component

### How To Test
1. Message me on Slack and I will share example links to example CSV files with you
1. Go to `/lists/recipient-lists/create`
1. Fill out the form, upload `over-12mb-recipient-list.csv`, submit the form, and verify that an error renders
1. Fill out the form, upload `under-12mb-recipient-list.csv`, submit the form, and verify that the recipient list submits successfully

### To Do
- [ ] HELP! I have tried to write unit test for this and nothing has worked. I even tried exporting the `asyncValidate` function and testing it directly, but passing in a `new File()` even with a defined `size` is not working. Does anyone have experience testing something like this? There was not test for this file size limit previously, unfortunately.


src/pages/recipientLists/components/RecipientListForm.js
```diff
@@ -1,130 +1,171 @@
...
+                  <span>
+                    You can download a{' '}
+                    <DownloadLink href={exampleRecipientListPath}>CSV template here</DownloadLink>{' '}
+                    to use when formatting your recipient list for upload.
+                  </span>
+                }
+                label={uploadHint}
+                name="csv"
+                validate={uploadValidators}
+                onBlur={() => asyncValidate()}
```
bkemper: This wouldn't work unless you passed the values.

src/pages/recipientLists/components/RecipientListForm.js
```diff
@@ -1,130 +1,171 @@
...
+      )}. Please upload a smaller file.`,
+    };
+    /* eslint-enable no-throw-literal */
+  }
+};
+
+const WrappedForm = reduxForm({
+  form: formName,
+  asyncValidate,
+  asyncBlurFields: ['username', 'password', 'csv'],
```
bkemper: Should we check after all fields?

src/pages/recipientLists/components/RecipientListForm.js
nicklemmon: Duh. Nice catch.

src/pages/recipientLists/components/RecipientListForm.js
nicklemmon: It couldn't hurt - now that you mention it, those `'username'` and `'password'` values stem from the Redux Form docs async validation example, so definitely need to remove those too (not that they are causing any harm, just useless code)

nicklemmon:
Closing this PR in favor of #1344

PR #1344, opened 2020-01-03
TR-2038 - Simpler implementation - sync validation of recipient list file size handling by updating config, add global 413 error handling

nicklemmon:
[TR-2038](https://jira.int.messagesystems.com/browse/TR-2038)

### What Changed
- Prettier formatting for multiple files
- Incorporate default message details content
- Updates sparkPostApi request tests so they no longer violate the ESLint `jest/no-try-expect` rule (https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/no-try-expect.md)

### How To Test
1. Reach out to me and I'll give access to example CSV files that can be used to verify the frontend validation behavior
1. Upload said files and encounter the new error

### To Do
- [x] Incorporate feedback
- [x] Incorporate error message content

### DIFF
--- a/src/actions/helpers/sparkpostApiRequest.js
+++ b/src/actions/helpers/sparkpostApiRequest.js
@@ -35,7 +35,7 @@ const sparkpostRequest = requestHelperFactory({
       type: types.SUCCESS,
       meta,
       payload: results,
-      extra: { links, total_count }
+      extra: { links, total_count },
     });
 
     return meta.onSuccess ? dispatch(meta.onSuccess({ results })) : results;
@@ -45,7 +45,7 @@ const sparkpostRequest = requestHelperFactory({
     // sparkpost axios instance, so all sparkpost API errors look the same
     // (even those that don't use this helper because the request is not authenticated)
     const apiError = new SparkpostApiError(err);
-    const { message, response = {}} = apiError;
+    const { message, response = {} } = apiError;
     const { auth } = getState();
     const { retries = 0, showErrorAlert = true } = meta;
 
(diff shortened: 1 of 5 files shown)


src/pages/recipientLists/components/RecipientListForm.js
```diff
@@ -59,72 +65,79 @@ export class RecipientListForm extends Component {
 
     let actionText = 'Create';
     let uploadHint = 'Upload a CSV file of recipients';
-    let uploadValidators = [required, maxFileSize(config.maxUploadSizeBytes)];
+    let uploadValidators = [required, maxFileSize(config.maxRecipListUploadSizeBytes)];
```
bkemper: I don't think it is worth keeping this `maxFileSize` check because it is really inaccurrate.

bkemper: For example, the bug was raised with a 8.5mb file which when converted to json payload was over 12mb.

src/actions/helpers/sparkpostApiRequest.js
```diff
@@ -99,26 +100,31 @@ const sparkpostRequest = requestHelperFactory({
...
     if (!get404 && showErrorAlert) {
       dispatch(
         showAlert({
           type: 'error',
           message: 'Something went wrong.',
-          details: message
-        }));
+          details:
+            response.status === 413
+              ? 'File size larger than the server allows. Try again.'
```
bkemper: Do you think this should be on a global level? Or just handled in the form?

nicklemmon: I think it makes sense to handle it globally - right now, the feedback isn't very helpful globally. It just says something like 'Status code 413'

src/pages/recipientLists/components/RecipientListForm.js
nicklemmon: fair enough!

nicklemmon: Alright, done.

src/actions/helpers/sparkpostApiRequest.js
nicklemmon: I'm working with Aubrey to get new content - she asked that I follow up with Ang whose team will apparently be handling content going forward? I started a thread with the two of them to see what a process might look like. For this small of a change, probably nothing formal is needed but we'll definitely want something in place.

nicklemmon:
@bkemper OK this one is ready for another pass - the content I have in there was approved by Ang so we can move forward there. It is fairly generic, so covers us better than the generic 'Status code 413' we had previously.

PR #1345, opened 2020-01-06
SA-827 Add Incident Details Page

gfriedmansp:
### What Changed
 - Added an Incident Details Page which includes:
   A Panel with:
   - Date Listed
   - Date Resolved (or yellow active tag if not resolved)
   - Days Listed
   - List of up to 6 most recent historical incidents (same blacklist, same resource)
   - Button to link to reports/summary with the given parameters:
     - Date range of from 7 days before listing date to 7 days after resolved date (or current day if unresolved)
     - Filtered by sending domain or IP based on the incident's resource
  A Panel with:
   - List of up to 3 most recent incidents on _this_ incident's blacklist
   - If there are none, an empty state says so
  A Panel with:
   - List of up to 3 most recent incidents of _this_ incident's resource
   - If there are none, an empty state says so

### How To Test
 - Ensure all AC is there
 - Ensure loading states are good (each panel has separate loading state, since they are different endpoints

### To Do
- [ ] Add some more tests for code coverage


src/pages/blacklist/IncidentsPage.js
```diff
@@ -44,27 +40,21 @@ export const IncidentsPage = props => {
...
+          />
+        ) : null}
         <IncidentsCollection incidents={incidents} />
       </div>
     );
   };
 
   return (
     <Page
-      empty={{
```
kchu1337: I think we still need this since this page is the default empty page when they have not added any blacklist monitors/resources.

src/pages/blacklist/IncidentsPage.js
```diff
@@ -44,27 +40,21 @@ export const IncidentsPage = props => {
 
     return (
       <div data-id="incidents-table">
+        {incidents.length === 0 ? (
```
kchu1337: You can use `incidents.length === 0 &&` instead of a ternary.

src/actions/blacklist.js
```diff
@@ -43,3 +43,67 @@ export function listIncidents(from = '2019-01-01') {
...
+    type: 'GET_INCIDENT',
+    meta: {
+      method: 'GET',
+      url: `/v1/blacklist-monitors/incidents/${incidentId}`,
+      showErrorAlert: false,
+    },
+  });
+}
+
+export function listIncidentsForResource(resource, from = '2019-01-01', limit = 4) {
```
kchu1337: You can use `/v1/blacklist-monitors/incidents` and filter by resource. Also consider making a helper function for the dispatch with the type and query params as arguments.

src/pages/blacklist/IncidentDetailsPage.js
```diff
@@ -0,0 +1,165 @@
...
+
+  const renderContent = () => {
+    if (error) {
+      return (
+        <div data-id="error-banner">
+          <ApiErrorBanner
+            message={'Sorry, we seem to have had some trouble loading your blacklist incidents.'}
+            errorDetails={error.message}
+            reload={() => {
+              getIncident(id);
```
kchu1337: Might also need to call the other list endpoints as well.

src/pages/blacklist/IncidentDetailsPage.js
```diff
@@ -0,0 +1,165 @@
...
+    </Page>
+  );
+};
+
+const mapStateToProps = (state, props) => {
+  const { id } = props.match.params;
+  return {
+    historicalIncidents: selectHistoricalIncidents(state),
+    historicalIncidentsPending: state.blacklist.historicalIncidentsPending,
+    historicalIncidentsError: state.blacklist.historicalIncidentsError,
```
kchu1337: You are getting the error from redux but not using it.

src/pages/blacklist/components/CongratsBanner.js
```diff
@@ -1,15 +1,15 @@
 import React from 'react';
 import { Banner } from '@sparkpost/matchbox';
 
-const CongratsBanner = ({ onDismiss }) => {
+const CongratsBanner = ({
+  onDismiss,
+  title = 'Congratulations! You are not currently on a Blacklist',
```
kchu1337: Hopefully Ang can think of wording that works for both instances.

src/pages/blacklist/components/IncidentDetails.js
```diff
@@ -0,0 +1,64 @@
...
+        <strong>Days Listed</strong>
+        <span className={styles.Value}>{daysListed}</span>
+      </div>
+    </Grid.Column>
+    <Grid.Column xl={5} xlOffset={1} lg={6} xs={12}>
+      <div className={cx(styles.DetailGroup, styles.HistoricalIncidents)}>
+        <strong>Historical Incidents</strong>
+        {historicalIncidents.length === 0 ? (
+          <span>{`No historical incidents for ${resourceName} on ${blacklistName}`}</span>
+        ) : null}
```
kchu1337: I think instead of null, you could display the the historical incidents as the false condition.

src/pages/blacklist/components/IncidentDetails.js
```diff
@@ -0,0 +1,64 @@
...
+    </Grid.Column>
+    <Grid.Column xs={12}>
+      <Button
+        size="small"
+        color="orange"
+        className={styles.EngagementButton}
+        to={`/reports/summary?from=${moment
+          .utc(listedTimestamp)
+          .subtract('7', 'days')
+          .format()}&to=${moment.utc().format()}&range=custom&`}
```
kchu1337: I feel like they want to filter by engagement (is the click and open?). And idk if they mean 7 days from the timestamp or 7 days from now?

src/pages/blacklist/components/RelatedIncidents.js
```diff
@@ -0,0 +1,39 @@
...
+import { AccessTime } from '@sparkpost/matchbox-icons';
+import { PageLink } from 'src/components';
+import styles from './RelatedIncidents.module.scss';
+
+export default ({ incidents = [], header = '', type = '' }) => (
+  <Table>
+    <tbody>
+      <Table.Row>
+        <Table.HeaderCell>{header}</Table.HeaderCell>
+        {incidents.length > 0 ? <Table.HeaderCell>Resolved</Table.HeaderCell> : null}
```
kchu1337: Can use && here

src/pages/blacklist/components/tests/IncidentDetails.test.js
```diff
@@ -0,0 +1,64 @@
...
+          id: 'abc',
+          occurred_at_formatted: 'Jan 1 2019, 5:35pm',
+          resolved_at_formatted: 'Jan 4 2019, 5:35pm',
+        },
+      ],
+    });
+    const anchor = queryByText('Listed Jan 1 2019, 5:35pm | Resolved Jan 4 2019, 5:35pm');
+    expect(anchor).toBeInTheDocument();
+    expect(anchor.getAttribute('href')).toBe('/blacklist/incidents/abc');
+  });
```
kchu1337: These tests don't check that the related incidents stuff is there. IDK the best way to test that though. Maybe either a happy path queryByText or a shallow render and assert that the 2 components are there?

src/selectors/tests/blacklist.test.js
```diff
@@ -1,9 +1,15 @@
 import * as blacklistSelectors from '../blacklist';
+const RealDate = Date;
 
 describe('Blacklist Selectors: ', () => {
   let incidents, monitors;
   let formattedIncidents;
-
```
kchu1337: Can you also add tests for your new selectors?

src/pages/blacklist/IncidentsPage.js
gfriedmansp: Can't do that because it will display a `0` as content

src/pages/blacklist/components/tests/IncidentDetails.test.js
kchu1337: Nvm I think I was looking at the wrong component.

src/pages/blacklist/tests/IncidentDetailsPage.test.js
```diff
@@ -0,0 +1,106 @@
...
+      loading: false,
+      incident: mockIncident,
+      incidentsForBlacklist: mockIncidentsForBlacklist,
+      incidentsForResource: mockIncidentsForResource,
+      historicalIncidents: mockHistoricalIncidents,
+    });
+
+    wrapper.update();
+
+    expect(wrapper).toMatchSnapshot();
```
kchu1337: Not sure if this test will work because I think SASS builder/loader/whatever randomly assigns classnames.

src/pages/blacklist/tests/__snapshots__/IncidentDetailsPage.test.js.snap
```diff
@@ -0,0 +1,322 @@
...
+        "content": "Blacklist Incidents",
+        "to": "/blacklist/incidents",
+      }
+    }
+    empty={Object {}}
+    title="Blacklist Incident | 1.2.3.4 | spamhaus.org - pbl"
+  >
+    <div>
+      <div
+        className="I3_I5"
```
kchu1337: For example, this might not be the same every build.

src/pages/blacklist/components/RelatedIncidents.js
```diff
@@ -0,0 +1,39 @@
...
+                  {type === 'blacklist' ? incident.resource : incident.blacklist_name}
+                </strong>
+              </PageLink>
+              <div>
+                <AccessTime className={styles.TimeIcon} />
+                <span>{`Listed ${incident.occurred_at_formatted}`}</span>
+              </div>
+            </div>
+          </Table.Cell>
+          <Table.Cell>
```
kchu1337: Might want to set a fixed width. When you make the screen slimmer, the 2 columns aren't in line horizontally.

package-lock.json
```diff
@@ -22288,28 +22288,28 @@
       "dependencies": {
         "abbrev": {
           "version": "1.1.1",
-          "resolved": false,
```
kchu1337: Do you know why this happened?

src/pages/blacklist/IncidentDetailsPage.js
```diff
@@ -0,0 +1,182 @@
...
+              historicalIncidents={historicalIncidents}
+            />
+          )}
+        </Panel>
+
+        <Grid>
+          <Grid.Column lg={6} xs={12}>
+            {incidentsForBlacklistPending ? (
+              <PanelLoading />
+            ) : incidentsForBlacklist.length ? (
```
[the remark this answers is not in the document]
kchu1337: This is kinda a personal preference so if you don't want to make any changes, that's fine too.

package-lock.json
gfriedmansp: To be honest, no idea. Maybe I have a different npm version?

src/pages/blacklist/IncidentDetailsPage.js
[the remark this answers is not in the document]
gfriedmansp: Reasonable preference :D

PR #1347, opened 2020-01-07
TR-2099 - Add UI support for new RV "Typo" and "Neutral" results

nicklemmon:
[TR-2099](https://jira.int.messagesystems.com/browse/TR-2099)

### What Changed
- Some awesome Prettier reformatting
- Fixed issue with custom `.toHaveTextContent` matcher that was not correctly passing the selected HTML element to the React Testing Library matcher
- Cleaned up a few things in the `<SingleResult/>` component including moving arrow functions to the bottom of the file (we don't tend to do this but I think it might be worth thinking about? All praise our JS Lord and Savior for the idea: https://kentcdodds.com/blog/newspaper-code-structure)
- Renames several components to improve clarity
- Refactors several class methods that were being used to render components with...components!
- Adds new results descriptions outlined in the Jira ticket
- Updated component styles in an attempt to avoid one-offs, leveraging design tokens instead
- Refactors `<SingleResult/>` from a Class to a functional component
- Refactors `<SingleResult/>` tests to use React Testing Library instead of Enzyme
- Removes icons and related components to match new mocks

### How To Test

**NOTE:** The API updates have not been deployed, so testing the new validation types will require a little manual effort. Once those updates are deployed to staging, we'll re-verify things work as expected.

- Go to `/recipient-validation/single/hello@me.com` to verify the risky state
- Go to `/recipient-validation/single/nick@sparkpost.com` to verify the valid state
- Go to `/recipient-validation/single/hello@m` to verify the missing result state - should render as "Undeliverable" with a `"reason"` in the code snippet
- On line 37, modify the `calculatedResult` variable to `'neutral'` to see the neutral state
- On line 37, modify the `calculatedResult` variable to `'typo'` to see the typo state

### To Do
- [x] Incorporate mocks from UX that should be available on 1/8/20
- [x] Incorporate feedback 
- [X] Should I get rid of the class component?


src/pages/recipientValidation/constants/index.js
```diff
@@ -1,20 +1,27 @@
-export const ROLE_TOOLTIP = 'Role email addresses represent a group or activity (sales, abuse) \
+export const ROLE_TOOLTIP =
+  'Role email addresses represent a group or activity (sales, abuse) \
```
bkemper: These could be string literals.

PR #1348, opened 2020-01-08
TR-2016 - Handle non content-related preview template errors

nicklemmon:
[TR-2016](https://jira.int.messagesystems.com/browse/TR-2016)

### What Changed
- Adds multiple checks for the error object structure before rendering specific error message content
- Adds supporting unit tests (refactored from Enzyme to React Testing Library 🐐 )
- Removes snapshot that is no longer in use

#### Syntax Error (custom error code 3000 from https://www.sparkpost.com/docs/tech-resources/extended-error-codes/)
- Go to `/templates` and create a new template
- In the editor, cause a syntax error by typing `{{hello` in the HTML tab. You should see the following in the preview panel:

Screen Shot 2020-01-08 at 10 51 05 AM

- The same result should occur in the Text, and AMP HTML tabs

#### Other Errors (any error code that is not 3000 from https://www.sparkpost.com/docs/tech-resources/extended-error-codes/)
- In a new or existing template, try to write a template `{{each}}` loop for a variable that is not an array:

```html
{{each fruit}}
  {{fruit}}
{{end}}
```

```json
{
  "substitution_data": {
    "fruit": "string"
  },
  "metadata": {},
  "options": {}
}
```

- The same result should occur in the Text, and AMP HTML tabs

- You should see the following error message returned from the server rendered in the preview panel:

Screen Shot 2020-01-08 at 11 23 43 AM

#### 413 Error (no error message is returned by the endpoint)
- In a new template or in the one you created previously
- Paste a HUGE amount of HTML in the HTML tab. **Let me know and I can provide a link to a huge CSV file you can paste in**. You should see the following in the preview panel (an no error boundary):

Screen Shot 2020-01-08 at 10 54 48 AM

- The same result should occur in the Text, and AMP HTML tabs

### To Do
- [x] Add testing steps for other errors
- [ ] Incorporate feedback


src/pages/templates/components/PreviewErrorFrame.js
```diff
@@ -3,17 +3,39 @@ import { Warning } from '@sparkpost/matchbox-icons';
...
-    <p>We are unable to load your template preview due to a {errors[0].message} on line {errors[0].line} of your {errors[0].part}.</p>
-    <p>
-      If you notice this happens often, check your substitution data or code syntax as
-      these are frequent causes of preview errors.
-    </p>
-  </div>
-);
+const PreviewErrorFrame = ({ errors }) => {
+  // instability of the data leads to some funky logic - checking that the errors array exists, and defaulting to an empty object if either the first index or the array doesn't exist
+  const error = errors ? errors[0] || {} : {};
```
bkemper: Could errors === null?

nicklemmon: I don't think so - but `null` is still falsy so this should still work, no? I could add another test to be safe. What do you think?

src/pages/templates/components/PreviewErrorFrame.js
```diff
@@ -3,17 +3,39 @@ import { Warning } from '@sparkpost/matchbox-icons';
...
+      {code === '3000' ? (
+        <>
+          {/* Defensively checking the error object's structure */}
+          {['message', 'line', 'part'].every(key => key in error) && (
+            <p>
+              We are unable to load your template preview due to a {message} on line {line} of your{' '}
+              {part}.
+            </p>
+          )}
+        </>
```
bkemper: I don't think the fragments are needed.

src/pages/templates/components/PreviewErrorFrame.js
```diff
@@ -3,17 +3,39 @@ import { Warning } from '@sparkpost/matchbox-icons';
...
+      ) : (
+        <>{error.hasOwnProperty('description') && <p>{description}</p>}</>
+      )}
+
+      <p>
+        If you notice this happens often, check your substitution data or code syntax as these are
+        frequent causes of preview errors.
+      </p>
```
bkemper: I think this message only applies in the first condition when we have a line number.  Do we need it for both?

nicklemmon: wellll, if you follow the directions for example #2, the error is because of bad substitution data as well, it's just a less common scenario. That being said, it clearly doesn't apply to the `413` response. We could just only render it in the first scenario to be extra safe, or we leave it in there to always offer some sort of general advice about where preview errors tend to come from.

src/pages/templates/components/PreviewErrorFrame.js
nicklemmon: I'm getting syntax errors without them. You seeing something different?

PR #1353, opened 2020-01-09
AC-1158 - Test new billing page (Fixes)

shrirupa:
AC-1158 - Test new billing page (Fixes)

### What Changed
1. Changing the wordings - “Your new plan doesn’t include dedicated IPs. Your current IP(s) will be removed at the end of your current billing cycle.”
2. Adding the filtering for secret bundles in PlanSelect
3. users were still able to "enable tfa required " after selecting a starter plan. Solved by - [an additional call to {{/billing/subscription/check}} after the form has been submitted when a user upgrade from free to paid (along with the call to {{/account/subscription/check}})]

### How To Test
 - enable ui flag - "account_feature_limits"
 - to test the first change downgrade from premier plan to starter/free and check the wording in Changes to Features Section
- to test the second change make sure 2.5M plan gets selected when you click on this url http://localhost:3001/account/billing/plan?code=2.5M-0817 and also select plan section you do not see the 2.5M plan(or any secret plan returned by the bundle endpoint).
- to test the third change upgrade to a starter plan and ensure that you are not able to turn on tfa_required toggle and ui throws an error if you try to enable it.

### To Do
- [ ] Address Feedback

### DIFF
--- a/src/__func__/billing/__snapshots__/changePlanFirstTime.test.js.snap
+++ b/src/__func__/billing/__snapshots__/changePlanFirstTime.test.js.snap
@@ -148,6 +148,18 @@ Array [
       "headers": Object {
         "Authorization": "mock-access-token",
       },
+      "method": "post",
+      "params": undefined,
+      "responseType": "json",
+      "url": "/v1/billing/subscription/check",
+    },
+  ],
+  Array [
+    Object {
+      "data": undefined,
+      "headers": Object {
+        "Authorization": "mock-access-token",
+      },
       "method": "get",
       "params": Object {
         "include": "usage",
--- a/src/actions/billing.js
+++ b/src/actions/billing.js
@@ -63,6 +63,17 @@ export function syncSubscription({ meta = {} } = {}) {
   });
 }
 
+export function syncBillingSubscription({ meta = {} } = {}) {
+  return sparkpostApiRequest({
+    type: 'SYNC_BILLING_SUBSCRIPTION',
+    meta: {
+      method: 'POST',
+      url: '/v1/billing/subscription/check',
+      ...meta,
+    },
+  });
+}
+
 /**
  * attempts to collect payments (like when payment method is updated) to make sure pending payments are charged
  */
--- a/src/pages/billing/components/PlanSelect.js
+++ b/src/pages/billing/components/PlanSelect.js
@@ -17,10 +17,9 @@ export const useModal = () => {
   }
   return {
     isShowing,
-    toggle
+    toggle,
   };
 };
-
 
 export const SelectedPlan = ({ bundle, onChange, promoCodeObj, handlePromoCode }) => {
   const { messaging: plan, tier } = bundle;
@@ -33,10 +32,14 @@ export const SelectedPlan = ({ bundle, onChange, promoCodeObj, handlePromoCode }
       title="Your New Plan"
       actions={[
         {
-          content: (<span>Compare Features <ViewModule /></span>),
+          content: (
+            <span>
+              Compare Features <ViewModule />
+            </span>
+          ),
           onClick: toggle,
-          color: 'orange'
-        }
+          color: 'orange',
+        },
       ]}
     >
       <FeatureComparisonModal open={isShowing} handleClose={toggle} />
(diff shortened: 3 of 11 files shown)


src/pages/billing/context/FeatureChangeContext.js
```diff
@@ -69,17 +72,11 @@ export const FeatureChangeProvider = ({
...
-                    Your new plan doesn't include dedicated IPs.
-                    Your current IPs will now be billed at $20/month each, or you can
-                  </span>
-                  <SupportTicketLink issueId="general_issue">
-                     &nbsp;submit a ticket&nbsp;
-                  </SupportTicketLink>
-                  <span>
-                     to delete them.
+                    {`Your new plan doesn't include dedicated IPs. 
+                    Your current IP(s) will be removed at the end of your current billing cycle.`}
```
bkemper: The string be able to be the children of the span.

bkemper:
> to test the third change upgrade to a starter plan and ensure that you are not able to turn on tfa_required toggle and ui throws an error if you try to enable it.

If you navigate to the account setting page after the upgrade, the toggle is still on, but if you refresh the page it is off.  So, the update is happening correctly, but the current state of our application is not updated.

bkemper:
After thinking about it, I think the correct user experience is to not update the toggle, but to hide it all together since the user on a starter plan does not provide account wide TFA enforcement.  In addition, I would hide the SSO section.

src/__func__/billing/__snapshots__/changePlanFirstTime.test.js.snap
```diff
@@ -142,6 +142,18 @@ Array [
...
+  Array [
+    Object {
+      "data": undefined,
+      "headers": Object {
+        "Authorization": "mock-access-token",
+      },
+      "method": "post",
+      "params": undefined,
+      "responseType": "json",
+      "url": "/v1/billing/subscription/check",
```
bkemper: Something feels off about this endpoint.  We have to manually request a sync between our API and Zuora?

src/actions/billing.js
```diff
@@ -63,6 +63,17 @@ export function syncSubscription({ meta = {} } = {}) {
   });
 }
 
+export function syncBillingSubscription({ meta = {} } = {}) {
+  return sparkpostApiRequest({
+    type: 'SYNC_BILLING_SUBSCRIPTION',
+    meta: {
+      method: 'POST',
+      url: '/v1/billing/subscription/check',
+      ...meta,
```
bkemper: We shouldn't spread, we should clearly define what we are expecting to be passed.  I think the only thing we need is `onSuccess`.  Right?

src/pages/billing/components/PlanSelect.js
```diff
@@ -45,83 +48,96 @@ export const SelectedPlan = ({ bundle, onChange, promoCodeObj, handlePromoCode }
...
 const PlanSelectSection = ({ bundles, currentPlan, onSelect }) => {
-
-  const bundlesByTier = useMemo(() =>
-    _.groupBy(bundles, 'tier')
-  , [bundles]);
-
+  const publicBundlesByTier = useMemo(
+    () =>
+      _.groupBy(
+        bundles.filter(x => x.status !== 'secret'),
```
bkemper: Could bundles ever be undefined?

src/pages/billing/context/FeatureChangeContext.js
```diff
@@ -16,19 +15,23 @@ export const FeatureChangeProvider = ({
...
+    getSubscription();
+  }, [getSubscription]);
 
   //Rechecks conditions on re-entering tab. Only initializes once
-  const checkConditions = useCallback(() => { getSubscription(); }, [getSubscription]);
+  const checkConditions = useCallback(() => {
+    getSubscription();
+  }, [getSubscription]);
   useEffect(() => {
     window.addEventListener('focus', checkConditions);
```
bkemper: This is completely unrelated to this PR, but why do we need to getSubscription with the window is focused?

src/pages/billing/components/PlanSelect.js
bkemper: I just checked.  The default value is empty array.

src/actions/billing.js
bkemper: Don't worry about this.

PR #1354, opened 2020-01-09
FE-786 - Add Cypress integration tests for Recipient Validation (RV)

nicklemmon:
[FE-786](https://jira.int.messagesystems.com/browse/FE-786)

### What Changed
- Adds `cy.stubRequest` command to help reduce the boilerplate required to intercept a network request - @bkemper (...you mentioned adding this last time and I disagreed then and since changed my mind 😅)
- Adds Cypress integration tests for RV features
- Adds a whole bunch of RV-related fixtures
- Adds `cypress-file-upload` package to help mock file uploading in a Cypress test
- Adds screen reader only form field label to the single recipient validation form both as an a11y and testing improvement (`/recipient-validation/single`)
- Some Prettier re-formatting that came along for the ride

### How To Test
1. Run through the RV feature and verify that the written tests cover common use-cases:
  - Try uploading a list of recipients to validate
  - Try validating a single email address
2. Verify that all tests are passing (Travis should do this for you)
3. Verify that test descriptions and assertions make sense and document expectations/requirements

### To Do
- [x] Add tests for the list validation confirmation page
- [x] Figure out a way to test polling (maybe using `cy.wait()`?) - **Decided against this** - the Jest tests are already testing this behavior, which tend to do a better job checking for network requests.
- [ ] Incorporate feedback


cypress/tests/integration/recipient-validation/list-page-routes/list-route.spec.js
```diff
@@ -0,0 +1,168 @@
...
+
+    cy.visit(ROUTE_URL);
+
+    const exampleCSV = 'recipient-validation/fake-list.csv';
+
+    cy.fixture(exampleCSV).then(fileContent => {
+      cy.get('[name="csv"]').upload({ fileContent, fileName: 'example.csv', mimeType: 'text/csv' });
+    });
+
+    cy.findAllByText('Something went wrong.').should('be.visible');
```
bkemper: How does Cypress know to wait for the fixture promise above before evaluating this assertion?

cypress/tests/integration/recipient-validation/list-page.spec.js
```diff
@@ -0,0 +1,46 @@
...
+
+  it('renders pricing information when clicking on the "See Pricing" button', () => {
+    cy.findByText('See Pricing').click();
+
+    cy.findByText('How was this calculated?').should('be.visible');
+  });
+
+  it('has a relevant page title', () => {
+    cy.visit('/recipient-validation/list');
+    cy.title().should('include', 'Recipient Validation');
```
bkemper: This is probably a bad idea, but should this be a single command?

```
Cypress.Commands.add('mySafeVisit', (path, title) => {
  cy.visit(path).then(() => {
    cy.title().should('include', title);
  });
});
```

cypress/tests/integration/recipient-validation/list-confirmation-page.spec.js
```diff
@@ -0,0 +1,79 @@
...
+    cy.stubRequest({
+      url: '/api/v1/recipient-validation/job/fake-list',
+      fixture: 'recipient-validation/job/fake-list/200.get.queued_for_batch.json',
+    });
+
+    cy.visit(PAGE_URL);
+
+    cy.findByText('fake-list.csv').should('be.visible');
+  });
+});
```
bkemper: Should we test the "How was this calculated?" button?

cypress/tests/integration/recipient-validation/list-page-routes/list-route.spec.js
nicklemmon: https://docs.cypress.io/api/commands/then.html#Promises Looks like it just does! Basically when a Promise is in a test, it won't proceed until it resolves. With a fixture, since it's just a local file it should resolve almost immediately.

You could also use then with `cy.request()` to make a real network request - that could cause an actual delay obviously.

cypress/tests/integration/recipient-validation/list-page.spec.js
nicklemmon: Maybe? There are some times you want to check for an exact match too. Mind I think about it while we scale up the tests?

cypress/tests/integration/recipient-validation/list-confirmation-page.spec.js
nicklemmon: Done.

nicklemmon: I just realized I've been naming all of these files in a way that is not consistent with our standard. I'm going to go ahead and rename them.

cypress/.eslintrc
```diff
@@ -1,6 +1,7 @@
 {
   "rules": {
     "no-undef": "off",
-    "jest/expect-expect": "off"
+    "jest/expect-expect": "off",
+    "jest/valid-expect-in-promise": "off"
```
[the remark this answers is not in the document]
nicklemmon: So the reason is this:
https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/valid-expect-in-promise.md

Basically, the rule is checking that a valid assertion exists inside the `.then() =>` block. With Cypress, sometimes all you do within the `.then() =>` or the `catch() =>` is another browser interaction instead of an assertion. That was the case with the file upload.

I'll add a clarifying comment.

cypress/tests/integration/recipient-validation/list-page.spec.js
bkemper: Of course.

PR #1359, opened 2020-01-13
FE-784: Add Cypress tests for the downgrade flow

shrirupa:
FE-784: Add Cypress tests for the downgrade flow

### What Changed 
- Added cypress tests for downgrading plan 
- Added a few tests for billing page related to downgrading plan.

### How To Test
- run the cypress tests locally

### To Do
- [ ] Address Feedback


cypress/tests/integration/accounts/change-billingplan-page.spec.js
```diff
@@ -0,0 +1,120 @@
...
+    });
+    cy.visit('/account/billing/plan');
+    cy.get('[data-id=select-plan-50K-starter-0519]').click();
+    cy.findAllByText('Got it')
+      .first()
+      .click();
+    cy.findAllByText('Got it')
+      .last()
+      .click();
+  });
```
kchu1337: Might be missing your assertion here

cypress/tests/integration/accounts/change-billingplan-page.spec.js
```diff
@@ -0,0 +1,120 @@
...
+
+  //downgrading, starter  => free plan
+  it('Change plan button is displayed only when all the "features have been updated" ', () => {
+    selectAFreePlan();
+    cy.get('body').then($body => {
+      if ($body.text().includes('features have been updated')) {
+        cy.get('a[type=button]')
+          .contains('Change Plan')
+          .should('be.visible');
```
kchu1337: I am not seeing this assertion in the cypress console.

src/actions/brightback.js
```diff
@@ -4,23 +4,30 @@ import { getRelativeDates } from 'src/helpers/date';
 import { safeRate } from 'src/helpers/math';
```
kchu1337: Didn't see that anything changed besides prettier. Let me know if anything was changed

cypress/tests/integration/accounts/billing-page.spec.js
```diff
@@ -0,0 +1,106 @@
...
+      fixture: 'account/200.get.include-pending-subscription.json',
+      fixtureName: 'accountPendingDowngradeGet',
+    });
+    stubRequest({
+      url: '/api/v1/billing/subscription',
+      fixture: 'billing/subscription/200.get.include-pending-downgrades.json',
+      fixtureName: 'subscriptionPendingDowngradeGet',
+    });
+    cy.visit('/account/billing');
+    cy.findByText('Pending Plan Change');
```
kchu1337: Is this enough? Don't you need to assert that it exists?

cypress/tests/integration/accounts/billing-page.spec.js
```diff
@@ -0,0 +1,106 @@
...
+    stubRequest({
+      url: '/api/v1/billing/subscription',
+      fixture: 'billing/subscription/200.get.json',
+      fixtureName: 'subscriptionGet',
+    });
+    cy.visit('/account/billing');
+    cy.contains('Change Plan').click();
+    cy.title().should('include', 'Change My Plan');
+  });
+});
```
kchu1337: Idk if this is important, but the snapshots at the end of each test only show the loading state.

cypress/tests/integration/accounts/billing-page.spec.js
shrirupa: you are right! Although the test does fail if it can't find the text

src/actions/brightback.js
shrirupa: Only prettier changes.

cypress/tests/integration/accounts/change-billingplan-page.spec.js
shrirupa: Thanks 👍

cypress/tests/integration/accounts/change-billingplan-page.spec.js
shrirupa: Thanks! Fixed it.

cypress/tests/integration/accounts/billing-page.spec.js
shrirupa: Not sure about this.

cypress/tests/integration/accounts/change-billingplan-page.spec.js
```diff
@@ -116,5 +120,6 @@ describe('Billing Page', () => {
     cy.findAllByText('Got it')
       .last()
```
kchu1337: Could just do `.click({multiple: true});`. Although since, you are stubbing the plans, it probably doesn't matter.

cypress/tests/integration/accounts/change-billingplan-page.spec.js
```diff
@@ -80,6 +83,7 @@ describe('Billing Page', () => {
           .contains('Change Plan')
           .should('be.visible');
       }
+      cy.findAllByText('Change Plan').should('not.be.visible');
```
kchu1337: It's not super clear to me based on the description what exactly this test does. Especially for someone not too familiar with changing plan. It might make it easier to understand if you either add an assertion that the `Got It` button still exists and that's why the plan change button isn't there. Or rename the test to something like `'Change plan button is displayed only when the user acknowledges all of the updated features`

cypress/tests/integration/accounts/billing-page.spec.js
```diff
@@ -0,0 +1,106 @@
+/// <reference types="Cypress" />
+
+function stubRequest({ method = 'GET', statusCode = 200, url, fixture, fixtureName }) {
```
kchu1337: Looks like Nick already wrote a helper in `cypress/support/commands.js` line ~124

cypress/tests/integration/accounts/billing-page.spec.js
```diff
@@ -0,0 +1,106 @@
...
+    response: `@${fixtureName}`,
+  });
+}
+
+describe('Billing Page', () => {
+  beforeEach(() => {
+    cy.stubAuth();
+    cy.login({ isStubbed: true });
+
+    cy.server();
```
kchu1337: Could use the helper instead.

cypress/tests/integration/accounts/change-billingplan-page.spec.js
```diff
@@ -0,0 +1,112 @@
+/// <reference types="Cypress" />
+
+function stubRequest({ method = 'GET', statusCode = 200, url, fixture, fixtureName }) {
```
kchu1337: Same here with existing helper function

cypress/tests/integration/accounts/change-billingplan-page.spec.js
```diff
@@ -0,0 +1,112 @@
...
+    response: `@${fixtureName}`,
+  });
+}
+
+describe('Billing Page', () => {
+  beforeEach(() => {
+    cy.stubAuth();
+    cy.login({ isStubbed: true });
+
+    cy.server();
```
kchu1337: Same here with using the helper

shrirupa: Thanks for all the above suggestions @kchu1337 👍

PR #1360, opened 2020-01-13
SA-1162: Fix days listed for resolved incidents

kchu1337:
### What Changed
 - Fixed days listed for resolved date

### How To Test
 - Go to blacklist/incidents
 - Click on any incident.
 - Verify that the days listed is correct.

### DIFF
--- a/src/selectors/blacklist.js
+++ b/src/selectors/blacklist.js
@@ -22,10 +22,8 @@ const getHistoricalIncidentsError = state => state.blacklist.historicalIncidents
 const getIncidentsForBlacklistError = state => state.blacklist.incidentsForBlacklistError || false;
 
 const getDaysListed = (resolvedDate, occurredDate) => {
-  const daysListed = resolvedDate
-    ? moment(resolvedDate).diff(moment(occurredDate), 'days')
-    : moment().diff(moment(occurredDate), 'days');
-  return daysListed || 1;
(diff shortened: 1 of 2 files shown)


src/selectors/blacklist.js
```diff
@@ -21,13 +21,20 @@ const getHistoricalIncidentsError = state => state.blacklist.historicalIncidents
 
 const getIncidentsForBlacklistError = state => state.blacklist.incidentsForBlacklistError || false;
 
+const getDaysListed = (resolvedDate, occurredDate) => {
+  const daysListed = resolvedDate
+    ? moment(resolvedDate).diff(moment(occurredDate), 'days')
+    : moment().diff(moment(occurredDate), 'days');
+  return daysListed || 1;
+};
```
bkemper: This is a nit, but it might be easier to read if the date being compared is the variable.

```
const getDaysListed = (resolvedDate, occurredDate) => {
  const resolvedMoment = resolvedDate === null ? moment() : moment(resolvedDate);
  return resolvedMoment.diff(moment(occurredDate), 'days') || 1;
};
```

bkemper: Also, I think we need to round up instead of setting a default.
```
Math.ceil(resolvedMoment.diff(moment(occurredDate), 'days', true))
```

PR #1364, opened 2020-01-15
SA-1153 Link Watchlist Page to Incidents Page with Filter

gfriedmansp:
### What Changed
 - Updated link on resources on the Watchlist page
 - Allow search value to come from URL on incidents page and update URL on blur
 - Fill search box with the resource from the URL

### How To Test
 - Click on an item in the watchlist, ensure this goes to incidents page, filtered by that resource
 - Type in the search box, click outside (blur) and ensure URL is updated with the new search value


src/components/collection/FilterBox.js
```diff
@@ -1,12 +1,21 @@
...
 import { getRandomExampleSearch } from './helpers/exampleSearch';
 import styles from './FilterBox.module.scss';
 
 export default function CollectionFilterBox(props) {
-  const { placeholder, wrapper, onChange } = props;
+  const { placeholder, wrapper, onChange, onBlur, defaultValue } = props;
   const placeholderText = placeholder || `Filter results e.g. ${getRandomExampleSearch(props)}`;
+
+  // If there is a defaultValue for filter box, call onChange with defaultValue
+  useEffect(() => {
```
gfriedmansp: I don't like the existence of this. If you can think of a better place/way to do this, I'm all ears. The goal is to set the value of the TextField below on mount based on a defaultValue, however it seems we need to keep the component uncontrolled, so we can't just set the value prop.

src/pages/blacklist/IncidentsPage.js
```diff
@@ -9,10 +9,15 @@ import { Users } from 'src/components/images';
 import { listMonitors, listIncidents } from 'src/actions/blacklist';
 import { selectIncidentsList } from 'src/selectors/blacklist';
 import { getRelativeDates } from 'src/helpers/date';
+import usePageFilters from 'src/hooks/usePageFilters';
 import IncidentsCollection from './components/IncidentsCollection';
 import styles from './IncidentsPage.module.scss';
 import CongratsBanner from './components/CongratsBanner';
 
+const filterWhitelist = {
+  search: {},
```
gfriedmansp: Since this is a just an arbitrary string parameter, we don't need a validate function and we don't need a default value

src/actions/blacklist.js
```diff
@@ -91,7 +90,6 @@ export function listIncidentsForBlacklist(blacklist, from = '2019-01-01', limit
       showErrorAlert: false,
       params: {
         blacklists: blacklist,
-        limit,
```
nicklemmon: Mind explaining why this was removed?

src/components/collection/FilterBox.js
```diff
@@ -19,6 +28,8 @@ export default function CollectionFilterBox(props) {
         suffix={<Search />}
         placeholder={placeholderText}
         onChange={e => onChange(e.target.value)}
+        onBlur={e => onBlur && onBlur(e.target.value)}
```
nicklemmon: I didn't think methods needed to be passed this way for events. Does the following not work?

```js
...
onChange={onChange}
onBlur={onBlur}
...
```

nicklemmon: Looks like no...the filtering hooks don't seem to play nicely with that implementation.

The only reason I bring it up, is React should handle this situation such that you don't have to check for the prop `onBlur` before invoking it.

src/pages/blacklist/components/tests/IncidentsCollection.test.js
```diff
@@ -44,4 +44,24 @@ describe('Blacklist Component: IncidentsCollection', () => {
...
+      updateTextField: mockUpdateTextField,
+    });
+    const target = queryByDisplayValue('test-search');
+    fireEvent.change(target, {
+      target: { value: 'new-test-search' },
+    });
+
+    fireEvent.blur(target);
+    expect(mockUpdateTextField).toBeCalledWith('new-test-search');
+  });
```
nicklemmon: So much better than Enzyme!

src/pages/blacklist/components/IncidentsCollection.js
```diff
@@ -100,7 +93,14 @@ export const IncidentsCollection = props => {
...
       pagination={true}
-      filterBox={filterBoxConfig}
+      filterBox={{
+        show: true,
+        exampleModifiers: ['resource', 'blacklist_name'],
+        itemToStringKeys: ['resource', 'blacklist_name'],
+        wrapper: props => <div className={styles.FilterBox}>{props}</div>,
+        defaultValue: search,
+        onBlur: value => updateTextField(value),
+      }}
```
nicklemmon: A little bit of jankiness when visiting a URL with a query string string - looks like all of the data loads, and is then filtered. If it isn't a *huge* amount of work, applying the filter before the table rows are rendered would be ideal. If you're out of runway for this kind of a thing, I understand.

https://drive.google.com/open?id=1Q9TzvEd-KBszgjxD6zAxbUav9X8a-fFA

src/selectors/blacklist.js
```diff
@@ -45,7 +45,11 @@ export const selectRelatedIncidentsForResource = createSelector(
   [getIncidentsForResource, getIncident],
   (incidents, currentIncident) =>
     incidents
-      .filter(incident => incident.id !== currentIncident.id)
+      .filter(
+        incident =>
+          incident.id !== currentIncident.id &&
+          incident.blacklist_name !== currentIncident.blacklist_name,
```
nicklemmon: Also, mind explaining this change? Not sure how these selector updates play in to the current ticket.

src/components/collection/FilterBox.js
nicklemmon: I can't think of a way off the top of my head if the goal is to keep the component uncontrolled

src/components/collection/FilterBox.js
bkemper: We could probably simply this a little, just by setting a default prop value.

```
const noop = () => {};
const { placeholder, wrapper, onChange, onBlur = noop, defaultValue } = props;
```

nicklemmon: At that point, it's just a matter of which is more clear! Totally subjective either way.

src/actions/blacklist.js
bkemper: This is related to the selector change.  There isn't a way to exclude the current incident, so we request more than 4 then in the selector filter out all that are not needed.

src/selectors/blacklist.js
bkemper: See — https://github.com/SparkPost/sparkpost-admin-api-documentation/blob/master/services/blacklist-monitoring-api.apib#L147

PR #1370, opened 2020-01-16
TR-2140 - Update HTTP status code and content for usage limit exceeded errors for Recipient Validation (RV)

nicklemmon:
[TR-2140](https://jira.int.messagesystems.com/browse/TR-2140)

### What Changed
- Updates check for HTTP status `420` vs `400`
- Updates content to `"Validation limit exceeded"` on both the single validation and list progress views
- Updates the details of the relevant alerts with a link to open the support modal in the app rather than the sales site

### How To Test
- Run Cypress integration tests locally and verify the UI renders as expected in the test cases

### To Do
- [ ] Incorporate feedback


src/pages/recipientValidation/SingleResult.js
```diff
@@ -25,18 +25,25 @@ export function SingleResult(props) {
   useEffect(() => {
     singleAddress(address).catch(({ response = {}, message }) => {
       const { status } = response;
+      console.log('status', status);
+      console.log('message', message);
```
shrirupa: we can remove these now

nicklemmon: We need to re-introduce that ESLint rule!

shrirupa: yes, I agree!

nicklemmon: Cut this ticket to address:
https://jira.int.messagesystems.com/browse/FE-837

PR #1374, opened 2020-01-21
FE-727 Add useABTest hook

gfriedmansp:
### What Changed
 - Added a hook to get a specific variant of an experiment AB Test based on UID

### How To Test
 - Ensure tests are passing and are good enough coverage

### To Do
- Determine if we should import the redux store in the hook so that we don't need to pass the customer_id from every use of this hook.

### DIFF
--- a/src/hooks/useABTest/hash.js
+++ b/src/hooks/useABTest/hash.js
@@ -1,3 +1,5 @@
+// Based on Java's hashcode function
+// https://werxltd.com/wp/2010/05/13/javascript-implementation-of-javas-string-hashcode-method/
 export const hash = str => {
   let hash = 0;
   if (str.length === 0) {
--- a/src/hooks/useABTest/hash.js
+++ b/src/hooks/useABTest/hash.js
@@ -1,0 +1,13 @@
+export const hash = str => {
+  let hash = 0;
+  if (str.length === 0) {
+    return hash;
+  }
+
+  for (let i = 0, l = str.length; i < l; i++) {
+    const char = str.charCodeAt(i);
+    hash = (hash << 5) - hash + char;
+    hash |= 0;
+  }
+  return Math.abs(hash);
+};
--- a/src/hooks/useABTest/hash.test.js
+++ b/src/hooks/useABTest/hash.test.js
@@ -1,0 +1,22 @@
+import { hash } from './hash';
+
+describe('hash', () => {
+  it('gets the same value on multiple calls to hash', () => {
+    expect(hash('abc123')).toEqual(hash('abc123'));
+  });
+
+  it('gets the different values for different string values', () => {
+    const cids = ['1', '2', '3', '4', '5', '6', '7', '8', '9', '10'];
+    for (let i = 0; i < cids.length; i++) {
+      for (let j = 0; j < cids.length; j++) {
+        if (i !== j) {
+          expect(hash(cids[i])).not.toEqual(hash(cids[j]));
+        }
+      }
+    }
+  });
+
+  it('gets 0 hash for empty string', () => {
+    expect(hash('')).toEqual(0);
+  });
+});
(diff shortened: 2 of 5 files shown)


src/hooks/useABTest/hash.js
```diff
@@ -0,0 +1,13 @@
+export const hash = str => {
```
nicklemmon: You grabbed this from here looks like? https://werxltd.com/wp/2010/05/13/javascript-implementation-of-javas-string-hashcode-method/

Might be worth crediting with a comment? I don't know how you feel about doing that.

src/hooks/useABTest/hash.test.js
```diff
@@ -0,0 +1,22 @@
...
+
+  it('gets the different values for different string values', () => {
+    const cids = ['1', '2', '3', '4', '5', '6', '7', '8', '9', '10'];
+    for (let i = 0; i < cids.length; i++) {
+      for (let j = 0; j < cids.length; j++) {
+        if (i !== j) {
+          expect(hash(cids[i])).not.toEqual(hash(cids[j]));
+        }
+      }
+    }
```
nicklemmon: You could use https://github.com/atlassian/jest-in-case instead of the for loop I would think - might make it a little more readable than a nested for loop.

src/hooks/useABTest/useABTest.js
```diff
@@ -0,0 +1,27 @@
...
+const logToPendo = ({ testName, variantName }) => {
+  if (window.pendo) {
+    window.pendo.track(`${testName} | ${variantName}`);
+  }
+};
+
+function useABTest(options = {}) {
+  const { testName, variants = [], onVariantLoad = logToPendo, uid } = options;
+
+  const index = useMemo(() => hash(uid) % variants.length, [variants.length, uid]);
```
nicklemmon: Will this generate an even split between the variants? 

You're probably future-proofing a bit, but technically the a/c only reference splitting users in to two groups. Would obviously simplify the code:

> Need a flexible way to provide one of two user experiences per user

src/hooks/useABTest/useABTest.js
```diff
@@ -0,0 +1,27 @@
+import { useMemo, useEffect } from 'react';
```
nicklemmon: Really well written. Very clear. 👍

src/hooks/useABTest/useABTest.test.js
```diff
@@ -0,0 +1,71 @@
+import React from 'react';
```
nicklemmon: Really nice tests here. A great example of a good set of unit tests.

src/hooks/useABTest/useABTest.js
```diff
@@ -0,0 +1,27 @@
...
+import { hash } from './hash';
+
+const logToPendo = ({ testName, variantName }) => {
+  if (window.pendo) {
```
nicklemmon: Should we throw an error somewhere if no `testName` is supplied to the hook?

src/hooks/useABTest/hash.js
gfriedmansp: Oh fun - this was in Jon's example hook, didn't know it was from somewhere else :o

src/hooks/useABTest/useABTest.js
gfriedmansp: Assuming the uids are roughly linear (have about the same even and odd numbers) and we use 2 variants, it should split into two groups-ish

src/hooks/useABTest/useABTest.js
gfriedmansp: We can and probably should. Definitely a required parameter.

src/hooks/useABTest/hash.test.js
gfriedmansp: I took a look at this and it seemed like it was more trouble (and actually less readable) to use this than the nested loop. Pretty much would just need to write out all of the combinations (45 in this case) and put them in objects.

PR #1375, opened 2020-01-21
SE-485 Remove Year from Events Page Date

gfriedmansp:
### What Changed
 - Added a new config for `dateFormatWithoutYear`
 - Added new date formatter function `formatDateTimeWithoutYear`
 - Use this new formatter on the messageEvents selector (but only for the events list, not the CSV)

### How To Test
 - Open /reports/message-events
 - Move to the next page so dates show up
 - Ensure there is no year on the dates

### DIFF
--- a/src/pages/reports/messageEvents/MessageEventsPage.js
+++ b/src/pages/reports/messageEvents/MessageEventsPage.js
@@ -3,11 +3,23 @@ import React, { Component } from 'react';
 import { connect } from 'react-redux';
 import { snakeToFriendly } from 'src/helpers/string';
 import { Button, Page } from '@sparkpost/matchbox';
-import { PanelLoading, TableCollection, CursorPaging, PerPageButtons, ApiErrorBanner, Empty } from 'src/components';
+import {
+  PanelLoading,
+  TableCollection,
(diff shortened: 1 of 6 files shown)


src/pages/reports/messageEvents/MessageEventsPage.js
```diff
@@ -3,11 +3,23 @@ import React, { Component } from 'react';
 import { connect } from 'react-redux';
```
gfriedmansp: This page was _only_ prettier changes

src/selectors/messageEvents.js
```diff
@@ -1,85 +1,95 @@
...
+const appendFormattedDate = event => ({
   ...event,
-  formattedDate: formatDateTime(event.timestamp)
+  formattedDate: formatDateTime(event.timestamp),
 });
 
-export const selectMessageEvents = createSelector(
-  [ getMessageEvents ],
-  (events) => _.map(events, appendFormattedDate)
+const appendFormattedDateWithoutYear = event => ({
```
gfriedmansp: This is the only new code, rest is prettier

PR #1376, opened 2020-01-22
SA-1093: Prevent submitting date picker update with only 1 date selected

kchu1337:
### What Changed
 - Disabled Date Picker submit button when only selected one date.

### How To Test
 - Go to signals page. Open date picker and select only 1 date. You should not be able to submit the change.
 - Check other date pickers to make sure nothing else breaks. Especially for inbox placement trends since it also uses validation

### Notes
 - This isn't quite the same as the AC but this was the most painless way to fix the bug.

- [ ] Get Product approval for this solution.

### DIFF
--- a/src/components/datePicker/DatePicker.js
+++ b/src/components/datePicker/DatePicker.js
@@ -1,7 +1,13 @@
 /* eslint-disable max-lines */
 import React, { Component } from 'react';
 import { subMonths, format } from 'date-fns';
-import { getStartOfDay, getEndOfDay, getRelativeDateOptions, getNextHour, isSameDate } from 'src/helpers/date';
+import {
+  getStartOfDay,
+  getEndOfDay,
+  getRelativeDateOptions,
+  getNextHour,
+  isSameDate,
+} from 'src/helpers/date';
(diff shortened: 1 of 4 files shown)


src/components/datePicker/DatePicker.js
```diff
@@ -223,14 +239,21 @@ export default class AppDatePicker extends Component {
           />
         )}
 
-        <Button primary onClick={this.handleSubmit} className={styles.Apply}>Apply</Button>
+        <Button
+          primary
+          onClick={this.handleSubmit}
+          disabled={this.state.selecting}
```
kchu1337: This is my change.

src/components/datePicker/tests/DatePicker.test.js
```diff
@@ -72,8 +74,17 @@ describe('Component: DatePicker', () => {
     expect(wrapper.find('ManualEntryForm')).not.toExist();
   });
 
-  describe('syncTimeToState', () => {
+  it('should disable submit button when selecting date', () => {
```
kchu1337: This is a test I added

PR #1377, opened 2020-01-22
AC-1124: Add Payment Functionality Recipient Validation - List Tab

shrirupa:
AC-1124: Add Payment Functionality Recipient Validation - List Tab

### What Changed
- Feature Flag to hide this new flow/page. (_standalone_rv_)
- Show the Add Credit Card Form with the validate button on List Tab.
- Once CC is added the details of the entered CC shows up along with validate button.
- If the user already has a CC on file then CC details show up along with validate button.
- You can use another CC if there a CC on file already by clicking on _Use Another Credit Card_ Button.

### How To Test
- Create an account and enable 'stanadlone_rv' flag.
- Navigate to /recipient-validation/list.
- Upload a list and see if it the works as described above.

### To Do
- [ ] Address Feedback


PatrickSison:
Currently some of the cypress tests are failling

src/pages/recipientValidation/RecipientValidationPage.js
```diff
@@ -103,7 +103,7 @@ export class RecipientValidationPage extends Component {
   };
 
   render() {
-    return (
+    return !this.props.isStandAloneRVSet ? (
```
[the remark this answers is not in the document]
shrirupa: yes, so after standalone rv goes out, all users should be able to preview the RV page based on the grant and RVDisabled page will no longer be needed.

shrirupa:
> Currently some of the cypress tests are failling

So it is weird as all cypress tests seem to be passing in `continuous-integration/travis-ci/push`.

shrirupa:
> Will still continue to go through and review more. As far as manually billed customers go, would we be stopping them at the initial page, or the uploaded list page? Because it might be confusing to upload a list and then get blocked later on.

🤔 I feel that it is similar experience for other users as they will be stopped till they enter Credit Card details, but worth asking UX about it.

src/constants/index.js
```diff
@@ -24,16 +24,10 @@ export const FORMATS = {
```
PatrickSison: A lot of prettier 😵

src/helpers/conditions/account.js
```diff
@@ -17,6 +17,7 @@ export const isBillingSubscriptionSelfServe = ({ subscription }) => {
   return _.includes(SELF_SERVE_ACCOUNT_TYPES, _.get(subscription, 'type'));
 };
 export const isAws = ({ account }) => _.get(account, 'subscription.type') === 'aws';
+export const isManuallyBilled = ({ account }) => _.get(account, 'subscription.type') === 'manual';
```
PatrickSison: It's a bit in a weird state, but we have 3 (i think) conditions to check for manually billed. `account.subscription.self_serve !== true`, `account.subscription.type === 'manual'`, and `billing.subscription.type = 'manual'`

PatrickSison: I think previously we checked `!isSelfServeBilling` but we should probably establish one as the source of truth.

shrirupa:
> > Will still continue to go through and review more. As far as manually billed customers go, would we be stopping them at the initial page, or the uploaded list page? Because it might be confusing to upload a list and then get blocked later on.
> 
> 🤔 I feel that it is similar experience for other users as they will be stopped till they enter Credit Card details, but worth asking UX about it.

Confirmed with Haley and Aubrey, they think it is better to have the Contact Sales button after the list is uploaded to keep the experience consistent.

src/helpers/conditions/account.js
shrirupa: I think may be we can just use isSelfServeBilling to keep it consistent, I was worried earlier about any state is which a user is not self serve but can still use RV, I can't think of any such cases though, so I will use the isSelfServeBilling then.

src/pages/recipientValidation/UploadedListPage.js
```diff
@@ -68,6 +79,9 @@ export class UploadedListPage extends Component {
             )}
           </Panel.Section>
         </Panel>
+        {isStandAloneRVSet && (
+          <ValidateSection credit_card={credit_card} handleValidate={() => {}} />
```
PatrickSison: It was a little bit confusing on first reading the relation between the page, the UploadedListForm, and the ValidateSection. Should this be consolidated or reconstructed in some way? We have a top level check for the condition to render ValidateSection which has its own submit button, and then check the inverse within UploadedListForm to see if we render its own submit button .

If it's because of the differences in behavior of the button between versions, we can probably break out that top section of UploadedListForm (which contains the price and info) and then have the 2 separate bottom parts of the form.

shrirupa: So the check in UploadedListForm is temporary till we have this feature in development, and once standalone rv is out, we won't need to do that check and can get rid of that whole section completely. Then only the UploadedListPage will have the ValidateSection. We can may be later rename Uploaded List Form to something else because its technically not a Form. Did i understand it right ?

shrirupa:
> Looks good. As discussed, we will do a little bit more clean up on release ticket.

Thanks. Added a note in AC-1122

PR #1379, opened 2020-01-24
AC-1126: Add Payment Functionality Recipient Validation - Single Address Tab

shrirupa:
AC-1126: Add Payment Functionality Recipient Validation - Single Address Tab

### What Changed
- Feature Flag to hide this new flow/page.(_standalone_rv_)
- (UX-Change - Show the Add Credit Card Form with the validate button instead of Add Payment method button and the modal.)
- Once CC is added the details of the entered CC shows up along with validate button.
- If the user already has a CC on file then CC details show up along with validate button.

### How To Test
- Create an account and enable 'standalone_rv' flag.
- Navigate to /recipient-validation/single.
- Check if it works as described above.
### To Do
- [ ] Address Feedback

### DIFF
--- a/src/pages/recipientValidation/RecipientValidationPage.js
+++ b/src/pages/recipientValidation/RecipientValidationPage.js
@@ -74,7 +74,7 @@ export class RecipientValidationPage extends Component {
 
   renderRecipientValidation = () => {
     const { selectedTab, showPriceModal } = this.state;
-    const { isStandAloneRVSet, billing } = this.props;
+    const { isStandAloneRVSet, billing, billingLoading } = this.props;
 
     return (
       <Page
@@ -102,7 +102,7 @@ export class RecipientValidationPage extends Component {
 
         {selectedTab === 0 && <JobsTableCollection />}
 
-        {selectedTab === 1 && isStandAloneRVSet && (
+        {selectedTab === 1 && isStandAloneRVSet && !billingLoading && (
           <ValidateSection credit_card={billing.credit_card} handleValidate={() => {}} />
         )}
 
@@ -133,6 +133,7 @@ const mapStateToProps = (state, props) => ({
   tab: tabs.findIndex(({ key }) => key === props.match.params.category) || 0,
   isStandAloneRVSet: isAccountUiOptionSet('standalone_rv')(state),
   billing: state.account.billing || {},
+  billingLoading: state.account.billingLoading,
 });
 
 export default withRouter(connect(mapStateToProps, { getBillingInfo })(RecipientValidationPage));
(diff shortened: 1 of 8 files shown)


src/pages/recipientValidation/components/SingleAddressForm.js
```diff
@@ -40,17 +41,25 @@ export class SingleAddressForm extends Component {
...
             />
           </div>
-          <div className={styles.Submit}>
-            <Button size="large" fullWidth primary submit disabled={submitDisabled}>
-              {buttonContent}
-            </Button>
-          </div>
+          {!this.props.isStandAloneRVSet && (
+            <div className={styles.Submit}>
+              <Button size="large" fullWidth primary submit disabled={submitDisabled}>
```
PatrickSison: There's not as clean a way to do it now since they exist as separate forms, but the button within ValidateSection wouldn't get the same behavior of disabling as this one does

PatrickSison: This extends to a bigger question about how the forms are separated. I'm not sure what the best practice would be since the forms would technically be accomplishing separate things (Updating payment vs actually validating), but the validate wouldn't be able to happen without the payment. Do you think we should still keep these forms separate?

shrirupa: I think once we remove this feature flag we can make  let's make RecipientValidationPage to be a form. And then ValidateSection, all other tabs don't need to be their own forms. But I don't think we can make that transition now.

src/pages/recipientValidation/components/SingleAddressForm.js
```diff
@@ -7,12 +7,14 @@ import { TextFieldWrapper } from 'src/components';
...
 import { singleAddress } from 'src/actions/recipientValidation';
 import styles from './SingleAddressForm.module.scss';
+import { isAccountUiOptionSet } from 'src/helpers/conditions/account';
 
 const formName = 'singleAddressForm';
 export class SingleAddressForm extends Component {
   singleAddressForm = values =>
     this.props.history.push(`/recipient-validation/single/${values.address}`);
 
+  getClassName = className => (!this.props.isStandAloneRVSet ? className : className + 'SRV');
```
PatrickSison: Rather than having a function like this, you could apply multiple class names and then control it within CSS to make it cleaner: 

SCSS example: 
```scss
.MyClass {
  text-align: center;

  .conditionalClass {
     text-align: right;
  }
}
```

src/pages/recipientValidation/RecipientValidationPage.js
```diff
@@ -26,6 +28,10 @@ export class RecipientValidationPage extends Component {
     showPriceModal: false,
   };
 
+  componentDidMount() {
+    this.props.getBillingInfo();
```
PatrickSison: We can probably add small test coverage to just check if this runs on mounting.

src/pages/recipientValidation/RecipientValidationPage.js
```diff
@@ -95,6 +102,10 @@ export class RecipientValidationPage extends Component {
 
         {selectedTab === 0 && <JobsTableCollection />}
 
+        {selectedTab === 1 && isStandAloneRVSet && !billingLoading && (
```
PatrickSison: Up to you, but we could add a small unit test to check if billingLoading that this isn't rendered.

PR #1380, opened 2020-01-24
FE-791 - Add tests for navigation content based on role

nicklemmon:
[FE-791](https://jira.int.messagesystems.com/browse/FE-791)

### What Changed
- Adds Cypress tests for navigation content based user role and grants

### How to Test
- Verify that tests are checking for intended behavior and continue to pass

### To Do
- [ ] Incorporate feedback


cypress/support/commands.js
```diff
@@ -144,3 +143,17 @@ Cypress.Commands.add(
...
+ *
+ * @param {string} content - The content within the link - if the content is broken up by multiple DOM elements, this may not work well
+ * @param {string} href - The expected value of the `href` attribute present on the link
+ *
+ */
+Cypress.Commands.add('assertLink', ({ content, href }) => {
+  cy.queryByText(content)
+    .should('be.visible')
+    .closest('a')
+    .should('have.attr', 'href', href);
```
nicklemmon: So I do think testing for the href is important to make sure that if we make global component link changes the core behavior of the link is still working (meaning the prob value is properly propagating).

nicklemmon: This could be helpful when updating `react-router-dom`, for example.

PR #1381, opened 2020-01-24
SE-510: Blacklist UI updates to match the API

bkemper:
Refer to [SE-510](https://jira.int.messagesystems.com/browse/SE-510)

### What Changed
 - When creating or updating a blacklist alert the `filter_type` should now be "blacklist_provider" or "blacklist_resource".

### How To Test

 - Enable "blacklist_monitors" account option
 - Navigate to /alerts/create
 - Fill-out alert form and click "Create Alert"


gfriedmansp:
Just something I noticed, but could potentially be a new ticket - The UI allows you to hit the "Create Alert" button without selecting _any_ filters at all, which ends up sending an invalid request.

The docs also seem to say that you should send an empty array for `blacklist_provider` and `blacklist_resource` when they are not being used. I saw your comment on the PR, but wanted to flag it in case that doesn't get implemented in the way you expected.

PR #1385, opened 2020-01-27
FE-840 Add HTTP response status and redux action to Sentry logs

gfriedmansp:
### What Changed
 - Added tags: httpResponseStatus and reduxActionType to sentry reports on requests

### How To Test
 - Cause an error in the UI (Make up a template id and try to view it, etc.)
 - Look in sentry log for the event, check that the tags are there and are complete.
 - Ensure you can filter by that httpResponseStatus key in sentry


src/actions/helpers/tests/requestHelperFactory.test.js
```diff
@@ -1,7 +1,13 @@
 import factory from '../requestHelperFactory';
+import ErrorTracker from 'src/helpers/errorTracker';
 
-describe('Helper: Request Helper Factory', () => {
+jest.mock('src/helpers/errorTracker');
+
+ErrorTracker.addRequestContextAndThrow = jest.fn((type, response, err) => {
```
bkemper: This is just a style question, but should we declare variables outside of describe?

bkemper: Also, a maintenance question... I saw that we don't test that this function is called.  This is safe because our current tests confirmed that there isn't a regression and the throwing of the error on failure is still working as expected.  However, if one day we remove this function call, a test won't fail to alert us that we need to update the tests and remove the stub.  Should it or should we rely on future engineer and code reviewer to do this manual check?

src/actions/helpers/tests/zuoraRequest.test.js
```diff
@@ -1,42 +1,43 @@
...
-        success: false,
-        reasons: [{ message: 'The credit card is bad and wrong' }]
-      }
-    }));
-
-    try {
-      await mockStore.dispatch(zuoraRequest(action));
-    } catch (err) {
-      expect(err.response).toBeDefined();
-      expect(mockStore.getActions()).toMatchSnapshot();
```
bkemper: The removal of this snapshot removes the test coverage that confirms that a pending action is dispatched before the action and a show global alert action is dispatched after.  Should we write explicit tests confirming these actions are dispatched?

src/helpers/errorTracker.js
```diff
@@ -205,9 +212,20 @@ class ErrorTracker {
...
-    if (!Raven.isSetup()) { return; }
+    if (!Raven.isSetup()) {
+      return;
+    }
     Raven.captureException(error, { logger, extra });
   }
+
+  addRequestContextAndThrow(reduxActionType, response = {}, error) {
+    if (!Raven.isSetup()) {
+      return;
```
bkemper: Should the default be to throw the error?

src/helpers/errorTracker.js
```diff
@@ -205,9 +212,20 @@ class ErrorTracker {
...
   report(logger, error, extra = {}) {
     // Silently ignore if Sentry is not setup
-    if (!Raven.isSetup()) { return; }
+    if (!Raven.isSetup()) {
+      return;
+    }
     Raven.captureException(error, { logger, extra });
   }
+
+  addRequestContextAndThrow(reduxActionType, response = {}, error) {
```
bkemper: Do we need a default for response?

src/helpers/tests/errorTracker.test.js
```diff
@@ -327,11 +368,68 @@ describe('.install', () => {
...
+  beforeEach(() => {
+    captureException = jest.spyOn(mockRaven, 'captureException');
+    context = jest.spyOn(mockRaven, 'context');
+    isSetup = jest.spyOn(mockRaven, 'isSetup');
+  });
+
+  afterEach(() => {
+    captureException.mockRestore();
+    context.mockRestore();
+    isSetup.mockRestore();
```
bkemper: This should be setup by our global configuration.

src/actions/helpers/tests/requestHelperFactory.test.js
gfriedmansp: Feels reasonable to add a check for calling the function - had never had that perspective on the reasoning for tests that just check for calling the function

gfriedmansp: I'm unaware of any standard that says to not declare variables outside of the describe scope, but I can move it in - I'd expect it doesn't change anything either way

src/actions/helpers/tests/zuoraRequest.test.js
gfriedmansp: I think I'll just add the snapshot test back in - missed that I removed that from the test.

src/helpers/errorTracker.js
gfriedmansp: YES! Nice catch

src/helpers/errorTracker.js
gfriedmansp: Probably not, but I don't see it hurting anything. It allows someone to pass undefined as the second param if there's no response object to pass 🤷‍♂

PR #1396, opened 2020-02-04
AC-1218: Don't show billing page for AWS market place users

PatrickSison:
### What Changed
 - AWS users are now shown a banner linking to a support ticket instead of the billing page
 - AWS users can no longer access billing change plan pages

### How To Test
 - Will need to ask Nancy for help with setting up an AWS account (or you can use mine on staging)
 - Check that `/account/billing` shows the banner without the AWS action
 - Check the page with a manually billed account and see the banner (ask me if you need help setting up account)
 -Check the page with a regular self serve

### To Do
- [x] Follow up with support actions for free AWS users (if there are any) as it leads to an incorrect support ticket + visible link to inaccessible page.

- [x] (?) Possibly rename from ManuallyBilledBanner to something more concise.


shrirupa:
Since AWS users can't sign up for new account now, can we also remove the aws conditions that we have on src/pages/onboarding/ChoosePlan.js

src/actions/billing.js
```diff
@@ -21,9 +21,7 @@ export function updateSubscription({ code, bundle = code, promoCode, meta = {} }
             type: 'UPDATE_SUBSCRIPTION',
             meta: {
               method: 'PUT',
-              url: isAws(getState())
-                ? '/v1/account/aws-marketplace/subscription'
-                : '/v1/billing/subscription/bundle',
+              url: '/v1/billing/subscription/bundle',
               data: {
                 promo_code: promoCode,
                 [isAws(getState()) ? 'code' : 'bundle']: bundle,
```
shrirupa: we can also remove this isAws check here on line 27.

src/pages/billing/SummaryPage.js
```diff
@@ -59,7 +60,7 @@ export class BillingSummaryPage extends Component {
       <ManuallyBilledBanner
         account={account}
         onZuoraPlan={billingInfo.onZuoraPlan}
```
shrirupa: is there no case in which aws users will have billingId on there plan ?

PatrickSison: Yeah, the AWS plans are kept. separately. The zuora billing id are plans in zuora, and those will be the self serve + manually billed ones

PR #1397, opened 2020-02-05
SA-1195 Hide Accounts Dropdown For Subaccount Reporting User

gfriedmansp:
### What Changed
 - Adjusted the logic for the `hasSubaccounts` selector to include the fact that subaccounts have `has_subaccounts` set to true regardless of their role. It needs to be false when the isSubaccountUser selector is true.

### How To Test
 - Point local dev to production environment
 - Log in as a subaccount user of account 108
 - Open Health Score page, see that the account dropdown is hidden and the single subaccount the user belongs to is the only one in the list.
 - Note there is a separate ticket to fix the graphs on the top to default to the subaccount in this same case


gfriedmansp:
> Looks good to me. One thing I noticed in src/pages/templates/components/create/CreateForm.js we handle the same thing differently, [ `selectCondition(not(isSubaccountUser))(state)` ] . May be we should make this consistent everywhere.

I would say let's worry about this when we're eventually going through and fixing the other areas around the app that don't work quite right for subaccount reporting users (and other similar niches).

PR #1398, opened 2020-02-05
FE-854 Codecov Integration

gfriedmansp:
### What Changed
 - Added Codecov integration as the last step (only on success) in travis process
 - Removed remnants of coveralls in repo

### How To Test
 - Navigate to https://codecov.io/gh/SparkPost/2web2ui/commits and see that this branch is there

### To Do
- Add Codecov integration (bot) to this repo so we get the automatic posts


src/toDelete/ToDelete.js
```diff
@@ -0,0 +1,24 @@
...
+    case 6:
+      return <span>{`Goodbye ${x}`}</span>;
+    case 7:
+      return <span>{`Okay ${x}`}</span>;
+    default:
+      return null;
+  }
+};
+
+export default ToDelete;
```
nicklemmon: So....do we delete this now? Kinda confused as to how to handle this.

.travis.yml
```diff
@@ -28,3 +28,6 @@ jobs:
       script: npm run test-integration-ci
 
 after_script: greenkeeper-lockfile-upload
+
+after_success:
+  - bash <(curl -s https://codecov.io/bash) -e TRAVIS_NODE_VERSION
```
nicklemmon: Is this working?

Also, do you have the docs where you go this from? Every example I'm encountering Googling around is different.

src/toDelete/ToDelete.js
gfriedmansp: Haha, yeah this will be deleted before merge. Was hopefully there to prove that it's checking things correctly, but decidedly it won't without a master branch to compare it to.

.travis.yml
gfriedmansp: https://docs.codecov.io/v4.3.0/docs/about-the-codecov-bash-uploader

src/toDelete/ToDelete.js
gfriedmansp: I think opening a new pull request might show this

PR #1399, opened 2020-02-06
SA-1196: Health Score display for subaccount reporting users

kchu1337:
### What Changed
 -Selects the correct health score data for subaccount reporting user.

### How To Test
 - Create suabccount Reporting user in an account w/ signals data
 - Go to `/signals/health-score`. Verify that the health score dashboard contains data calculated for that subaccount only.

### DIFF
--- a/src/selectors/currentUser.js
+++ b/src/selectors/currentUser.js
@@ -1,7 +1,9 @@
+import { isSubaccountUser } from 'src/helpers/conditions/user';
+
 export const userCookieConsentFlagSelector = state => !!state.currentUser.cookie_consent;
 
 export const usernameSelector = state =>
   state.currentUser.username || state.tfa.username || state.auth.username;
 
 export const subaccountReportingIdSelector = ({ currentUser = {} }) =>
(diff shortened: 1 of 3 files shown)


src/selectors/currentUser.js
```diff
@@ -1,3 +1,7 @@
-export const userCookieConsentFlagSelector = (state) => !!state.currentUser.cookie_consent;
+export const userCookieConsentFlagSelector = state => !!state.currentUser.cookie_consent;
 
-export const usernameSelector = (state) => state.currentUser.username || state.tfa.username || state.auth.username;
+export const usernameSelector = state =>
+  state.currentUser.username || state.tfa.username || state.auth.username;
+
+export const subaccountReportingIdSelector = ({ currentUser = {} }) =>
+  currentUser.access_level === 'subaccount_reporting' ? currentUser.subaccount_id : -1;
```
gfriedmansp: We could use the `isSubaccountUser` condition here - don't think this would be specific to subacccount reporting users, but just to subaccounts users (which happen to be the same subset of users right now).

PR #1410, opened 2020-02-10
SE-531: Fix filtered by message for blacklist alerts

PatrickSison:
### What Changed
 - Blacklist alerts will not rendered "Filtered By parts if 

### How To Test
 - Create a blacklist alert (available on appteam account)
 - Create alert with no filters no filters, or only filters on one of blacklist provider or blacklist resources

### DIFF
--- a/src/pages/alerts/components/AlertDetails.js
+++ b/src/pages/alerts/components/AlertDetails.js
@@ -76,8 +76,11 @@ export const AlertDetails = ({ alert, id, subaccountIdToString, hasSubaccounts }
       undefined
     );
 
+    console.log(filters);
     //Removes filter which contain string 'any'. Currently only appears on blacklist alerts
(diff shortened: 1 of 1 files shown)


src/pages/alerts/components/AlertDetails.js
```diff
@@ -76,8 +76,11 @@ export const AlertDetails = ({ alert, id, subaccountIdToString, hasSubaccounts }
       undefined
     );
 
+    //Removes filter which contain string 'any'. Currently only appears on blacklist alerts
+    const cleanedFilters = filters.filter(({ filter_values }) => filter_values[0] !== 'any');
```
kchu1337: Although extremely unlikely, it's possible for a user to name their ip pool 'any', and if they use that as a filter, it will get removed. See https://app-staging.sparkpost.com/alerts/details/13702

PR #1411, opened 2020-02-10
AC-1235: Biling uploads first 6 digits of card for bin number

PatrickSison:
### What Changed
 - Spaces are removed from card number before pulling first 6 characters

### How To Test
 - Create a new account
 - Select a paid plan and keep an eye on the network requests
 - Check that in the /cors-data request, the bin includes 6 characters with no spaces

You can also check master branch to see that spaces were previously being sent.

### DIFF
--- a/src/helpers/billing.js
+++ b/src/helpers/billing.js
@@ -14,7 +14,7 @@ export function formatDataForCors(values) {
     state: billingAddress.state,
     country: billingAddress.country,
     zip_code: billingAddress.zip,
-    bin: card.number.slice(0, 6),
+    bin: card.number.replace(/\s/g, '').slice(0, 6),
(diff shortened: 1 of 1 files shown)


src/helpers/billing.js
```diff
@@ -14,7 +14,7 @@ export function formatDataForCors(values) {
     state: billingAddress.state,
     country: billingAddress.country,
     zip_code: billingAddress.zip,
-    bin: card.number.slice(0, 6),
+    bin: card.number.replace(/\s/g, '').slice(0, 6),
```
shrirupa: not related to this PR but cors endpoint doesn't send bin and other data when CreditCard is updated, do we need to update that ?

PR #1413, opened 2020-02-11
AC-1197: "Create API Key" button on RV API view subscribes to RV bundle

shrirupa:
AC-1197: "Create API Key" button on RV API view subscribes to RV bundle

### What Changed
 - added Create API Key functionality to the API Integration tab on RV page.
    "Create API Key" button on the API Integration tab 
               - checks if there is a Zuora account
               - If there is no Zuora account, create an account and add the RV product to the subscription
               - If there is a Zuora account, check if RV is on the subscription. If not, add the RV bundle to the subscription.
                - Redirect to the API Key page.

### How To Test
 - enable `standalone_rv` flag.
 - Navigate to API Intergration tab - /recipient-validation/api

### To Do
- [ ] Address Feedback


src/pages/recipientValidation/RecipientValidationPage.js
```diff
@@ -208,7 +210,9 @@ export class RecipientValidationPage extends Component {
           <ValidateSection
             credit_card={billing.credit_card}
             submitButtonName={selectedTab === 2 ? 'Create API Key' : 'Validate'}
-            submitDisabled={submitDisabled}
+            submitDisabled={
+              selectedTab === 2 ? submitDisabled && !this.state.useSavedCC : submitDisabled
```
PatrickSison: Not sure what the condition for the tab is needed for

PatrickSison: The saved cc condition should probably be on both tab 2 and tab 1

src/pages/recipientValidation/RecipientValidationPage.js
```diff
@@ -82,40 +82,35 @@ export class RecipientValidationPage extends Component {
...
-      return action.then(() => this.validate(values));
+    if (this.state.useSavedCC && isRVonSubscription) {
+      return this.redirectToNextStep(formValues);
     }
+
+    const values = formValues.card
+      ? { ...formValues, card: prepareCardInfo(formValues.card) }
+      : formValues;
+
+    addRVtoSubscription({
```
PatrickSison: I think we need to return the promise so that redux form can properly set prop `submitting`

src/actions/billingUpdate.js
```diff
@@ -19,15 +19,25 @@ export default function billingUpdate(values) {
...
-      values.planpicker || values.bundle
-        ? updateSubscription({
-            bundle: values.bundle,
-            code: values.planpicker.code,
-            promoCode: values.promoCode,
-            meta,
-          })
-        : meta.onSuccess({ meta }); // bypass
+    const maybeUpdateSubscription = ({ meta }) => {
+      if (values.planpicker) {
```
PatrickSison: This is getting bigger, but the issue is that there is a planpicker initilalized in the `NewChangePlanForm` but it's not needed now. Should we file a bug for this?

PatrickSison: Newchangeplanform should look at the values.bundle

bkemper: I reverted it and used lodash to get the planpicker.code when this is called from RV to avoid breaking.

PR #1416, opened 2020-02-12
FE-858 - Cypress tests for the Alerts list and details pages

nicklemmon:
[FE-858](https://jira.int.messagesystems.com/browse/FE-858)

### What Changed

- Updates Cypress NPM scripts to use the [Cypress module API](https://docs.cypress.io/guides/guides/module-api.html) instead of the CLI. This allows for a distinct (themed!) configuration of test batches
- Adds new tests for the details and list page for alerts:
  * `/alerts/details/<id>`
  * `/alerts`
- Adds `data-id` attributes to various components

### How To Test
- Verify tests are passing
- Visit the aforementioned app routes and verify that tests properly cover feature functionality and requirements (including edge cases)

### To Do
- [ ] Incorporate feedback


cypress/tests/integration/alerts/listPage.spec.js
```diff
@@ -0,0 +1,282 @@
...
+      url: API_URL,
+      fixture: 'alerts/200.get.all-muted.json',
+    });
+
+    cy.visit(PAGE_URL);
+
+    cy.queryByText('Recent Incidents').should('not.be.visible');
+    cy.get('[data-id="recent-incidents"]').should('not.be.visible');
+  });
+
```
kchu1337: Could also have a test for showing only the 4 most recent. Might be overkills on this section tho?

cypress/tests/integration/alerts/detailsPage.spec.js
```diff
@@ -0,0 +1,265 @@
...
+    });
+
+    // Using score data to assert against table ordering
+    // Using dates within Travis CI is a little fickle at this point
+    // and some failures occur due to time zone differences
+    it('is sortable by "Triggered"', () => {
+      cy.visit(PAGE_URL);
+
+      cy.findByText('Triggered').click();
+
```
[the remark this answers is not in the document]
kchu1337: Also, not sure if these tests are even necessary since they test functionality of the `TableCollection` component rather than page or multi-component interactions.

src/pages/alerts/ListPage.js
```diff
@@ -58,36 +52,43 @@ export class ListPage extends Component {
     return (
       <>
```
kchu1337: Could replace this fragment with `<div data-id="recent-incidents">` instead of below the header.

cypress/tests/integration/alerts/listPage.spec.js
```diff
@@ -0,0 +1,282 @@
...
+        url: `${API_URL}/2`,
+        fixture: 'alerts/2/400.delete.json',
+      });
+
+      cy.get('#modal-portal').within(() => cy.findByText('Delete').click());
+
+      cy.findByText('Something went wrong.').should('be.visible');
+      cy.findByText('View Details').click();
+      cy.findByText('This is an error').should('be.visible');
+    });
```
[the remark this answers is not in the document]
nicklemmon: Nice, I didn't notice that. Good catch.

cypress/tests/integration/alerts/detailsPage.spec.js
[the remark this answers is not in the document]
nicklemmon: I think you're right - this is really more of a unit test than anything, well covered by other testing.

cypress/tests/integration/alerts/listPage.spec.js
```diff
@@ -0,0 +1,282 @@
...
+    cy.visit(PAGE_URL);
+
+    cy.findByText('Recent Incidents').should('be.visible');
+
+    cy.get('[data-id="recent-incidents"]').within(() => {
+      // 'Alert 1' and 'Alert 4' are not rendered since they are muted
+      cy.queryByText('Alert 1').should('not.be.visible');
+      cy.findByText('Alert 2').should('be.visible');
+      cy.findByText('Alert 3').should('be.visible');
+      cy.queryByText('Alert 4').should('not.be.visible');
```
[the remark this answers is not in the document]
nicklemmon: https://testing-library.com/docs/dom-testing-library/api-queries#queryby

`queryBy` is better used when you're checking for something _not_ to exist. `findBy` in most every other scenario.

cypress/tests/integration/alerts/listPage.spec.js
nicklemmon: Isn't the previous test showing that?

src/pages/alerts/ListPage.js
nicklemmon: That's a good idea.

nicklemmon:
@kchu1337 Ready for another pass when you have time

cypress/tests/integration/alerts/listPage.spec.js
kchu1337: I meant a test to show that even if there are 8 recent alert incidents, only 4 are displayed. Might be overkill though. Or you could combine it with the above and use a fixture with more incidents.

cypress/tests/integration/alerts/listPage.spec.js
```diff
@@ -241,6 +241,8 @@ describe('The alerts list page', () => {
     it('opens a delete confirmation modal when clicking the delete button', () => {
       cy.visit(PAGE_URL);
 
+      cy.findAllByText('Alert 2').should('have.length', 2);
```
kchu1337: Could add a comment about why its' 2. I assume the table + incident cards?

cypress/tests/integration/alerts/listPage.spec.js
nicklemmon: Yeeah I feel like that might be overkill? A good thing to check for, but not a deal breaker if that didn't work.

cypress/tests/integration/alerts/listPage.spec.js
nicklemmon: Good call. Done.

PR #1417, opened 2020-02-13
AC-1238: Update the Recipient Validation Page for manually billed customers

shrirupa:
AC-1238: Update the Recipient Validation Page for manually billed customers

### What Changed
- Update the RV page for manually billed customers who have rv on their subscription. 
      -Do not show the Credit Card Section
      -Show the Validate and Create API key buttons.
      -Added loading state for UploadedListPage and RecipientValidationPage as discussed on [uxfe](https://msys.slack.com/archives/GLLAW6UAX/p1582658131001000)

### How To Test
- Enable 'standalone_rv' ui flag.
- Create a manually billed account using https://github.com/SparkPost/sparkpost-admin-api- 
   documentation/blob/master/services/billing_api.md#create-a-manually-billed-account-post .
- Add rv to user subscription (use PUT /billing/subscription/bundle)

### To Do
- [ ] Address Feedback

### DIFF
--- a/src/pages/recipientValidation/RecipientValidationPage.js
+++ b/src/pages/recipientValidation/RecipientValidationPage.js
@@ -137,7 +137,14 @@ export class RecipientValidationPage extends Component {
 
   renderRecipientValidation = () => {
     const { selectedTab, showPriceModal } = this.state;
-    const { isStandAloneRVSet, billing, billingLoading, valid, submitting } = this.props;
+    const {
+      isStandAloneRVSet,
+      billing,
+      billingLoading,
+      valid,
+      submitting,
+      isRVonSubscription,
+    } = this.props;
 
     return (
       <Page
@@ -206,6 +213,7 @@ export class RecipientValidationPage extends Component {
             formname={FORMNAME}
             handleCardToggle={this.handleToggleCC}
             defaultToggleState={!this.state.useSavedCC}
+            isProductOnSubscription={isRVonSubscription}
           />
         )}
 
--- a/src/pages/recipientValidation/RecipientValidationPage.js
+++ b/src/pages/recipientValidation/RecipientValidationPage.js
@@ -8,6 +8,7 @@ import _ from 'lodash';
 import { hasAccountOptionEnabled, isAccountUiOptionSet } from 'src/helpers/conditions/account';
 import { prepareCardInfo, isProductOnSubscription } from 'src/helpers/billing';
 import { rvAddPaymentFormInitialValues } from 'src/selectors/recipientValidation';
+import { selectIsSelfServeBilling } from 'src/selectors/accountBillingInfo';
 import { getBillingInfo } from 'src/actions/account';
 import addRVtoSubscription from 'src/actions/addRVtoSubscription';
 import { getSubscription as getBillingSubscription } from 'src/actions/billing';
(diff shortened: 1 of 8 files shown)


src/pages/recipientValidation/RecipientValidationPage.js
```diff
@@ -96,9 +100,9 @@ export class RecipientValidationPage extends Component {
   };
 
   onSubmit = formValues => {
-    const { addRVtoSubscription, isRVonSubscription } = this.props;
-
-    if (this.state.useSavedCC && isRVonSubscription) {
+    const { addRVtoSubscription, isRVonSubscription, isManuallyBilled } = this.props;
+    this.setState({ submitStatus: 'pending' });
+    if ((this.state.useSavedCC && isRVonSubscription) || (isRVonSubscription && isManuallyBilled)) {
```
kchu1337: You could do `(isRVonSubscription && (this.state.useSavedCC || isManuallyBilled))`.
That way the logic flows like do they have RV plan and if so, do they have either saved CC or are manually billed? But it's not a big deal either way.

src/pages/recipientValidation/RecipientValidationPage.js
```diff
@@ -110,7 +114,9 @@ export class RecipientValidationPage extends Component {
       values,
       updateCreditCard: !this.state.useSavedCC,
       isRVonSubscription: isRVonSubscription,
-    }).then(() => this.redirectToNextStep(values));
+    })
+      .then(() => this.redirectToNextStep(values))
```
kchu1337: One rare issue that could occur is that the user leaves the page while the promise is getting resolved. When it gets resolved, then it will immediately run the redirect even of they are no longer on that page. You can test this by throttling chrome network response in dev tools, trying to validate with new CC and while it's resolving, move to another page.

src/pages/recipientValidation/UploadedListPage.js
```diff
@@ -46,14 +50,26 @@ export class UploadedListPage extends Component {
...
+      isStandAloneRVSet,
+      isManuallyBilled,
+    } = this.props;
     const { useSavedCC } = this.state;
+    this.setState({ submitStatus: 'pending' });
 
-    if (!isStandAloneRVSet || (useSavedCC && isRVonSubscription)) {
+    if (
+      !isStandAloneRVSet ||
+      (useSavedCC && isRVonSubscription) ||
```
kchu1337: Could do `(isRVonSubscription && (this.state.useSavedCC || isManuallyBilled))` if it makes sense to do so here. Not a big deal.

src/pages/recipientValidation/RecipientValidationPage.js
```diff
@@ -217,6 +231,9 @@ export class RecipientValidationPage extends Component {
   };
 
   render() {
+    if (this.state.submitStatus === 'pending' && this.props.isStandAloneRVSet) {
```
kchu1337: why do you need to check this.props.isStandAloneRVSet? Not saying its wrong, just wondering why.

src/pages/recipientValidation/RecipientValidationPage.js
shrirupa: nice catch! Added a condition in redirectToNext to avoid this edge case.

src/pages/recipientValidation/RecipientValidationPage.js
shrirupa: its not necessary at all but I just wanted to be sure nothing related to standalone rv effects the current rv flow.

PR #1418, opened 2020-02-18
TR-2221 - Filter out falsy values from the templates content object

nicklemmon:
[TR-2221](https://jira.int.messagesystems.com/browse/TR-2221)

### What Changed
- Adds unit test cases for the templates selector
- Adds Cypress test to check for exception
- Resolves bug by filtering out falsy values from the `template.content` object received from a request

### How To Test
- Run Cypress integration tests locally - `npm run test-integration`
- Comment out line 41 of `src/selectors/templates.js`
- Run the `editDraftPage.spec.js` tests in Cypress and see the test fail due to the uncaught exception
- Re-introduce line 41 of `src/selectors/templates.js`
- Run the `editDraftPage.spec.js` tests in Cypress again and see everything pass

### To Do
- [ ] Incorporate feedback

### DIFF
--- a/cypress/tests/integration/templates/editDraftPage.spec.js
+++ b/cypress/tests/integration/templates/editDraftPage.spec.js
@@ -65,7 +65,7 @@ describe('The templates edit draft page', () => {
   it('redirects to the list page if the template fails to load', () => {
     cy.stubRequest({
       statusCode: 400,
-      url: '/api/v1/templates/stubbed-template-1*',
+      url: '/api/v1/templates/stubbed-template-1',
       fixture: 'templates/stubbed-template-1/400.get.json',
     });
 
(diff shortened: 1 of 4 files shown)


cypress/tests/integration/templates/editDraftPage.spec.js
```diff
@@ -76,6 +76,22 @@ describe('The templates edit draft page', () => {
     cy.findByText('Stubbed Template 1').should('not.be.visible');
   });
 
+  it.only('does not render an error if the server returns content with null values', () => {
```
bkemper: Should this be `it.only`?

src/selectors/templates.js
```diff
@@ -8,78 +8,91 @@ import { filterTemplatesBySubaccount } from 'src/helpers/templates';
...
+  // eslint-disable-next-line lodash/prop-shorthand
+  _.filter(state.templates.list, template => template.has_published);
+
+const createTemplateSelector = getter =>
+  createSelector([getter, selectDefaultTemplateOptions], (template, defaultOptions) => {
+    if (!template) {
+      return;
+    }
+
+    const templateContent = _.isEmpty(template.content)
```
bkemper: I don't think we need this check, but it doesn't hurt to have it.

PR #1422, opened 2020-02-19
AC-1119: New sign up flow for RV

PatrickSison:
### What Changed
 - New page after joining which directs to recipient validation

### How To Test
 - Log out and go to path /join?product=rv
 - Sign up and should navigate to /onboarding/recipient-validation
 - Verify that copy is correct and matches mocks
 - Verify that link navigates to RV page after

### To Do
- [ ] Fix some of the styling
 - [ ] Verify copy as it it is a bit large

### DIFF
--- a/src/config/routes/index.js
+++ b/src/config/routes/index.js
@@ -719,7 +719,7 @@ const routes = [
   {
     path: '/onboarding/recipient-validation',
     component: onboarding.RVBundlePage,
-    condition: all(configFlag('featureFlags.has_signup')),
+    condition: configFlag('featureFlags.has_signup'),
     title: 'Recipient Validation | Onboarding',
     layout: LargeForm,
   },
--- a/src/config/routes/index.js
+++ b/src/config/routes/index.js
@@ -717,6 +717,13 @@ const routes = [
     supportDocSearch: 'upgrade account',
   },
   {
+    path: '/onboarding/recipient-validation',
+    component: onboarding.RVBundlePage,
+    condition: all(configFlag('featureFlags.has_signup')),
+    title: 'Recipient Validation | Onboarding',
+    layout: LargeForm,
+  },
+  {
     path: '/recipient-validation',
     redirect: '/recipient-validation/list',
   },
--- a/src/pages/onboarding/RVBundlePage.js
+++ b/src/pages/onboarding/RVBundlePage.js
@@ -1,0 +1,68 @@
+import React from 'react';
+import { CenteredLogo } from 'src/components';
+import { Panel, UnstyledLink, Button } from '@sparkpost/matchbox';
+import { CheckCircle, Send } from '@sparkpost/matchbox-icons';
+import { Link } from 'react-router-dom';
+import { Card } from 'src/components';
+import styles from './RVBundlePage.module.scss';
+
+const PANEL_TITLE =
+  "Take a look at what's included in our Recipient Validation Service and then Start Validating.";
+
+const RVBundlePage = () => {
+  return (
+    <div>
+      <CenteredLogo />
+      <div className={styles.header}>You're Almost There...</div>
+      <div className={styles.subheader}>
+        Here at SparkPost we do it all! As part of our Recipient Validation service we also provide
+        a SparkPost sending account. Make every email count and start verifying.
+      </div>
+
+      <Panel>
+        <Panel.Section>{PANEL_TITLE}</Panel.Section>
+        <Panel.Section>
+          <div className={styles.grid}>
+            <Card>
+              <h2>
+                <CheckCircle className={styles.icon} />
+                <span>&nbsp;Recipient Validation</span>
+              </h2>
+              <div className={styles.text}>
+                With our monthly pay-as-you-go plan using tiered pricing, the more you validate, the
+                less you pay per message.
+              </div>
+              <div className={styles.linkcontainer}>
+                <UnstyledLink className={styles.link}>See Pricing</UnstyledLink>
+              </div>
+            </Card>
+
+            <Card>
+              <h2>
+                <Send className={`${styles.icon} ${styles.send}`} />
+                <span>&nbsp;Email Messaging</span>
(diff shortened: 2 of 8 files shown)


src/config/routes/index.js
```diff
@@ -716,6 +716,13 @@ const routes = [
     title: 'Choose Your Plan | Onboarding',
     supportDocSearch: 'upgrade account',
   },
+  {
+    path: '/onboarding/recipient-validation',
+    component: onboarding.RVBundlePage,
+    condition: all(configFlag('featureFlags.has_signup')),
```
bkemper: Do we need `all`?

src/pages/onboarding/RVBundlePage.js
```diff
@@ -0,0 +1,68 @@
+import React from 'react';
+import { CenteredLogo } from 'src/components';
+import { Panel, UnstyledLink, Button } from '@sparkpost/matchbox';
+import { CheckCircle, Send } from '@sparkpost/matchbox-icons';
+import { Link } from 'react-router-dom';
+import { Card } from 'src/components';
+import styles from './RVBundlePage.module.scss';
+
+const PANEL_TITLE =
+  "Take a look at what's included in our Recipient Validation Service and then Start Validating.";
```
bkemper: Recipient Validation Service has service capitalized here, but not in the subheader.  We should pick one and make it consistent.  Also, does start validating need to be in caps?
#nit

src/pages/onboarding/RVBundlePage.js
```diff
@@ -0,0 +1,68 @@
+import React from 'react';
+import { CenteredLogo } from 'src/components';
+import { Panel, UnstyledLink, Button } from '@sparkpost/matchbox';
+import { CheckCircle, Send } from '@sparkpost/matchbox-icons';
+import { Link } from 'react-router-dom';
+import { Card } from 'src/components';
+import styles from './RVBundlePage.module.scss';
+
+const PANEL_TITLE =
```
bkemper: Do we need to make this a const?  It is only being used one time.  It might read better in JSX.

src/pages/onboarding/RVBundlePage.js
```diff
@@ -0,0 +1,68 @@
...
+      </div>
+
+      <Panel>
+        <Panel.Section>{PANEL_TITLE}</Panel.Section>
+        <Panel.Section>
+          <div className={styles.grid}>
+            <Card>
+              <h2>
+                <CheckCircle className={styles.icon} />
+                <span>&nbsp;Recipient Validation</span>
```
bkemper: The `&nbsp;` should be replaced with css.

```
.icon {
  margin-right: spacing();
}
```

src/pages/onboarding/RVBundlePage.js
```diff
@@ -0,0 +1,68 @@
...
+              </div>
+              <div className={styles.linkcontainer}>
+                <UnstyledLink className={styles.link}>See Pricing</UnstyledLink>
+              </div>
+            </Card>
+
+            <Card>
+              <h2>
+                <Send className={`${styles.icon} ${styles.send}`} />
+                <span>&nbsp;Email Messaging</span>
```
bkemper: Same here.

src/pages/onboarding/RVBundlePage.module.scss
```diff
@@ -0,0 +1,46 @@
...
+.icon {
+  transform: translate(0, -10%);
+
+  &.send {
+    transform: translate(0, -25%) rotate(-45deg);
+  }
+}
+
+.text {
+  font-size: font-size(300);
```
bkemper: Is this necessary?

src/pages/onboarding/RVBundlePage.js
```diff
@@ -0,0 +1,68 @@
...
+            <Card>
+              <h2>
+                <CheckCircle className={styles.icon} />
+                <span>&nbsp;Recipient Validation</span>
+              </h2>
+              <div className={styles.text}>
+                With our monthly pay-as-you-go plan using tiered pricing, the more you validate, the
+                less you pay per message.
+              </div>
+              <div className={styles.linkcontainer}>
```
bkemper: If these were `<p>`, we wouldn't need to add a margin here.

src/pages/onboarding/RVBundlePage.js
```diff
@@ -0,0 +1,68 @@
...
+import styles from './RVBundlePage.module.scss';
+
+const PANEL_TITLE =
+  "Take a look at what's included in our Recipient Validation Service and then Start Validating.";
+
+const RVBundlePage = () => {
+  return (
+    <div>
+      <CenteredLogo />
+      <div className={styles.header}>You're Almost There...</div>
```
bkemper: This is a better question for Nick, but we should discuss when to use which tag.  This should probably be a `<h1>`?  Then the subheader could be `<p>`?

src/pages/onboarding/RVBundlePage.js
```diff
@@ -0,0 +1,68 @@
...
+              <h2>
+                <CheckCircle className={styles.icon} />
+                <span>&nbsp;Recipient Validation</span>
+              </h2>
+              <div className={styles.text}>
+                With our monthly pay-as-you-go plan using tiered pricing, the more you validate, the
+                less you pay per message.
+              </div>
+              <div className={styles.linkcontainer}>
+                <UnstyledLink className={styles.link}>See Pricing</UnstyledLink>
```
bkemper: This could be a reusable component, especially, if we will continue to see this style of link.

src/pages/onboarding/RVBundlePage.module.scss
PatrickSison: Makes the text a little smaller to match the mocks a bit closer. I could swap to <small> tags?

PatrickSison: Actually swapping to p tags works fine

PR #1424, opened 2020-02-20
AC-1252: Onboarding updates

PatrickSison:
### What Changed
 - Removed the welcome modal
 - Remove the features step in the onboarding

### How To Test
 - Create a new account
 - Verify that modal no longer shows when getting to /dashboard
 - Go to `/dashboard` and verify that the Features step is removed

### Notes
 - Added in a catch for users who may have had guide step as "Features" and would catch as "Sending"


src/pages/dashboard/components/GettingStartedGuide.js
```diff
@@ -57,7 +56,10 @@ export const GettingStartedGuide = ({
           },
         ];
   //stepName could be Features,Sending,Show Me Sparkpost, Let's Code
-  const [stepName, setStepName] = useState(active_step || 'Features');
+  const [stepName, setStepName] = useState(
+    //FIXME: Check in case the active step was set to features and step was already removed
+    active_step === 'Features' || !active_step ? 'Sending' : active_step,
```
shrirupa: Can we maintain a list of possible Steps instead, so if later Some other step gets removed, we don't have to add the conditions for that.

`const possibleSteps = ['Sending', "Let's Code", "Show Me SparkPost"]`
`const [stepName, setStepName] = useState(possibleSteps.find(x => x === activeStep) || 'Sending')`

PR #1426, opened 2020-02-20
AC-1177: Billing page updates

PatrickSison:
### What Changed
 - Removed indent from "Changes to Features" section
 - Hide Changes to features if no features being removed
 - Changed features from having to click "Got it" to already being checked

### How To Test
  - Go to `/account/billing/plan` and check the Changes to Features has removed indents + prechecked
   - To see all conditions, do Free -> Starter, Starter to Premier, and Premier to starter

### To Do
 - [ ] Potentially remove some of the complexity of FeatureChangeContext

### Notes
 FeatureChangeContext is more complex because it combines quantity changes (where users would have to leave the page) and acknowledgements (where features will be dropped automatically)


src/pages/billing/context/FeatureChangeContext.js
```diff
@@ -18,26 +17,21 @@ export const FeatureChangeProvider = ({
...
-  const checkConditions = useCallback(() => {
-    getSubscription();
-  }, [getSubscription]);
-  useEffect(() => {
-    window.addEventListener('focus', checkConditions);
-    return () => {
-      window.removeEventListener('focus', checkConditions);
-    };
-  }, [checkConditions]);
+  // TODO: Used to recheck subscription on refocusing tab. Doesn't matter now since there's no products that can be checked
```
shrirupa: we can remove this commented lines of code

src/pages/billing/components/FeatureChangeSection.js
```diff
@@ -39,33 +35,33 @@ const FeatureChangeSection = () => {
...
+      </>
     ) : (
       <>
         <Warning className={cx(styles.FeatureListIcon, styles.danger)} />
-        <div name='status-description'>
-          <span>Your new plan has additional limits on features you currently use. See the list below to </span>
+        <div name="status-description">
+          <span>
+            Your new plan has additional limits on features you currently use. See the list below to{' '}
+          </span>
```
shrirupa: I think we can remove this condition as these changes are always acknowledged.

src/pages/billing/context/FeatureChangeContext.js
```diff
@@ -106,11 +100,10 @@ export const FeatureChangeProvider = ({
       _.map(actions, ({ action, condition, ...rest }, key) => ({
         ...rest,
         key,
-        value: condition !== undefined ? condition : confirmations[key],
-        action:
-          condition !== undefined ? action : <Button onClick={() => onConfirm(key)}>Got it</Button>,
+        value: condition !== undefined ? condition : true,
+        action,
       })),
```
shrirupa: we don't need to pass isReady in Context anymore, so we can remove all code that is there is calculate that.

shrirupa:
Cypress tests failing, since "Got it " buttons got removed.

PR #1429, opened 2020-02-20
FE-853 Fix ComboBoxTypeahead to use redux-form initalValues.

kchu1337:
### What Changed
 - Updated ComboboxTypeahead to use redux-form initial values.
 - Removed the the selectedMap prop and isExclusiveItem check.
 - Added isExclusiveItem function prop.
 - ComboboxTypeahead only accepts strings, not objects.

### How To Test
 - Check that the comboboxs in alerts works correctly.
 - Run Cypress tests to verify.

### DIFF
--- a/src/pages/alerts/components/fields/SubaccountsField.js
+++ b/src/pages/alerts/components/fields/SubaccountsField.js
@@ -1,60 +1,61 @@
-import React, { Component } from 'react';
+import React, { useCallback, useEffect, useMemo } from 'react';
 import { connect } from 'react-redux';
-import { Field, formValueSelector } from 'redux-form';
+import { Field } from 'redux-form';
 import ComboBoxTypeaheadWrapper from 'src/components/reduxFormWrappers/ComboBoxTypeaheadWrapper';
 import { getSubaccounts } from 'src/selectors/subaccounts';
 import { list as listSubaccounts } from 'src/actions/subaccounts';
-import { FORM_NAME } from '../../constants/formConstants';
 
-const subaccountToString = (subaccount) => {
-  if (!subaccount) {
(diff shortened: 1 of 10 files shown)


src/pages/alerts/components/fields/SubaccountsField.js
```diff
@@ -1,60 +1,61 @@
...
+  useEffect(() => {
+    listSubaccounts();
+  }, [listSubaccounts]);
 
-  render() {
-    const { disabled, subaccounts, subaccountsFieldValue } = this.props;
-    const subaccountItems = [
-      { id: -1, name: 'Master and all subaccounts', isExclusiveItem: true },
-      { id: -2, name: 'Any subaccount', isExclusiveItem: true },
+  const subaccountItems = useMemo(() => {
```
shrirupa: I couldn't understand, why do we need to use useMemo here

kchu1337: I thought it would re-render and run this function every time the combobox value changes, so I memoized it since subaccounts wasn't changing. If this is recreated every time, then the list of options on line 39 would get recreated as an array with a different pointer. So the combobox component thinks that it's a completely different list of items. Then that will trigger the useEffect in ComboBoxTypeahead that has `results` as a dependency.

kchu1337: ^ In theory at least.

kchu1337: Looks like you are right. It is not re-rendering when the field value changes. It was probably a holdover from when I was passing in the field value as a prop. I will go ahead and remove the memoization parts

PR #1433, opened 2020-02-21
FE-859 - Blacklist incident and watchlist pages

nstanard:
[FE-858](https://jira.int.messagesystems.com/browse/FE-859)

### What Changed

- Adds new tests for the blacklist incidents and watchlist pages
  * `/blacklist/incidents`
  * `/blacklist/watchlist`

### How To Test
- Verify tests are passing
- Visit the aforementioned app routes and verify that tests properly cover feature functionality and requirements (including edge cases)

### To Do
- [x] Incorporate feedback


cypress/support/commands.js
```diff
@@ -161,3 +161,23 @@ Cypress.Commands.add('assertLink', ({ content, href }) => {
...
+
+/**
+ * Used to check within the global alert view
+ *
+ * @param {function} assertMethod - The function that has assertions to run within the alert container
+ *
+ */
+Cypress.Commands.add('withinAlert', assertMethod => {
+  cy.get('#alert-portal').within(assertMethod);
+});
```
nicklemmon: Is this for the [Snackbar component](https://sparkpost.github.io/matchbox/?selectedKind=Feedback%7CSnackbar&selectedStory=Success&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel)?

If so, I might call this command `withinSnackbar` or `withinSnackbars` or something - the only reason I say that is that's what we tend to call the component on the team (even though the portal clearly has an ID of `alert-portal`)

cypress/tests/integration/blacklists/incidents.spec.js
```diff
@@ -0,0 +1,269 @@
...
+    case 1:
+      return 'st';
+    case 2:
+      return 'nd';
+    case 3:
+      return 'rd';
+    default:
+      return 'th';
+  }
+};
```
nicklemmon: I see what you're doing here, but the function name is a little confusing to me. Could it have a more action-oriented name by any chance? Like `getDaySuffix` or something?

src/components/collection/FilterSortCollection.js
```diff
@@ -75,6 +75,9 @@ const FilterSortCollection = ({
                   {filterBox}
                 </Grid.Column>
                 <Grid.Column xs={12} md={3}>
+                  <label className={styles.Label} htmlFor={`sortSelect`}>
+                    Sort Select
+                  </label>
```
nicklemmon: `htmlFor` can just be simple string - `htmlFor="sortSelect"` - no need for anything fancy!

cypress/fixtures/blacklists/incident/200-resources.get.json
```diff
@@ -0,0 +1,64 @@
...
+			"watched_at": "2019-12-24T20:01:42.621Z",
+			"last_listed_at": "2019-09-09T00:00:00.000Z",
+			"total_listing_count": 1,
+			"active_listing_count": 1
+		},
+		{
+			"resource": "12.12.12.",
+			"currently_blacklisted_on": [],
+			"watched_at": "2020-01-07T21:47:40.824Z",
+			"last_listed_at": null,
```
nicklemmon: I would re-structure your fixture names just a little bit - the other fixtures are structured like this:

`<status-code>.<method>.<description>.json`

And yours appear to be

`<status-code>-<description>.<method>.json`

Just trying to be consistent!

cypress/fixtures/blacklists/incident/200.incident-blacklist-blacklist.lashback.com.json
```diff
@@ -0,0 +1,12 @@
+{
+	"results": [
+		{
+			"id": "44",
+			"resource": "127.0.0.2",
+			"blacklist_name": "blacklist.lashback.com",
```
nicklemmon: And this fixture doesn't have the method at all, so that's a little confusing. Hopefully these aren't too tricky to clean up a bit from a file name perspective.

cypress/fixtures/blacklists/incident/200.incident-resources-127.0.0.2.json
```diff
@@ -0,0 +1,84 @@
+{
+	"results": [
+		{
+			"id": "42",
+			"resource": "127.0.0.2",
+			"blacklist_name": "blacklist.lashback.com",
```
nicklemmon: Formatting on this file looks really weird...? I would think Prettier should have run on this though.

cypress/tests/integration/blacklists/addPage.spec.js
```diff
@@ -18,6 +18,8 @@ describe('Blacklist - Add to Watchlist Page', () => {
   });
 
   it('successfully adds a single resource and redirects to watchlist page', () => {
+    cy.findByText('Save').should('be.disabled');
+    cy.findByText('Save and Add Another').should('be.disabled');
```
nicklemmon: 🎉

cypress/tests/integration/blacklists/incidents.spec.js
```diff
@@ -0,0 +1,269 @@
...
+
+  it('navigates to the blacklist state and renders with a relevant page title/text for the empty state.', () => {
+    cy.visit('/dashboard');
+
+    cy.wait(500);
+    cy.get('a')
+      .findByText('Blacklist')
+      .click(); // use the nav to get there
+
+    cy.url().should('include', `${PAGE_BASE_URL}`);
```
nicklemmon: I don't think actually navigating to the page is necessary for this feature - what you have written here is actually testing the navigation rather than the blacklist incidents page.

I _believe_ I pointed you to these two videos, but he goes over test isolation approaches pretty well here:
https://www.youtube.com/watch?v=5XQOK0v_YRE
https://www.youtube.com/watch?v=5FnalKRjpZk

cypress/tests/integration/blacklists/incidents.spec.js
```diff
@@ -0,0 +1,269 @@
...
+    cy.findByText(
+      'Keep an eye on your Domains and IPs and maintain a healthy sender reputation and improve your deliverability',
+    ).should('be.visible');
+  });
+
+  it(`goes to add to watchlist page from the call to action on ${PAGE_BASE_URL} page`, () => {
+    cy.visit(`${PAGE_BASE_URL}`);
+    cy.findByText('Add to Watchlist').click();
+    cy.url().should('include', 'blacklist/watchlist/add');
+  });
```
nicklemmon: In this case, I think you could just check that the 'Add to Watchlist' element has the `href` to the add page rather than testing the add page by actually clicking on the link.

cypress/tests/integration/blacklists/incidents.spec.js
```diff
@@ -0,0 +1,269 @@
...
+    cy.stubRequest({
+      method: 'GET',
+      url: 'api/v1/blacklist-monitors/incidents*',
+      statusCode: 200,
+      fixture: 'blacklists/incident/200.get.json',
+    });
+
+    cy.visit(`${PAGE_BASE_URL}`);
+
+    cy.url().should('include', `${PAGE_BASE_URL}`);
```
nicklemmon: In these cases, no need to use template literals in these sorts of scenarios!

```
cy.visit(PAGE_BASE_URL)
```

cypress/tests/integration/blacklists/incidents.spec.js
```diff
@@ -0,0 +1,269 @@
...
+      statusCode: 200,
+      fixture: 'blacklists/incident/200.get.json',
+    });
+
+    cy.visit(`${PAGE_BASE_URL}`);
+
+    cy.get('tbody > tr:first-child').then(el => {
+      cy.findAllByText('127.0.0.2', { container: el }).should('be.visible');
+      cy.findAllByText('blacklist.lashback.com', { container: el }).should('be.visible');
+      cy.findAllByText('Nov 20 2019, 11:14am', { container: el }).should('be.visible');
```
nicklemmon: It looks like these are failing due to time zone differences between you and me! This has actually been a problem I hadn't figured out yet, but it looks like there is a solution:
https://github.com/cypress-io/cypress/issues/1043

We can set `TZ` when we run the script. I'm going to commit this to your branch and then you can change your assertions accordingly. I _believe_ that accounts for some of the failures in Travis as well.

cypress/tests/integration/blacklists/watchlist.spec.js
```diff
@@ -0,0 +1,153 @@
...
+    cy.login({ isStubbed: true });
+  });
+
+  it('shows title and info', () => {
+    cy.stubRequest({
+      url: 'api/v1/blacklist-monitors',
+      fixture: 'blacklists/incident/200-watchlist.get.json',
+    });
+
+    cy.visit(`${PAGE_BASE_URL}`);
```
nicklemmon: Same as in other places - no need to use a template literal (just `cy.visit(PAGE_BASE_URL)`)

cypress/tests/integration/blacklists/watchlist.spec.js
```diff
@@ -0,0 +1,153 @@
...
+
+    cy.get('tbody > tr').should('have.length', 6);
+    cy.findByLabelText('Filter By').type('2.2.8');
+    cy.get('tbody > tr').should('have.length', 1);
```
nicklemmon: You shouldn't need to use the `>` immediate child selector so frequently - `tbody tr` should get the job done I think?

cypress/tests/integration/blacklists/watchlist.spec.js
```diff
@@ -0,0 +1,153 @@
...
+      method: 'DELETE',
+      statusCode: 204,
+      url: '/api/v1/blacklist-monitors/*',
+      fixture: 'blacklists/incident/204.delete.json',
+    });
+
+    cy.visit(`${PAGE_BASE_URL}`);
+
+    cy.get('tbody > tr').then(val => {
+      cy.findByText('Stop Monitoring', { container: val[0] }).click();
```
nicklemmon: In this case, I don't think you need to supply a `container` option

cypress/tests/integration/blacklists/watchlist.spec.js
```diff
@@ -0,0 +1,153 @@
...
+      cy.withinModal(modalContainer => {
+        cy.findByText('Stop Monitoring', { container: modalContainer }).click();
+      });
+    });
+
+    cy.withinAlert(alertContainer => {
+      cy.findByText('Stopped Monitoring 12.12.12..', { container: alertContainer }).should(
+        'be.visible',
+      );
+    });
```
nicklemmon: 🎉

cypress/tests/integration/blacklists/watchlist.spec.js
```diff
@@ -0,0 +1,153 @@
...
+      url: 'api/v1/blacklist-monitors',
+      fixture: 'blacklists/incident/200-search.get.json',
+    });
+
+    cy.visit(`${PAGE_BASE_URL}`);
+
+    cy.get('tbody > tr').then(val => {
+      cy.findByText('2.2.8', { container: val[0] }).click();
+      cy.url().should('include', 'blacklist/incidents?search=2.2.8');
+    });
```
nicklemmon: Again, you could probably just check the `href` rather than actually going to the new page

src/components/collection/FilterSortCollection.module.scss
```diff
@@ -17,3 +17,7 @@
     margin-bottom: 0;
   }
 }
+
+.Label {
+  @include visually-hidden();
+}
```
nicklemmon: 👍 👍 👍

cypress/tests/integration/blacklists/incidents.spec.js
kchu1337: We can probably just use moment.js instead of writing our own function.

kchu1337:
```js
import moment from 'moment';
import { FORMATS } from '../../../../src/constants' ///<-Probably better way to import this
const formatDate = (date) => moment(date).format(FORMATS.DATE)
```
Should work

cypress/tests/integration/blacklists/incidents.spec.js
```diff
@@ -0,0 +1,269 @@
...
+      statusCode: 200,
+      fixture: 'blacklists/incident/200.get.json',
+    });
+    cy.visit(`${PAGE_BASE_URL}`);
+
+    const today = new Date();
+    const date = today.getDate();
+    const month = months[today.getMonth()];
+
+    cy.findByLabelText('Broad Date Range').select('Last 24 Hours');
```
kchu1337: I don't think we need all the tests for different date ranges. Since that would be testing the date-picker drop-down rather than this page. And I think it would be good to add a test for the corresponding API call. We can do something like 
```js
 cy.stubRequest({
      method: 'GET',
      url: 'api/v1/blacklist-monitors/incidents*',
      statusCode: 200,
      fixture: 'blacklists/incident/200.get.json',
      requestAlias: 'getIncidents', ///You already have this but need to add alias
    });
    cy.wait('@nextCohortRequest').then(({ url }) => {
          cy.wrap(url).should('include', '2019-2-24');
        });
```
Although the date might not work, depending on time zone. Might need regex or set cy.Clock();

kchu1337: @nicklemmon Thoughts on ^?

nicklemmon: That makes a lot of sense to me. Totally right there.

cypress/tests/integration/blacklists/incidents.spec.js
nicklemmon: `moment` is also exposed on the `Cypress` object - https://docs.cypress.io/api/utilities/moment.html#Syntax - the benefit of using the NPM package is the version will be consistent with what's used in the app!

cypress/tests/integration/blacklists/watchlist.spec.js
nstanard: Will remove 👍 - Technically it'll always be a direct child so either will work in any situation.

cypress/tests/integration/blacklists/watchlist.spec.js
nstanard: It needs to click the button in the first row of the table so I think this one is needed.

cypress/tests/integration/blacklists/incidents.spec.js
nstanard: Removing the date checks anyway so shouldn't be a problem.

cypress/fixtures/blacklists/incident/200.incident-blacklist-blacklist.lashback.com.json
nstanard: Pretty sure I have them all cleaned up locally - can push here in a bit.

cypress/tests/integration/blacklists/incidents.spec.js
nstanard: This used to have a path on the end - was just lazy when converting - will change

cypress/tests/integration/blacklists/incidents.spec.js
nstanard: Did not completely remove date checks - checked url on request instead like comment below

cypress/tests/integration/blacklists/incidents.spec.js
```diff
@@ -0,0 +1,230 @@
...
+
+    const filterInput = cy.findByLabelText('Filter By');
+    filterInput.should('be.visible');
+    filterInput.type('2.2.8').blur();
+    cy.url().should('include', '2.2.8');
+    cy.get('tbody > tr').should('have.length', 1);
+    cy.get('tbody > tr').within(el => {
+      cy.findByText('2.2.8', { container: el }).should('be.visible');
+      cy.findByText('new.spam.dnsbl.sorbs.net', { container: el }).should('be.visible');
+      cy.findByText('Sep 8 2019, 7:00pm', { container: el }).should('be.visible');
```
kchu1337: It looks like timezone stuff is making your tests fail. I used regex in this file `cypress/tests/integration/blacklists/detailsPage.spec.js` so it would pass both local and Travis.

package.json
```diff
@@ -19,8 +19,8 @@
     "test-e2e": "npm run test-cy-env && cypress open --config integrationFolder=cypress/tests/e2e",
     "test-e2e-headless": "cypress run --config integrationFolder=cypress/tests/e2e",
     "test-e2e-ci": "npm run start-app & wait-on http://localhost:3100 && npm run test-e2e-headless",
-    "test-integration": "npm run test-cy-env && cypress open --config integrationFolder=cypress/tests/integration",
-    "test-integration-headless": "node scripts/cypressIntegration.js",
+    "test-integration": "TZ=UTC npm run test-cy-env && cypress open --config integrationFolder=cypress/tests/integration",
```
kchu1337: Not sure this is working. I still see time in my local timezone.

nicklemmon: Yeeeah it doesn't seem to be. @nstanard would you mind taking this out? Kevin's earlier advice, re: using regex is the right route to take at this point in time. We _may_ have to wait for Cypress to allow timezone configuration.

cypress/tests/integration/blacklists/incidents.spec.js
nstanard: Sorry guys - wasn't paying close enough attention to all the spots I was checking dates - this simply shouldn't be hard-coded and I've changed and pushed.

nstanard: Sorry guys - wasn't paying close enough attention to all the spots I was checking dates - this simply shouldn't be hard-coded and I've changed and pushed.

package.json
nstanard: Should be able to keep this around if it's not hurting anything - The date tests should be fixed just waiting on travis.

nstanard: Or not.... :(

PR #1435, opened 2020-02-24
FE-783 - Adding additional Cypress tests to the billing page

nicklemmon:
[FE-783](https://jira.int.messagesystems.com/browse/FE-783)

### What Changed
- Updates default account fixture to higher level plan to show additional UI states
- Updates Cypress to `4.0.2`
- Fleshes out billing page Cypress integration tests
- updates `cy.assertLink` method to use `cy.findBy*` instead of `cy.queryBy*` to take advantage of baked in timeout

### How To Test
- Run tests locally via `npm run test-integration`
- Go through the billing page and verify page states are well tested by new changes

### To Do
- [ ] Incorporate feedback


cypress/tests/integration/accounts/billingPage.spec.js
```diff
@@ -36,46 +38,511 @@ describe('Billing Page', () => {
...
   });
 
-  it('redirects to Change My Plan page when Change Plan button is clicked', () => {
-    cy.stubRequest({
-      url: '/api/v1/account',
-      fixture: 'account/200.get.json',
-      fixtureAlias: 'accountGet',
+  it('renders with a link to the change plan page', () => {
+    cy.visit(PAGE_URL);
+    cy.assertLink({ content: 'Change Plan', href: '/account/billing/plan' });
```
shrirupa: we should also add a check that this link doesn't show up when there is a pending banner.

cypress/tests/integration/accounts/billingPage.spec.js
```diff
@@ -36,46 +38,511 @@ describe('Billing Page', () => {
...
+  });
+
+  it('renders with the dedicated IPs section if the user is transitioning to self serve (they have no credit card on file and the default subscription)', () => {
```
shrirupa: rendering Dedicated Ips section depends only on whether user canPurchaseIps.

cypress/tests/integration/accounts/billingPage.spec.js
```diff
@@ -1,18 +1,20 @@
-/// <reference types="Cypress" />
+const PAGE_URL = '/account/billing';
+const ACCOUNT_API_BASE_URL = '/api/v1/account';
```
shrirupa: Few more tests scenarios for this page are for: (ref - src/pages/billing/SummaryPage)
 - Manually billed banner
 - Suspended billing banner.

cypress/tests/integration/accounts/billingPage.spec.js
nicklemmon: I'm going to add another test that checks that section _does not_ render when they can't purchase IPs! Adding that now.

PR #1443, opened 2020-02-25
FE-904 - Implement Hibana theme context in the app

nicklemmon:
[FE-904](https://jira.int.messagesystems.com/browse/FE-904)

### What Changed
- Adds `HibanaContext` that exposes a custom provider, consumer, and hook:
  + `<HibanaProvider>` can be used to wrap UI and provide Hibana-related context to child components
  + `<HibanaConsumer>` can be used by class components that need access to Hibana context
  + `useHibana` can be used by functional components to retrieve and send events (actions) that result in Hibana state changes
- Renders `<HibanaControls/>` components behind an account UI options flag:
- Adds supporting unit tests for Hibana context providers and consumers
- Adds supporting Cypress integration tests for the Hibana UI

### How To Test
1. Sign up for a new account and verify the "Enable Hibana" button does _not_ render in the bottom right of the app
2. Then, enable the UI flag via the `/account/control` endpoint and verify the "Enable Hibana" button renders in the bottom right of the app:
```json
{
  "options": {
    "ui": {
      "hibana": {
        "hasThemeControls": true
     }
  }
}
```
3. Import the `useHibana` hook in to any functional component in the app and verify that said component can both access Hibana state. The following would do it:
```javascript
import React from 'react'
import { useHibana } from 'src/context/HibanaContext'

export default function MyFunctionalComponent() {
  const [state] = useHibana();
  const { isHibanaEnabled } = state

  return (
    <>
      { isHibanaEnabled ? (
        <h1>Hibana Enabled!</h1>
      ) : (
        <h1>Hibana Disabled!</h1>
      )}
    </>
  )
}
```
4. Import the `HibanaConsumer` component in to any class component in the app and verify that said component can access Hibana state. The following should work:
```javascript
import React from 'react'
import { HibanaConsumer } from 'src/context/HibanaContext'

export default class MyClassComponent extends React.Component {
  render() {
    <HibanaConsumer>
      {state => (
        <>
          {state.isHibanaEnabled ? (
            <h1>Hibana Enabled!</h1>
          ) : (
            <h1>Hibana Disabled!</h1>
          )}
        </>
      )}
    </HibanaConsumer>
  }
}
```

### To Do
- [ ] Incorporate feedback
- [ ] Document feature flag internally once approved
- [ ] ❓**Question for the reviewer** ❓ -> Should I document the Hibana hook and consumer components? If so, where in the project should that go?


src/context/HibanaContext.js
```diff
@@ -0,0 +1,62 @@
...
+      return { isHibanaEnabled: false };
+    }
+    default: {
+      throw new Error(`Unhandled Hibana action type: ${action.type}`);
+    }
+  }
+}
+
+function HibanaProvider({ children }) {
+  const [state, dispatch] = useReducer(hibanaReducer, initialState);
```
shrirupa: do we foresee more states to be added to this context? just curious about the use of useReducer instead of useState.

nicklemmon: I suppose I could just simplify it to using `useState`, but eventually we may track more things here? Not really sure though!

nicklemmon: Actually, yeah, I'm pretty sure we will eventually add events/state for dismissing the Hibana banner entirely if that makes sense. So this allows for that in the future.

PR #1446, opened 2020-02-26
AC-1263: Add data privacy page

PatrickSison:
### What Changed
 - Added page as `/account/data-privacy`

### How To Test
 - Enable UI flag `data_privacy`
 - Check that top right dropdown includes link for `Data and Privacy`


PatrickSison:
@nicklemmon What I will probably too to alieviate some work from the other tickets is get the tabs + panel set up in this one + unit tests for the tabs

PR #1448, opened 2020-02-26
SE-566: Added default filters

PatrickSison:
### What Changed
 - Message events, add filters
  - Now has default 3 filter categories + 1 empty select
  - Validation now only checks for typed fields. Empty field (regardless of select) will be removed
  - Default filters do not appear if other filters modify the select

### How To Test
 - Navigate to `/message-events`
 - Go to "Add Filters"
 - Verify that the bottom section defaults to multiple filters.
 - Add any filter (preferably just 1) and save
 - Reopen modal
 - Verify that the defaults do not show up
 - Verify that the validation does not pop up on empty text input fields

### DIFF
--- a/src/pages/reports/messageEvents/helpers/transformData.js
+++ b/src/pages/reports/messageEvents/helpers/transformData.js
@@ -10,11 +10,15 @@ import { getEmptyFilters } from 'src/helpers/messageEvents';
  * @returns {Object} An object containing the search parameters as keys and the search terms as values.
  */
 export function getFiltersFromSearchQueries(searchQueries = []) {
-
   // Build a single object containing a key for each filter, initialised to an empty array
   const emptyFilters = getEmptyFilters(EVENTS_SEARCH_FILTERS);
   // Collect queries into an array of objects of form {key: value}
-  const queries = searchQueries.map(({ key, value }) => ({ [key]: stringToArray(value) }));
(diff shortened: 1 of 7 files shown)


src/pages/reports/messageEvents/helpers/transformData.js
```diff
@@ -10,11 +10,15 @@ import { getEmptyFilters } from 'src/helpers/messageEvents';
...
   // Build a single object containing a key for each filter, initialised to an empty array
   const emptyFilters = getEmptyFilters(EVENTS_SEARCH_FILTERS);
   // Collect queries into an array of objects of form {key: value}
-  const queries = searchQueries.map(({ key, value }) => ({ [key]: stringToArray(value) }));
+  const queries = searchQueries.reduce((queryArray, { key, value }) => {
+    if (key && value) {
+      queryArray.push({ [key]: stringToArray(value) });
+    }
+    return queryArray;
+  }, []);
```
kchu1337: Instead of here, you can change the filter on line 25 in `src/pages/reports/messageEvents/components/AdvancedFiltersModal.js`
That line removes empty filters still at `Select a Filter`. You can change that logic to just remove any empty filters.
Something like
```js
    const filteredSearchQuery = searchQuery.filter(({value}) => Boolean(value))
```

PR #1451, opened 2020-03-02
FE-901 - Resolve <title> not updating when new tab is opened

nicklemmon:
[FE-901](https://jira.int.messagesystems.com/browse/FE-901)

### What Changed
- Adds `defer={false}` prop to `<Helmet>` instance that controls the `<title>`
- Adds new commands to Cypress to ensure `cy.log()` outputs to the Node console during headless runs
- Cleans up deep equality check by stringifying the network response
- Updates obsolete snapshots for the Layout component

### How To Test
1. Log in to the app in staging or prod and open a new tab
1. Verify the bug exists by verifying the `<title>` only updates when the relevant tab receives focus
1. Log in to the app locally and open a new tab
1. Verify the page title shows in the new tab without focusing the tab

### To Do
- [ ] Incorporate feedback

### DIFF
--- a/src/components/layout/Layout.js
+++ b/src/components/layout/Layout.js
@@ -15,9 +15,12 @@ export const Layout = ({ children, location }) => {
   return (
     <LayoutComponent>
       {route.title && (
-        <Helmet>
-          <title>{route.title} | SparkPost</title>
-        </Helmet>
+        <>
+          {/* `defer` solves issue with <title> not updating when a new tab is opened: https://github.com/nfl/react-helmet/issues/315 */}
+          <Helmet defer={false}>
+            <title>{route.title} | SparkPost</title>
+          </Helmet>
+        </>
       )}
 
       {children}
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,16 +20,16 @@ before_script: greenkeeper-lockfile-update
 
 jobs:
   include:
-    # - stage: test-and-build
-    #   name: 'Unit Testing'
-    #   script: npm run test-ci
+    - stage: test-and-build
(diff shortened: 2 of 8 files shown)


src/components/layout/Layout.js
```diff
@@ -15,9 +15,12 @@ export const Layout = ({ children, location }) => {
   return (
     <LayoutComponent>
       {route.title && (
-        <Helmet>
-          <title>{route.title} | SparkPost</title>
-        </Helmet>
+        <>
+          {/* `defer` solves issue with <title> not updating when a new tab is opened: https://github.com/nfl/react-helmet/issues/315 */}
+          <Helmet defer={false}>
```
bkemper: > executes synchronously when defer={true} and async otherwise
> — https://github.com/nfl/react-helmet/blob/1d21159dbaf0638388c1a81e7e4a60c3fdd18ef9/test/HelmetTest.js#L2373

For what we use Helmet for, maybe we should be using defer=false everywhere?

nicklemmon: not sure? Might be a negative perf impact but I don't know.

bkemper: This prop is a little mysterious to me.  The test description was the most helpful so far, but from our side conversations, sounds misleading and risky.  So, probably good to do nothing.

nicklemmon:
> Feels odd to need to use JSON.stringify to compare objects, but I trust you already looked for alternatives.

I did 😅- it was tough to do deep comparison with what Cypress offers out of the box. Worth further exploration, but this at least preserves the purpose of the test and should resolve the flakiness.

PR #1453, opened 2020-03-02
FE-884 - Component Wrapper for Matchbox ProgressBar

shrirupa:
FE-884 - Component Wrapper for Matchbox ProgressBar

### What Changed
 - ProgressBar has a single, globally accessible point of entry within the app
 - ProgressBar is only imported from Matchbox in one place - replaced all instances of the ProgressBar directly imported from matchbox with this new global component
- New props available to the Hibana version of the component are exposed - and subsequently ignored by the original component (https://jira.int.messagesystems.com/browse/FE-812)
- Ensure no unusual positioning UI defects are introduced as a result of this change - both for when using Hibana and when using the default theme

### How To Test
 - enable hibana ui flag 
`{
	"options": {
		"ui":{
"hibana":{
	"hasThemeControls": true
}
		}
	}
 }`
- Check all instances of ProgressBar for ui defects with and without hibana enabled

### To Do
- [ ] Address feedback


package.json
```diff
@@ -79,6 +79,7 @@
     "date-fns": "^1.28.5",
     "downshift": "^3.2.10",
     "email-addresses": "^3.0.1",
+    "hibana": "npm:@sparkpost/matchbox@^4.0.0-hibana.3",
```
kchu1337: Hibana is already imported in line 72

src/components/matchboxWrapper/ProgressBar.container.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { ProgressBar as PreHibanaProgressBar } from '@sparkpost/matchbox';
+import { ProgressBar as HibanaProgressBar } from 'hibana';
+import { useHibana } from 'src/context/HibanaContext';
+import _ from 'lodash';
+
+export default function ProgressBar(props) {
+  const [state, _dispatch] = useHibana();
```
kchu1337: You don't actually need to deconstruct the 2nd argument.

src/components/matchboxWrapper/tests/ProgressBar.container.test.js
```diff
@@ -0,0 +1,43 @@
...
+        .last()
+        .props(),
+    ).not.toHaveProperty('size');
+    expect(
+      instance
+        .find('ProgressBar')
+        .last()
+        .props(),
+    ).not.toHaveProperty('my');
+  });
```
kchu1337: Could also add a shallow test to check that hibana progressbar is returned vs old progressbar

src/components/matchboxWrapper/tests/ProgressBar.container.test.js
```diff
@@ -0,0 +1,43 @@
...
+describe('ProgressBar', () => {
+  const subject = (props, context) => {
+    return mount(
+      <HibanaProvider>
+        <ProgressBar {...props} />
+      </HibanaProvider>,
+      context,
+    );
+  };
+  it('should only pass pre-hibana when hibana is not enabled', () => {
```
kchu1337: This test is checking that the progress bar only contains old props right? Might want to rename the test to something like `should only pass pre-hibana props when hibana is not enabled`

src/components/matchboxWrapper/ProgressBar.container.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
```
nicklemmon: I don't think we want to use the `.container` suffix - that implies this is related to Redux? We are having that same convo over with Kevin's PR.

What do you think?

shrirupa: sure, i just renamed it 👍 .

package-lock.json
```diff
@@ -3888,13 +3888,14 @@
       }
     },
     "@sparkpost/matchbox-hibana": {
-      "version": "npm:@sparkpost/matchbox@4.0.0-hibana.3",
-      "resolved": "https://registry.npmjs.org/@sparkpost/matchbox/-/matchbox-4.0.0-hibana.3.tgz",
-      "integrity": "sha512-TuXvCJxOkN4VKPTxO1Bzm0oAnuPnT9nYcZ48ehY0KJJarz+PxtzRzQRjfQYrV7aO89hIYhPDeo90Pw0noTAAsg==",
+      "version": "npm:@sparkpost/matchbox@4.0.0-hibana.4",
```
kchu1337: Could revert these diffs. I assume these were from when you installed hibana?

src/components/matchbox/ProgressBar.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { ProgressBar as PreHibanaProgressBar } from '@sparkpost/matchbox';
```
kchu1337: Could rename to `OGProgressBar` but meh.

package-lock.json
shrirupa: let's leave this one, since version of hibana seems to have updated from 3 to 4

PR #1455, opened 2020-03-02
FE-866: Add matchbox wrapper for new Box component

kchu1337:
### What Changed
 - Added wrapper component for new Hibana Box component

### How To Test
 - Does it look right? Could import it somewhere to test.

### Notes:
 - There is no need to use Hibana Context since there is no Box component in the OG matchbox


src/components/matchboxWrapper/Box.container.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import { Box as HibanaBox } from '@sparkpost/matchbox-hibana';
+
+const Box = props => {
+  return <HibanaBox {...props} />;
+};
+
+export default Box;
```
bkemper: Is this a container?

src/components/matchboxWrapper/index.js
```diff
@@ -0,0 +1 @@
+export { default as Box } from './Box.container.js';
```
bkemper: Do we want this directory?  I think it would be fine to add these components to `src/components`.

src/components/matchboxWrapper/Box.container.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import { Box as HibanaBox } from '@sparkpost/matchbox-hibana';
+
+const Box = props => {
+  return <HibanaBox {...props} />;
```
bkemper: This was a good call instead of just exporting `HibanaBox`!

src/components/matchboxWrapper/tests/Box.test.js
```diff
@@ -0,0 +1,19 @@
+import { shallow } from 'enzyme';
```
bkemper: 😄

bkemper:
Last thing, should we add docs for this component?

src/components/matchboxWrapper/index.js
kchu1337: I think it makes sense because for matchbox components we import from '@sparkpost/matchbox'. So now we switch that directory to `src/components/matchboxWrapper'.

src/components/matchboxWrapper/Box.container.js
nicklemmon: Yeah I feel like `.container` implies this has something to do with Redux...

src/components/matchboxWrapper/index.js
nicklemmon: yeah I like this direction - could it just be `src/components/matchbox` maybe? Would that be confusing?

bkemper: I don't think so, the import would be very explicit.

`import Blah from 'src/components/matchbox';`

src/components/matchboxWrapper/Box.container.js
nicklemmon: @kchu1337  - Critical thing I forgot about! Doesn't this need to return an empty fragment with `{props.children}` if Hibana is not enabled? We will be using this component in the app to help update presentation, but it should do nothing when Hibana is not enabled.

bkemper: Hmm... I figured this would only be used in places that Hibana was expected, but this is a good precaution.  Thanks, @nicklemmon.

PR #1469, opened 2020-03-04
AC-1256: Label for button changes

PatrickSison:
### What Changed
 - Button for label depends on window variable
 - Also refactored the code a bit to make it easier to digest and consolidate some of the logic

### How To Test
  - TBD (not sure with VWO yet)
Temp: 
 - Navigate to /dashboard, and go to the Sending step (top level of breadcrumb)
  - Open console and set `window.onboardingStartSendingTest = true`
 - Trigger rerender by going through step and back

### Notes
 - We could also change up the bread crumb text if needed


src/pages/dashboard/components/SendingStep.js
```diff
@@ -33,7 +33,7 @@ export const SendingStepListItem = ({ setAndStoreStepName, name, content }) => (
             onClick={() => setAndStoreStepName(name)}
             className={styles.SendingStepButtons}
           >
-            {name}
+            {label}
```
shrirupa: When I set `window.onboardingStartSendingTest = false`  
this is what I see 
Screen Shot 2020-03-05 at 10 21 37 AM



we might want to make sure that label is displayed only when it is not a falsy value label ? label : name.

PR #1474, opened 2020-03-05
FE-888 Hibana Snackbar and Utility Function

gfriedmansp:
### What Changed
 - Added a utility function to remove styled system props before displaying OGMatchbox components
 - Added a wrapper for the hibana Snackbar component
 - Replaced the usages of Snackbar with the wrapper in:
   - ConsentBar
   - Alert
 - Updated the ProgressBar wrapper to use the utility function and updated tests

### How To Test
 - Ensure Alerts look good (can trigger those in various places throughout the app, success, danger, etc.)
 - Test that the consent bar looks right - The hibana view will never be seen in production, technically because you can't login without consenting and the default is that hibana is off, but you can test it manually by editing the `HibanaContext.isHibanaEnabled` default to be true locally


src/components/cookieConsent/components/ConsentBar.js
```diff
@@ -1,11 +1,21 @@
 import React from 'react';
-import { UnstyledLink, Snackbar } from '@sparkpost/matchbox';
+import { UnstyledLink } from '@sparkpost/matchbox';
+import { Snackbar } from 'src/components/matchbox';
```
gfriedmansp: Changed this

src/components/cookieConsent/components/ConsentBar.js
```diff
@@ -1,11 +1,21 @@
...
 
-export const ConsentBar = ({ onDismiss }) => <div className={styles.CookieConsent}>
-  <div className={styles.ConsentBar}>
-    <Snackbar maxWidth={700} onDismiss={onDismiss}>
-      We use cookies to optimize your experience, analyze traffic, and personalize content. To learn more, please visit our <UnstyledLink external to={'https://www.sparkpost.com/policies/privacy/'}>Cookie Policy</UnstyledLink>. By using our site without disabling cookies, you consent to our use of them.</Snackbar>
+export const ConsentBar = ({ onDismiss }) => (
+  <div className={styles.CookieConsent}>
+    <div className={styles.ConsentBar}>
+      <Snackbar maxWidth={700} onDismiss={onDismiss} status="default">
+        <span>
```
gfriedmansp: Wrapped text in spans

src/components/matchbox/tests/Snackbar.test.js
```diff
@@ -0,0 +1,65 @@
...
+
+  it('renders the Hibana version of the Snackbar component correctly when hibana is enabled', () => {
+    useHibana.mockImplementationOnce(() => [{ isHibanaEnabled: true }]);
+
+    const wrapper = subject({ mb: 0, pt: 5 });
+
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('only renders passed in children when hibana is not enabled', () => {
```
nicklemmon: I don't think this is an accurate test name - maybe copy+pasted from another file?

src/helpers/tests/hibana.test.js
```diff
@@ -0,0 +1,48 @@
...
+      py: 5,
+      color: 'red',
+    };
+
+    expect(omitSystemProps(props, ['color'])).toEqual({ className: 'test', color: 'red' });
+  });
+
+  // This might not actually be a thing that happens, but wanted to prove that it's not
+  // a generic truthy check to add props back in
+  it('removes styled system props except whitelisted ones that are set, regardless if null', () => {
```
nicklemmon: I think this is a good idea 👍

src/components/matchbox/tests/Snackbar.test.js
gfriedmansp: Whoops, I'm not sure if this test is even necessary anyways. Copied that one over before doing the one testing that individual props aren't passed.

gfriedmansp: Indeed, re-looking at these tests - feels like snapshot is the better way to go than the first test.

src/components/cookieConsent/components/tests/ConsentBar.test.js
```diff
@@ -1,23 +1,26 @@
...
   });
 
-  it('should wire up onDismiss', () => {
-    const wrapper = mount(<ConsentBar {...props} />);
-    wrapper.find('a').at(1).simulate('click');
-    expect(props.onDismiss).toHaveBeenCalled();
+  it('should render correctly in hibana', () => {
+    useHibana.mockImplementationOnce(() => [{ isHibanaEnabled: true }]);
+    const wrapper = subject();
+    expect(wrapper).toMatchSnapshot();
```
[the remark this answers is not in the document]
gfriedmansp: Very much agree - meant to remove this test.

PR #1478, opened 2020-03-05
FE-848 Add upgrade plan button for free users.

kchu1337:
### What Changed
 - Added link to change plan page in the top nav for free users.
 - Added upgrade label to the billing section of the account nav dropdown.

### How To Test
 - Login to a free user account (or create a new acount).
 - Check that there is a upgrade plan link on the top nav in orange. 
 - Check that there is an upgrade label in the account dropdown.
 - Login as a paid user. Check that the above do no appear for the paid user. 

### To Do
- [x] UX sign off

### DIFF
--- a/cypress/tests/integration/navigation/navigation.spec.js
+++ b/cypress/tests/integration/navigation/navigation.spec.js
@@ -92,7 +92,7 @@ describe('Mobile Navigation', () => {
     cy.wait('@stubbedAccountRequest'); // ...the request is made twice! So wait again for the second
   }
 
-  describe.only('navigation behavior', () => {
+  describe('navigation behavior', () => {
     beforeEach(() => {
       cy.viewport(500, 1000);
       cy.stubAuth();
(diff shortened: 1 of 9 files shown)


gfriedmansp:
Do we want the upgrade plan button to be visible in the mobile views (mobile nav or perhaps a banner across the top)?

cypress/tests/integration/navigation/navigation.spec.js
```diff
@@ -92,7 +92,7 @@ describe('Mobile Navigation', () => {
     cy.wait('@stubbedAccountRequest'); // ...the request is made twice! So wait again for the second
   }
 
-  describe('navigation behavior', () => {
+  describe.only('navigation behavior', () => {
```
gfriedmansp: Think we need to remove this

kchu1337:
> Do we want the upgrade plan button to be visible in the mobile views (mobile nav or perhaps a banner across the top)?

According to the AC in the ticket, it is only for Desktop users.

PR #1482, opened 2020-03-10
SA-1249: Add option to use inbox-placement-ea endpoint with user flag

kchu1337:
### What Changed
 - Added a helper function that dispatches a thunk that switches out the inbox placement url based on a user UI flag. 
 - Updated tests to match the changes.
 - Removed the async await in the tests. 

### How To Test
 - Create new user or login to a user under appteam account.
 - Add the user ui 'use_inbox_placement_ea' flag through the /users/{username}/control endpoint. 
 - Go to any inbox placement route in the UI. Check that the api calls use `/v1/inbox-placement-ea`

### TODO
- [ ] Add Feature flag to confluence Doc
- [ ] Add instructions on how to create new user with flag to ticket and inform SA team

### DIFF
--- a/src/actions/inboxPlacement.js
+++ b/src/actions/inboxPlacement.js
@@ -5,7 +5,7 @@ import { isUserUiOptionSet } from 'src/helpers/conditions/user';
 export const inboxPlacementApiRequest = ({ type, meta }) => {
   return (dispatch, getState) => {
     const { url } = meta;
-    const newUrl = isUserUiOptionSet('inbox_placement_v2_api')(getState())
+    const newUrl = isUserUiOptionSet('use_inbox_placement_ea')(getState())
       ? url.replace('/inbox-placement', '/inbox-placement-ea')
       : url;
     return dispatch(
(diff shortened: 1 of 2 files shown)


src/actions/inboxPlacement.js
```diff
@@ -1,8 +1,27 @@
 import sparkpostApiRequest from 'src/actions/helpers/sparkpostApiRequest';
 import { PLACEMENT_FILTER_TYPES } from 'src/pages/inboxPlacement/constants/types';
+import { isUserUiOptionSet } from 'src/helpers/conditions/user';
+
+export const inboxPlacementApiRequest = ({ type, meta }) => {
```
bkemper: This is a really awesome way to handle this feature toggle!

src/actions/tests/inboxPlacement.test.js
```diff
@@ -1,12 +1,15 @@
-import sparkpostApiRequest from 'src/actions/helpers/sparkpostApiRequest';
 import * as inboxPlacement from '../inboxPlacement';
-
+import { isUserUiOptionSet } from 'src/helpers/conditions/user';
+jest.mock('src/helpers/conditions/user');
+isUserUiOptionSet.mockImplementation(() => () => false);
```
bkemper: Test setup should be done inside `describe` (unless it can't like `import`, `jest.mock`, etc.).  Jest protects us by not allowing this setup to leak into other tests, but there might be other modern day reasons for scoping it correctly (memory leak, concurrent tests, ???).

Also, if this is the default setup for the mock, it can be declared with the mock.

```
jest.mock('src/helpers/conditions/user', () => ({ isUserUiOptionSet: jest.fn(() => () => false) }));
```

PR #1483, opened 2020-03-10
AC-1230: Changed ChoosePlan to use bundles endpoints

PatrickSison:
### What Changed
- Changed ChoosePlan and PlanPicker to be functional components
- Changed ChoosePlan to use new endpoints and depend on joined data
- PlanPicker now accepts bundles instead of account plans

### How To Test
 - Check that onboarding billing flow works as intended. 
   - New to free
   - New to paid
   - [Marketing Site](https://www.sparkpost.com/pricing/) with preselected plan to app
 - Test with slow connection to verify that it won't render form without data being ready.

### Notes
 With the joining of data that isn't called on app load, we have to now wait for data + joining of the data by the selector.
  - Selector now returns empty if both parts of data aren't ready
 -  Part of the loading state checks that bundles is empty (if data not loaded, or both parts not loaded).

### To Do
- [ ] Fix tests
- [ ] Handle error states


src/selectors/onboarding.js
```diff
@@ -1,20 +1,29 @@
...
+  createStructuredSelector({
+    loading: state =>
+      Boolean(
+        state.account.loading ||
+          state.billing.plansLoading ||
+          state.billing.countriesLoading ||
+          state.billing.bundlesLoading ||
+          state.billing.bunldePlansLoading,
+      ),
+    bundlesLoading: state => state.billing.bundlesLoading,
```
shrirupa: Are we using is anywhere ?

src/selectors/accountBillingInfo.js
```diff
@@ -118,6 +118,9 @@ export const selectTieredVisiblePlans = createSelector([selectVisiblePlans], pla
 export const selectAvailableBundles = createSelector(
   [bundleSelector, bundlePlanSelector, selectIsSelfServeBilling],
   (bundles, plans, isSelfServeBilling) => {
+    if (!plans.length) {
```
shrirupa: is this line required? 🤔

shrirupa: looks like even when plans are not loaded this selector would still return [] without this condition

shrirupa:
Few unit tests seem to be failing.

PR #1485, opened 2020-03-10
FE-922 - Adds new styles for Desktop Navigation

nicklemmon:
[FE-922](https://jira.int.messagesystems.com/browse/FE-922)

### What Changed
- Implements new navigation design for Hibana and adds some supporting tests:

Screen Shot 2020-03-11 at 8 54 39 AM

### How To Test
1. Sign in to the app, swap to Hibana using the `appteam` account
1. Resize the window, verify the desktop nav renders at 960px and above, verify the mobile nav (just a placeholder at the moment) renders below 960px
1. Verify nav items render appropriately according to user permissions (toggling between the OG theme and the Hibana theme is a good way to check)
1. Verify that nav links with children nav items render those children in a sub navigation section with a dark background
1. Verify that the active state for links renders appropriately for each link in the app (in the case of subnav items, verify the primary nav
1. Verify the navigation is keyboard navigable - tab through the elements and use the enter key to navigate - also verify the skip navigation link moves the user to the main
1. Verify Heroku banner doesn't overlap the navigation. _Ask me for details on how to get this to render_

**NOTE:** Some navigation items are being moved elsewhere, and others exist in the navigation only temporarily:
- 'Alerts' and 'Configuration' are moving to the accounts dropdown (FE-924)
- Many subnav items will eventually be removed from the subnav as part of a re-architecture of the navigation (UX is not yet ready with that design, so it hasn't been implemented)
- Nav items currently have some unusual wrapping behavior due to the current navigation architecture - the plan is to resolve this issue as part of the updated inbox tracker rollout

### To Do
- [X] Double check alternate layouts
- [X] Fix loading positioning introduced by flexbox changes on the dashboard - **will be addressed as part of <Loading/> component update**
- [X] Check conditional routing to RV
- [X] Fix vertical alignment issues (go to `/dashboard` on a large-ish viewport)
- [ ] Incorporate feedback
- [x] Figure out how to offset `<BoomerangBanner/>`
- [ ] Does `<PendingCancelGlobalBanner />` need to be handled?


shrirupa:
On Recipient Validation Page when I switch tabs the subnav disappears 
Screen Shot 2020-03-12 at 2 57 44 PM

shrirupa:
I don't know if it is in the scope of this ticket, but all pages look fine with new nav except the templates edit page, since the top nav completely disappears in that case.
Screen Shot 2020-03-12 at 3 07 10 PM

shrirupa:
Developer role should be able to see the recipients tab, but in hibana nav clicking on recipients redirects to 404

src/components/hibana/DesktopNavigation.module.scss
```diff
@@ -0,0 +1,128 @@
...
+.PrimaryNavLayout {
+  display: grid;
+  grid-template-columns: 175px auto auto;
+  grid-gap: spacing(600);
+  padding-top: spacing(400);
+  padding-bottom: spacing(400);
+}
+
+// Some specificity issues with links
+.SkipLink.SkipLink.SkipLink {
```
shrirupa: 😨

src/components/layout/LayoutHibana.module.scss
```diff
@@ -0,0 +1,17 @@
+@import '~@sparkpost/design-tokens-hibana/tokens';
```
shrirupa: just curious but what all is being imported from here ? `media-query(lg)`

src/components/loading/Loading.js
```diff
@@ -25,29 +24,33 @@ Loading.displayName = 'Loading';
...
-export const LoadingSVG = ({ className = '', size = '' }) => (
-  <svg
-    xmlns="http://www.w3.org/2000/svg"
-    viewBox="25 25 50 50"
-    className={classnames(styles.CircleWrapper, styles[size], className)}
-  >
-    <circle className={styles.Circle} cx="50" cy="50" r="20" vectorEffect="non-scaling-stroke" />
-  </svg>
-);
+export const LoadingSVG = ({ className = '', size = '' }) => {
```
shrirupa: only prettier changes, right ?

src/config/navItems/index.js
```diff
@@ -91,7 +91,7 @@ export default [
   },
   {
     label: 'Recipients',
-    to: '/lists',
+    to: '/recipient-validation/list', // TODO: How can this be handled using grants?
```
shrirupa: this comment still needs to be handled ?

src/components/layout/LayoutHibana.module.scss
nicklemmon: and `spacing()` and `color()`!

src/components/loading/Loading.js
nicklemmon: Let me comment on the things that are actual changes

src/components/sparkPost/SparkPost.js
```diff
@@ -9,19 +9,27 @@ import styles from './SparkPost.module.scss';
...
 
-const Logo = ({ type, ...rest }) => (
+const Logo = ({ type, className, ...rest }) => (
   <svg
     {...rest}
     role="img"
     aria-labelledby="sparkpost-full-logo-title"
-    className={classnames(styles.Logo, type && styles[type])}
-    height='100%' xmlns="http://www.w3.org/2000/svg" viewBox="0 0 499 130"
+    className={classnames(className, styles.Logo, type && styles[type])}
```
nicklemmon: Actual change here

src/config/navItems/index.js
nicklemmon: It looks like everyone has access to this route now? Maybe? There aren't many grants in our navigation at this point so I'm hoping you can answer that question.

nicklemmon:
> On Recipient Validation Page when I switch tabs the subnav disappears
> 
> * on List Tab
> 
> Screen Shot 2020-03-12 at 2 59 20 PM
> 
> * on Single Address Tab
> 
> Screen Shot 2020-03-12 at 2 57 44 PM

I'm going to need to re-work how I was handling this - I think I know a way!

nicklemmon:
> I don't know if it is in the scope of this ticket, but all pages look fine with new nav except the templates edit page, since the top nav completely disappears in that case.
> Screen Shot 2020-03-12 at 3 07 10 PM

Yep, this will have to be addressed when we get to the templates page

nicklemmon:
@shrirupa I believe I addressed the issue, but the way it's handled isn't ideal...I basically made a note that we'll need a sort of technical debt cleanup ticket for the nav once the swap is made permanent. What do you think?

Also, it looks like there is this `<PendingCancellationBanner/>` component in the old nav? I'm not really sure why it's a part of the nav, but we probably need to handle that here as well? Are we still using that component in the app?

PR #1496, opened 2020-03-16
FE-923 - Adds Hibana mobile navigation

nicklemmon:
[FE-923](https://jira.int.messagesystems.com/browse/FE-923)

### What Changed
- Adds new mobile navigation
- Adds `cypress-plugin-tab` which provides the `cy.tab()` command for keyboard interaction testing
- Adds `react-focus-lock` for managing tab traps within modal-like interfaces

### How To Test
1. Reduce the screen width below 960px
1. Open the navigation using the button in the top right
1. Verify the appropriate links are present and that subsections render as expandable/collapsible sections
1. Verify that all portions of the navigation are accessible using a keyboard only
1. Verify that when the navigation is open, a tab trap is implemented - when using `tab` and `shift + tab` the user cannot tab outside of the navigation - but _only_ when the navigation is open!

### To Do
- [x] Incorporate feedback
- [x] Should I break `MobileNavigation.js` in to multiple files? I've been leaning towards larger files with multiple components lately, but that might just be me!


src/components/hibana/Navigation.js
```diff
@@ -32,12 +32,18 @@ function Navigation(props) {
     })
     .filter(Boolean); // Removes undefined entries
 
+  const mobileNavItems = navItems.filter(item => {
+    if (item.label !== 'Dashboard' && item.label !== 'Alerts') {
+      return item;
```
bkemper: 😄 This is always misleading to me.  It works because you are returning an object that is truthy or undefined which is falsy, but `filter` is really expecting a boolean.

src/components/hibana/Navigation.js
```diff
@@ -32,12 +32,18 @@ function Navigation(props) {
     })
     .filter(Boolean); // Removes undefined entries
 
+  const mobileNavItems = navItems.filter(item => {
+    if (item.label !== 'Dashboard' && item.label !== 'Alerts') {
```
bkemper: Should we add this to config instead of matching on label?

src/components/hibana/MobileNavigation.js
```diff
@@ -1,11 +1,251 @@
...
+          {navItems.map((item, index) => {
+            const isExpandable = item => {
+              if (item.children && item.children.length) {
+                return true;
+              }
+
+              return false;
+            };
+
+            if (isExpandable(item)) {
```
bkemper: This is clearly a preference, but it might be easier to read if it was a variable instead of a fuction.

```
const isExpandable = Boolean(item.children && item.children.length);
```

src/components/hibana/MobileNavigation.js
```diff
@@ -1,11 +1,251 @@
...
+              return (
+                <Menu.Item
+                  onClick={() => setIsOpen(false)}
+                  variant="primary"
+                  id={item.label.toLowerCase().replace(/\s/g, '')}
```
bkemper: This could be the key too.

src/components/hibana/MobileNavigation.js
```diff
@@ -1,11 +1,251 @@
...
+                to={item.to}
+                key={`mobile-nav-item-${index}`}
+              >
+                {item.label}
+              </Menu.Item>
+            );
+          })}
+
+          <div className={styles.Footer}>
+            {/* TODO: Populate these as part of FE-924 */}
```
bkemper: I guess this will be addressed in FE-924, but when a primary menu item is expanded, I was unable to scroll to see the entire footer.

Screen Shot 2020-03-17 at 11 25 28 PM

src/components/hibana/MobileNavigation.js
```diff
@@ -1,11 +1,251 @@
...
+            <Menu.Item variant="secondary" to="/templates">
+              Placeholder
+            </Menu.Item>
+          </div>
+        </Menu.Wrapper>
```
bkemper: 👏 For use of compound components!

src/components/hibana/MobileNavigation.js
```diff
@@ -1,11 +1,251 @@
...
+      </FocusLock>
+    </div>
+  );
+}
+
+function Wrapper(props) {
+  const { children, onOverlayClick, isOpen } = props;
+  const { motionDuration_medium } = tokens;
+  const menuEl = useRef(null);
+  const motionDurationMS = Number(motionDuration_medium.replace(/\D/g, '')) * 100;
```
bkemper: Should this live here or should it be abstracted into a helper?

nicklemmon:
> I left a bunch of notes and recommendations, but nothing that would stop this from moving forward. Address what you want and merge when ready.

Nice - definitely some good things to look at. I'll address today and hopefully we can deploy this afternoon?

bkemper:
Sounds good

cypress/support/commands.js
```diff
@@ -9,6 +9,7 @@
 
 import '@testing-library/cypress/add-commands';
 import 'cypress-file-upload';
+import 'cypress-plugin-tab';
```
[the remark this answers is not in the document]
nicklemmon: It looks like it's authored by a Cypress core team contributor? They do have it on their roadmap to implement in Cypress, so should only be temporary. It's been in Matchbox for several months now and haven't seen any problems!

src/components/hibana/MobileNavigation.js
```diff
@@ -1,11 +1,251 @@
...
     <div className={styles.MobileNavigation} data-id="mobile-navigation">
-      {/* markup is entirely placeholder for FE-923 */}
-      <p>Mobile Navigation</p>
+      {/* Prevents scrolling behind the navigation UI when opened */}
+      {isOpen
+        ? document.body.setAttribute('style', 'overflow-y: hidden')
+        : document.body.setAttribute('style', 'overflow-y: initial')}
+
+      {/* Keyboard focus is trapped within the element when opened */}
+      <FocusLock disabled={!isOpen}>
```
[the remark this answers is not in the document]
nicklemmon: That's a good idea - I think I know a slightly better article on the topic to include.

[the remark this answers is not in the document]
nicklemmon: Actually this is pretty good compared to what I had...good find

src/components/hibana/MobileNavigation.js
nicklemmon: That is cleaner!

src/components/hibana/MobileNavigation.js
nicklemmon: Yep, will address in FE-924

src/components/hibana/MobileNavigation.js
nicklemmon: Probably a good idea to make a helper

src/components/hibana/Navigation.js
nicklemmon: I'm actually re-working this for FE-924 - I'll end up updating how it's handled for the desktop nav as well

src/components/hibana/Navigation.js
nicklemmon: oh dang you're right. I always get this mixed up with how `.map` works. let me update it.

src/components/layout/LayoutHibana.module.scss
```diff
@@ -7,7 +7,11 @@
 }
 
 .content.content {
-  padding: spacing(400);
+  padding: spacing(800) spacing(400);
```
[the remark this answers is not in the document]
nicklemmon: Yep - I would like to address with the global loading component ticket when that day comes.

nicklemmon:
@bkemper Updated a few things - if you want to take another look let me know. We can roll it out this afternoon.

src/components/hibana/MobileNavigation.js
```diff
@@ -45,15 +45,9 @@ export default function MobileNavigation({ navItems }) {
...
-              if (item.children && item.children.length) {
-                return true;
-              }
+            const isExpandable = Boolean(item.children && item.children.length);
 
-              return false;
-            };
-
-            if (isExpandable(item)) {
+            if (isExpandable) {
```
bkemper: 🙏

PR #1497, opened 2020-03-16
FE-886 - Component wrapper for Matchbox Select

shrirupa:
FE-886 - Component wrapper for Matchbox Select

### What Changed
 - Select has a single, globally accessible point of entry within the app
 - Select is only imported from Matchbox in one place - replaced all instances of the Select directly imported from matchbox with this new global component
- New props available to the Hibana version of the component are exposed - and subsequently ignored by the original component (https://jira.int.messagesystems.com/browse/FE-816)
- Ensure no unusual positioning UI defects are introduced as a result of this change - both for when using Hibana and when using the default theme

### How To Test
 - enable hibana ui flag 
`{
	"options": {
		"ui":{
"hibana":{
	"hasThemeControls": true
}
		}
	}
 }`
- Check all instances of Select for ui defects with and without hibana enabled

### To-do
- [ ] Address Feedback

### DIFF
--- a/src/components/reduxFormWrappers/SelectWrapper.js
+++ b/src/components/reduxFormWrappers/SelectWrapper.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Select } from '@sparkpost/matchbox';
+import { Select } from 'src/components/matchbox';
 
 export default function SelectWrapper({ input, meta, ...rest }) {
   const { active, error, touched } = meta;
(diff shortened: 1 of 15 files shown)


src/components/reduxFormWrappers/SelectWrapper.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Select } from '@sparkpost/matchbox';
+import { Select } from 'src/components/matchbox';
```
nicklemmon: Screen Shot 2020-03-16 at 8 50 21 PM

Could we see about handling the space between these rows when using Hibana? This is the "Events" advanced search page.

PR #1498, opened 2020-03-16
AC-1264 - (UI) Single address tab

shrirupa:
AC-1264 - (UI) Single address tab

### What Changed
- Tab 1 for the single address form is created
- The form changes endpoints based on Right to Be forgotten or opting out
- Subaccount is assigned to header
- Cypress tests are added for Single Recipient Page

### How To Test
 - Enable 'data_privacy' flag.
 - Navigate to /account/data-privacy/single-recipient.
 - Check if it matched mocks https://sparkpost.invisionapp.com/share/BCVKAJRH4E9#/screens/400849700
 - Submit request for rtbf and Opt out check if requests are sent as doucmented - https://github.com/SparkPost/sparkpost-admin-api-documentation/pull/452/files

### To Do
- [ ] Address Feedback

### DIFF
--- a/src/pages/dataPrivacy/components/SingleRecipientTab.js
+++ b/src/pages/dataPrivacy/components/SingleRecipientTab.js
@@ -16,7 +16,7 @@ const requestTypes = [
     label: 'Right to be forgotten',
   },
   {
-    value: 'oo-third-party',
+    value: 'opt-out',
     label: 'Opt-out of third-party use',
   },
 ];
@@ -25,17 +25,21 @@ export function SingleRecipientTab(props) {
   const onSubmit = values => {
     switch (values.requestType) {
       case 'rtbf':
-        props.submitRTBFRequest({
-          recipients: [values.address],
-          request_date: new Date().toISOString(),
-        });
+        props
+          .submitRTBFRequest({
+            recipients: [values.address],
+            request_date: new Date().toISOString(),
+          })
+          .then(() => props.reset());
         break;
-
+      case 'opt-out':
       default:
-        props.submitOptOutRequest({
-          recipients: [values.address],
-          request_date: new Date().toISOString(),
-        });
+        props
+          .submitOptOutRequest({
+            recipients: [values.address],
+            request_date: new Date().toISOString(),
+          })
+          .then(() => props.reset());
         break;
     }
   };
--- a/src/pages/dataPrivacy/components/SingleRecipientTab.js
+++ b/src/pages/dataPrivacy/components/SingleRecipientTab.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import { connect } from 'react-redux';
 import { reduxForm } from 'redux-form';
 import { RadioGroup } from 'src/components';
 import { Field } from 'redux-form';
@@ -7,6 +8,7 @@ import { TextFieldWrapper } from 'src/components';
 import { Label } from 'src/components/matchbox';
 import { required, email, maxLength } from 'src/helpers/validation';
 import { Button } from '@sparkpost/matchbox';
+import { submitRTBFRequest, submitOptOutRequest } from 'src/actions/dataPrivacy';
 
 const requestTypes = [
   {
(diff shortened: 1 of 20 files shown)


src/pages/snippets/EditPage.js
```diff
@@ -128,7 +126,10 @@ export default class EditPage extends React.Component {
                   />
                 </Panel.Section>
                 {canViewSubaccountSection && (
-                  <SubaccountSection newTemplate={false} disabled={disabled} />
+                  <Panel.Section>
+                    {' '}
```
PatrickSison: Nitpick, but we should probably remove this space

src/pages/dataPrivacy/components/SingleRecipientTab.js
```diff
@@ -0,0 +1,118 @@
...
+            include_subaccounts: include_subaccounts,
+          })
+          .then(() => props.reset());
+        break;
+      case 'opt-out':
+      default:
+        props
+          .submitOptOutRequest({
+            recipients: [values.address],
+            request_date: new Date().toISOString(),
```
PatrickSison: Same thing with this in regards to date

src/pages/dataPrivacy/components/SingleRecipientTab.js
```diff
@@ -0,0 +1,118 @@
...
+        <form onSubmit={props.handleSubmit(onSubmit)}>
+          <Field
+            component={RadioGroup}
+            name="requestType"
+            title="Select Compliance Type"
+            options={requestTypes}
+            disabled={false}
+            validate={[required]}
+          />
+          <Label id="email-address-field">Email Address</Label>
```
PatrickSison: Should this be recipient email address?

PatrickSison: I think if we were doing the prop for this, it should also be `htmlFor` instead of id for the label, but I think we should swap to using the prop `label` on Field, which gets passed to TextFieldWrapper

src/pages/dataPrivacy/components/SingleRecipientTab.js
```diff
@@ -0,0 +1,118 @@
...
+            validate={[required]}
+            createOptions={createOptions}
+          />
+
+          <ButtonWrapper>
+            <Button color="orange" type="submit">
+              Submit Request
+            </Button>
+            <div className={styles.CancelButtonContainer}>
+              <Button onClick={props.reset}>Cancel</Button>
```
PatrickSison: Was the decision that the cancel button would reset the form or navigate away?

src/pages/dataPrivacy/components/SingleRecipientTab.js
```diff
@@ -0,0 +1,118 @@
...
+          />
+          <SubaccountSection
+            newTemplate={true}
+            disabled={false}
+            validate={[required]}
+            createOptions={createOptions}
+          />
+
+          <ButtonWrapper>
+            <Button color="orange" type="submit">
```
PatrickSison: We should probably implement a disable in case the request takes a long time. Same with the other fields.

src/pages/dataPrivacy/components/SingleRecipientTab.js
```diff
@@ -0,0 +1,118 @@
...
+      : values.assignTo === 'subaccount'
+      ? values.subaccount.id
+      : subaccountItems[values.assignTo];
+    const include_subaccounts = values.assignTo === 'shared';
+    switch (values.requestType) {
+      case 'rtbf':
+        props
+          .submitRTBFRequest({
+            recipients: [values.address],
+            request_date: new Date().toISOString(),
```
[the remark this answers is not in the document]
shrirupa: missed that comment. Good catch!

src/pages/dataPrivacy/components/SingleRecipientTab.js
shrirupa: I think the decision was to reset the form and also confirmed on uxfe to call it "Clear"

PatrickSison:
There's a couple of tests that need to be updated as well

PR #1505, opened 2020-03-19
FE-894: Component Wrapper for Matchbox TextField

bkemper:
### What Changed

Here is a list of all the components that were switched to use the wrapper.

Screen Shot 2020-03-18 at 9 51 12 PM

The id prop was added to all instances that were missing it.  Also, it is important to note that none of the components were using styled props.

### How To Test

 * Only plan picker page, select a new plan, enter "blah" into the promo code text field, and click "Apply" [should get a correctly styled text field and error message]
Screen Shot 2020-03-18 at 10 18 28 PM

 * On account setting page, click "Provision SSO" button [should have a correctly styled copy field]
Screen Shot 2020-03-18 at 10 29 50 PM

Test Considerations

- [x] Confirm it is alright `resize="both"` and `type="text"` are missing from snapshots. (This is expected because the wrapper doesn't have defaultProps)
- [x] Confirm nothing wrong is being spread in src/pages/templates/components/FromEmailInput.js
- [x] Confirm data-id is still set correctly on src/pages/inboxPlacement/components/TestDetails.js
- [x] Confirm data-lpignore prop still works in src/components/enableTfaForm/EnableTfaForm.js
- [x] Confirm prefix, suffix, and readonly props work in FilterDropdown.js
- [x] Confirm wrappers still work correctly, SliderFieldWrapper, TextFieldWrapper, etc.
- [x] Confirm src/components/typeahead/Typeahead.js and src/pages/reports/components/Typeahead.js
- [x] Confirm type="password" is set correctly for DisableTfaModal
- [x] Confirm labelHidden is set correctly for CustomReports
- [x] Confirm onKeyPress prop works in 
- [x] rows and resize for RequestBlock and ResponseBlock
- [x] multiline src/pages/inboxPlacement/SeedListPage.js


bkemper:
@shrirupa Good catch on the stories.  I updated them.  Plus, I just resolved all the merge conflicts by rebasing on FE-881 branch.  This should be ready to go.

PR #1507, opened 2020-03-19
AC-1286: Tabs for join page

PatrickSison:
### What Changed
 - Add tabs for the join page

### How To Test
 - Log out and go to join
 - Modify JoinPage code by modifying `const brand` to `CROSS_LINK_MAP['spc']` or `CROSS_LINK_MAP['spceu']`
 - Verify the link navigates to the appropriate app.


src/pages/join/components/SignUpTabs.js
```diff
@@ -0,0 +1,23 @@
...
+      tabs={[
+        {
+          content: 'SparkPost',
+          to: isSPCEU && `${brand.url}${location.pathname}${location.search}`,
+        },
+        {
+          content: 'SparkPost EU',
+          to: !isSPCEU && `${brand.url}${location.pathname}${location.search}`,
+        },
+      ]}
```
shrirupa: Conditions for redirection on line 11 and 15 do not seem to redirect correctly,
So this is how it can be reproduced, so when I set `config.crossLinkTenant` to spc and set isSPCEU to false, then the SparkPost tab gets highlighted correctly but clicking on SparkPost EU Tab redirects to 'app.sparkpost.com' instead of app.eu.sparkpost.com.
This is happening because of following - 
When we are on app.sparkpost.com and clicking on Sparkpost EU tab the brand object that get passed is 
`{ label: '', url: LINKS.SPC_US_URL }` and hence it gets redirected incorrectly.

src/pages/join/components/SignUpTabs.module.scss
```diff
@@ -0,0 +1,20 @@
+@import '~@sparkpost/matchbox/src/styles/config.scss';
+
+.Tabs {
+  margin-bottom: spacing();
+
+  > div {
+    text-align: center;
+    height: 80px;
+    padding: 0;
```
shrirupa: Can we add 
```
a:visited {
        color: #828287;
    }
```
to avoid getting the blue color when link is clicked, I am not sure though that this will happen in prod as they are different builds 🤷‍♀ .

PR #1510, opened 2020-03-20
AC-1299 -  (UI) Join page should submit email_opt_in flag in creation_params to POST /account

shrirupa:
AC-1299 -  (UI) Join page should submit email_opt_in flag in creation_params to POST /account

### What Changed
 - Added email_opt_in flag in creation_params to POST /account

### How To Test
 - Create a account and check if email_out_in flag is sent in the body for POST /account

### To Do
- [ ] Address feedback

### DIFF
--- a/src/pages/join/JoinPage.js
+++ b/src/pages/join/JoinPage.js
@@ -56,7 +56,7 @@ export class JoinPage extends Component {
       ...accountFields,
       sfdcid,
       salesforce_data: { ...attributionData, email_opt_out: !values.email_opt_in },
-      creation_params: creationParams,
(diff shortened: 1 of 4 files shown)


src/pages/join/JoinPage.js
```diff
@@ -56,7 +56,7 @@ export class JoinPage extends Component {
       ...accountFields,
       sfdcid,
       salesforce_data: { ...attributionData, email_opt_out: !values.email_opt_in },
-      creation_params: creationParams,
+      creation_params: { ...creationParams, email_opt_in: !!values.email_opt_in },
```
bkemper: Instead of double bang, can we use `Boolean`?

PR #1511, opened 2020-03-20
FE-867: [follow-up]

bkemper:
Some follow-up items that may or may not have been introduced in FE-867

 - Remove additional padding from SummaryTable cells 
 - Left align Subaccount dropdown text
Screen Shot 2020-03-20 at 1 00 45 PM

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -8,12 +8,12 @@ import configureStore from './store';
 import config from './config';
 import ErrorTracker from './helpers/errorTracker';
 
-import Providers from './Providers';
-
+// note, these must be imported before App and Providers
 import './critical.scss';
 import './index.scss';
 
 import App from './App';
+import Providers from './Providers';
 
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -8,12 +8,12 @@ import configureStore from './store';
...
 import ErrorTracker from './helpers/errorTracker';
 
-import Providers from './Providers';
-
+// note, these must be imported before App and Providers
 import './critical.scss';
 import './index.scss';
 
 import App from './App';
+import Providers from './Providers';
```
nicklemmon: This alone should solve the problem - you can remove the other CSS changes.

nicklemmon:
Just pushed removing those `!important`s. Should be good to go.

bkemper:
@nicklemmon Looks good to me.

bkemper:
Teamwork!

PR #1525, opened 2020-03-25
SA-1272 Add timezone and precision to report URLs

gfriedmansp:
### What Changed
 - Added timezone and precision URL params to the report URLs for sharing

### How To Test
 - Open a report (/reports/summary as example)
   - See that your local timezone (based on browser/system clock) is added to the query params
   - See that precision (based on the to and from dates) is added to the query params
   - See that metrics API requests send that timezone along with the request
- Change the timezone in the query params to UTC
  - See that metrics API requests send that timezone along with the request

### DIFF
--- a/src/actions/brightback.js
+++ b/src/actions/brightback.js
@@ -32,7 +32,13 @@ export function prepBrightback(accountDetails) {
 
 export function loadPrerequisiteMetrics() {
   const { from, to } = getRelativeDates('7days', { roundToPrecision: false });
-  const metricsRequest = getQueryFromOptions({ from, to, metrics: METRICS });
+  const metricsRequest = getQueryFromOptions({
+    from,
+    to,
+    precision: 'day',
+    timezone: 'UTC',
+    metrics: METRICS,
+  });
   return sparkpostApiRequest({
     type: 'BRIGHTBACK_METRICS',
     meta: {
--- a/src/actions/reportOptions.js
+++ b/src/actions/reportOptions.js
@@ -1,16 +1,18 @@
+import moment from 'moment';
+
 import {
   fetchMetricsDomains,
   fetchMetricsCampaigns,
   fetchMetricsSendingIps,
   fetchMetricsIpPools,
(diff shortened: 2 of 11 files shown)


src/actions/brightback.js
```diff
@@ -32,7 +32,13 @@ export function prepBrightback(accountDetails) {
 
 export function loadPrerequisiteMetrics() {
   const { from, to } = getRelativeDates('7days', { roundToPrecision: false });
-  const metricsRequest = getQueryFromOptions({ from, to, metrics: METRICS });
+  const metricsRequest = getQueryFromOptions({
+    from,
+    to,
+    precision: 'day',
+    timezone: 'UTC',
```
gfriedmansp: Manually setting these defaults since they now come from URL on reports.

src/actions/reportOptions.js
```diff
@@ -120,14 +125,24 @@ export function refreshReportOptions(update) {
...
 
-    // calculate relative dates if range is not "custom"
-    if (update.relativeRange && update.relativeRange !== 'custom') {
-      update = { ...update, ...getRelativeDates(update.relativeRange) };
+    if (update.relativeRange) {
+      if (update.relativeRange !== 'custom') {
+        const { from, to } = getRelativeDates(update.relativeRange);
+        const precision = getPrecision(from, moment(to));
+        update = { ...update, from, to, precision };
+      } else {
```
gfriedmansp: This entire else will be removed once we have a precision dropdown in the date picker. For now, it will always choose precision based on smallest precision based on the from/to dates.

src/helpers/metrics.js
```diff
@@ -2,37 +2,46 @@ import moment from 'moment';
 import _ from 'lodash';
```
gfriedmansp: Only non-prettier changes in this file are in `getQueryFromOptions`

PatrickSison:
@gfriedmansp Nvm, realized I was on old branch, and not the `SA-1272-1` 🤦‍♂

PR #1536, opened 2020-03-31
FE-928: Re-style wrapper global components

shrirupa:
FE-928: Re-style wrapper global components

### What Changed

**There are some changes to CardTitle original css, switched the color to #55555a or color('gray',2) or sparkpost brand color, which seems to be the one used in mocks and also some adjustments to font size to accommodate different font sizes**
link to mocks-
https://sparkpost.invisionapp.com/share/BMU7BR332NK#/screens/386841857
https://sparkpost.invisionapp.com/share/XDSMDD078FT#/screens

wrapper global components for Hibana updated:
```
<ButtonWrapper/>
<Callout/>
<Card/>
<Empty/>
```
- Updated component styles for Hibana

### How To Test
 - `<Card>`
    - /onboarding/recipient-validation
Screen Shot 2020-03-31 at 1 14 54 PM
    

- /dashboard

Screen Shot 2020-03-31 at 11 42 33 AM

Screen Shot 2020-03-31 at 1 19 06 PM

- `<Empty />`
Screen Shot 2020-03-31 at 5 07 05 PM

### To Do
- [ ] Address feedback


src/components/card/Card.js
```diff
@@ -1,15 +1,43 @@
...
+    <Box display="inline-block" fontSize="500">
+      {children}
+    </Box>
+  );
+};
+
+export const CardTitle = ({ children }) => {
+  const [state] = useHibana();
+  const { isHibanaEnabled } = state;
+  if (!isHibanaEnabled) return <h2 className={styles.CardTitle}>{children}</h2>;
```
bkemper, line 36: Should we make any changes to the current component?

src/components/buttonWrapper/ButtonWrapper.js
```diff
@@ -1,10 +1,13 @@
...
-const ButtonWrapper = ({ children }) => (
-  <div className={styles.ButtonWrapper}>
-    {children}
-  </div>
-);
+export const ButtonWrapper = ({ children }) => {
+  const [state] = useHibana();
+  const { isHibanaEnabled } = state;
+  if (!isHibanaEnabled) return <div className={styles.ButtonWrapper}>{children}</div>;
+  return <Box marginTop="500">{children}</Box>;
```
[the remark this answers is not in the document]
nicklemmon: This is a token, actually - passing in these values as strings is how you access the tokens from the theme provider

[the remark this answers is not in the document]
shrirupa: Tokens, I think 🤷‍♀ .

src/components/buttonWrapper/ButtonWrapper.js
```diff
@@ -1,10 +1,13 @@
...
+};
 
 export default ButtonWrapper;
```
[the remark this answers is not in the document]
shrirupa: I can add a snapshot test.

src/components/card/Card.js
shrirupa: I wasn't completely sure of this. but did this for RVBundlePage. 
We are using <CardTitle> in only two places in the app (**RVBundlePage** and **IpForms**, FeatureStepList is disabled so not in use anymore).

src/components/callout/Callout.js
```diff
@@ -1,20 +1,37 @@
...
+        <div>
+          <h3 className={styles.Title}>{title}</h3>
+          {children && <p className={styles.Content}>{children}</p>}
+        </div>
+      </div>
+    );
+  return (
+    <Box height={height} display="flex" justifyContent="center" alignItems="center">
+      <Box>
+        <Text as="h3" lineHeight="2">
```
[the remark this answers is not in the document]
nicklemmon: Agreed with Brian here!

src/components/callout/Callout.js
```diff
@@ -1,20 +1,37 @@
...
+      <div className={styles.Callout} style={{ height }}>
+        <div>
+          <h3 className={styles.Title}>{title}</h3>
+          {children && <p className={styles.Content}>{children}</p>}
+        </div>
+      </div>
+    );
+  return (
+    <Box height={height} display="flex" justifyContent="center" alignItems="center">
+      <Box>
```
nicklemmon, line 21: Is this extra `<Box/>` doing anything?

src/components/callout/Callout.test.js
```diff
@@ -1,20 +1,27 @@
 import React from 'react';
-import { shallow } from 'enzyme';
+import { mount } from 'enzyme';
```
nicklemmon, line 2: 👏

src/components/callout/Callout.test.js
```diff
@@ -1,20 +1,27 @@
...
-      </Callout>
+    const wrapper = mount(
+      <TestApp>
+        <Callout title="Example">Here is the deal...</Callout>
+      </TestApp>,
     );
 
-    expect(wrapper).toMatchSnapshot();
+    expect(wrapper.find('h3').text()).toEqual('Example');
+    expect(wrapper.find('p').text()).toEqual('Here is the deal...');
```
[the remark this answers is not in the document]
nicklemmon: Agreed!

src/components/card/Card.js
nicklemmon: Yeah the original semantics were more appropriate, actually - `<div role="heading" aria-level="3"/>` is the equivalent of an `<h3>` not an `<h2>`, but I think in general we don't want to change the original components.

src/components/card/Card.module.scss
```diff
@@ -11,9 +11,7 @@
 
 .CardTitle {
   display: block;
-  color: color(gray, 3);
```
nicklemmon, line 14: Are these styles not needed for the original component?

src/components/empty/Empty.js
```diff
@@ -1,24 +1,36 @@
...
       <Panel sectioned title={title}>
         <h6 className={styles.Center}>{message}</h6>
       </Panel>
     );
   }
+  return (
+    <Panel sectioned title={title}>
+      <Box textAlign="center" color="gray.400">
+        <Block size={28} />
+        <Text as="h3" color="gray.400">
```
nicklemmon, line 23: 👍 This is a better choice semantically. Nice.

src/components/card/Card.js
```diff
@@ -1,15 +1,43 @@
...
-    {children}
-  </div>
-);
+export const Card = ({ children, textAlign }) => {
+  const [state] = useHibana();
+  const { isHibanaEnabled } = state;
+  if (!isHibanaEnabled)
+    return <div className={classNames(styles.CardContainer, styles[textAlign])}>{children}</div>;
+  return (
+    <Box border="400" padding="400" textAlign={textAlign}>
```
nicklemmon, line 13: I think the Hibana color styles need to be applied to these components - right now those colors are being inherited from global styles - not ideal!

src/components/empty/Empty.js
```diff
@@ -1,24 +1,36 @@
...
+  if (!isHibanaEnabled) {
     return (
       <Panel sectioned title={title}>
         <h6 className={styles.Center}>{message}</h6>
       </Panel>
     );
   }
+  return (
+    <Panel sectioned title={title}>
+      <Box textAlign="center" color="gray.400">
```
nicklemmon, line 21: This color does not meet color contrast requirements! Even if the mocks indicate a lighter color, we should use the lightest possible value that still passes. You can use this to check the values, or install the Axe Chrome extension:
https://color.review/

Screen Shot 2020-04-01 at 12 34 50 PM

src/components/empty/Empty.js
nicklemmon: Same as above - the color is too light

src/pages/onboarding/RVBundlePage.js
```diff
@@ -55,33 +61,35 @@ const RVBundlePage = () => {
...
-              <p>
-                With our monthly pay-as-you-go plan using tiered pricing, the more you validate, the
-                less you pay per message.
-              </p>
-              <p>
-                <UnstyledLink onClick={() => setModalOpen(true)} className={styles.link}>
-                  See Pricing
-                </UnstyledLink>
-              </p>
+              </CardTitle>
```
nicklemmon, line 67: If this one _needs_ to be an `<h2>` then you could make a `level` prop that defaults to `"3"`?

src/components/empty/tests/Empty.test.js
```diff
@@ -1,10 +1,16 @@
...
   it('should render', () => {
-    const wrapper = shallow(<Empty title={'Empty Title'} message={'nothing here to see'} />);
-    expect(wrapper).toMatchSnapshot();
+    const wrapper = mount(
+      <TestApp>
+        <Empty title={'Empty Title'} message={'nothing here to see'} />
+      </TestApp>,
+    );
+    expect(wrapper.find('Panel')).toHaveLength(1);
+    expect(wrapper.find('h6')).toHaveLength(1);
```
nicklemmon, line 14: Might be better to use `.toHaveTextContent()` again

src/components/buttonWrapper/ButtonWrapper.js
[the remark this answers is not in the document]
bkemper: My question wasn't very good.  Take two...

Do we want to define style-system props with static values verse explicitly with the token?

```
<Box marginTop="500">
```

vs 

```
<Box marginTop={tokens.spacing_500}>
```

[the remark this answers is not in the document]
nicklemmon: The static values are actually some syntactic sugar that `styled-system` provides - it directly hooks in to the theme provider so we don't have to import the tokens in to every component, reducing a lot of boilerplate.

Not only that, but it's a form of restriction that prevents someone from passing in the wrong token. If you pass in tokens explicitly, technically this works:
```
<Box marginTop={tokens.borderRadius_200}/>
```

Furthermore, if the token name changes, it will break - using `"200"` keeps those two things separate!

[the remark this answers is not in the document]
bkemper: Cool, can you summarize this decision and share with everyone in public guild channel?

src/components/card/Card.module.scss
shrirupa: Since I had changed the `<CardTitle>` component so I updated these to make sure they look same visually. Reverting it.

src/components/card/Card.js
```diff
@@ -1,15 +1,63 @@
...
+const HibanaCardTitle = ({ children }) => {
+  return (
+    <Text as="h3" mb="100" color="gray.900">
+      {children}
+    </Text>
+  );
+};
+
+export const Card = ({ children, textAlign }) => {
+  return useHibanaToggle(OGCard, HibanaCard)({ children, textAlign });
```
nicklemmon, line 54: It's probably simpler to just pass in `(props)` instead of doing the destructuring an extra time. Just my two cents - not a critical problem or anything.

src/components/buttonWrapper/ButtonWrapper.js
```diff
@@ -1,10 +1,17 @@
...
+};
+
+const HibanaButtonWrapper = ({ children }) => {
+  return <Box marginTop="500">{children}</Box>;
+};
+export const ButtonWrapper = ({ children }) => {
+  return useHibanaToggle(OGButtonWrapper, HibanaButtonWrapper)({ children });
+};
 
 export default ButtonWrapper;
```
nicklemmon, line 17: Do both need to be exported in this way? I would think you could just stick with the default export.

src/components/callout/Callout.js
```diff
@@ -1,8 +1,25 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import styles from './Callout.module.scss';
+import useHibanaToggle from 'src/components/matchbox/useHibanaToggle';
+import { Box, Text } from 'src/components/matchbox';
 
-const Callout = ({ children, height = '220px', title }) => (
+const HibanaCallout = ({ children, height = '220px', title }) => (
+  <Box height={height} display="flex" justifyContent="center" alignItems="center">
+    <Box>
```
nicklemmon, line 9: Is the extra `<Box/>` necessary? Maybe the flexbox styles aren't necessary? I don't know, just seems a little unusual! I could be wrong.

src/components/callout/Callout.test.js
```diff
@@ -1,20 +1,27 @@
...
 
   it('renders title and details', () => {
-    const wrapper = shallow(
-      <Callout title="Example">
-        Here is the deal...
-      </Callout>
+    const wrapper = mount(
+      <TestApp>
+        <Callout title="Example">Here is the deal...</Callout>
+      </TestApp>,
```
nicklemmon, line 21: 👏 Death to shallow mounting! Nice work.

src/components/card/Card.js
```diff
@@ -1,15 +1,63 @@
...
+const OGCardTitle = ({ children, level = '3' }) => (
+  <div
+    className={classNames(
+      styles.CardTitle,
+      level === '2' && styles.level2,
+      level === '3' && styles.level3,
+    )}
+    role="heading"
+    aria-level={level}
+  >
```
nicklemmon, line 24: 👏

src/components/card/Card.js
```diff
@@ -1,15 +1,63 @@
...
+  return (
+    <Box display="inline-block" fontSize="500" color="gray.900">
+      {children}
+    </Box>
+  );
+};
+
+const HibanaCardTitle = ({ children }) => {
+  return (
+    <Text as="h3" mb="100" color="gray.900">
```
nicklemmon, line 47: Probably worth supporting `level` here as well - might even mean doing `<Text as="div" role="heading" level={level}>`

src/components/matchbox/useHibanaToggle.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import { useHibana } from 'src/context/HibanaContext';
+
+export default function useHibanaToggle(OGComponent, HibanaComponent) {
+  const [state] = useHibana();
+  const { isHibanaEnabled } = state;
+  return props => (isHibanaEnabled ? <HibanaComponent {...props} /> : <OGComponent {...props} />);
+}
```
nicklemmon, line 8: Code looks good - I would think this should either live with Hibana code or in `src/hooks`? This doesn't really have much to do with Matchbox.

src/pages/onboarding/RVBundlePage.js
```diff
@@ -55,33 +61,35 @@ const RVBundlePage = () => {
         <Panel.Section>
           <div className={styles.grid}>
             <Card>
-              <h2>
-                <CheckCircle className={styles.icon} />
+              <CardTitle level="2">
```
nicklemmon, line 64: 👍 Nice and clean.

src/components/callout/Callout.js
```diff
@@ -1,8 +1,25 @@
...
+      {children && (
+        <Text as="p" color="gray.800">
+          {children}
+        </Text>
+      )}
+    </Box>
+  </Box>
+);
+
+const OGCallout = ({ children, height = '220px', title }) => (
```
nicklemmon, line 22: Screen Shot 2020-04-03 at 9 47 44 AM

It looks like `height` isn't making it for some reason?

shrirupa: 😱

src/components/card/Card.js
shrirupa: looks like aria-level does't work for `<Text>`

PR #1537, opened 2020-04-02
TR-2344: Links in template editor need noreferrer attribute

bkemper:
### What Changed
 - Set rel attribute on anchors in template preview

### How To Test
 - Create a new template
 - Set "HTML" content to `<a href="http://google.com">Click Here</a>`
 - Confirm link rendered in the preview has "rel" attribute set and when clicked it opens a new tab

### DIFF
--- a/src/pages/templates/components/PreviewFrame.js
+++ b/src/pages/templates/components/PreviewFrame.js
@@ -62,8 +62,10 @@ export default class PreviewFrame extends Component {
     // ...because DOM collections only array-like
     const anchors = [...anchorHTMLCollection];
     anchors.forEach(a => {
+      // must overwrite "rel" value to avoid a security vulnerability
       a.setAttribute('rel', 'noopener noreferrer');
-      a.setAttribute('target', '_blank');
+      // must use _parent because Firefox and other browsers do not support _blank from an iframe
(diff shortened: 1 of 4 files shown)


src/pages/templates/components/PreviewFrame.js
```diff
@@ -56,20 +53,25 @@ export default class PreviewFrame extends Component {
...
     );
 
     // Avoid loading links in iframe
     const anchorHTMLCollection = this.iframe.contentDocument.getElementsByTagName('a');
 
     // ...because DOM collections only array-like
     const anchors = [...anchorHTMLCollection];
-    anchors.forEach((a) => a.setAttribute('target', '_parent'));
+    anchors.forEach(a => {
+      a.setAttribute('rel', 'noopener noreferrer');
```
gfriedmansp, line 65: Not sure if it's anything we need to worry about, but this will completely overwrite the rel attribute if there was one set in the template.

bkemper: Good point.  I thought about it last night.  Since it is a security vulnerability and merging user set rel values might negate the effectiveness of ours, I decide to overwrite.  I'll add a comment.

PR #1543, opened 2020-04-06
AC-1267 - Release UI for data privacy

shrirupa:
AC-1267 - Release UI for data privacy

### What Changed
 - Removed ui flag data_privacy
 - Added Data and Privacy Page to hibana nav
 - Replaced title with label in SubaccountSection, MultipleRecipientsTab, SingleRecipientsTab

### How To Test
 - Check if Data and Privacy page is visible without enabling feature flag.(only to admin users)
 - Check if SubaccountSection used in templates, snippets and Data Privacy page looks the same after the replacement of title with label.
 - Check if the Data and Privacy Page is visible under Hibana Nav.

### To Do
- [ ] Address Feedback

### DIFF
--- a/src/config/navItems/account.js
+++ b/src/config/navItems/account.js
@@ -2,8 +2,8 @@ import { OpenInNew, ExitToApp } from '@sparkpost/matchbox-icons';
 import { LINKS } from 'src/constants';
 import { openSupportPanel } from 'src/actions/support';
 import { isHeroku } from 'src/helpers/conditions/user';
-import { all, hasGrants, not, any } from 'src/helpers/conditions';
-import { isAccountUiOptionSet, onPlan } from 'src/helpers/conditions/account';
+import { hasGrants, not, any } from 'src/helpers/conditions';
+import { onPlan } from 'src/helpers/conditions/account';
 
 /***
  * These values are pulled in through the accountNavItems selector, which will map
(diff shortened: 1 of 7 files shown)


shrirupa:
something went wrong with travis and it did not run the unit tests for last commit (all seem to be passing locally )

src/config/navItems/account.js
```diff
@@ -131,6 +131,7 @@ export const hibanaAccountNavItems = [
   // usage,  // In the mock, but doesn't exist yet?
   ...billingLinks,
   users,
+  dataAndPrivacy,
```
bkemper, line 134: 👍

shrirupa:
> Is the documentation going to be ready for this launch or should we hide the link?
> 
> Screen Shot 2020-04-06 at 1 52 20 PM

this documentation is going out with AC-1303, it should be out today in some time.

bkemper:
Another item, not related to this code, but the feature in general, the subaccount field is not initialized with a default value and doesn't require the user to select an option.

bkemper:
I just kicked off another build to see if it passes.

shrirupa:
> Another item, not related to this code, but the feature in general, the subaccount field is not initialized with a default value and doesn't require the user to select an option.

🤔 when we don't send the subaccountid, api treats it as master and all subaccounts. but may be we should have the default state to be Master, like other forms ?

shrirupa:
Added in a fix as redux-form breaks when a empty form when it is 
submitted -> cleared -> submitted.
ref - https://github.com/redux-form/redux-form/issues/2971

PR #1547, opened 2020-04-06
SA-1216 Add "Per day/minute/hour" for unique metrics based on precision

gfriedmansp:
### What Changed
 - Added a `uniqueLabel` for each item in `metricsRollupPrecisionMap`
 - Added a `isUniquePerTimePeriod` option for metrics
 - If the metric is `isUniquePerTimePeriod`, the Legend will display the current precision's `uniqueLabel` i.e. "Unique clicks per minute"

### How To Test
 - Add metrics to the summary report that have the word Unique in them. ie. Unique confirmed opens
 - Add metrics that are not unique
 - Ensure the Legend displays the correct "Unique Confirmed Opens per hour" based on precision.
 - Ensure the Legend does not display the "per" precision times for non unique metrics
 - Ensure these only are when you have the metrics rollup feature flag turned on.


src/pages/reports/summary/components/ChartHeader.js
```diff
@@ -38,10 +38,10 @@ export default class ChartHeader extends Component {
 
     return (
       <Grid className={styles.ChartHeader}>
-        <Grid.Column xs={12} md={5} lg={5}>
+        <Grid.Column xs={12} md={9}>
```
kchu1337, line 41: This and connecting to redux for Legend changed the snapshot

src/pages/reports/summary/components/Legend.js
```diff
@@ -1,20 +1,51 @@
...
+    const metricsRollupPrecisionMap = getConfig('metricsRollupPrecisionMap');
+    const precisionObj = metricsRollupPrecisionMap.find(p => p.value === precision);
+    return (
+      <div className={styles.Legend}>
+        {metrics &&
+          metrics.map(metric =>
+            renderRollupMetric(metric, precisionObj ? precisionObj.uniqueLabel : ''),
+          )}
+      </div>
+    );
```
kchu1337, line 40: You can do something like 
```js
    const metricsRollupPrecisionMap = getConfig('metricsRollupPrecisionMap');
    const precisionObj = metricsRollupPrecisionMap.find(p => p.value === precision);
    return (
      <div className={styles.Legend}>
        {metrics &&
          metrics.map(metric =>
            renderRollupMetric(metric, (precisionObj && useMetricsRollup) ? precisionObj.uniqueLabel : ''),
          )}
      </div>
    );
```
And then you can remove the other return and only have 1 render helper function.
Up to you though.

src/pages/reports/summary/components/tests/Legend.test.js
```diff
@@ -1,18 +1,86 @@
...
+    return shallow(<Legend {...defaults} {...props} />);
+  };
+
+  it('should not render with "per" labels without metrics rollup', () => {
+    const wrapper = subject();
+    expect(wrapper).not.toHaveTextContent('B per minute');
+    expect(wrapper).not.toHaveTextContent('C per minute');
+  });
+
+  it('should render using "per minute" labels for unique metrics', () => {
```
kchu1337, line 61: Could name the test a bit more specific like ```should render using "per minute" labels for unique metrics for minute precision``` Same with the other 2 tests

src/pages/reports/summary/components/ChartHeader.js
gfriedmansp: Looks like I had stashed this change and never unstashed it :/

PR #1550, opened 2020-04-07
FE-882: Component Wrapper for Matchbox Popover

bkemper:
### What Changed
 - Swapped Popover for wrapper
 - Set `id` and `aria-controls` for a11y (to guarantee ids are unique, add useUniqueId) H/T @nicklemmon 
 - Upgraded to 4.0.0-hibana.13

### How To Test
 - Confirm ActionPopover works correctly on Snippet and User list pages
 - Confirm action popover on snippet index page
 - Confirm version selection on AB Test edit page of cancelled test

### To Do


src/components/datePicker/DatePicker.js
```diff
@@ -255,6 +254,7 @@ export default class DatePicker extends Component {
 
         <TextField
           id={`date-field-${id}`}
+          aria-describedby={`popover-${id}`}
```
bkemper, line 257: @nicklemmon What if you have two triggers?

src/components/matchbox/tests/Tooltip.test.js
```diff
@@ -9,7 +9,7 @@ describe('Tooltip Matchbox component wrapper', () => {
   const subject = props => {
     const defaultProps = { width: '50px' };
     return shallow(
-      <Tooltip {...defaultProps} {...props}>
+      <Tooltip id="test-tooltip" {...defaultProps} {...props}>
```
bkemper, line 12: 🤷‍♂ This test was failing. #LIB

src/components/datePicker/DatePicker.js
nicklemmon: @bkemper - so `aria-describedby` isn't really appropriate for the popover component - that makes more sense for tooltips. My comments from that other PR only mention using `aria-expanded`:

> 1. It looks like aria-expanded is missing from the example on the button element...I know this won't be supplied by the component (though I think ideally it would - maybe when we update component APIs)

The problem is, a `<TextField/>` triggering a popover doesn't really make sense semantically - the triggering element for a popover should be a button.

src/pages/abTesting/components/StatusPanel.js
```diff
@@ -47,6 +49,7 @@ export const StatusPanel = ({ test, version, id, subaccountId, latest, subaccoun
           <VersionSelector current={version} id={id} latest={latest} subaccountId={subaccountId} />
         ),
         color: 'orange',
+        'aria-describedby': 'popover-version-selector',
```
bkemper, line 52: Not my favorite way to accomplish this, but see todo comment above.

src/components/actionPopover/ActionPopover.js
```diff
@@ -1,15 +1,15 @@
...
 
-const ActionPopover = ({ actions }) => (
+const ActionPopover = ({ actions, id }) => (
   <div style={{ textAlign: 'right' }}>
     <Popover
+      id={id}
       left
       trigger={
-        <Button flat>
+        <Button aria-describedby={id} flat>
```
nicklemmon, line 12: Again, doesn't really make sense to use `aria-describedby` - `aria-controls` would make more sense.

src/components/matchbox/tests/Tooltip.test.js
nicklemmon: I believe the wrapper was completed ahead of the published component - one of the downsides of doing that!

src/components/datePicker/DatePicker.js
```diff
@@ -269,6 +268,7 @@ export default class DatePicker extends Component {
 
     return (
       <Popover
+        id={`popover-${id}`}
```
bkemper, line 271: Tagged with an id to satisfy the requirement, but didn't set aria-controls as advised.

src/components/filterDropdown/FilterDropdown.js
```diff
@@ -73,6 +72,7 @@ export class FilterDropdown extends Component {
     return (
       <div>
         <Popover
+          id={`filter-dropdown-popover-${id}`}
```
bkemper, line 75: Tagged with an id to satisfy the requirement, but didn't set aria-controls as advised.

src/components/hibana/AccountPopover.js
```diff
@@ -18,6 +17,7 @@ function AccountPopover(props) {
   return (
     <div className={styles.AccountPopover} data-id="desktop-navigation-account-popover">
       <Popover
+        id="desktop-navigation-account-popover-contents"
```
bkemper, line 20: 🤷‍♂ Couldn't think of a better name.

bkemper:
@nicklemmon This should be ready for review.  The only thing I didn't add that we discussed earlier is we need to follow-up to add aria-expanded props to the trigger.

nicklemmon:
> @nicklemmon This should be ready for review. The only thing I didn't add that we discussed earlier is we need to follow-up to add aria-expanded props to the trigger.

nicklemmon:
Whoops, closed this PR accidentally. Re-opened.

src/components/hibana/AccountPopover.js
nicklemmon: didn't want to use your hook here?

src/components/filterDropdown/FilterDropdown.js
nicklemmon: You could go ahead and use your unique ID hook here. Every ID needs to be unique whether or not you're using any `aria-` stuff with it!

src/components/matchbox/tests/Popover.test.js
```diff
@@ -0,0 +1,31 @@
...
+import Popover from '../Popover';
+
+jest.mock('src/context/HibanaContext');
+
+describe('Popover Matchbox component wrapper', () => {
+  const subject = props =>
+    shallow(
+      <Popover
+        id="test-popover"
+        trigger={<button aria-describedby="test-popover">Click</button>}
```
nicklemmon, line 13: Might be good to remove the `aria-describedby` here just to reduce the likelihood of confusion for future devs.

src/components/tooltip/Tooltip.js
```diff
@@ -1,22 +1,30 @@
...
-import { Popover } from '@sparkpost/matchbox';
+import { Popover } from 'src/components/matchbox';
+import useUniqueId from 'src/hooks/useUniqueId';
 import style from './Tooltip.module.scss';
 
 // note, Matchbox's Tooltip has static dimensions that prevent us from doing fun things
 // todo, should use Matchbox's Tooltip
 const Tooltip = ({ children, content, initialOpen = false }) => {
   const [isOpen, setOpen] = useState(initialOpen);
+  const uniqueId = useUniqueId('tooltip-');
```
nicklemmon, line 10: No need for the `-` here - the hook should already be handling it.

src/pages/abTesting/components/StatusPanel.js
```diff
@@ -25,6 +26,7 @@ const VersionSelector = ({ current, latest, id, subaccountId }) => {
 
   return (
     <Popover
+      id="popover-version-selector"
```
nicklemmon, line 29: Should probably use the unique ID here - are there not multiple instances of the component?

bkemper:
Both `<AccountPopover />` and `<StatusPanel />` are only used once, so I'm going to leave the static assignment.  I would change `<FilterDropdown />`, but it is not a functional component which might be a little more work than I want to take on for this PR.

PR #1551, opened 2020-04-07
FE-930: Table Components

PatrickSison:
### What Changed
 - Updated table components hibana styling

### How To Test
 - Verify that tables still work in default styling
 - Verify that tables match snapshots in Hibana
 Unique Cases
 - Message Events (separate pagination)
 - Templates (no save CSV button)

### To Do
- [ ] ~Fix message events table~
- [x] Fix padding on PerPageButtons
- [x] Fix SortLabel


package.json
```diff
@@ -71,7 +71,7 @@
     "@sparkpost/boomerang": "^1.0.6",
     "@sparkpost/design-tokens-hibana": "npm:@sparkpost/design-tokens@^1.0.0-beta.16",
     "@sparkpost/matchbox": "^3.9.1",
-    "@sparkpost/matchbox-hibana": "npm:@sparkpost/matchbox@^4.0.0-hibana.10",
+    "@sparkpost/matchbox-hibana": "npm:@sparkpost/matchbox@^4.0.0-hibana.11",
```
nstanard, line 74: looks like this already got bumped to 12

src/components/collection/SortLabel.js
```diff
@@ -25,4 +26,32 @@ const SortLabel = ({ label, direction, ...rest }) => {
...
 };
 
+const HibanaSortLabel = ({ label, direction, ...rest }) => {
+  return (
+    <UnstyledLink {...rest}>
+      <Text fontWeight="500" color="gray.800">
+        <Box alignItems="center" display="flex">
+          <span>{label}</span>
+          <Box>
+            {direction !== 'desc' && (
```
PatrickSison, line 36: Need to figure out a better pattern for this. This is very off mock, but working on fixing right now

nstanard: Are you trying to avoid something like this:
```
<Box>
    {direction === 'desc' ? <ArrowDropDown />  : <ArrowDropUp />}
</Box>
```
?

src/components/collection/NameCell.js
```diff
@@ -16,4 +17,19 @@ const NameCell = ({
...
 
+const HibanaNameCell = ({
+  id,
+  name = slugToFriendly(id), // this default is an edge case, most resources will include a name
+  to,
+}) => [
+  <p key="name">
+    <PageLink to={to}>{name}</PageLink>
+  </p>,
+  <p key="id">ID: {id}</p>,
```
nicklemmon, line 28: I don't think a `<p>` is semantically appropriate here - better to just use `<Text>` in this case.

src/components/collection/Pagination.js
```diff
@@ -1,11 +1,12 @@
...
-import { Pagination } from 'src/components/matchbox';
+import { Box, Pagination } from 'src/components/matchbox';
 import styles from './Pagination.module.scss';
 import { DEFAULT_PER_PAGE_BUTTONS } from 'src/constants';
 import PerPageButtons from './PerPageButtons';
 import SaveCSVButton from './SaveCSVButton';
+import useHibanaToggle from 'src/hooks/useHibanaToggle';
 
-class CollectionPagination extends Component {
+class OGCollectionPagination extends Component {
```
nicklemmon, line 9: I don't think duplicating all of the component logic is a great idea - it might be better to just use the `useHibana` hook and check for `isHibanaEnabled` in the render method instead.

src/components/collection/PerPageButtons.js
```diff
@@ -31,4 +32,42 @@ const PerPageButtons = ({
...
 
+const HibanaPerPageButtons = ({
+  perPage,
+  perPageButtons = DEFAULT_PER_PAGE_BUTTONS,
+  onPerPageChange,
+  totalCount,
+}) => {
+  if (totalCount <= Math.min(...perPageButtons)) {
+    return null;
+  }
```
nicklemmon, line 43: Same here - there has to be a way to not duplicate the component logic.

src/components/collection/SortLabel.js
nicklemmon: If you need custom CSS, check out the `useHibanaOverride` hook. It is used in the app in `src/components/layout/App.js` if you want to see an example.

src/components/collection/SortLabel.js
```diff
@@ -25,4 +26,32 @@ const SortLabel = ({ label, direction, ...rest }) => {
   );
 };
 
+const HibanaSortLabel = ({ label, direction, ...rest }) => {
+  return (
+    <UnstyledLink {...rest}>
```
nicklemmon, line 31: This technically isn't a link - from an HTML perspective, it's a `<button>` - right now these elements are not keyboard accessible because they can't be focused. If you change the markup to be a button, that should handle it.

Could do `<Box as="button">`? Or you could try leveraging `<Button/>` but overriding some of the sizing/spacing?

src/components/collection/PerPageButtons.js
```diff
@@ -31,4 +32,42 @@ const PerPageButtons = ({
...
+            flat={perPage !== buttonAmount}
+            key={buttonAmount}
+            name="collection-per-page-button"
+            onClick={() => onPerPageChange(buttonAmount)}
+            size="small"
+            width="600"
+          >
+            {buttonAmount}
+          </Button>
+        ))}
```
nicklemmon, line 63: Could you add some `<ScreenReaderOnly>` content inside the `<Button>`?
```
<Button>
  {buttonAmount}
  <ScreenReaderOnly>per page</ScreenReaderOnly>
</Button>
```

src/components/collection/FilterBox.js
```diff
@@ -28,3 +29,34 @@ export default function CollectionFilterBox(props) {
...
+  const placeholderText = placeholder || `Filter results e.g. ${getRandomExampleSearch(props)}`;
+
+  const text = (
+    <>
+      <Label id="collection-filter-box">
+        <ScreenReaderOnly>Filter By</ScreenReaderOnly>
+      </Label>
+
+      <TextField
+        label="Filter By"
```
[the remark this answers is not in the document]
nicklemmon: Actually, ignore this comment. Things are fine as they are until Matchbox is updated.

src/components/collection/SortLabel.js
nstanard: Are you actively changing this area? I'm getting this currently while testing:

```
Warning: validateDOMNesting(...): <div> cannot appear as a descendant of <p>.
    in div (created by Context.Consumer)
    in StyledComponent (created by Box)
    in Box (at Box.js:9)
    in Box (at SortLabel.js:33)
    in p (created by Context.Consumer)
    in StyledComponent (created by HibanaText)
    in HibanaText (at Text.js:11)
    in Text (at SortLabel.js:32)
    in a (created by Context.Consumer)
    in StyledComponent (created by HibanaText)
    in HibanaText (created by HibanaUnstyledLink)
    in HibanaUnstyledLink (at UnstyledLink.js:14)
    in UnstyledLink (at SortLabel.js:31)
    in HibanaSortLabel (at useHibanaToggle.js:7)
    in SortLabel (at HeaderLabel.js:36)
    in HeaderLabel (at Head.js:19)
    in th (created by Context.Consumer)
    in StyledComponent (created by styled.th)
    in styled.th (created by HibanaTableHeaderCell)
```

nstanard: Nick made some really great comments, if you wanna let me know when you change this specific stuff around I can take another look. Or if you wanna rubber duck debug/collaborate just let me know!

PatrickSison: > Are you trying to avoid something like this:
> 
> ```
> <Box>
>     {direction === 'desc' ? <ArrowDropDown />  : <ArrowDropUp />}
> </Box>
> ```
> 
> ?

I was trying to follow the mocks which had one arrow if it was being sorted on, but both arrows if it wasn't being sorted on.

src/components/collection/SortLabel.js
```diff
@@ -1,11 +1,16 @@
 import React from 'react';
```
PatrickSison, line 1: Let me know if there's a better pattern @nicklemmon and @nstanard . 2 things to accomplish with this was:

1.  Have the styles be updated (and keep icons working)
2. Swap the UnstyledLink to button

I didn't want to introduce the button into the original so I used the useHibanaToggle to keep separate components, but I still needed some original styles as is currently, so I also used useHibanaOverride

nstanard: LGTM! @nicklemmon thoughts?

src/components/collection/CursorPaging.js
```diff
@@ -40,4 +41,39 @@ export const CursorPaging = ({
...
+  nextDisabled,
+  handleFirstPage,
+  perPage,
+  totalCount,
+}) => (
+  <Box display="flex" alignItems="center">
+    <Button flat disabled={currentPage === 1} key="rewind" onClick={handleFirstPage}>
+      <FirstPage />
+      <ScreenReaderOnly>Return to First Page</ScreenReaderOnly>
+    </Button>
```
nicklemmon, line 58: The hover state is kind of weird here - no horizontal padding on either side. Any idea why?

Screen Shot 2020-04-10 at 3 53 21 PM

PatrickSison: Currently with mocks + original look, the label lines up with the column as it's left justified. Should we add a negative margin to the left of the button and keep the padding inside the button?

src/components/collection/FilterBox.js
```diff
@@ -28,3 +29,34 @@ export default function CollectionFilterBox(props) {
...
+
+function HibanaCollectionFilterBox(props) {
+  const { initialValue, placeholder, wrapper, onChange, onBlur = () => {} } = props;
+  const placeholderText = placeholder || `Filter results e.g. ${getRandomExampleSearch(props)}`;
+
+  const text = (
+    <>
+      <Label id="collection-filter-box">
+        <ScreenReaderOnly>Filter By</ScreenReaderOnly>
+      </Label>
```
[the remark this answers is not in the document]
nicklemmon: Actually, I'm not sure this even needs to have two components? There basically aren't any style differences? Just the use of `label` props? Might be fun to just use the `useHibana` hook and conditionally render that one prop rather than the entire component.

src/components/collection/SortLabel.js
nicklemmon: I think this is a fine way to handle it - I left a comment about how the hover state horizontal padding is weird. The text goes to the edge of the button.

src/components/collection/TableCollection.js
```diff
@@ -5,14 +5,28 @@ import Collection from './Collection';
...
+      <Table>{props.children}</Table>
+    </div>
+  </Panel>
+);
+
+const TableWrapper = props => {
+  return useHibanaToggle(OGTableWrapper, HibanaTableWrapper)(props);
+};
+
 const TableBody = props => <tbody>{props.children}</tbody>;
```
nicklemmon, line 30: This is a great way to handle it

src/components/collection/CursorPaging.js
nicklemmon: The code might be ugly, but would look a lot nicer!

nstanard: This seems to be updated/correct on my local - marking this as resolved

src/components/collection/SortLabel.js
nstanard: This looks to be updated.

src/components/collection/SortLabel.js
nstanard: I'm fairly certain these comments are all resolved. Everything looks good locally.

src/components/collection/SortLabel.js
```diff
@@ -25,4 +30,37 @@ const SortLabel = ({ label, direction, ...rest }) => {
...
+  const styles = useHibanaOverride(OGStyles, HibanaStyles);
+
+  const classes = cx(
+    styles.SortLabel,
+    (direction === 'asc' || direction === 'desc') && styles.selected,
+  );
+
+  return (
+    <Button className={classes} flat {...rest}>
+      <Text fontWeight="500" color="gray.800">
```
nicklemmon, line 43: This renders as a `<p>` tag which isn't right - should be a `<span>`

src/components/collection/SortLabel.js
```diff
@@ -25,4 +30,37 @@ const SortLabel = ({ label, direction, ...rest }) => {
...
+        {direction !== 'desc' && (
+          <span className={styles.Up}>
+            <ArrowDropUp />
+          </span>
+        )}
+        {direction !== 'asc' && (
+          <span className={styles.Down}>
+            <ArrowDropDown />
+          </span>
+        )}
```
nicklemmon, line 56: The `<SortLabel/>` padding is still off - padding on the righthand side isn't the same as on the left. I think for Hibana maybe abandoning the absolute positioning in favor of flexbox would make more sense?

Screen Shot 2020-04-13 at 5 07 15 PM

src/components/collection/tests/CursorPaging.test.js
```diff
@@ -1,7 +1,9 @@
 import React from 'react';
 import { shallow } from 'enzyme';
 import CursorPaging from '../CursorPaging';
-
+import { render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import TestApp from 'src/__testHelpers__/TestApp';
 describe('CursorPaging', () => {
```
nicklemmon, line 7: 🥳 YAay

src/components/collection/tests/SortLabel.test.js
```diff
@@ -1,18 +1,23 @@
...
   // Can't test more than this
   // Need to find out how to get css-modules rendered in enzyme
   it('should render correctly', () => {
-    const wrapper = shallow(<SortLabel {...props}/>);
+    const wrapper = shallow(
+      <TestApp>
+        <SortLabel {...props} />
+      </TestApp>,
+    );
     expect(wrapper).toMatchSnapshot();
```
nicklemmon, line 21: You can use `.dive()` I believe to keep the snapshot the same. The same should apply for `CursorPaging.test.js`

src/components/collection/FilterBox.js
```diff
@@ -3,18 +3,26 @@ import { Label, Panel, TextField, ScreenReaderOnly } from 'src/components/matchb
...
-      </Label>
+      {!isHibanaEnabled && (
+        <Label id="collection-filter-box" className={styles.FilterBoxLabel}>
+          <ScreenReaderOnly>Filter By</ScreenReaderOnly>
+        </Label>
+      )}
 
       <TextField
+        labelHidden
+        label={isHibanaEnabled ? 'Filter By' : undefined}
```
nicklemmon, line 25: Nice, thanks for doing that.

src/components/collection/tests/SortLabel.test.js
PatrickSison: I think this ends up with the same issue where the root component is the only one that's rendered deeper. I can rewrite the test to no longer use snapshots

nicklemmon: This looks good! Nice.

PR #1552, opened 2020-04-07
Test Hibana

bkemper:
This branch will give us an idea of what our tests look like when Hibana is enabled.


nicklemmon:
This can be closed now that we have everything running in CircleCI!

PR #1556, opened 2020-04-08
AC-1289: Added company name field and show password on join page

PatrickSison:
### What Changed
 - Added company name field
 - Added field view password field

### How To Test
 - Create account
 - Verify that show password works correctly
 -Add company name
 - Verify company name exists on the account information on logging in the app (network request)

### DIFF
--- a/src/pages/join/components/JoinForm.js
+++ b/src/pages/join/components/JoinForm.js
@@ -35,6 +35,7 @@ export class JoinForm extends Component {
     reCaptchaReady: false,
     reCaptchaInstance: null,
     recaptcha_response: null,
+    showPassword: false,
   };
 
   reCaptchaLoaded = () => {
@@ -65,7 +66,7 @@ export class JoinForm extends Component {
   render() {
     const { loading, pristine, invalid, submitting } = this.props;
 
-    const { reCaptchaReady } = this.state;
+    const { reCaptchaReady, showPassword } = this.state;
 
     const pending = loading || submitting || !reCaptchaReady;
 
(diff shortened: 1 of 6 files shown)


src/pages/join/components/JoinForm.js
```diff
@@ -95,6 +97,14 @@ export class JoinForm extends Component {
             />
           </Grid.Column>
         </Grid>
+        <Field
+          name="company_name"
+          component={TextFieldWrapper}
+          label="Company"
+          disabled={pending}
+          autoComplete="organization"
```
shrirupa, line 105: just curious, what does this autoComplete attribute do? https://www.w3schools.com/tags/att_input_autocomplete.asp looks like it should either be `on|off`

src/pages/join/components/JoinForm.js
```diff
@@ -157,6 +177,25 @@ export class JoinForm extends Component {
...
+
+function HibanaShowPasswordButton({ children, onClick }) {
+  return (
+    <Button primary outline onClick={onClick} data-id="show-password-button">
+      {children}
+    </Button>
+  );
+}
+
+function ShowPasswordButton(props) {
```
shrirupa, line 196: 🎉

src/pages/join/components/JoinForm.js
PatrickSison: I went by this: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete where it seems like it can have other attributes as well

PR #1557, opened 2020-04-08
FE-871: Empty State hibana wrapper

kchu1337:
### What Changed
 - Added EmptyState hibana wrapper.
 - Prettier changes.

### How To Test
 - Enable Hibana (or use appteam 105 staging).
 - Go to a url that doesn't exist, IE: /foo.
 - Check that everything looks alright.

### Note
 - Matchbox has not updated empty state yet but we can fix formatting later.


src/components/errorBoundaries/tests/ErrorBoundary.test.js
```diff
@@ -14,10 +14,14 @@ describe('Component: ErrorBoundary', () => {
   beforeEach(() => {
     props = {
       showAction: false,
-      children: jest.fn()
+      children: jest.fn(),
     };
 
-    wrapper = shallow(<ErrorBoundary {...props}><div>Children</div></ErrorBoundary>);
+    wrapper = shallow(
```
nstanard, line 20: I wonder - if ErrorBoundary is in the TestApp/Provider layer - would it be better to use that here instead?

PR #1561, opened 2020-04-10
FE-931 - Re-style date picker components

nicklemmon:
[FE-931](https://jira.int.messagesystems.com/browse/FE-931)

### What Changed

- Adds new `getVariantProps` function to the `<Button/>` component that maps the `"variant"` prop to a series of Matchbox `Button` props
- Adds new supporting documentation for the 2web2ui `<Button/>` component
- Update date picker components to use new, Hibana design tokens
- Incorporates default prop for the `<DatePicker/>` component for `id` using `prop-types` instead of assigning a value when destructuring
- Adds a new wrapping `<DatePicker/>` component that passes in `style` props to existing class components while using the `useHibanaOverride` hook
- Updates `<Navbar/>` component in the `<DatePicker/>` component to leverage HTML `<button>` elements instead of clickable icons (as well as `<ScreenReaderOnly/>` content) - makes these controls meet a11y standards
- Updates `<ManualEntryForm/>` component from a style perspective, adding a `style` prop that accepts passed in styles derived from the `useHibanaOverride` hook
- Updates relevant unit tests to match updated components - typically by mocking or passing in `styles` as a prop

### How To Test
- Verify `<Button/>` components in the OG version of the app were not negatively impacted by the introduction of the `variant` prop
- Verify components in the Jira ticket have been updated and leverage the updated Matchbox design tokens
- Verify date pickers in the app continue to behave in the same manor as their counterparts currently in production
- Routes to check:
  * `/reports/summary`
  * `/reports/bounce`
  * `/reports/rejections`
  * `/reports/accepted`
  * `/reports/delayed`
  * `/signals/health-score`
  * `/signals/spam-traps`
  * `/signals/engagement`
  * `/reports/engagement`
  * `/reports/message-events`
  * `/lists/suppressions`
  * `/inbox-placement`
  * `/blacklist/incidents`

### To Do

- [ ] Incorporate feedback


src/components/datePicker/DatePicker.js
```diff
@@ -1,5 +1,7 @@
 /* eslint-disable max-lines */
 import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import moment from 'moment';
```
nicklemmon, line 4: Just moved an external import above app imports

src/components/datePicker/DatePicker.js
```diff
@@ -10,14 +12,16 @@ import {
 } from 'src/helpers/date';
 import { roundBoundaries, getRollupPrecision, getPrecision } from 'src/helpers/metrics';
 import { Button, Error, Popover, Select, TextField, WindowEvent } from 'src/components/matchbox';
+import ButtonWrapper from 'src/components/buttonWrapper';
```
nicklemmon, line 15: Used this component to add spacing above the buttons in a standard way - works in both version of the app

src/components/datePicker/tests/DatePicker.test.js
```diff
@@ -23,6 +25,7 @@ describe('Component: DatePicker', () => {
     mockNow = utc({ year: 2017, month: 1, day: 15, hour: 12 });
     mockFrom = utc({ year: 2017, month: 1, day: 15, hour: 11 });
     props = {
+      id: 'date-picker',
```
bkemper, line 28: This should be set by defaultProps

src/components/matchbox/Button.js
```diff
@@ -26,10 +33,61 @@ const Group = props => {
...
+          outlineBorder: true,
+        };
+      case 'tertiary':
+        return {
+          color: 'blue',
+          flat: true,
+        };
+      case 'destructive':
+        return {
+          color: 'red',
```
bkemper, line 58: What if this needs to be flat?

src/components/matchbox/Button.js
```diff
@@ -7,12 +8,18 @@ import { omitSystemProps } from 'src/helpers/hibana';
 const Button = props => {
   const [state] = useHibana();
   const { isHibanaEnabled } = state;
+  const { variant } = props;
 
   if (!isHibanaEnabled) {
-    return <OGButton {...omitSystemProps(props, ['size', 'color'])} />;
+    return (
+      <OGButton
+        {...getVariantProps({ variant, isHibanaEnabled })}
```
bkemper, line 16: So, we will have to add "primary" to every form submit button?

Screen Shot 2020-04-13 at 3 40 42 PM

src/components/dateSelector/tests/DateSelector.test.js
```diff
@@ -38,28 +43,35 @@ describe('dateselector', () => {
 });
 
 describe('Navbar', () => {
+  beforeEach(() => {
+    useHibanaOverride.mockImplementationOnce(() => styles);
+  });
 
   it('should render correctly', () => {
-    expect(shallow(<Navbar/>)).toMatchSnapshot();
+    useHibanaOverride.mockImplementationOnce(() => styles);
```
bkemper, line 51: This looks like a duplicate.

src/components/dateSelector/tests/DateSelector.test.js
```diff
@@ -38,28 +43,35 @@ describe('dateselector', () => {
 });
 
 describe('Navbar', () => {
+  beforeEach(() => {
+    useHibanaOverride.mockImplementationOnce(() => styles);
```
bkemper, line 47: Is this needed?

src/components/datePicker/tests/DatePicker.test.js
nicklemmon: It's a default prop of the default export - but there is no default prop for the class component. Removing it and the snapshot tests fail (just one example):

```
       <label
          className="Label"
    -     htmlFor="date-field-date-picker"
    +     htmlFor="date-field-undefined"
        >
```

src/components/matchbox/Button.js
nicklemmon: I believe those additional props still work as overrides for the variants. Let me try it out.

nicklemmon: Yeah, this works:
```
<Button variant="destructive">DESTROY</Button>
<Button variant="destructive" flat>
  DESTROY
</Button>
```

Screen Shot 2020-04-13 at 9 01 58 PM

src/components/matchbox/Button.js
nicklemmon: I think so - I made the default that muted/tertiary/flat style as it seems by count that's the most common style in the app. Primary buttons are quite common - but probably the second most common.

src/components/dateSelector/tests/DateSelector.test.js
nicklemmon: Without it, the `styles` object is undefined and the component doesn't render.

nicklemmon:
@bkemper Ready for another pass when you have time. Or if you want to chat about some of the above, let me know

PR #1563, opened 2020-04-13
FE-937: Re-style Support global components

shrirupa:
FE-937: Re-style Support global components

### What Changed
 - Updated styles for `<HerokuMessage>, <NoIsuues>, <SearchPanel>, <AlgoliaResults>`
 - Added `<Stack>` to `SupportForm`. 

### How To Test
 - Check Support Modal.

### To Do
- [ ] Address Feedback


src/components/reduxFormWrappers/FileFieldWrapper.js
```diff
@@ -50,7 +50,7 @@ export default class FileFieldWrapper extends Component {
 
     return (
       <fieldset className={styles.Field}>
-        <Label id={input.id}>
+        <label id={input.id}>
```
PatrickSison, line 53: Should we not be using the label component? I do see though that the label hides.

src/components/support/components/SearchPanel.module.scss
```diff
@@ -10,3 +10,9 @@
   padding: 0;
   margin: 0;
 }
+
+.ResultsHibana ul {
```
PatrickSison, line 14: Isn't this a duplicate class of the previous? `styles.Results`

src/components/reduxFormWrappers/FileFieldWrapper.js
PatrickSison: Also we should be using `htmlFor={input.id}`

shrirupa: 🤔 looks like this is already being handled in FE-935 https://github.com/SparkPost/2web2ui/pull/1555. I can may be remove this change from here. What do you think ?

src/components/support/components/SearchPanel.module.scss
shrirupa: I duplicated this since in Hibana Version I didn't have to use `.Results {
  height: rem(457); // must exclude search input
  overflow-y: scroll;
}` only wanted to override the default styles for ul.

PR #1574, opened 2020-04-15
FE-1015: Error: useHibanaState must be used within a HibanaProvide

bkemper:
### What Changed
 - Moved HibanaProvider and HibanaTheme above ErrorBoundary
 - Added two new routes /error-boundary and /private-error-boundary for testing ErrorBoundary's

### How To Test
 - Open /error-boundary to confirm app error boundary renders EmptyState without error

### DIFF
--- a/src/Providers.js
+++ b/src/Providers.js
@@ -14,13 +14,13 @@ const reloadApp = () => {
 
 const Providers = ({ store = {}, children }) => (
   <Provider store={store}>
-    <ErrorBoundary onCtaClick={reloadApp} ctaLabel="Reload Page">
-      <HibanaProvider>
-        <HibanaTheme>
+    <HibanaProvider>
+      <HibanaTheme>
+        <ErrorBoundary onCtaClick={reloadApp} ctaLabel="Reload Page">
           <Poll>{children}</Poll>
-        </HibanaTheme>
-      </HibanaProvider>
-    </ErrorBoundary>
+        </ErrorBoundary>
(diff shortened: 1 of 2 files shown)


src/Providers.js
```diff
@@ -14,13 +14,13 @@ const reloadApp = () => {
 
 const Providers = ({ store = {}, children }) => (
   <Provider store={store}>
-    <ErrorBoundary onCtaClick={reloadApp} ctaLabel="Reload Page">
-      <HibanaProvider>
-        <HibanaTheme>
+    <HibanaProvider>
+      <HibanaTheme>
+        <ErrorBoundary onCtaClick={reloadApp} ctaLabel="Reload Page">
```
PatrickSison, line 19: Wasn't the original intention to have the ErrorBoundary catch any errors with HibanaProvider as well? I suppose we can't have both though if we're styling it. Would the error catching in the index.html count it as not being able to load the app?

src/config/routes/index.js
```diff
@@ -897,6 +897,20 @@ const routes = [
     supportDocSearch: 'upgrade plan',
     category: 'Account',
   },
+  {
+    path: '/error-boundary',
```
PatrickSison, line 901: Good idea for these

src/Providers.js
bkemper: It is a risk to move the provider above the ErrorBoundary.  I'll cut a follow-up ticket after Hibana.

PR #1576, opened 2020-04-15
FE-936 - Re-style SummaryTable component for Hibana

nicklemmon:
[FE-936](https://jira.int.messagesystems.com/browse/FE-936)

### What Changed

- Upgrades to Hibana Matchbox version 15
- Re-styles `<SummaryTable/>` components for Hibana

### How To Test
- The component is used on:
  + `/signals/health-score`
  + `/signals/spam-traps`
  + `/signals/engagement`
- Make sure they continue to behave as they do in production for both the OG app and Hibana
- **Note:** A lot of styles for the tables are derived from the features in which they are used - this includes the use of `<Panel/>` in some spots as well as passed in padding for table cells - this will be addressed during feature cleanup

### To Do
- [ ] Incorporate feedback


src/components/summaryTable/Billboard.js
```diff
@@ -1,15 +1,21 @@
...
-  </tbody>
-);
+function Billboard({ children, colSpan }) {
+  const [state] = useHibana();
+  const { isHibanaEnabled } = state;
+
+  return (
+    <tbody>
+      <Table.Row>
+        <Table.Cell className={!isHibanaEnabled ? styles.Billboard : undefined} colSpan={colSpan}>
```
shrirupa, line 13: may be this `className = classNames(!isHibanaEnabled && styles.Billboard)` instead of specifying undefined.

src/components/summaryTable/Body.js
```diff
@@ -39,11 +42,13 @@ const Body = ({ columns, data, empty, error, loading, perPage }) => {
...
-                [styles.DataCell]: !Component,
-                [styles.CenterAlign]: align === 'center',
-                [styles.RightAlign]: align === 'right',
-              })}
+              className={
+                !isHibanaEnabled &&
+                classNames(styles.Cell, {
+                  [styles.DataCell]: !Component,
+                })
+              }
```
shrirupa, line 50: this should be 
`className = classNames(!isHibanaEnabled && classNames(styles.Cell, {[styles.DataCell]: !Component})) )`

Since the above is causing a warning in the console _index.js:1446 Warning: Failed prop type: Invalid prop `className` of type `boolean` supplied to `HibanaTableCell`, expected `string`._

nicklemmon: You're totally right - I've definitely made this mistake before! Good catch.

PR #1580, opened 2020-04-17
FE-1022 - Hide month pagination buttons with CSS instead of conditional rendering

nicklemmon:
[FE-1022](https://jira.int.messagesystems.com/browse/FE-1022)

### What Changed
- Updates `<Navbar/>` component in `DateSelector.js` to conditionally show and hide components instead of using conditional rendering
- Adds `type="button"` to next and previous buttons in case this component is ever used inside a `<form>`
- Updates unit tests to match new implementation

### How To Test
- Go through signals pages and open the date picker - verify that clicking the previous button and then the next button does not close the popover
- Verify this fix works both for the OG theme and the Hibana theme

### To Do
- [ ] Incorporate feedback


src/components/dateSelector/tests/DateSelector.test.js
```diff
@@ -73,4 +66,28 @@ describe('Navbar', () => {
...
     expect(handleNextClick).toHaveBeenCalled();
   });
+
+  it('renders the next and previous buttons without the `isHidden` styles when controlled via the `showNextButton` and `showPreviousButton` props', () => {
+    const wrapper = shallow(<Navbar showNextButton showPreviousButton />);
+
+    expect(wrapper).toHaveTextContent('Previous Month');
+    expect(wrapper).toHaveTextContent('Next Month');
+    expect(wrapper.find('.isHidden')).not.toExist();
+  });
```
nicklemmon, line 76: Just moved this one down, updated the description, and added an extra check for the className

PR #1584, opened 2020-04-20
AC-1298: Changed manually billed condition to look at billing subscri…

PatrickSison:
### What Changed
 - Changed condition for manually billed to look at subscription type

### How To Test
Via Holmes for most steps
 - Create a free account, and set to manually billed (Holmes)
 - Add R to subscription (Holmes)
 - Change the message bundle to free (Holmes)
- Verify that RV pages do not request for Credit card


PatrickSison:
I'd like to consolidate some of the conditionals for the account. We have a lot of one off selectors and selectors that seem duplicated. We should either deprecate some or clarify which ones are to be used

PR #1587, opened 2020-04-21
FE-962: Suppressions page updates

PatrickSison:
### What Changed
 - Updated suppressions page and create page (under `/lists/suppressions` and `/lists/suppressions/create`
- Updated suppressions modal and 

### How To Test
 - Go to `/lists/suppressions` and `/lists/suppressions/create`
 - Check that search still works on page
 - Check that creation still works correctly
 - Check that pages match mocks
 - Check that details and delete modal matches mocks


src/components/matchbox/Button.js
```diff
@@ -72,7 +77,7 @@ function getVariantProps({ variant, isHibanaEnabled }) {
       };
     case 'destructive':
       return {
-        color: 'red',
+        destructive: true,
```
PatrickSison, line 80: Original destructive was this type. Changed it unless important elsewhere

src/pages/suppressions/CreatePage.js
```diff
@@ -43,12 +43,17 @@ export class CreatePage extends Component {
         {(parseError || persistError) && (
           <ErrorBanner parseError={parseError} persistError={persistError} />
         )}
-        <Tabs
-          selected={selectedTab}
-          connectBelow={true}
-          tabs={tabs.map(({ content }, idx) => ({ content, onClick: () => this.handleTabs(idx) }))}
-        />
-        <Panel>{selectedTab === 1 ? <UploadForm /> : <AddForm />}</Panel>
+        <Panel>
```
PatrickSison, line 46: Tabs moved inside panel

src/pages/suppressions/ListPage.js
```diff
@@ -48,13 +48,16 @@ export class ListPage extends Component {
...
         }}
       >
-        <Tabs
-          selected={selectedTab}
-          connectBelow={true}
-          tabs={tabs.map(({ content }, idx) => ({ content, onClick: () => this.handleTabs(idx) }))}
-        />
-
         <Panel>
+          <Tabs
```
PatrickSison, line 52: Tabs moved inside panel

src/pages/suppressions/components/AddForm.js
```diff
@@ -43,40 +46,46 @@ export class AddForm extends Component {
...
-                type="checkbox"
-                validate={this.atLeastOne}
-              />
-            </Checkbox.Group>
-          </Panel.Section>
-          <Panel.Section>
-            <Button primary disabled={pristine || submitting} type="submit">
-              Add / Update
-            </Button>
+            <Stack>
```
PatrickSison, line 49: Added stack + box with max width

src/pages/suppressions/components/AddForm.js
```diff
@@ -43,40 +46,46 @@ export class AddForm extends Component {
...
+                />
+                <Field
+                  component={CheckboxWrapper}
+                  name="types.non_transactional"
+                  label="Non-Transactional"
+                  type="checkbox"
+                  validate={this.atLeastOne}
+                />
+              </Checkbox.Group>
+              <Field name="description" component={TextFieldWrapper} label="Description" />
```
PatrickSison, line 82: Moved below to match mocks, although mocks seem to use a textArea

src/pages/suppressions/components/AddForm.js
```diff
@@ -43,40 +46,46 @@ export class AddForm extends Component {
...
+            </Stack>
+            <ButtonWrapper>
```
PatrickSison, line 84: Button moved inside panel section

src/components/matchbox/Button.js
```diff
@@ -57,6 +57,11 @@ function getVariantProps({ variant, isHibanaEnabled }) {
         return {
           color: 'red',
         };
+      case 'destructive-secondary':
+        return {
+          color: 'red',
+          outlineBorder: true,
+        };
```
nicklemmon, line 64: This is a great idea.

src/pages/suppressions/CreatePage.js
nicklemmon: Two things
- It looks like a box shadow was introduced in to the tabs here in the OG version of the app
- And the Panel looks like it has more padding than it should - looks like 2 `<Panel.Section>` components?

Screen Shot 2020-04-23 at 1 12 25 PM

src/pages/suppressions/components/AddForm.js
```diff
@@ -3,15 +3,18 @@ import React, { Component, Fragment } from 'react';
...
 import { TextFieldWrapper, CheckboxWrapper } from 'src/components';
+import ButtonWrapper from 'src/components/buttonWrapper';
 
 import { showAlert } from 'src/actions/globalAlert';
 import { createOrUpdateSuppressions } from 'src/actions/suppressions';
 import SubaccountTypeaheadWrapper from 'src/components/reduxFormWrappers/SubaccountTypeaheadWrapper';
 import { required, email } from 'src/helpers/validation';
 const FORM_NAME = 'addSuppression';
 
+const maxWidth = '860px';
```
nicklemmon, line 16: Would you mind adding a TODO to update this when we have relevant design tokens from Matchbox?

src/pages/suppressions/components/AddForm.js
nicklemmon: My recipient list ticket had the same thing - I just ignored that and kept it as a regular input since that's a fairly significant UX change.

src/pages/suppressions/components/Detail.js
```diff
@@ -46,12 +45,9 @@ export default class Detail extends Component {
     const { open, onCancel, suppression } = this.props;
 
     return (
-      <Modal open={open} onClose={onCancel}>
+      <Modal open={open} showCloseButton onClose={onCancel}>
         <Panel title="Suppression Details" sectioned>
           {suppression && this.renderContents()}
-          <div className={styles.Buttons}>
-            <Button onClick={onCancel}>Close</Button>
-          </div>
```
nicklemmon, line 54: So glad we'll be removing these

src/pages/suppressions/ListPage.js
nicklemmon, line 52: It might be a bug with Matchbox, but the 'Sources' field is not filling the grid appropriately in Hibana only:

Screen Shot 2020-04-23 at 1 20 41 PM
Screen Shot 2020-04-23 at 1 21 52 PM

nicklemmon: Also looks like the 'Supression Details' modal could use some more vertical padding inside there - `<Box/>` should do the trick I would think?

Screen Shot 2020-04-23 at 1 25 00 PM

nicklemmon:
@PatrickSison - Oh yeah! Double check the console in your browser. I'm seeing a few errors that should be addressed.

src/pages/suppressions/ListPage.js
PatrickSison: @nicklemmon Not sure if it's exclusively Matchbox, but because the field is a trigger for a popover in the first issue, it is wrapped in a span so doesn't go full width. We could fix in matchbox, but this is a weird custom select that we can probably change

PR #1590, opened 2020-04-22
FE-971: Hibana style cleanup for Alerts pages

kchu1337:
FE-970 - Hibana style cleanup for AB Testing pages

### What Changed
- Everything
- List Page
   - Made the page component it's own functional component in order to utilize hooks.
   - Added Box components to fix Hibana spacing.
   - Switched to Box component from Panel.Section in Hibana for the recently triggered cards.
- Alerts Collection
  - Added a wrapper to switch between styles
- Details Page
   - Switched to Box component from Panel.Section in Hibana.
   - Changed all `<h6>` to` <p>` tags and handled bolding with css. This is because in Hibana, they are no longer bolded.
   - Switched the edit and duplicate+delete positions
- Alert Form
   - Changed to a functional component.
   - Fixed spacing in the various form fields.
- SliderFieldWrapper
   - Made a max width for the text box
   - Fixed the label id for Hibana to match OG 

### How To Test
- Test the flow for Hibana and OG version for following routes - 
     - /alerts
     - /alerts/create
     - /alerts/{id}

### To Do
- [ ] Address Feedback


src/pages/alerts/components/AlertDetails.js
```diff
@@ -84,13 +93,13 @@ export const AlertDetails = ({ alert, id, subaccountIdToString, hasSubaccounts }
         !(blacklistFilters.includes(filter_type) && filter_values[0] === 'any'),
     );
 
-    //Only show 'and' if it's not the first filter or if there's a subaccounts filter
+    //Only show 'and' if it's not the first filter or if there's a subaccounts filter (linebreak for Hibana)
     const filtersTags = cleanedFilters.map((filter, i) => (
-      <span key={filter.filter_type}>
-        {(shouldShowSubaccounts || i > 0) && 'and'}{' '}
-        <h6 className={styles.BoldInline}>{FILTERS_FRIENDLY_NAMES[filter.filter_type]}</h6>{' '}
+      <React.Fragment key={filter.filter_type}>
```
kchu1337, line 98: This is because <span> was adding extra spacing

src/pages/alerts/components/AlertForm.js
```diff
@@ -112,30 +118,34 @@ export class AlertForm extends Component {
...
-                <label htmlFor="name">Alert Name</label>
-                <Field
-                  name="name"
-                  component={TextFieldWrapper}
-                  disabled={submitting}
-                  validate={[required, maxLength(50)]}
-                />
-                <div className={styles.MetricSelector}>
-                  <label>Alert Metric</label>
+                <Box maxWidth="699px">
```
kchu1337, line 128: Temporary value to be removed later.

PatrickSison: We should possibly move this to a variable or add a TODO to remove this later on so we don't miss it

src/pages/alerts/components/AlertForm.js
```diff
@@ -112,30 +118,34 @@ export class AlertForm extends Component {
...
+                  <label htmlFor="name">Alert Name</label>
```
PatrickSison, line 129: Would we be able to just change this to just use the label prop in the field?

src/pages/alerts/components/AlertForm.js
```diff
@@ -112,30 +118,34 @@ export class AlertForm extends Component {
...
+                      name="metric"
+                      component={SelectWrapper}
+                      options={visibleMetricOptions}
+                      onChange={resetFormValues}
+                      validate={required}
+                      disabled={submitting || !isNewAlert}
+                    />
+                  </div>
+                </Box>
                 {metric !== '' && !formSpec.hideEvaluator && (
```
PatrickSison, line 149: Mocks are kind of inconsistent. Sometimes they show some fields as shorter, and then others full width, like the Subaccounts typeahead. Not sure what the solution is

src/pages/alerts/ListPage.js
```diff
@@ -140,4 +104,66 @@ export class ListPage extends Component {
...
+                    <Box height={'100px'}>
+                      <Stack>
+                        <PageLink className={styles.AlertName} to={`/alerts/details/${alert.id}`}>
+                          <strong data-id="link-alert-name">{alert.name}</strong>
+                        </PageLink>
+                        <Tag>{METRICS[alert.metric]}</Tag>
+                      </Stack>
+                    </Box>
+                  </CardSectionComponent>
+                  <CardSectionComponent className={styles.Footer} padd={'400'}>
```
PatrickSison, line 138: Is this a typo with the `padd`?

src/pages/alerts/components/AlertDetails.js
PatrickSison: Not sure if it's easier, but at this point, we could make a separate render so that the value label is a span inside a div, and then each section is a separate div instead of a fragment to keep them separated further

src/pages/alerts/components/AlertDetails.js
```diff
@@ -184,20 +193,37 @@ export const AlertDetails = ({ alert, id, subaccountIdToString, hasSubaccounts }
...
       return (
-        <Panel.Section key={i}>
+        <SectionComponent key={i} paddingLeft="500" paddingBottom="400">
           <LabelledValue label={label}>{renderedValue}</LabelledValue>
-        </Panel.Section>
+        </SectionComponent>
       );
     });
 
+  const HibanaPrimaryArea = () => (
```
PatrickSison, line 202: I think this might have fallen under a UX change so I don't think we needed to change this part. It might just be a little bit more confusing as we cut out code later

src/pages/alerts/components/AlertForm.js
```diff
@@ -170,9 +182,14 @@ export class AlertForm extends Component {
                       error="At least one notification channel must be not empty"
                     />
                   )}
-                  {this.renderNotificationChannels()}
+                  {renderNotificationChannels()}
                 </div>
-                <Button submit primary disabled={isSubmitDisabled} className={styles.SubmitButton}>
+                <Button
```
PatrickSison, line 187: We could remove this class and use the ButtonWrapper component

src/pages/alerts/components/fields/FilterFields.js
```diff
@@ -147,12 +146,14 @@ export class FilterFields extends Component {
             {...extraProps[value]}
           />
         ));
+      return <Stack space="500">{multiFilters}</Stack>;
```
PatrickSison, line 149: Noice 👍

src/pages/alerts/components/AlertForm.js
PatrickSison: Not super necessary though

PatrickSison: Actually seeing more recently this pattern of using the label so either way. This is a non issue

src/pages/alerts/components/AlertForm.js
kchu1337: I'll add a TODO. It will eventually be made in a token by matchbox. I just picked an arbitrary unique number to use as a placeholder.

src/pages/alerts/components/AlertDetails.js
kchu1337: I thought these individual page tickets were the time to change UX requests? Although I have an idea on how to make it easier to manage and cut out.

src/pages/alerts/components/AlertForm.js
kchu1337: 🤷

src/pages/alerts/components/AlertForm.js
kchu1337: Just gonna leave it as is because ButtonWrapper adds some extra spacing which this doesn't need and it'll be a hassle to fix.

src/components/hibana/OGHibanaSwitch.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
```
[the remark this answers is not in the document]
kchu1337: Sorta, but the useHibanaToggle is a hook, which I can't use since the component I am using it in is a class component.

PR #1591, opened 2020-04-23
FE-927: Updated billing components

PatrickSison:
### What Changed
Updated:
- Billing address form
- Payment Form
 - Plan price (on plan select)

### How To Test
 - Go to /account/billing
 - Verify that the update billing info and payment information match styles
 - Go to /account/billing/plan
 - Verify that the styles are matching correctly

### To Do
- [ ] Update the global banner
- [ ] Possibly update card summary? Not sure if anything needs to be done here


src/pages/billing/components/PendingCancelGlobalBanner.js
```diff
@@ -1,55 +1,59 @@
...
-    }
-
-    return (
-      <div className={styles.banner} style={style}>
-        <span className={styles.CenterContainer}>
-          <AccessTime />
-          <span> Your account will be cancelled on </span>
-          {formatDate(pending_cancellation.effective_date)}
-          <span>, and you will no longer be able to send email or login. Changed your mind? </span>
-          <Button className={styles.RenewButton} flat onClick={this.handleRenewAccount}>
```
kchu1337, line 41: Looks like the new mocks no longer have it as just a button, but as a link. Also, it looks like the banner goes on top of the navigation.

kchu1337: Looks like it will be a simple change to `src/components/hibana/Navigation.js`

src/pages/billing/components/PlanSelectHibana.module.scss
```diff
@@ -0,0 +1,5 @@
+@import '~@sparkpost/design-tokens-hibana/tokens';
+
+.SelectedPlan {
+  color: color(green, 700);
+}
```
kchu1337, line 5: Also might need to override the `.tierPlans {` color picker and use the style tokens for `.tierLabel {
  font-weight: 500;
}`

src/pages/billing/components/tests/PendingCancelGlobalBanner.test.js
```diff
@@ -1,43 +1,39 @@
...
+          {...props}
+        />
+      </TestApp>,
+    );
 
   it('should call renewAccount, showAlert, and fetchAccount on renewAccount', async () => {
-    const wrapper = subject();
-    await wrapper.find('Button').at(0).simulate('click');
+    const { queryAllByRole } = subject();
+    await userEvent.click(queryAllByRole('button')[0]);
```
kchu1337, line 28: I think it would be more robust to use `queryByText("Don't Cancel")`

src/pages/billing/components/tests/__snapshots__/PlanSelect.test.js.snap
```diff
@@ -1,232 +1,230 @@
...
+    class="Jl_Jn"
   >
     <div
-      className="tierLabel"
-    >
-      Test Account
-    </div>
-    <div
-      className="tierPlans"
+      class="Jl_CE"
```
kchu1337, line 9: I don't think these snapshots will work as these classnames are randomly generated

src/pages/billing/components/PlanSelect.js
```diff
@@ -109,6 +114,7 @@ const PlanSelectSection = ({ bundles, currentPlan, onSelect }) => {
                       onClick={() => onSelect(bundleCode)}
                       data-id={`select-plan-${bundleCode}`}
                       size="small"
+                      variant="secondary"
```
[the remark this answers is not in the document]
PatrickSison: No, we talked about it before where we don't want to touch any UX such as the price being on the right as of now

src/pages/billing/components/tests/__snapshots__/PlanSelect.test.js.snap
PatrickSison: Yeah, for some reason it's still passing, but it'd probably be better to avoid them. I'll rewrite rewrite this test

src/pages/billing/components/PendingCancelGlobalBannerHibana.module.scss
```diff
@@ -2,4 +2,15 @@
 
 .banner {
   background-color: color(blue, 700);
+  height: 3rem;
+  line-height: 3rem;
```
kchu1337, line 6: Could use the line-height token if applicable.

src/pages/billing/components/PlanSelectHibana.module.scss
```diff
@@ -3,3 +3,7 @@
 .SelectedPlan {
   color: color(green, 700);
 }
+
```
kchu1337, line 6: Still need to replace the color selector from the original stylesheet since (gray,8) is no longer a valid choice.

PR #1592, opened 2020-04-23
AC-1388 -  (UI)Testing asking for domain during set up (like we used to)

shrirupa:
AC-1388 -  (UI)Testing asking for domain during set up (like we used to)

### What Changed
- Added back following 
      - SendingDomain Page
      - Cypress tests and unit tests related to SendingDomainPage
      - Added in a variable (`window.skipSendingDomain `) to control whether user sees the Sending Domain step or not.
    

### How To Test
- Set window.skipSendingDomain to false in console, then this should be the signup flow
     signup > planpicker > DOMAIN > Dashboard

- Set window.skipSendingDomain to true in console, then this should be the signup flow
     signup > planpicker > Dashboard

### To Do
- [ ] Address Feedback


src/pages/onboarding/ChoosePlan.js
```diff
@@ -40,6 +38,7 @@ export function OnboardingPlanPage({
   hasError,
   bundles,
 }) {
+  const NEXT_STEP = Boolean(window.skipSendingDomain) ? '/dashboard' : '/onboarding/sending-domain';
```
kchu1337, line 41: I don't think it needs to be in allcaps since it's a variable that changes, but it doesn't really matter.

src/pages/onboarding/tests/SendingDomain.test.js
```diff
@@ -0,0 +1,50 @@
+import { shallow } from 'enzyme';
+import React from 'react';
+import { SendingDomainPage } from '../SendingDomainPage';
```
kchu1337, line 3: Can you rename this file to `SendingDomainPage.test.js`. To match the filename of the component.

PR #1593, opened 2020-04-23
FE-1064: CircleCI Project Setup

nicklemmon:
### What Changed
- Adds Circle CI config
- Removes `travis` command from our NPM scripts
- Replaces Travis badge in Readme with CircleCI badge
- Breaks up navigation tests in to two files, simplifies navigation tests by removing extraneous usage of `it()`. See: https://docs.cypress.io/guides/references/best-practices.html#Creating-%E2%80%9Ctiny%E2%80%9D-tests-with-a-single-assertion
- There are still ways to make this more efficient!
- Adds support for launching the app in Hibana by default

### To Do
- [X] Get Hibana batches running
- [x] Incorporate feedback


config/cypressBatches.js
```diff
@@ -1,15 +1,15 @@
...
+  paulWalker: [
     'cypress/tests/integration/accounts/**/*',
     'cypress/tests/integration/alerts/**/*',
     'cypress/tests/integration/events/**/*',
     'cypress/tests/integration/navigation/**/*',
     'cypress/tests/integration/blacklists/**/*',
-    'cypress/tests/integration/hibana/**/*',
   ],
-  'Vin Diesel': [
+  vinDiesel: [
```
nicklemmon, line 9: CircleCI has problems handling strings with spaces as command arguments for some reason - this is an easy solution.

config/cypressBatches.js
```diff
@@ -1,15 +1,15 @@
...
     'cypress/tests/integration/recipients/**/*',
     'cypress/tests/integration/signals-analytics/**/*',
+    'cypress/tests/integration/hibana/**/*',
```
nicklemmon, line 12: Moved these specs in to the other group to try and bring the duration of each batch in to balance. It was getting our of whack.

config/cypressBatches.js
```diff
@@ -1,13 +1,13 @@
...
     'cypress/tests/integration/alerts/**/*',
     'cypress/tests/integration/events/**/*',
-    'cypress/tests/integration/navigation/**/*',
-    'cypress/tests/integration/blacklists/**/*',
     'cypress/tests/integration/hibana/**/*',
+    'cypress/tests/integration/navigation/**/*',
   ],
-  'Vin Diesel': [
+  vinDiesel: [
+    'cypress/tests/integration/blacklists/**/*',
```
nicklemmon, line 10: CircleCI has trouble with passing space-separated strings as arguments - this is the easy way out.

package.json
```diff
@@ -14,18 +14,17 @@
...
+    "test-ci": "CI=true npm test -- --coverage --colors --maxWorkers=2",
     "test-cy-env": "node ./scripts/generateCypressEnv",
     "test-cy-run": "cypress run --config integrationFolder=cypress/tests/integration",
     "test-e2e": "npm run test-cy-env && cypress open --config integrationFolder=cypress/tests/e2e",
     "test-e2e-headless": "cypress run --config integrationFolder=cypress/tests/e2e",
     "test-e2e-ci": "npm run start-app & wait-on http://localhost:3100 && npm run test-e2e-headless",
     "test-integration": "npm run test-cy-env && cypress open --config integrationFolder=cypress/tests/integration",
     "test-integration-headless": "node scripts/cypressIntegration.js",
-    "test-integration-ci": "npm run start-app & wait-on http://localhost:3100 && sleep 90 && node scripts/cypressIntegration.js",
+    "test-integration-ci": "npm run start-app & wait-on http://localhost:3100 && sleep 30 && node scripts/cypressIntegration.js",
```
nicklemmon, line 25: Faster runner means we can reduce the duration of `sleep` without causing problems

scripts/cypressIntegration.js
```diff
@@ -2,13 +2,14 @@ const cypress = require('cypress');
...
 
+/* eslint-disable no-console */
 function runCypress({ specs }) {
   console.log('Running specs:', specs || 'All');
 
   return cypress
     .run({
       config: {
-        video: false,
+        numTestsKeptInMemory: 1,
```
nicklemmon, line 12: This prevents potential memory leaks - keeping tests in memory isn't important in a CI environment.

nicklemmon: Note that re-introducing video recording slows down Travis tremendously - but this is one of the main reasons we wanted to use CircleCI instead!

bkemper: How does this affect running locally?  Should we make this a conditional and only set for CI?

cypress/tests/integration/navigation/navigationBehavior.spec.js
```diff
@@ -0,0 +1,92 @@
...
+    cy.visit(PAGE_URL);
+    /*
+        Wait is included to wait for window resize event as mobile navigation
+        is controlled via window resize events instead of CSS. Ideally, CSS media
+        queries could be used to show and hide the two version of the navigation
+        to improve performance and reduce the likelihood of unusual
+        application states
+      */
+    /* eslint-disable-next-line */
+    cy.wait(500);
```
bkemper, line 51: 💸

.circleci/config.yml
```diff
@@ -0,0 +1,69 @@
...
+            - v1-deps-{{ .Branch }}-{{ checksum "package-lock.json" }}
+            - v1-deps-{{ .Branch }}-
+            - v1-deps-
+      - run: npm ci
+      - save_cache:
+          key: v1-deps-{{ .Branch }}-{{ checksum "package-lock.json" }}
+          paths:
+            - ~/.npm
+            - ~/.cache
+      - run: npm run test-ci
```
bkemper, line 49: Don't we want to run a build?

nicklemmon: We could! We aren't doing it in Travis currently but it's a good idea.

Though we do get a development build as part of the integration testing, which should catch any compilation errors?

scripts/cypressIntegration.js
nicklemmon: This script is only used in CI, really - the main benefit of having tests in memory is the ability to view snapshots of the page in headed mode. To be honest, I'm not sure why this isn't a default for headless runs?

bkemper:
I don't think so.  Running webpack with dev vs prod configurations is much
different.

bkemper:
Cool! I didn't race back to where it was used.

bkemper:
This is a dumb question, but do "on_hold" builds count against our total billable minutes?  The UI is freaking me out with the duration continuing to run. 

Screen Shot 2020-04-26 at 8 27 07 AM

.circleci/TODO.md
```diff
@@ -0,0 +1,7 @@
+# TODO Items for CircleCI
+
+- [] Reduce `sleep` duration back to `30` once Travis is fully removed in the NPM script
+  `test-integration-ci`
+- [] Should we use gating? i.e., `Build` and `Unit Test` must pass before integration tests run.
+- [] Use parallelization to split up Jest tests -
+  https://dev.to/drazisil/test-splitting-with-jest-on-circleci-1i0c
```
bkemper, line 7: Can we move this to Jira tickets, so they get prioritized?

bkemper: And we should probably do this in other "todo" in code?

.circleci/config.yml
```diff
@@ -0,0 +1,154 @@
...
+    - image: circleci/node:10.17.0
+
+commands:
+  # NPM caching commands based on: https://github.com/cypress-io/cypress-example-docker-circle/blob/master/circle.yml
+  restore_npm_cache:
+    description: 'Restores cached Node modules'
+    steps:
+      - restore_cache:
+          keys:
+            - v1-deps-{{ .Branch }}-{{ checksum "package-lock.json" }}
```
bkemper, line 14: Shouldn't this be "v2" to match the [circleci config version](https://github.com/SparkPost/2web2ui/pull/1593/files#diff-1d37e48f9ceff6d8030570cd36286a61R1)?

.circleci/config.yml
```diff
@@ -0,0 +1,154 @@
+version: 2.1
+
+defaults: &defaults
+  docker:
+    - image: circleci/node:10.17.0
+
+commands:
+  # NPM caching commands based on: https://github.com/cypress-io/cypress-example-docker-circle/blob/master/circle.yml
```
bkemper, line 8: This is an old example last updated on Jun 20, 2018 — https://circleci.com/docs/2.0/caching/#restoring-cache might be a better link to share.

> Automatic dependency caching is not available in CircleCI 2.0, so it is important to plan and implement your caching strategy to get the best performance. Manual configuration in 2.0 enables more advanced strategies and finer control.

.circleci/config.yml
```diff
@@ -0,0 +1,154 @@
...
+  restore_npm_cache:
+    description: 'Restores cached Node modules'
+    steps:
+      - restore_cache:
+          keys:
+            - v1-deps-{{ .Branch }}-{{ checksum "package-lock.json" }}
+            - v1-deps-{{ .Branch }}-
+            - v1-deps-
```
bkemper, line 16: > CircleCI restores caches in the order of keys listed in the restore_cache step

This is really cool.  🚀

.circleci/config.yml
bkemper: > If you would like to invalidate your cache manually, you can bump v1 to v2 in your config.yml file.
NM, I found my answer, but definitely worth documenting.

.circleci/config.yml
```diff
@@ -0,0 +1,154 @@
...
+
+  save_npm_cache:
+    description: 'Saves Node modules for later re-use'
+    steps:
+      - save_cache:
+          key: v1-deps-{{ .Branch }}-{{ checksum "package-lock.json" }}
+          paths:
+            - ~/.npm
+            - ~/.cache
```
bkemper, line 25: As noted in our Travis config, do we want to cache entire directory?

https://github.com/SparkPost/2web2ui/blob/master/.travis.yml#L13

.circleci/config.yml
```diff
@@ -0,0 +1,154 @@
...
+    parameters:
+      spec_batch:
+        type: string
+      default_to_hibana:
+        type: boolean
+        default: false
+    steps:
+      - setup
+      - run:
+          sudo apt-get install libgtk2.0-0 libgtk-3-0 libnotify-dev libgconf-2-4 libnss3 libxss1
```
bkemper, line 48: Shouldn't these be cached?

bkemper: I think I found my answer which is a custom docker image for cypress, https://circleci.com/docs/2.0/custom-images/. Does one exist?

bkemper: https://github.com/cypress-io/cypress-docker-images

.circleci/TODO.md
bkemper: +1 using parallelize for both Jest and Cypress tests.

.circleci/config.yml
```diff
@@ -0,0 +1,154 @@
...
+      - run: npm run build
+
+  'Unit Testing':
+    <<: *defaults
+    parallelism: 4
+    steps:
+      - setup
+      - run:
+          command:
+            npm run test-ci $(circleci tests glob "./src/**/*.test.js" "./scripts/**/*.test.js" |
```
bkemper, line 80: It is a bummer this glob can't be pulled from Jest config.

nicklemmon: yeeeeah - this command works with any test harness, so I can understand why they don't want to read all of those different configs.

.circleci/config.yml
nicklemmon: I suppose I could tweak it to `'$HOME/.cache/Cypress'` - let me try it.

.circleci/TODO.md
nicklemmon: > And we should probably do this in other "todo" in code?

Not sure what you mean...put it in comments in the config file?

nicklemmon:
> This is a dumb question, but do "on_hold" builds count against our total billable minutes? The UI is freaking me out with the duration continuing to run.
> 
> Screen Shot 2020-04-26 at 8 27 07 AM

Fortunately no...but it is freaky. They need to update their UI to not show that timer.

.circleci/config.yml
nicklemmon: We could try those? But they don't come with the build in CircleCI dependencies I don't believe...

It looks like the install takes 5 seconds for those items, so I would argue it wouldn't be worth migrating to a non-supported docker image.

.circleci/TODO.md
nicklemmon: Unfortunately that first `TODO` related to the `sleep` duration would belong in our `package.json` - no comments allowed 😢

bkemper: > Not sure what you mean...put it in comments in the config file?
Sorry, I meant to say the first time, we should probably create Jira tickets for follow-up items instead of adding todo comments in code.

package.json
```diff
@@ -66,12 +68,15 @@
...
+  "jest-junit": {
+    "outputDirectory": "test_results/junit"
+  },
   "dependencies": {
     "@css-modules-theme/core": "^2.2.0",
     "@sparkpost/boomerang": "^1.0.6",
     "@sparkpost/design-tokens-hibana": "npm:@sparkpost/design-tokens@^1.0.0-beta.16",
     "@sparkpost/matchbox": "^3.9.1",
-    "@sparkpost/matchbox-hibana": "npm:@sparkpost/matchbox@^4.0.0-hibana.16",
+    "@sparkpost/matchbox-hibana": "npm:@sparkpost/matchbox@4.0.0-hibana.16",
```
nicklemmon, line 79: This was causing an unintentional upgrade otherwise

nicklemmon:
@bkemper Ready for another pass - I wouldn't jump through the commits since your last review - they are a big mess because I suck at bash!

Everything is working as expected, now CircleCI is splitting up our tests and we aren't doing it manually so less maintenance for us. Right now I'm using 4 runners on each step, still requiring a manual approval to run the Hibana-specific tests.

.circleci/config.yml
```diff
@@ -0,0 +1,127 @@
...
+      - cypress_dependencies
+      - run:
+          name: 'Run Cypress Integration Tests'
+          command: |
+            function join { local IFS="$1"; shift; echo "$*"; }
+            SPECS=$(circleci tests glob "./cypress/tests/integration/**/*.spec.js" | circleci tests split --split-by=timings);spec="${SPECS[@]}"
+            echo $SPECS
+            FORMATTED_SPECS=$(join , ${SPECS[@]})
+            echo $FORMATTED_SPECS
+            npm run test-integration-ci-hibana -- --spec $FORMATTED_SPECS
```
bkemper, line 106: > Error autodetecting timing type, falling back to weighting by name
I'm assuming this error is there isn't data yet and it is not an error in our script?

bkemper: Also, it might be worth building all this into a node script that can power, test-integration-ci-hibana.

nicklemmon: 🤷 I feel like it would be worth doing if this were going to live longer than it will, but not for now

> I'm assuming this error is there isn't data yet and it is not an error in our script?

That's right. Should clear up. Even if it doesn't, it still splits automatically by filename (even though that's an arbitrary way to split, still better than manually doing it)

PR #1594, opened 2020-04-24
FE-960: Hibana style cleanup for snippets pages

bkemper:
### What Changed
 - Touched up Snippet pages to work with both OG and Hibana
 - Added a new button variant, "connected" and used it everywhere

### How To Test
 - Confirm Snippet list, create, and edit pages looks good in both OG and Hibanan


src/components/contentEditor/ContentEditorHibana.module.scss
```diff
@@ -0,0 +1,27 @@
+@import '~@sparkpost/design-tokens-hibana/tokens';
```
nicklemmon, line 1: I've been avoiding having _all_ of the styles in the Hibana override stylesheet - I'm a little nervous about maintaining both. Instead, I've just been overriding the old design tokens.

Even when it comes time to remove the old theme, replacing those particular properties should be pretty easy to do.

src/pages/snippets/components/IdentifierHelpText.js
```diff
@@ -1,16 +0,0 @@
...
-    Use this unique identifier to{' '}
-    {
-      <ExternalLink to={LINKS.SNIPPET_SUBSTITUTION_REFERENCE}>
-        reference your snippet in a template
-      </ExternalLink>
-    }
-  </React.Fragment>
-);
-
-export default IdentifierHelpText;
```
nicklemmon, line 16: 👏

src/components/contentEditor/ContentEditor.js
```diff
@@ -1,11 +1,13 @@
 import React from 'react';
 import { Field } from 'redux-form';
-import { Panel, Tabs } from 'src/components/matchbox';
+import { Box, Panel, Tabs } from 'src/components/matchbox';
 import { json } from 'src/helpers/validation';
+import useHibanaOverride from 'src/hooks/useHibanaOverride';
 import AceWrapper from './AceWrapper';
 
 import './ContentEditor.scss';
```
nicklemmon, line 8: Looks like this file needs Hibana overrides as well - a good many design tokens from the old version of Matchbox are still in use here.

bkemper:
> @bkemper Overall things look good - two CSS-related comments to look at.
> 
> I don't think the syntax highlighting in the mocks is going to be possible until we migrate to Monaco.

Yep, I didn't even attempt it based on your comment in standup last week.

src/components/contentEditor/AceWrapper.js
```diff
@@ -5,48 +5,53 @@ import 'brace/mode/text';
...
+
+  return (
+    <div>
+      {submitFailed && !active && error && (
+        <div className={styles.ErrorWrapper}>
+          <Error error={error} />
+        </div>
+      )}
+      <AceEditor
+        className={styles.TemplateEditor}
```
bkemper, line 30: To contain this global style, I switched static class name with a module.

bkemper:
@nicklemmon This is ready for another pass.

PR #1608, opened 2020-04-30
FE-964: Updated API keys and tracking domains pages

PatrickSison:
### What Changed
 - Tracking domains pages (List, create)
- API keys pages (LIst, create/edit, view)

### How To Test
- Navigate to all these routes:
   - /account/api-keys
   - /account/api-keys/create
   - /account/api-keys/edit/:id
   - /account/api-keys/view/:id
   - /account/tracking-domains
   - /account/tracking-domains/create
 - Check that all the pages visually look correct

### DIFF
--- a/src/components/matchbox/Button.js
+++ b/src/components/matchbox/Button.js
@@ -62,11 +62,6 @@ function getVariantProps({ variant, isHibanaEnabled }) {
           color: 'red',
           outlineBorder: true,
         };
-      case 'destructive-secondary':
-        return {
-          color: 'red',
-          outlineBorder: true,
-        };
       // use when connecting a button to a TextField
       case 'connected':
         return {
--- a/src/pages/api-keys/components/ApiKeyForm.js
+++ b/src/pages/api-keys/components/ApiKeyForm.js
@@ -23,7 +23,6 @@ import { required } from 'src/helpers/validation';
 import GrantsCheckboxes from 'src/components/grantBoxes/GrantsCheckboxes';
 
 const formName = 'apiKeyForm';
-const maxWidth = '860px'; //TODO: Remove once Hibana tokens are available
 
 export class ApiKeyForm extends Component {
   get availableGrants() {
@@ -67,7 +66,7 @@ export class ApiKeyForm extends Component {
       <form onSubmit={handleSubmit(this.onSubmit)}>
         <Panel.Section>
           <Stack>
-            <Box maxWidth={maxWidth}>
+            <Box maxWidth="1200">
               <Field
                 name="label"
                 component={TextFieldWrapper}
@@ -76,7 +75,7 @@ export class ApiKeyForm extends Component {
                 disabled={isReadOnly}
               />
             </Box>
-            <Box maxWidth={maxWidth}>
+            <Box maxWidth="1200">
               <Field
                 name="subaccount"
                 helpText={
--- a/src/pages/api-keys/components/ApiKeyForm.js
+++ b/src/pages/api-keys/components/ApiKeyForm.js
@@ -102,19 +102,21 @@ export class ApiKeyForm extends Component {
               show={showGrants}
               disabled={isReadOnly}
             />
-            <Field
-              name="validIps"
-              component={TextFieldWrapper}
-              label="Allowed IPs"
-              helpText={
-                isReadOnly
-                  ? ''
-                  : 'Leaving the field blank will allow access by valid API keys from any IP address.'
-              }
-              placeholder={isReadOnly ? '' : '10.20.30.40, 10.20.30.0/24'}
-              validate={validIpList}
-              disabled={isReadOnly}
-            />
+            <Box maxWidth="1200">
+              <Field
+                name="validIps"
+                component={TextFieldWrapper}
+                label="Allowed IPs"
+                helpText={
+                  isReadOnly
+                    ? ''
+                    : 'Leaving the field blank will allow access by valid API keys from any IP address.'
+                }
+                placeholder={isReadOnly ? '' : '10.20.30.40, 10.20.30.0/24'}
+                validate={validIpList}
+                disabled={isReadOnly}
+              />
+            </Box>
           </Stack>
         </Panel.Section>
         {!isReadOnly && (
--- a/src/pages/api-keys/components/ApiKeyForm.js
+++ b/src/pages/api-keys/components/ApiKeyForm.js
@@ -3,7 +3,7 @@ import React, { Component } from 'react';
 import { connect } from 'react-redux';
 import { withRouter } from 'react-router-dom';
 import { Field, reduxForm, formValueSelector } from 'redux-form';
-import { Button, Panel } from 'src/components/matchbox';
+import { Box, Button, Panel, Stack } from 'src/components/matchbox';
 
 import {
   RadioGroup,
@@ -23,6 +23,7 @@ import { required } from 'src/helpers/validation';
 import GrantsCheckboxes from 'src/components/grantBoxes/GrantsCheckboxes';
 
 const formName = 'apiKeyForm';
+const maxWidth = '860px'; //TODO: Remove once Hibana tokens are available
 
 export class ApiKeyForm extends Component {
   get availableGrants() {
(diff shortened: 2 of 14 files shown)


src/pages/api-keys/components/ApiKeyForm.js
```diff
@@ -65,49 +66,61 @@ export class ApiKeyForm extends Component {
...
+            <Box maxWidth={maxWidth}>
+              <Field
+                name="label"
+                component={TextFieldWrapper}
+                validate={required}
+                label="API Key Name"
+                disabled={isReadOnly}
+              />
+            </Box>
+            <Box maxWidth={maxWidth}>
```
PatrickSison, line 79: Mock has this similar to SubaccountSection, but this field doesn't currently have an all subaccounts assignment

src/pages/api-keys/components/ApiKeyForm.js
```diff
@@ -3,7 +3,7 @@ import React, { Component } from 'react';
 import { connect } from 'react-redux';
```
PatrickSison, line 3: Mocks also have this form split into more sections, but might work more appropriately for when the layout of the page changes to have a sidebar description

src/pages/api-keys/components/ApiKeyForm.js
```diff
@@ -65,49 +66,61 @@ export class ApiKeyForm extends Component {
...
+              component={RadioGroup}
+              label="API Permissions"
+              options={this.getGrantOptions()}
+            />
+            <GrantsCheckboxes
+              grants={this.availableGrants}
+              show={showGrants}
+              disabled={isReadOnly}
+            />
+            <Field
```
PatrickSison, line 106: Didn't add max width on this field. Let me know if you feel like this should have the same max widths as the other fields

src/pages/trackingDomains/components/TrackingDomainRow.js
```diff
@@ -83,7 +82,7 @@ export class TrackingDomainRow extends Component {
     const verifyText = verifying.indexOf(domain) >= 0 ? 'Verifying...' : 'Retry Verification';
     return (
       <Button size="small" onClick={this.retryVerification}>
-        <Refresh /> {verifyText}
```
PatrickSison, line 86: This ends up messing with the button group alignment. Might be an with the button group, but removed since it doesn't seem like other pages are using button with icon pattern. Let me know if I should revert this

src/pages/trackingDomains/components/TrackingDomainRow.js
```diff
@@ -138,7 +137,7 @@ export class TrackingDomainRow extends Component {
     return (
       <Panel.Section>
         <Grid>
-          <Grid.Column xs={12} md={9}>
+          <Grid.Column xs={12} md={8}>
```
PatrickSison, line 140: Changed the column width because the buttons slightly bleed over into the column at max length

src/pages/api-keys/components/ApiKeyForm.js
```diff
@@ -65,49 +66,61 @@ export class ApiKeyForm extends Component {
...
+          <Stack>
+            <Field
+              name="grantsRadio"
+              component={RadioGroup}
+              label="API Permissions"
+              options={this.getGrantOptions()}
+            />
+            <GrantsCheckboxes
+              grants={this.availableGrants}
+              show={showGrants}
```
nicklemmon, line 103: For some reason in Hibana only, the "All" and "Select" radio buttons aren't showing and hiding the checkboxes that follow:
Screen Shot 2020-04-30 at 1 06 40 PM

src/components/matchbox/Button.js
```diff
@@ -62,11 +62,6 @@ function getVariantProps({ variant, isHibanaEnabled }) {
           color: 'red',
           outlineBorder: true,
         };
-      case 'destructive-secondary':
-        return {
-          color: 'red',
-          outlineBorder: true,
-        };
```
nicklemmon, line 69: I thought this was useful! Should it not exist?

PatrickSison: There was a duplicate with the merges. Just removed one of them

src/pages/api-keys/components/ApiKeyForm.js
```diff
@@ -23,6 +23,7 @@ import { required } from 'src/helpers/validation';
 import GrantsCheckboxes from 'src/components/grantBoxes/GrantsCheckboxes';
 
 const formName = 'apiKeyForm';
+const maxWidth = '860px'; //TODO: Remove once Hibana tokens are available
```
bkemper, line 26: Isn't this available now?

bkemper:
Shouldn't the collection be striped? 

Screen Shot 2020-04-30 at 2 23 06 PM

src/pages/api-keys/components/ApiKeyForm.js
bkemper: Assuming it will be sizing-1200.

— https://github.com/SparkPost/matchbox/blob/rc/4.0.0/packages/design-tokens/tokens/sizing.yml#L18

bkemper:
Should the tracking domain list page buttons match blacklist?

Screen Shot 2020-04-30 at 2 42 23 PM

src/pages/api-keys/components/ApiKeyForm.js
PatrickSison: Was hiding through css. Instead, changed so it's completely removed

src/pages/api-keys/components/ApiKeyForm.js
PatrickSison: What would be the way to access these tokens? I didn't see this in the token api docs

src/pages/api-keys/components/ApiKeyForm.js
bkemper: Yep and we probably should tweak the other fields.

```
<Box maxWidth="1200">
```

PatrickSison:
> Should the tracking domain list page buttons match blacklist?
> 
> Screen Shot 2020-04-30 at 2 42 23 PM

Is that the pattern we're following? Other tables show the buttons without the outline. @bkemper Like the suppressions page had the mock for buttons with no outline

src/pages/api-keys/components/ApiKeyForm.js
bkemper: Do we want to just add this to matchbox/TextField when rows prop is undefined?

src/pages/api-keys/components/ApiKeyForm.js
bkemper: Good call!

bkemper: Oh Stack with your empty element.

Screen Shot 2020-04-30 at 4 09 06 PM

src/pages/api-keys/components/ApiKeyForm.js
bkemper: Oops, I answered this on the other reply.

— https://github.com/SparkPost/2web2ui/pull/1608#discussion_r418255272

bkemper:
> > Should the tracking domain list page buttons match blacklist?
> > Screen Shot 2020-04-30 at 2 42 23 PM
> 
> Is that the pattern we're following? Other tables show the buttons without the outline. @bkemper Like the suppressions page had the mock for buttons with no outline

I think we should use the same button styles, but like you just pointed out, this is probably better left as follow-up since we have other inconsistencies.

nicklemmon:
> > > Should the tracking domain list page buttons match blacklist?
> > > Screen Shot 2020-04-30 at 2 42 23 PM
> > 
> > 
> > Is that the pattern we're following? Other tables show the buttons without the outline. @bkemper Like the suppressions page had the mock for buttons with no outline
> 
> I think we should use the same button styles, but like you just pointed out, this is probably better left as follow-up since we have other inconsistencies.

I highlighted this as an inconsistency with UX yesterday. The two big ones are:

- Handling `Button` components within table rows
- Handling `Button` components with `Banner` components

As of right now, each mock is pretty unique. This was a problem that pre-dates Hibana, actually, but Hibana is a good opportunity to actually address it.

src/pages/api-keys/components/ApiKeyForm.js
```diff
@@ -65,49 +65,63 @@ export class ApiKeyForm extends Component {
...
-            helpText={
-              isReadOnly
-                ? ''
-                : 'This assignment is permanent. Leave blank to assign to master account.'
-            }
-            component={SubaccountTypeaheadWrapper}
-            disabled={!isNew}
-          />
+          <Stack>
+            <Box maxWidth="1200">
```
bkemper, line 69: You didn't go crazy :)

PR #1609, opened 2020-04-30
FE-978 Dashboard Hibana Updates

gfriedmansp:
### What Changed
 - Updated Hibana styles for dashboard page

### How To Test
 - Ensure spacing, colors, etc. are not broken on both hibana and OG


gfriedmansp:
Things left out that were in the mocks: 

- The Panel title's actions do not have the styles (I think this would be a matchbox update)
- The border bottom on the Panel title isn't there - I'd also expect that to be a matchbox update for consistency

src/components/matchbox/Button.js
```diff
@@ -73,6 +73,11 @@ function getVariantProps({ variant, isHibanaEnabled }) {
           color: 'blue',
           outline: true,
         };
+      case 'outline':
+        return {
+          color: 'black',
```
bkemper, line 78: Should this be black or gray?

bkemper: The ApiErrorBanner uses "gray".

src/components/usageReport/UsageReport.module.scss
```diff
@@ -11,8 +11,8 @@
   font-weight: 400;
 }
 
-.orange {
-  color: color(orange)
+.primary {
```
bkemper, line 14: 👏

src/pages/dashboard/components/SendingStep.js
```diff
@@ -2,23 +2,26 @@ import React from 'react';
...
 
 export default function SendingStep() {
   return (
     <Panel.Section>
       <GuideBreadCrumbs />
-      <p className={styles.SendingStepHeading} role="heading" aria-level="4" tabIndex={-1}>
-        Where Would You Like to Begin?
-      </p>
+      <Box mb="400">
+        <p className={styles.SendingStepHeading} role="heading" aria-level="4" tabIndex={-1}>
```
bkemper, line 15: This class uses an old token, `font-size`, that needs to be replaced.

src/pages/dashboard/components/SendingStep.js
```diff
@@ -39,13 +42,46 @@ export const SendingStepListItem = ({ setAndStoreStepName, name, label = name, c
...
+    <Box mb={mb}>
+      <Card textAlign="center">
+        <CardContent>
+          <Text>{content}</Text>
+        </CardContent>
+        <CardActions>
+          <Button
+            mt="400"
+            variant="primary"
+            outlineBorder
```
bkemper, line 62: Do we want to use `variant` and `outlineBorder` or just make a new variant?  I vote for a new variant and we can consolidate the list of variants later.

gfriedmansp: I was hesitant about making a bunch of variants of variants hah. I'll go ahead and make this a variant and as you said, we can go back later and work with design if there's too many. It might actually help us determine that there _are_ too many if there are.

bkemper: That is what I am hoping, but sadly I think we already have too many.

nicklemmon: Just jumping in here randomly - I think Kevin made a new variant you can use in this scenario - `monochrome`. Try it!

gfriedmansp:
I was looking at FeaturesStepList and can't see that it's being used anywhere. Can you confirm @bkemper ? Otherwise, this should be ready for re-review.

bkemper:
> I was looking at FeaturesStepList and can't see that it's being used anywhere. Can you confirm @bkemper ? Otherwise, this should be ready for re-review.

Good catch.  `<FeaturesStepList />` is only used in `<FeaturesStep />` and as far as I can see, `<FeaturesStep />` is not being used.  Should would check with Shrirupa and Patrick, but it looks like more code that we can remove.

Screen Shot 2020-05-05 at 11 55 00 AM

shrirupa:
> > I was looking at FeaturesStepList and can't see that it's being used anywhere. Can you confirm @bkemper ? Otherwise, this should be ready for re-review.
> 
> Good catch. `<FeaturesStepList />` is only used in `<FeaturesStep />` and as far as I can see, `<FeaturesStep />` is not being used. Should would check with Shrirupa and Patrick, but it looks like more code that we can remove.
> 
> Screen Shot 2020-05-05 at 11 55 00 AM

Yes, its not being used anymore( ongoing experiments with onboarding) we can remove it.

src/components/breadcrumbs/BreadCrumbs.js
```diff
@@ -15,7 +18,38 @@ export const BreadCrumbs = ({ children }) => (
   </div>
 );
 
-export const BreadCrumbsItem = ({ children, active, onClick }) => (
+export const HibanaBreadCrumbs = ({ children }) => (
+  <Box pr="200" pb="400">
```
bkemper, line 22: nit: I think these abbreviated props are unnecessary complexity.  They require a familiarity with styled-system.

src/components/usageReport/SendMoreCTA.js
```diff
@@ -32,4 +34,50 @@ export const SendMoreCTA = ({ hasSendingLimits }) => (
...
+                  Switch to a new plan.
+                </PageLink>
+              }
+            />
+
+            {/* regardless of self serve billing */}
+            <Case
+              condition={defaultCase}
+              children={
+                <PageLink as={Button} variant="monochrome" to="/account/billing">
```
bkemper, line 58: 🚀

src/components/usageReport/SendMoreCTA.js
```diff
@@ -32,4 +34,50 @@ export const SendMoreCTA = ({ hasSendingLimits }) => (
...
+                  Upgrade your account.
+                </PageLink>
+              }
+            />
+          </ConditionSwitch>{' '}
```
bkemper, line 63: I've noticed lots of trailing spaces.  This one specifically shouldn't be needed.

src/pages/dashboard/components/GuideListItem.js
```diff
@@ -1,8 +1,46 @@
...
+      {itemCompleted ? (
+        <CheckBox size={20} color={tokens.color_blue_700} />
+      ) : (
+        <CheckBoxOutlineBlank size={20} />
+      )}
+    </Box>
+    <Box pl="200" flexGrow="1">
+      {children}
+    </Box>
+    <Box float="right" ml="200">
```
bkemper, line 29: It looks like this float prop is being ignored.

src/pages/dashboard/components/GuideListItem.js
```diff
@@ -1,8 +1,46 @@
...
 import styles from './GuideListItem.module.scss';
-export const GuideListItem = ({
+
+export const HibanaGuideListItem = ({
+  itemCompleted,
+  children,
+  action: { name, key = name, onClick, to, external },
+}) => (
+  <Box display="flex">
+    <Box width="20px" maxHeight="20px">
```
bkemper, line 19: Do we need these widths?

src/components/breadcrumbs/BreadCrumbs.js
gfriedmansp: It looks like we're using these already throughout. I think if we're moving towards using styled system (as opposed to css files), it makes sense to use them

bkemper: I hope we are talking about the same thing.  My nit is I think "paddingRight" is better than "pr".

gfriedmansp: We are talking about the same thing. There's like 50 occurrences of each of "mt", "mr", etc. in the repo already. Might be worth bringing up as a team, though

src/pages/dashboard/components/GuideListItem.js
gfriedmansp: Not sure why it was even there in the first place 🤷

PR #1611, opened 2020-04-30
FE-965 - Hibana style cleanup for account settings, cancel, profile, and SMTP pages

nicklemmon:
[FE-965](https://jira.int.messagesystems.com/browse/FE-965)

### What Changed
- Updated components and styles to look AMAZING for Hibana for the following routes:
  - /account/settings
  - /account/cancel - **be careful with this one - I wouldn't use `appteam` to test this - create a new account**
  - /account/profile
  - /account/smtp
  - _AND_ the support modal

### How To Test
- Visit the above routes and verify things look good on both the OG and Hibana versions of the app

### To Do
- [x] Fix out of date unit tests
- [x] Waiting on merging the RV restyle ticket for some updates to the `ExternalLink` component
- [ ] Incorporate feedback


src/components/layout/LayoutHibana.module.scss
```diff
@@ -15,6 +15,16 @@
...
+
+  .app & {
+    &.bannerOpen {
+      padding: spacing(800) spacing(400);
+
+      @media (min-width: media-query(md)) {
+        padding: spacing(400);
+      }
+    }
+  }
```
nicklemmon, line 27: Needed to override some OG styles here

src/components/smtpDetails/SmtpDetails.js
```diff
@@ -34,16 +35,21 @@ class SmtpDetails extends Component {
...
-          <LabelledValue label="Alternative Port" value={`${smtpAuth.alternativePort}`} />
-        )}
-        <LabelledValue label="Authentication" value="AUTH LOGIN" />
-        <LabelledValue label="Encryption" value="STARTTLS" />
-        <LabelledValue label="Username" value={smtpAuth.username} />
-        <LabelledValue label="Password">{passwordContent}</LabelledValue>
+        <Stack space="500">
+          <p>{smtpDescContent}</p>
+
+          <div>
```
nicklemmon, line 41: Just added the `Stack` and the `<div>` here

src/components/links/ExternalLink.js
```diff
@@ -3,10 +3,6 @@ import PropTypes from 'prop-types';
 import { OpenInNew } from '@sparkpost/matchbox-icons';
 import { Button, UnstyledLink } from 'src/components/matchbox';
 
-let asButtonStyles = {
-  textDecoration: 'underline',
-};
-
```
nicklemmon, line 9: Removed this - I was about to use the component, but then didn't end up doing so - after looking at several spots where we'll need this, the underline isn't necessary - and we don't use it in `<Button/>`s throughout the app, so I noticed it was inconsistent.

src/pages/account/components/UIOptionsPanel.js
```diff
@@ -28,19 +27,19 @@ export class UIOptionsPanel extends Component {
...
-                    onChange={() => this.setUIOption(key, !value)}
-                  />
-                </div>
-              </div>
-            </LabelledValue>
+            <Grid key={`ui-option-${key}`}>
+              <Grid.Column xs={12} md={11}>
+                <LabelledValue label={label}>{description}</LabelledValue>
+              </Grid.Column>
+              <Grid.Column xs={12} md={1} style={{ textAlign: 'right' }}>
```
bkemper, line 34: It matches TFA, but now they both hang over 🤷‍♂️ 

Screen Shot 2020-05-05 at 1 33 19 PM

src/pages/account/constants/index.js
```diff
@@ -0,0 +1 @@
+export const PANEL_LOADING_HEIGHT = '200px';
```
bkemper, line 1: Can you add a comment to this to move logic to PanelLoading?

nicklemmon: Hmm, not sure I follow? This is actually a height that works well for this feature but might not across the board.

nicklemmon:
@bkemper Ready for another pass when you have time 👍

src/pages/account/constants/index.js
bkemper: In the spirit of our conversation around consistency and remove one-offs, I don't know that we need this many different sized PanelLoading components.  I think it might benefit us to consolidate and see what we could do with say three variations.

 * AcceptedPage - 120px
 * BouncePage - 115px
 * ConfirmationModal - 200px
 * CurrentHealthGauge - 407px
 * DelayReport - 115px
 * DeleteTemplateModal - 190px
 * DuplicateTemplateModal - 330px
 * EngagementRecencyPreview - 170px
 * EngagementSummary - 115px
 * HealthScoreChart - 407px
 * HealthScorePreview - 170px
 * IncidentDetailPage - 150px
 * RejectionPage - 115px
 * SingleSignOnPanel - 130px
 * SpamTrapsPreview - 170px
 * StopMonitoringModal - 175px
 * TfaManager - 100px
 * TrendsChart - 280px

nicklemmon: I suspect the solution here is actually by changing how PanelLoading works (or find another way to implement it) - what the sizing is trying to get at is matching the size of the Panel with content and when in the loading state.

One potential route is even adding a prop on the `<Panel isLoading>` or something like that. That way, we could still render the DOM in the background and use CSS to overlay the loading UI such that it matches the height of its content! Then, no need for any height setting at all.

bkemper: It might be hard to render the panel without the data, but I love this train of thought.

Now seeing how many different sizes we have, don't worry about the comment.

PR #1616, opened 2020-05-04
SA-1328: Summary Report - Hibana Copy

PatrickSison:
### What Changed
 - Made a copy of the summary report page
 - Label changed to Analytics report

### How To Test
 - On account with Hibana flag enabled, check summary page
 - Verify it's copy of page by checking title. Should be `Analytics Report`


src/config/routes/index.js
```diff
@@ -180,197 +177,7 @@ const routes = [
...
-  {
-    path: '/signals/integration',
-    component: signals.IntegrationPage,
-    condition: hasAccountOptionEnabled('allow_events_ingest'),
-    layout: App,
-    title: 'Signals | Integration',
-    supportDocSearch: 'signals',
-    category: 'Configuration',
-  },
+  ...signalsRoutes,
```
kchu1337, line 180: Possible move them down to the end where all the other composite routes are, just so it's easier to find.

src/pages/reports/summary/ReportBuilder.js
```diff
@@ -0,0 +1,105 @@
+import React, { useEffect, useState } from 'react';
```
kchu1337, line 1: Depending on how much we refactor, we could put it in it's own folder so that it's easier to remove all of the other metrics pages later.

PR #1618, opened 2020-05-06
SA-1324 Allow from and to params in health score dashboard page

kchu1337:
### What Changed
 - If to and from are given in the health score dashboard URL, update redux with new from/to and update the url to remove the from and to params.

### How To Test
 - Go to /signals/health-score?from={date}&to={date}
 - Verify that the correct dates are used.

### To Do
- [ ] Possible always have to and from params showing?
- [ ] Can't seem to stop the table from fetching data twice.?


src/pages/signals/dashboards/HealthScoreDashboard.js
```diff
@@ -10,24 +12,66 @@ import DateFilter from '../components/filters/DateFilter';
...
+  } = props;
 
   // Gets subaccount info on mount
   useEffect(() => {
     getSubaccounts();
   }, [getSubaccounts]);
 
+  //If To and From are given as URL params, first update redux with the new date range and then remove
+  //the params.
+  const { from: urlFrom, to: urlTo } = qs.parse(search);
```
bkemper, line 42: This can be replace by `useRouter`.  Here is an example.

https://github.com/SparkPost/2web2ui/blob/c45737b0d2b55d31f9bbad9bd617cf2cbe95ba7a/src/pages/templates/context/EditorContext.js#L23

PR #1619, opened 2020-05-06
AC-1417 Add Segment Scripts to Head

gfriedmansp:
### What Changed
 - Added the Segment.io scripts

### How To Test
 - Check the <head> tag and find the segment script added to the head. Check that the correct key is there (based on value in config files).
 - Ensure logs are being propagated into https://app.segment.com/sparkpost-daniel-chalef/sources/webui_dev/debugger


src/components/segment/Segment.js
```diff
@@ -0,0 +1,24 @@
...
+const getScript = key => `!function(){var analytics=window.analytics=window.analytics||[];if(!analytics.initialize)if(!analytics.invoked){analytics.invoked=!0;analytics.methods=["trackSubmit","trackClick","trackLink","trackForm","pageview","identify","reset","group","track","ready","alias","debug","page","once","off","on","addSourceMiddleware","addDestinationMiddleware"];analytics.factory=function(e){return function(){var t=Array.prototype.slice.call(arguments);t.unshift(e);analytics.push(t);return analytics}};for(var e=0;e<analytics.methods.length;e++){var t=analytics.methods[e];analytics[t]=analytics.factory(t)}analytics.load=function(e,t){var n=document.createElement("script");n.type="text/javascript";n.async=!0;n.src="https://cdn.segment.com/analytics.js/v1/"+e+"/analytics.min.js";var a=document.getElementsByTagName("script")[0];a.parentNode.insertBefore(n,a);analytics._loadOptions=t};analytics.SNIPPET_VERSION="4.1.0";
+analytics.load("${key}");
+analytics.page();
+}}();`;
+
+export default () => {
+  const enabled = getConfig('segment.enabled') || false;
+  if (enabled) {
+    const key = getConfig('segment.publicKey') || '';
```
bkemper, line 15: Should we continue if we don't have a key?

src/config/env/test-config.js
```diff
@@ -2,6 +2,10 @@ const testConfig = {
   apiBase: 'http://fake-api-test-host.com',
   splashPage: '/dashboard',
   release: 'test-release',
+  segment: {
+    enabled: true,
+    publicKey: 'LR3TCDRu7vPZfty8FPZsg4Oe1WTmpAtt',
+  },
```
bkemper, line 8: I don't think we want it enabled when running automated tests.

gfriedmansp: Oh wow - I thought this was a test environment (like uat, etc.) That makes a whole lot more sense.

src/components/segment/Segment.js
```diff
@@ -0,0 +1,24 @@
...
+import getConfig from 'src/helpers/getConfig';
+
+// Removed the following from the script because it was interfering with tests
+// if (analytics.invoked) window.console && console.error && console.error("Segment snippet included twice.")
+const getScript = key => `!function(){var analytics=window.analytics=window.analytics||[];if(!analytics.initialize)if(!analytics.invoked){analytics.invoked=!0;analytics.methods=["trackSubmit","trackClick","trackLink","trackForm","pageview","identify","reset","group","track","ready","alias","debug","page","once","off","on","addSourceMiddleware","addDestinationMiddleware"];analytics.factory=function(e){return function(){var t=Array.prototype.slice.call(arguments);t.unshift(e);analytics.push(t);return analytics}};for(var e=0;e<analytics.methods.length;e++){var t=analytics.methods[e];analytics[t]=analytics.factory(t)}analytics.load=function(e,t){var n=document.createElement("script");n.type="text/javascript";n.async=!0;n.src="https://cdn.segment.com/analytics.js/v1/"+e+"/analytics.min.js";var a=document.getElementsByTagName("script")[0];a.parentNode.insertBefore(n,a);analytics._loadOptions=t};analytics.SNIPPET_VERSION="4.1.0";
```
bkemper, line 7: Are we concerned with `window.analytics` colliding with another service?

src/components/segment/Segment.js
gfriedmansp: Seems fair, given we control both

src/components/segment/Segment.js
gfriedmansp: I don't think so, but if we did I think we'd need to do some hacking on segment's internals

bkemper: I just confirmed our current setup with GA/GTM is not using `window.analytics`.

Screen Shot 2020-05-11 at 10 02 56 AM

gfriedmansp: 👍

PR #1627, opened 2020-05-13
FE-1074 - eslint plugin for no matchbox import

nstanard:
Example output:
image

### What Changed
 - Created a custom plugin with a rule to not allow direct matchbox imports.

### How To Test
 - Add a direct matchbox import to a file that is not in the `src/components/matchbox` folder.
 - Stage and commit - it should error via the husky `"pre-commit": "lint-staged"` step.

ExternalLink.js (what I used to test with - copy this import block into that component and stage/commit)
```
import React from 'react';
import { Button } from '@sparkpost/matchbox';
import { UnstyledLink } from 'src/components/matchbox';

```

### To Do
- [x] Address feedback


eslint-plugins/index.js
```diff
@@ -0,0 +1 @@
+export
```
bkemper, line 1: oops

PR #1628, opened 2020-05-13
AC-1446 - Show the Subaccounts Section while changing plans

shrirupa:
AC-1446 - Show the Subaccounts Section while changing plans

**Note** We had implemented this functionality earlier but removed it as api did not support limits

### What Changed
 - Added the functionality to bring back the subaccounts section when a user changes plan
 - Shows a warning to user and prevents them from changing plan if the current quantity of subaccounts is over the limit of the plan that they have selected
- if the current Limit is more than the future limit than than shows up as plain text **without** "Update account" button

### How To Test
 - Run the following cypress tests to see how this functions as the api as still not returning limits
       - Upgrades free account to premier 1M with previous active subaccounts, current quantity of subaccounts doesn't exceed new limit
       - Upgrades free account to starter 100K with previous active subaccounts, current quantity of subaccounts exceeds new limit

- We should be able to get this ticket out without the need of limit implementation in api, check if it allows for upgrade and downgrade without any issue

- link to old mocks: https://sparkpost.invisionapp.com/public/share/DGTCT4JW6JN#/screens/380587108

### To Do
- [ ] Address feedback

### DIFF
--- a/cypress/tests/integration/accounts/billing/changeBillingPlanPage.spec.js
+++ b/cypress/tests/integration/accounts/billing/changeBillingPlanPage.spec.js
@@ -581,6 +581,51 @@ describe('Change Billing Plan Page', () => {
     cy.queryAllByText('Premier Plans').should('not.be.visible');
     cy.queryAllByText('Close').should('not.be.visible');
   });
+  it('Upgrades from a starter plan to premier plan, with subaccounts limit_override higher than premier subaccount limit', () => {
+    // user is on the test plan
+    cy.stubRequest({
+      url: '/api/v1/account',
+      fixture: 'account/200.get.100k-starter-plan.json',
+    });
+    cy.stubRequest({
+      url: '/api/v1/billing',
+      fixture: 'billing/200.get.json',
+      fixtureAlias: 'billingGet',
+    });
+    cy.stubRequest({
+      url: '/api/v1/billing/subscription',
+      fixture: 'billing/subscription/200.get.starter-plan-higher-subaccounts-limitoverride.json',
+    });
+
+    cy.visit('/account/billing/plan');
+
+    cy.get('[data-id=select-plan-1M-premier-0519]').click();
+
+    cy.findByText('Your new plan only allows for 15 active subaccounts.').should('not.be.visible');
+
+    cy.stubRequest({
+      url: '/api/v1/billing/subscription/bundle',
+      fixture: 'billing/subscription/bundle/200.put.json',
+      method: 'PUT',
+    });
+
+    mockCommonHttpCalls();
+
+    // Then server returns the new plan info that was saved...
+    cy.stubRequest({
+      url: '/api/v1/account',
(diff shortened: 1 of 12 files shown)


cypress/tests/integration/accounts/billing/changeBillingPlanPage.spec.js
```diff
@@ -413,6 +443,47 @@ describe('Change Billing Plan Page', () => {
     cy.url().should('equal', `${Cypress.config().baseUrl}/account/billing`);
   });
 
+  it("Upgrades free account to premier 1M with previous active subaccounts, current quatity of subaccounts doesn't exceed new limit", () => {
```
kchu1337, line 446: typo in quantity

src/pages/billing/context/FeatureChangeContext.js
```diff
@@ -33,56 +43,105 @@ export const FeatureChangeProvider = ({
     );
   }, [plans, selectedBundle]);
 
+  // console.log(selectedPlansByProduct);
```
kchu1337, line 46: Probably leftover console.log

src/pages/billing/context/FeatureChangeContext.js
```diff
@@ -21,6 +21,16 @@ export const FeatureChangeProvider = ({
...
   }, [getSubscription]);
 
+  const checkConditions = useCallback(() => {
+    getSubscription();
+  }, [getSubscription]);
+  useEffect(() => {
+    window.addEventListener('focus', checkConditions);
+    return () => {
+      window.removeEventListener('focus', checkConditions);
+    };
```
kchu1337, line 31: Nothing wrong with it, just curious to what this does.

cypress/tests/integration/accounts/billing/changeBillingPlanPage.spec.js
```diff
@@ -196,6 +196,36 @@ describe('Change Billing Plan Page', () => {
...
+    cy.findAllByText('$20').should('be.visible');
+
+    cy.get('[data-id=select-plan-100K-starter-0519]').click();
+
+    cy.findByText("Your new plan doesn't include subaccounts.").should('be.visible');
+    cy.findByText('Update Status').should('be.visible');
+    cy.findByText('Update Status')
+      .should('have.prop', 'href')
+      .and('include', 'account/subaccounts');
+  });
```
kchu1337, line 228: Could you also check that the change plan button is disabled?

src/pages/billing/context/FeatureChangeContext.js
shrirupa: So when a new tab opens on clicking `Update Status` for Subaccounts, and then user comes back to this tab this reloads the subscription to check updated quantity of subaccounts.

shrirupa:
@kchu1337 Thanks for the feedback! This should be ready for another pass.

PR #1631, opened 2020-05-14
SA-1336 - Metrics Display for Report Builder

kchu1337:
### What Changed
 - Moved the <Legend> component that has the metrics into ReportOptions. But kept the old select metrics button for now.
 - Updated `<Legend>` component for new style of metrics tags. 
 - Added new colors for Hibana
 - Allows unlimited picking of metrics.
 - metrics variable in ReportOptions store always stores it as an array of strings now.
 - Added selector to get processed metrics object.

### How To Test
 -Go to `/reports/summary` and try everything. Use OG and Hibana.

### To Do
- [ ] Possibly rename `<Legend>` to `<MetricsDisplay>` or something.


src/pages/reportBuilder/components/tests/Legend.test.js
```diff
@@ -0,0 +1,86 @@
+import React from 'react';
```
kchu1337, line 1: This is copy paste from OG component

src/helpers/metrics.js
```diff
@@ -258,19 +259,23 @@ export function getValidDateRange({
...
 }
 
-export function _getMetricsFromKeys(keys = []) {
+export function _getMetricsFromKeys(keys = [], isHibanaEnabled) {
+  const metricsColors = isHibanaEnabled ? HIBANA_METRICS_COLORS : chartColors;
   return keys.map((metric, i) => {
     const found = METRICS_LIST.find(M => M.key === metric || M.key === metric.key);
 
     if (!found) {
       throw new Error(`Cannot find metric: ${JSON.stringify(metric)}`);
```
PatrickSison, line 268: Outside of this PR, but I wonder if it would be better if the metric is not found that it just gets stripped. this would be an issue if the URL gets modified manually only.

src/pages/reportBuilder/components/Legend.js
```diff
@@ -0,0 +1,59 @@
...
+import { Inline, Tag, Box } from 'src/components/matchbox';
+
+import { selectFeatureFlaggedMetrics } from 'src/selectors/metrics';
+import getConfig from 'src/helpers/getConfig';
+
+function renderMetric(metric, uniqueLabel, removeMetric) {
+  return (
+    <Tag key={metric.name} onRemove={() => removeMetric(metric.name)}>
+      <Inline>
+        <Box //The color circle
```
PatrickSison, line 12: We should double check with Jon if there's a better way to do the circle color. I imagine it might be reused later on so we should make it a reusable component if it's not built in by default

PatrickSison: Mainly checking with Jon the circle doesn't end up being in the round part of the left side of the tag

src/helpers/metrics.js
kchu1337: Maybe both? Remove the metric but show a snackbar that an invalid metric was in the URL.

src/pages/reportBuilder/components/Legend.js
kchu1337: I don't think the circle needs it's own component since it would be the same as <Box> except `borderRadius` and one of `height`, `width`.

PR #1634, opened 2020-05-15
FE-1076 - Address intermittently crashing Cypress runner in CircleCI

nicklemmon:
[FE-1076](https://jira.int.messagesystems.com/browse/FE-1076)

### What Changed
- I _tried_ using the Cypress Docker image to address the browser crashing issue, but no luck - not to mention they don't support all Node versions (in particular, the version our application is pegged to)
- I _tried_ using the Cypress Circle CI orb - this defaulted to the Cypress Docker image that caused the same Node version incompatibility issue
- Instead, I updated a few tests by removing unnecessary network requests
- Updated a flaky templates test to use a different means of checking local storage that *should* be more flake-proof
- Removed extraneous request stubbing in `changeBillingPlanPage.spec.js`
- Removed test for required field requiring validation on each form field
- Updated health score dashboard tests to assert more frequently, reducing test flakiness. See: https://docs.cypress.io/guides/core-concepts/retry-ability.html#Alternate-commands-and-assertions
- Updated summary reports tests to use new `findByDataId` method. Leveraged `DEFAULT_TO_HIBANA` env variable to check for different data ID due to changing location of elements

### How To Test
- Run this branch repeatedly in CircleCI
- Verify that the browser doesn't crash in CircleCI

### To Do
- [X] Run the pipeline with all steps enabled


cypress/tests/integration/accounts/billing/changeBillingPlanPage.spec.js
```diff
@@ -102,29 +91,20 @@ describe('Change Billing Plan Page', () => {
...
-    cy.findAllByText('Required').should('have.length', 4);
-    cy.findByLabelText('Cardholder Name').type('Test Account');
-    cy.findAllByText('Required').should('have.length', 3);
-    cy.findByLabelText('Expiration Date').type('03/33');
-    cy.findAllByText('Required').should('have.length', 2);
-    cy.findByLabelText('Security Code').type('123');
-    cy.findAllByText('Required').should('have.length', 1);
-    cy.findByLabelText('Country').select('United States');
-    cy.findByLabelText('State').select('Maryland');
-    cy.findByLabelText('Zip Code').type('12345');
```
nicklemmon, line 127: Testing each individual interaction probably isn't the best idea - better to just throw them all at once.

cypress/tests/integration/accounts/billing/changeBillingPlanPage.spec.js
```diff
@@ -102,29 +91,20 @@ describe('Change Billing Plan Page', () => {
...
     });
 
     // they visit the billing page and change to the starter 50k plan
-    cy.visit('/account/billing/plan');
+    cy.visit(PAGE_URL);
     cy.get('[data-id=select-plan-50K-starter-0519]').click();
 
-    cy.findByText('Your features have been updated').should('be.visible');
+    cy.findByText(/Your features have been updated/g)
+      .scrollIntoView()
```
shrirupa, line 102: interesting, what does scrollIntoView ensure here? Is it equivalent to adding a delay ?

cypress/tests/integration/signals-analytics/summaryReport.spec.js
```diff
@@ -30,22 +32,28 @@ describe('Summary Report page', () => {
...
-    cy.visit('/reports/summary');
+    cy.visit(PAGE_URL);
 
     const dayAgo = momentDateTime
       .subtract(1, 'day')
       .local()
       .format('MMM Do YYYY h:mma');
     cy.url().should('not.include', 'range=7days');
-    cy.get('[data-id="report-options"]').within(() => {
+    cy.findByDataId('report-options').within(() => {
```
shrirupa, line 56: findByDataId 🎉 👏

nicklemmon:
> Saw this error on the rerunning integration tests.
> 
> ```
> 1) The log in page
>  
>        "before each" hook for "renders a "Required" error message when no email address is entered":
>  
>      CypressError: `cy.visit()` failed trying to load:
>  
> 
>  
> http://127.0.0.1:3100/auth
>  
> 
>  
> We attempted to make an http request to this URL but the request failed without a response.
>  
> 
>  
> We received this error at the network level:
>  
> 
>  
>   > Error: ESOCKETTIMEDOUT
>  
> 
>  
> Common situations why this would fail:
>  
>   - you don't have internet access
>  
>   - you forgot to run / boot your web server
>  
>   - your web server isn't accessible
>  
>   - you have weird network configuration settings on your computer
>  
> 
>  
> The stack trace for this error is:
>  
> 
>  
> Error: ESOCKETTIMEDOUT
>  
>     at ClientRequest.<anonymous> (/home/circleci/.cache/Cypress/4.4.0/Cypress/resources/app/packages/server/node_modules/@cypress/request/request.js:816:19)
>  
>     at Object.onceWrapper (events.js:299:28)
>  
>     at ClientRequest.emit (events.js:210:5)
>  
>     at Socket.emitRequestTimeout (_http_client.js:690:9)
>  
>     at Object.onceWrapper (events.js:299:28)
>  
>     at Socket.emit (events.js:210:5)
>  
>     at Socket._onTimeout (net.js:468:8)
>  
>     at listOnTimeout (internal/timers.js:531:17)
>  
>     at processTimers (internal/timers.js:475:7)
>  
> 
>  
> 
>  
> Because this error occurred during a `before each` hook we are skipping the remaining tests in the current suite: `The log in page`
> ```

Hmm, interesting. I'll take a look - think I know why!

nicklemmon:
@shrirupa Ready for another pass finally - I think I got it - though had to address one flaky test with a fairly hacky solution. Better than not having the test, I think.

PR #1635, opened 2020-05-18
FE-1019 - Address UX feedback for Hibana navigation

nicklemmon:
[FE-1019](https://jira.int.messagesystems.com/browse/FE-1019)

### What Changed
- Adds new `FullBanner` component - along with supporting tests and documentation
- Adds `GlobalBanner` component that handles positioning for the rendered `FullBanner`
- Adds `GlobalBanner` Redux container component and selector to handle business logic
- Updates logic in `AccountPopover` to render the customer ID by default, email as backup
- Updates design for 'Upgrade' secondary label in `AccountPopover`
- Updates `AccountPopover` meta data color to brand orange from purple
- Adds icons to `AccountPopover` component - *some of them will need to be updated later as they have not yet been introduced in to Matchbox*
- Addresses some alignment feedback in the navigation
- Renames Hibana `Navigation` to `Header` (more accurate component name!)
- Adds new `ButtonLink` component and supporting documentation - semantically a `<button>` but looks like an `<a>` - use this instead of `javascript:void(0);`!
- Adds console warning to `PageLink` component when an `onClick` prop is passed
- Adds supporting Cypress/unit tests

### How To Test

1. Log in to a relevant account with a free plan (PM for a username or sign up for a new account)
1. Verify the 'upgrade' banner renders
1. Verify the 'Upgrade' secondary label renders in the account popover
1. Navigate to `/account/cancel` - verify on `/account/settings` and `/account/billing` the pending cancellation banner
1. Navigate to any other page and verify the 'upgrade' banner renders
1. Log in to a paid account
1. Very the upgrade banner does *not* render
1. Verify the `Upgrade` secondary label on the account popover does *not* render
1. Make sure automated testing captures relevant changes

**Note: The old pending cancellation banner for the OG theme survives - though it was technically possible to update this design, the amount of effort required seemed to exceed the value such a change would provide. If you think I'm wrong let's chat about it.** 

### To Do
- [x] Add additional test for mutual exclusivity of cancellation vs. upgrade banners
- [X] Waiting on some new icons from UX - _moving forward for now so this PR doesn't get stale_
- [x] Waiting on meeting on 5/19/20 to chat about some of the UX feedback received, particularly regarding expanding the secondary section of the navigation on hover - **UX agreed we should proceed with the current design, re: showing and hiding the secondary navigation**


src/selectors/globalBanner.js
```diff
@@ -0,0 +1,29 @@
...
+    (moment.duration(
+      moment(state.account.pending_cancellation.effective_date).diff(moment.utc(), 'days'),
+    ) < 7 ||
+      // ...and only on certain paths
+      cancelBannerPaths.includes(props.location.pathname))
+  ) {
+    return 'pending-cancellation';
+  }
+
+  if (
```
kchu1337, line 21: If a cancellation is pending but they are not on the account pages, this banner will show instead. Probably not a good look to ask them to upgrade while they are considering cancelling their account.

src/components/globalBanner/GlobalBanner.js
```diff
@@ -0,0 +1,59 @@
+import React from 'react';
+import { formatDate } from 'src/helpers/date';
+import { Text } from 'src/components/matchbox';
+import { PageLink } from 'src/components/links';
+import { FullBanner } from 'src/components/fullBanner';
+import styles from './GlobalBanner.module.scss';
+
+export default function GlobalBanner(props) {
```
kchu1337, line 8: Can you add a close/X icon to hide the banner? Or was that discouraged?

nicklemmon: Wasn't in the mocks! Seems like it should be though.

src/selectors/globalBanner.js
nicklemmon: Good call for sure.

src/components/links/ButtonLink.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { UnstyledLink } from 'src/components/matchbox';
+
+export default function ButtonLink(props) {
```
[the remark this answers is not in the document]
nicklemmon: Basically it's semantically an HTML `<button>` but looks like an HTML `<a>` - doesn't necessarily do a redirect. Could open a modal, for example.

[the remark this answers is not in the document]
nicklemmon: Imagine a paragraph of text, for example, that has a clickable link inside that opens a modal. Though it should look like a link (having a button in the middle of a paragraph would be weird), it should semantically be a `<button>`.

src/selectors/globalBanner.js
nicklemmon: My other question was, should this logib be in a selector? Should it be in the container component instead? Wasn't really sure.

nicklemmon:
@kchu1337 I believe I addressed your feedback - let me know what questions you have.

PR #1636, opened 2020-05-19
FE-1073 - Cypress tests should pass for both the original and Hibana versions of the app

nicklemmon:
[FE-1073](https://jira.int.messagesystems.com/browse/FE-1073)

### What Changed
- Updated Cypress tests to pass in both the Hibana and OG themes
- Invalidates CircleCI cache to resolve problem with unpackaging tarball
- Disabled `no-matchbox-import` ESLint rule within the `/cypress` dir
- Renamed `cy.assertLink()` => `cy.verifyLink()` - a more accurate name!
- Fixed env variable detection in Cypress that was broken following a prior refactor
- Resolved a bug for within the `ReportOptions` component in which a Redux action was not properly passed to child components. Unclear as to why this was working at all in the OG theme and not Hibana...

### How To Test
- Verify both sets of tests are passing in CI
- Verify I didn't introduce any problems with updated tests
- Verify I didn't introduce bugs as a result of my updates (particularly by verifying the filtering on `/reports/summary`)

### To Do
- [ ] Incorporate feedback


src/pages/reportBuilder/components/ReportOptions.js
```diff
@@ -36,6 +36,7 @@ const PanelContent = ({
   handleTypeaheadSelect,
   typeaheadCache,
   searchOptions,
+  refreshReportOptions,
```
nicklemmon, line 39: This was a bug caught by Cypress - very unclear why the bug was only showing up with Hibana enabled.

cypress/.eslintrc
```diff
@@ -8,6 +8,7 @@
     "jest/valid-expect-in-promise": "off",
     "jest/valid-expect": "off",
     // Does not apply to Cypress tests
```
bkemper, line 10: I'm going to fix this in FE-1078.  We shouldn't have to turn things off.

.circleci/config.yml
```diff
@@ -15,15 +15,15 @@ commands:
             #   "Caches are immutable so it is useful to start all your cache keys with a version prefix,
             #    for example v1-.... This enables you to regenerate all of your caches by incrementing
             #    the version in this prefix."
-            - v1-deps-{{ .Branch }}-{{ checksum "package-lock.json" }}
-            - v1-deps-{{ .Branch }}-
-            - v1-deps-
+            - v2-deps-{{ .Branch }}-{{ checksum "package-lock.json" }}
+            - v2-deps-{{ .Branch }}-
+            - v2-deps-
```
bkemper, line 20: This fixes the immediate problem, but do we know why we were getting `Error untarring cache`?

src/pages/reportBuilder/components/ReportOptions.js
bkemper: 👏

.circleci/config.yml
nicklemmon: I *think* it's because I was messing with a different Docker image in another branch (`FE-1076`). In the future, I may disable caching when doing that sort of work in the future:
https://support.circleci.com/hc/en-us/articles/360004250693-Restoring-cache-fails-with-Permission-Denied-

bkemper:
Oh yeah, that has got to be it.  I'm glad it is not a service bug.

PR #1641, opened 2020-05-26
SA-1327: Report Builder Metrics Drawer

kchu1337:
### What Changed
 - Added Metrics Drawer.

### How To Test
 - go to /reports/summary.
 - enable hibana
 - Open metrics drawer and select some metrics
     - verify that more than 5 can be selected
 - Click `clear metrics`. verify that no metrics are selected.
 - Select some metrics and click 'Apply Metrics`.
 - Verify that the drawer does not close but the summary report has been updated.

### To Do
- [ ] Some metrics differ from mocks. Get UX eyes.


src/pages/reportBuilder/components/ReportOptions.js
```diff
@@ -242,6 +243,18 @@ export function ReportOptions(props) {
     refreshReportOptions({ metrics: updatedMetrics });
   };
 
+  const renderTabContent = () => {
```
PatrickSison, line 246: Should be good for now, but I was considering writing an in app compound component for Tabs so that the children don't get derendered and we don't have to do any of the switch case stuff for tabs. Main reason is for this one, we don't want to derender the children and lose the temp state of checkboxes or the filters form if they switch tabs

src/components/matchbox/Drawer.js
```diff
@@ -50,4 +51,5 @@ Drawer.Footer = Footer;
 Drawer.Content = Content;
 Drawer.Header = Header;
 
+export const useDrawer = matchboxUseDrawer;
```
PatrickSison, line 54: Nice

src/pages/reportBuilder/components/ReportOptions.js
kchu1337: Ooh; yeah, didn't think of that. Wasn't sure how you wanted to handle it so I just wrote something simple, but that's a good idea.

src/pages/reportBuilder/components/MetricsDrawer.js
```diff
@@ -0,0 +1,88 @@
...
+          </Box>
+          <Inline space="100">{renderMetrics(metrics)}</Inline>
+        </div>
+      );
+    });
+  };
+  const renderMetrics = metrics =>
+    metrics.map(metric => {
+      return (
+        <Box marginRight="300" width="200px" key={metric.key}>
```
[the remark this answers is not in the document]
kchu1337: I noticed that too but changing margin/padding or even repositioning it didn't seem to help. I'll try again but I had a hard time fixing it.

PR #1642, opened 2020-05-26
FE-1080 - For Hibana only, reposition Snackbar global alerts to the bottom left of the viewport

nicklemmon:
[FE-1080](https://jira.int.messagesystems.com/browse/FE-1080)

### What Changed
- Refactored `Animator` component from class to functional in order to leverage the `useHibanaOverride` hook
- Added new styles to reposition global alerts

### How To Test
- Verify global alerts render in the right spot according to the a/c
- Verify they continue to render in the top right in the OG theme

### To Do
- [ ] Incorporate feedback


src/components/globalAlert/Animator.js
```diff
@@ -1,34 +1,32 @@
...
-            <div className={classes} style={{ height }} ref={this.transitionNode}>
-              {children}
-            </div>
-          );
-        }}
-      </Transition>
-    );
-  }
+        return (
+          <div className={classes} style={{ height }} ref={transitionNode}>
```
bkemper, line 23: This works, but it should be using useRef, not useState.

nicklemmon: 🤷‍♂️ Probably yes - is there a perf benefit or just more of a code smell thing?

PR #1646, opened 2020-05-28
SA-1329: Filters Drawer for Report Builder

PatrickSison:
### What Changed
 - Filters changed to be a drawer
- Filters are selected based on type selection

### How To Test
 - With a hibana account, go to Summary Report
 - Click "Add Filters" button
- Drawer should open

### To Do
- [ ] Drawer should be preserved when switching to another tab
- [ ] Move the query paramater behavior out of the share button
- [ ] Confirm the clear behavior with Jon


src/components/typeahead/ComboBoxTypeahead.js
```diff
@@ -124,14 +126,19 @@ export const ComboBoxTypeahead = ({
 
     return (
       <ComboBox {...getRootProps({ refKey: 'rootRef' })}>
-        <ComboBoxTextField {...inputProps} />
+        <ComboBoxTextField {...inputProps} id={id} />
         <ComboBoxMenu {...getMenuProps({ items, isOpen: isMenuOpen, refKey: 'menuRef' })} />
```
PatrickSison, line 130: Need to add a loading or empty state possibly for this typeahead

src/pages/reportBuilder/components/ReportOptions.js
```diff
@@ -339,9 +260,17 @@ export function ReportOptions(props) {
             <Drawer.Header showCloseButton />
             <Drawer.Content p="0">
               <Tabs selected={drawerTab} fitted tabs={tabs} />
+              {drawerTab === 1 && (
```
PatrickSison, line 263: I can probably write a new Tabs component that doesn't derender the tab content and also would be a compound component

src/pages/reportBuilder/components/ReportOptions.js
```diff
@@ -339,9 +260,17 @@ export function ReportOptions(props) {
...
+                <AddFiltersSection
+                  addFilters={addFilters}
+                  closeDrawer={closeDrawer}
```
PatrickSison, line 266: If I rewrite this, it would prevent having to pass closedrawer down to the filters section

src/selectors/reportFilterTypeaheadCache.js
```diff
@@ -60,9 +59,30 @@ const selectCache = createSelector(
     selectPreparedRecipientDomains,
     selectPreparedCampaigns,
     selectPreparedSendingIps,
-    selectPreparedIpPools
+    selectPreparedIpPools,
+  ],
+  (...caches) => _.flatten(caches),
+);
+
+export const selectCacheReportBuilder = createSelector(
```
PatrickSison, line 67: Was a little lazy with this implementation, but the typeahead results are bundled into a single object. I can convert to be properly connected to the add filters section for the cache content

src/components/matchbox/Drawer.js
```diff
@@ -49,5 +49,6 @@ Header.displayName = 'Drawer.Header';
 Drawer.Footer = Footer;
 Drawer.Content = Content;
 Drawer.Header = Header;
+Drawer.useDrawer = useDrawer;
```
PatrickSison, line 52: Matchbox has a usedrawer hook that is now exported with the drawer

src/pages/reportBuilder/components/AddFiltersSection.js
```diff
@@ -0,0 +1,227 @@
...
+                    <Button
+                      size="small"
+                      onClick={() => {
+                        dispatch({ type: 'REMOVE_FILTER_TYPE', index });
+                      }}
+                    >
+                      Remove
+                    </Button>
+                  </Inline>
+                  <Box>equals</Box>
```
PatrickSison, line 173: Not sure if this is the best look for this, or if we should wrap this in extra tags or add extra attributes. I'll ask Nick what we should do with this

src/components/typeahead/ComboBoxTypeahead.js
PatrickSison: Especially with modifying this to a cached typeahead like I ended up doing

src/pages/reportBuilder/components/ReportOptions.js
```diff
@@ -45,197 +37,134 @@ const PRECISION_OPTIONS = metricsRollupPrecisionMap.map(({ value }) => ({
...
-    featureFlaggedMetrics,
     location,
-    initTypeaheadCache,
+    history,
   } = props;
 
+  // Updates the query params with incoming search option changes
+  useEffect(() => {
+    const search = qs.stringify(searchOptions);
+    history.replace({ pathname: location.pathname, search });
```
kchu1337, line 138: You can use `useRouter`

src/pages/reportBuilder/components/ReportOptions.js
PatrickSison: Can't remove tabs component because of the behavior needed with Drawer

src/pages/reportBuilder/components/ReportOptions.js
PatrickSison: Can't change tabs because of necessary behavior with Drawer

src/pages/reportBuilder/components/AddFiltersSection.js
```diff
@@ -0,0 +1,236 @@
...
+  fetchMetricsIpPools,
+  fetchMetricsTemplates,
+} from 'src/actions/metrics';
+import { list as listSubaccounts } from 'src/actions/subaccounts';
+import { list as listSendingDomains } from 'src/actions/sendingDomains';
+import { initTypeaheadCache, setFilters, clearFilters } from 'src/actions/reportOptions';
+import { connect } from 'react-redux';
+import { selectCacheReportBuilder } from 'src/selectors/reportFilterTypeaheadCache';
+
+const FILTER_TYPES = [
```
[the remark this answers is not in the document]
PatrickSison: I suppose not. I'd like to keep the values together so if we need to add more, we don't have to dig in multiple places. I'll add this in 👍

src/pages/reportBuilder/components/AddFiltersSection.js
```diff
@@ -0,0 +1,227 @@
...
+    //TODO: Change structure of filters in redux to be filterType: [filter, filter] instead
+    filters: initialFilters.length
+      ? _.map(
+          _.reduce(
+            //Groups and cleans up value array
+            initialFilters,
+            (acc, { type: key, value }) => {
+              acc[key] = acc[key] || [];
+              acc[key].push(value);
+              return acc;
```
kchu1337, line 89: Not really important, but you are already modifying the accumulator and then you are also returning it. For the sake of Immutability, you can do `return [...acc[key], value];`. But it's not necessary to change unless you really want to.

src/pages/reportBuilder/components/AddFiltersSection.js
```diff
@@ -0,0 +1,227 @@
...
+        </Button>
+      </Box>
+      <Drawer.Footer>
+        <Inline space="300">
+          <Button onClick={handleSubmit} variant="primary">
+            Apply Filters
+          </Button>
+          <Button
+            onClick={() => {
+              clearFilters();
```
kchu1337, line 199: Not sure if intentional, but the clear filters does not immediate clear the filters in drawer, but does clear the filters. I guess we need to update the drawer to match redux? Re-set the redux filters when they change?

src/components/typeahead/ComboBoxTypeahead.js
```diff
@@ -124,14 +130,22 @@ export const ComboBoxTypeahead = ({
 
     return (
       <ComboBox {...getRootProps({ refKey: 'rootRef' })}>
-        <ComboBoxTextField {...inputProps} />
-        <ComboBoxMenu {...getMenuProps({ items, isOpen: isMenuOpen, refKey: 'menuRef' })} />
+        <ComboBoxTextField {...inputProps} id={id} />
+        <ComboBoxMenu
+          {...getMenuProps({ items, isOpen: isMenuOpen, refKey: 'menuRef' })}
+          emptyMessage={loading ? <Loading /> : undefined}
```
kchu1337, line 136: I'm not seeing any indication of loading. How is it supposed to look?

src/pages/reportBuilder/components/AddFiltersSection.js
```diff
@@ -0,0 +1,236 @@
...
+  const handleSubmit = () => {
+    //Formats the filters for the query
+    const filters = state.filters.reduce((acc, { key, value }) => {
+      if (!key) {
+        return acc;
+      }
+      return [...acc, ...value.map(filter => ({ type: key, value: filter }))];
+    }, []);
+    setFilters(filters);
+    closeDrawer();
```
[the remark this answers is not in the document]
PatrickSison: Yeah, double checked with UX, we close on submission, but not on clearing

src/pages/reportBuilder/components/Typeahead.js
```diff
@@ -16,7 +15,7 @@ function Typeahead(props) {
     [index, dispatch],
   );
 
-  const updateLookahead = debounce(pattern => {
```
kchu1337: I think you need to use a useCallback here. I think it's creating a new function every time, passing it into ComboBoxTypeahead and triggering the useEffect

src/components/tabs/Tabs.js
```diff
@@ -0,0 +1,38 @@
+import React, { useState } from 'react';
+import { Tabs as TabsComponent } from 'src/components/matchbox';
+
+function Tabs(props) {
+  const { children, defaultTab = 0, forceRender = false, tabs, ...rest } = props;
+
+  const [selected, setSelected] = useState(defaultTab);
```
kchu1337, line 7: Oh, and you can use `useTabs` here if you want. There's not too many LOC differences from that but it's up to you.

src/context/HibanaContext.js
```diff
@@ -7,7 +7,7 @@ const HibanaStateContext = createContext();
 const HibanaDispatchContext = createContext();
 
 const initialState = {
-  isHibanaEnabled: process.env.REACT_APP_DEFAULT_TO_HIBANA === 'true' ? true : false,
+  isHibanaEnabled: process.env.REACT_APP_DEFAULT_TO_HIBANA === 'true' ? true : true,
```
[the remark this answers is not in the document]
PatrickSison: Whoops. LOL

src/pages/reportBuilder/components/Typeahead.js
```diff
@@ -1,12 +1,14 @@
-import React, { useCallback, useState } from 'react';
+import React, { useCallback, useState, useMemo } from 'react';
 import { ComboBoxTypeahead } from 'src/components/typeahead/AsyncComboBoxTypeahead';
 import { METRICS_API_LIMIT } from 'src/constants';
+import sortMatch from '../../../helpers/sortMatch';
```
kchu1337, line 4: I guess we should use the full path but 🤷

PR #1647, opened 2020-05-28
Sa 1350 - Add aggregates summary section

kchu1337:
### What Changed
 - Created new action reducer pair for getting aggregates.
 - Added the aggregate section below the summary chart.

### How To Test
 -  go to report builder, `/reports/summary` with Hibana enabled.
 -  select various metrics and check the section below the chart.
 - Make sure to select many metrics to check the wrapping.
### To Do
- [ ] Possibly some sorting for how each metric is presented.


src/pages/reportBuilder/ReportBuilder.js
```diff
@@ -33,7 +36,27 @@ export function ReportBuilder({
...
+  };
+  const renderAggregateMetric = useCallback(({ label, value, unit }) => {
+    return (
+      <Definition>
+        <Definition.Label>
+          <Box color={tokens.color_gray_600}>{label}</Box>
+        </Definition.Label>
+        <Definition.Value>
+          <Box color={'#ffffff'}>
+            {' '}
```
PatrickSison, line 52: Could get rid of this empty string bracket

src/actions/summaryChart.js
```diff
@@ -26,14 +30,51 @@ export function refreshSummaryReport(
...
     return Promise.all([
       dispatch(getChartData({ params, metrics: merged.metrics })),
-      dispatch(getTableData({ params, metrics: merged.metrics })),
+      dispatch(getAggregateData({ params, metrics: merged.metrics, isCurrentGroupingAggregates })),
+      !isCurrentGroupingAggregates && dispatch(getTableData({ params, metrics: merged.metrics })),
     ]);
   };
 }
 
+export function _getAggregateData({ params, metrics, isCurrentGroupingAggregates }) {
```
PatrickSison, line 42: Wouldn't this be duplicate data as the main summary table data? Couldn't we have a selector that does the aggregating of the same data rather than pulling the same data? I might be missing something though

kchu1337: Oh yeah, that's true.  I just did it this way since there is an endpoint and it might be more accurate but your way sounds like less hassle.

kchu1337: It looks like the calculations don't always match the aggregate API result. Also for recipient domain, the top domains checkbox throws away some data as well. I'll just keep the aggregate stuff as an API request.

src/actions/summaryChart.js
```diff
@@ -26,14 +30,51 @@ export function refreshSummaryReport(
 
     // convert new meta data into query param format
     const params = getQueryFromOptions(merged);
-
+    const isCurrentGroupingAggregates = summaryChart.groupBy === 'aggregate';
     return Promise.all([
       dispatch(getChartData({ params, metrics: merged.metrics })),
-      dispatch(getTableData({ params, metrics: merged.metrics })),
+      dispatch(getAggregateData({ params, metrics: merged.metrics, isCurrentGroupingAggregates })),
+      !isCurrentGroupingAggregates && dispatch(getTableData({ params, metrics: merged.metrics })),
```
PatrickSison, line 37: I got the gist of it now. Chart + aggregate data is in new flow, and then the table pulls data that is sometimes not aggregate data and avoids duplicate calls. Got it 👍

src/pages/reportBuilder/ReportBuilder.js
```diff
@@ -33,7 +36,26 @@ export function ReportBuilder({
...
+    value: `${moment(from).format('MMM Do')} - ${moment(from).format('MMM Do, YYYY')}`,
+  };
+  const renderAggregateMetric = useCallback(({ label, value, unit }) => {
+    return (
+      <Definition>
+        <Definition.Label>
+          <Box color={tokens.color_gray_600}>{label}</Box>
+        </Definition.Label>
+        <Definition.Value>
+          <Box color={'#ffffff'}>
```
PatrickSison, line 51: The token `color_white` also exists, but this is a nitpick.

PR #1653, opened 2020-06-05
SA-1351: Report Builder Table

kchu1337:
### What Changed
 - Bumped Hibana to v25
 - Copied over many summary report files and made them functional components
 - Updated the summary table.

### How To Test
 - Go to '/reports/summary'. Change the break down selector.


src/pages/reportBuilder/components/GroupByOption.js
```diff
@@ -0,0 +1,79 @@
+import React, { useState } from 'react';
```
kchu1337, line 1: This is mostly the same as from summary report. The main difference is changing from class component -> functional and changing some options in the <Select>

src/pages/reportBuilder/components/ReportTable.js
```diff
@@ -0,0 +1,142 @@
+import React from 'react';
```
kchu1337, line 1: Mostly the same, just removed some aggregates conditional rendering.

src/pages/reportBuilder/components/tests/GroupByOption.test.js
```diff
@@ -0,0 +1,80 @@
...
+  });
+
+  it('should correctly show watched domain in the select options when "Only Top Domains" is checked', () => {
+    expect(wrapper.find('Checkbox')).toBeChecked();
+    expect(wrapper.find('Select').prop('options')).toContainEqual({
+      label: 'Recipient Domain',
+      value: 'watched-domain',
+    });
+  });
+
```
kchu1337, line 53: Mostly the same tests, just adding Checkbox checking and unchecking since I can no longer use setState

src/pages/reportBuilder/constants/tableConfig.js
```diff
@@ -0,0 +1,34 @@
+export const GROUP_CONFIG = {
```
kchu1337, line 1: Copy+paste + remove aggregates

src/pages/reportBuilder/components/ReportTable.js
```diff
@@ -0,0 +1,142 @@
...
+        <Panel.Section>
+          <GroupByOption
+            _getTableData={_getTableData}
+            groupBy={groupBy}
+            hasSubaccounts={hasSubaccounts}
+            tableLoading={tableLoading}
+          />
+        </Panel.Section>
+      </Panel>
+      <div data-id={'summary-table'}>{renderTable()}</div>
```
PatrickSison, line 131: There's a double border + corner radius between the group by select and the table, but I don't see another pattern 
 used in app right now that puts tables and panels cleanly together

src/pages/reportBuilder/components/ReportTable.js
```diff
@@ -0,0 +1,142 @@
...
+    _getTableData,
+    hasSubaccounts,
+    metrics,
+    tableData = [],
+    tableLoading,
+    typeaheadCache,
+  } = props;
+
+  //TODO RB CLEANUP: Change value. Default has to be 'aggregate' for now due to sharing reducer with non-hibana summary report
+  const groupBy = props.groupBy === 'aggregate' ? 'placeholder' : props.groupBy;
```
PatrickSison, line 37: It looks a bit odd with just a single dropdown at the bottom with no table when it's a placeholder. Should we double check with UX if we want to have a default option for the select?

kchu1337: It's what's in the mocks on page 1. I'd say just leave it as is for now and if they want to change it, we can address the feedback in another epic.

PatrickSison: Gotcha 👍 Missed that mock. I'm gonna cut a ticket for UX feedback anyways. They have a few things

PR #1655, opened 2020-06-08
FE-1084 - Redesign <Empty></Empty> component to match updated mocks and introduce in to empty, filterable tables

nicklemmon:
[FE-1084](https://jira.int.messagesystems.com/browse/FE-1084)

### What Changed
- Redesigns `<Empty/>` component for Hibana using new a/c (see ticket)
- Removes baked-in wrapping `<Panel/>` within the `<Empty/>` component (better to use composition than a boolean prop!)
- Incorporate `<Empty/>` component in to the `<Collection/>` component when no `EmptyComponent` is passed in
- `<Empty/>` component renders in `<Collection/>` when filtering results in 0 rows rendered
- Replace almost all instances of the `Callout` component with `Empty` for improved UX consistency - previously, we basically had two very similar designs for empty states
- Updated a few loading states to use the new `PanelSectionLoading` component
- Updates Signals `<ChartHeader/>` component styles to help accommodate empty states - ideally this would be something more generic like `Panel.Header` in the future?
- Deleted unused preview components whose references were actually removed in a previous

### How To Test

- [x] Check out a page using the `Collection` component with a `FilterBox` and filter the results until 0 are returned to verify the empty state (`/templates` should do)
- [x] Go to `/signals/health-score` and modify the code to render the empty state for the `SummaryTable` component
- [x] Check out the empty state for the Engagement Recency tabs:
  + `/signals/engagement/cohorts/sid/<id>`
  + `/signals/engagement/engagement-rate/sid/<id>`
  + `/signals/engagement/unsubscribes/sid/<id>`
  + `/signals/engagement/complaints/sid/<id>`
- [x] Check out the empty states for `/signals/health-score/sid/<id>`
- [x] Check out the empty states for `/signals/spam-traps/sid/<id>`
- [x] Check out the empty state for `/signals/integration`
- [x] Check out the empty state for `/reports/bounce`
- [x] Check out the empty state for `/reports/rejections`
- [x] Check out the empty state for `/reports/accepted`
- [x] Check out the empty state for `/reports/delayed`

### To Do
- [ ] Incorporate feedback


src/components/empty/Empty.module.scss
```diff
@@ -1,8 +1,14 @@
 @import '~@sparkpost/matchbox/src/styles/config.scss';
 
-.Center {
+.Message {
   @include text-style-heading;
   text-align: center;
   padding: rem(80) 0;
-  color: color(gray, 6);
+  color: color(gray, 4);
```
nicklemmon, line 7: Old color wasn't meeting a11y standards.

src/pages/inboxPlacement/components/TrendsChart.js
```diff
@@ -130,12 +124,10 @@ export const TrendsChart = props => {
...
     <>
       {hasNoData ? (
-        <div>
-          <Callout height={panelHeight}>Inbox Placement Trends Not Available</Callout>
-        </div>
+        <Empty message="Inbox Placement Trends Not Available" />
       ) : (
         <>
-          {/*float:right does't work for some reason. Causes tooltip to not show when hovering bottom of chart*/}
+          {/* float:right doesn't work for some reason. Causes tooltip to not show when hovering bottom of chart */}
```
nicklemmon, line 130: Had to fix the typo!

src/pages/reports/bounce/BouncePage.js
```diff
@@ -182,8 +182,12 @@ export class BouncePage extends Component {
...
         <ReportOptions reportLoading={chartLoading} searchOptions={bounceSearchOptions} />
         {this.renderTopLevelMetrics()}
-        {this.renderTabs()}
-        {this.renderChart()}
+
+        <Panel>
+          <TabsWrapper>{this.renderTabs()}</TabsWrapper>
+
+          <Panel.Section>{this.renderChart()}</Panel.Section>
+        </Panel>
```
nicklemmon, line 190: Needed to do this to address some double border issues with Hibana

src/pages/signals/components/Actions.module.scss
```diff
@@ -28,7 +23,8 @@
   border: none;
   margin-bottom: spacing();
   height: 1px;
-  background: center left / 10px 1px  linear-gradient(to right, transparent 40%, color(gray, 8) 40%) repeat-x;
+  background: center left / 10px 1px linear-gradient(to right, transparent 40%, color(gray, 8) 40%)
```
nicklemmon, line 26: Just Prettier here

bkemper:
> Removes baked-in wrapping <Panel/> within the <Empty/> component (better to use composition than a boolean prop!)
👍

bkemper:
It looks like sometimes there is a border...

Screen Shot 2020-06-09 at 11 56 30 PM

and sometimes not.

Screen Shot 2020-06-09 at 11 56 43 PM

Should we make these all the same?

src/components/collection/Collection.js
```diff
@@ -153,17 +155,27 @@ export class Collection extends Component {
     } = this.props;
 
     if (!rows.length) {
-      return EmptyComponent ? <EmptyComponent /> : null;
+      return EmptyComponent ? <EmptyComponent /> : <Empty message="No results found." />;
```
bkemper, line 158: Do we need this case?  It looks like this prop was created for Blocklist incidents and it is doing way more than just rendering an empty message.  I would vote for deleting it and the use of it in Blocklist incidents.

src/components/summaryTable/Billboard.js
```diff
@@ -1,18 +1,14 @@
...
-  const { isHibanaEnabled } = state;
+import { Table } from 'src/components/matchbox';
 
+function Billboard({ children }) {
   return (
     <tbody>
       <Table.Row>
-        <Table.Cell className={classNames(!isHibanaEnabled && styles.Billboard)} colSpan={colSpan}>
-          <Box height="220px">{children}</Box>
+        <Table.Cell colSpan="100%" style={{ padding: 0 }}>
```
bkemper, line 8: I thought `colSpan` had to be an integer.
— https://developer.mozilla.org/en-US/docs/Web/HTML/Element/td#attr-colspan

src/components/text/tests/Heading.test.js
```diff
@@ -1,23 +1,34 @@
...
+
     return render(<OGHeading {...defaultProps} {...props} />);
   };
 
   describe('with Hibana enabled', () => {
     it('renders the passed in heading level controlled via the "as" prop', () => {
-      const { container } = hibanaSubject();
+      const { debug, container } = hibanaSubject();
+
+      debug();
```
bkemper, line 31: This needs to be removed.

src/pages/reports/delays/components/DelaysDataTable.js
```diff
@@ -39,7 +40,11 @@ export class DelaysDataTable extends Component {
     const { rows, title } = this.props;
 
     if (_.isEmpty(rows)) {
-      return <Empty title={title} message={'No delayed messages to report'} />;
+      return (
+        <Panel>
+          <Empty title={title} message={'No delayed messages to report'} />
```
bkemper, line 45: The `title` needs to be moved to `<Panel>` or removed.

src/components/summaryTable/Billboard.js
nicklemmon: Technically it ends up being rendered as having 100 columns, but that works for all tables/browsers so long as they don't have the CSS `table-layout: fixed;`

https://stackoverflow.com/questions/398734/colspan-all-columns

nicklemmon: Basically we get the same result as dynamically setting the `colspan` in practice without having to manage the code to do it. It's a _little_ hacky, but I would argue it's easier to do this way and it reads well.

src/components/collection/Collection.js
nicklemmon: sounds good! I can do that.

nicklemmon:
@bkemper Ready for another pass

src/components/collection/Collection.js
```diff
@@ -141,29 +143,32 @@ export class Collection extends Component {
...
       <OuterWrapper>
         <HeaderComponent />
         <BodyWrapper>
-          {this.getVisibleRows().map((row, i) => (
-            <RowComponent key={`${row[rowKeyName] || 'row'}-${i}`} {...row} />
-          ))}
+          {visibleRows.length === 0 ? (
+            <tr>
+              <td colSpan="100%" style={{ padding: 0 }}>
+                <Empty message="No results found." />
```
bkemper, line 164: 👏

src/pages/blacklist/components/IncidentsCollection.js
```diff
@@ -73,22 +72,6 @@ export const IncidentsCollection = props => {
...
   );
 
-  const EmptyComponent = () => {
-    const textFieldComponent = (
-      <div className={styles.FilterBox}>
-        <TextField disabled id="disabled-blacklist-search" suffix={<Search />} />
-      </div>
-    );
-    return (
-      <Panel>
```
bkemper, line 83: 👏

PR #1658, opened 2020-06-09
SA-1439: Report Builder tabs scaffolding

kchu1337:
### What Changed
 - Added `tab` prop to the metrics configs.
 - Changed `Tab` component to only render the tabs if there is >1.
 - Added the tabs into Report Builder
 - Changed the styling of the reportbuilder so it matches the mocks. Basically moved the 2 panels into a single panel.
 - Added state variable to determine whether or not to show the breakdown table.
 -  Fixed some older metrics stuff that was overlooked before.

### How To Test
 - Go to `/reports/summary`
 - You should see a bounce reasons tab by default. 
 - Click on the tab. Both the chart and table should be hidden
 - Remove bounce metric. The bounce reasons tab should not be visible.
 - Add rejection, click, and delayed metric. The respective tabs should be visible.


src/components/tabs/Tabs.js
```diff
@@ -3,13 +3,12 @@ import { Tabs as TabsComponent } from 'src/components/matchbox';
 import useTabs from 'src/hooks/useTabs';
 
 function Tabs(props) {
-  const { children, defaultTab = 0, forceRender = false, tabs: setTabs, ...rest } = props;
-
-  const [selectedTabIndex, tabs] = useTabs(setTabs, defaultTab);
+  const { children, defaultTabIndex = 0, forceRender = false, tabs: setTabs, ...rest } = props;
```
kchu1337, line 6: Just reworded defaultTab -> defaultTabIndex

src/pages/reportBuilder/components/MetricsDrawer.js
```diff
@@ -80,7 +80,9 @@ export default function MetricsDrawer(props) {
   const { DrawerFooter = Drawer.Footer } = props;
   return (
     <>
-      <Box margin="400">{renderMetricsCategories()}</Box>
+      <Box margin="400" paddingBottom={'90px'}>
```
kchu1337, line 83: Added this because the drawer footer was cutting off some metrics options.

src/pages/reportBuilder/components/ReportOptions.js
```diff
@@ -64,6 +64,12 @@ export function ReportOptions(props) {
   );
 
   const handleRemoveMetric = selectedMetric => {
+    const { metrics } = reportOptions;
+    if (metrics.length <= 1) {
```
kchu1337, line 68: Prevent users from removing all the metrics on the chart and causing an error. Could fix it the other way by changing how it's rendered but this was just a quick fix

src/pages/reportBuilder/ReportBuilder.js
```diff
@@ -39,7 +76,7 @@ export function ReportBuilder({
   const { to, from } = summarySearchOptions;
   const dateLabelValue = {
     label: 'Date',
-    value: `${moment(from).format('MMM Do')} - ${moment(from).format('MMM Do, YYYY')}`,
+    value: `${moment(from).format('MMM Do')} - ${moment(to).format('MMM Do, YYYY')}`,
```
kchu1337, line 79: Typo from previous ticket.

src/pages/reportBuilder/ReportBuilder.js
```diff
@@ -60,39 +97,50 @@ export function ReportBuilder({
...
-                  })}
-                </Inline>
-              </Grid.Column>
-            </Grid>
-          </Box>
+      <Panel>
+        <ReportOptions reportLoading={chart.chartLoading} searchOptions={summarySearchOptions} />
+        <hr className={styles.Line} />
+        <div data-id="summary-chart">
+          <Tabs defaultTabIndex={0} forceRender tabs={tabs}>
```
PatrickSison, line 104: If we want to avoid making too many API requests, we might have to remove forceRender for this set of tabs so the tables pull data only when viewed. I wonder though if that's going to negatively affect rendering the chart.

PatrickSison: An alternative is we can change how force render works by either passing in a boolean (where it renders everything), or an array of booleans, where the array dictates which indices are forced rendered. Then we can have the first tab always force rendered, and the rest are not?

PatrickSison: Or actually, on Tab.Item itself

kchu1337: I think it would be good for each item. Let me see if I can do that.

src/pages/reportBuilder/components/ReportOptions.js
```diff
@@ -16,6 +16,7 @@ import AddFiltersSection from './AddFiltersSection';
 import DateTimeSection from './DateTimeSection';
 import useRouter from 'src/hooks/useRouter';
 
+const drawerTabs = [{ content: 'Metrics' }, { content: 'Filters' }];
```
kchu1337, line 19: Because of my changes in useTabs, I need the tabs to be constant or else the useEffect would run on each re-render

src/components/tabs/Tabs.js
```diff
@@ -20,13 +19,21 @@ function Tabs(props) {
 function Item(props) {
   const { forceRender, selected, children } = props;
 
+  //Prevent rendering the tab children unless tab has already been selected before
+  const [hasRendered, setHasRendered] = useState(false);
```
PatrickSison, line 23: I was thinking an override flag on the item itself like 
```javascript
   <Tab.Item forceRender />
   <Tab.Item />
```
but this is pretty neat. Didn't think of this way 👍

PR #1660, opened 2020-06-12
SA-1335: Updating chart styling for hibana

PatrickSison:
### What Changed
 - Updating charts + tooltip to match mocks and new hibana style

### How To Test
 - On an account with Hibana enabled, open Report builder and access the chart
 - Select several various metrics
 - Check that styles match correctly

### Notes
- Bar used as a background. We could try searching for max value, but that seems like expensive operation
- Bar doesn't have background hover operation


src/pages/reportBuilder/components/async/LineChart.js
```diff
@@ -0,0 +1,165 @@
...
+          />
+          <Tooltip
+            cursor={{ fill: '#ff0' }}
+            content={<CustomTooltip />}
+            isAnimationActive={false}
+            itemSorter={orderDesc}
+            labelFormatter={tooltipLabelFormatter}
+            formatter={tooltipValueFormatter}
+          />
+          <Bar background={{ fill: tokens.color_gray_100 }} />
```
jonambas, line 157: I remember I did bar hover in recharts at some point, not sure if its anywhere in app. I think what I did was i saved selected / hovered `date` or `tick`. So this bar should have `onMouseOver={(e) => setHoveredDate(e.name)} ` < I know e will include the tick info

And you can also do something like this i think
```js
shape={(props) => (
   <Rectangle fill={props.??? === hoveredDate} />
)}
```

bkemper:
I saw the note about axis labeling, but I think we should do something instead of nothing.  I don't know which day it is until I hover.

Screen Shot 2020-06-17 at 7 32 20 AM

bkemper:
Shouldn't we set a max height on the chart?

Screen Shot 2020-06-17 at 7 36 54 AM

src/pages/reportBuilder/ReportBuilder.module.scss
```diff
@@ -2,19 +2,13 @@
 
 .ChartSection {
   position: relative;
-  padding: spacing(300);
-  min-height: 350px;
+  min-height: 250px;
 
   opacity: 1;
   pointer-events: auto;
```
bkemper, line 8: Can't this be removed since we aren't toggling between auto and none?

jonambas:
> Shouldn't we set a max height on the chart?
> 
> Screen Shot 2020-06-17 at 7 36 54 AM

👋  Chart heights are supposed to be dynamic too. I think 400 for one chart, 200 for two charts, and 150 for 3+ charts

src/pages/reportBuilder/components/Charts.js
```diff
@@ -0,0 +1,53 @@
...
+
+function getUniqueUnits(metrics) {
+  return _.uniq(metrics.map(({ unit = DEFAULT_UNIT }) => unit));
+}
+
+export default function Charts(props) {
+  const { chartData = [], metrics, chartLoading, precision, yScale, to } = props;
+
+  if (!chartData.length || !metrics) {
+    return null;
```
bkemper, line 17: Should we provide a slightly different experience if all metrics are removed?  

Maybe display an Empty component with a friendly message and putting a guard in place to avoid calling action when no metrics are selected.


Screen Shot 2020-06-17 at 4 20 34 PM

src/pages/reportBuilder/components/Charts.js
```diff
@@ -0,0 +1,53 @@
...
+  const charts = getUniqueUnits(metrics).map(unit => ({
+    metrics: metrics.filter(metric => metric.unit === unit),
+    ...METRICS_UNIT_CONFIG[unit],
+  }));
+
+  return (
+    <div>
+      {charts.map((chart, i) => (
+        <LineChart
+          key={`chart=${i}`}
```
bkemper, line 32: nit: Maybe replace "=" with "-" and `i` with `unit`.

src/pages/reportBuilder/components/Charts.js
```diff
@@ -0,0 +1,53 @@
...
+          syncId="summaryChart"
+          data={chartData}
+          precision={precision}
+          lines={chart.metrics.map(({ name, label, stroke }) => ({
+            key: name,
+            dataKey: name,
+            name: label,
+            stroke: chartLoading ? '#f8f8f8' : stroke,
```
bkemper, line 40: This is really cool!  We should use a design_token.

src/pages/reportBuilder/components/async/LineChart.js
```diff
@@ -0,0 +1,178 @@
...
+            content={<CustomTooltip />}
+            isAnimationActive={false}
+            itemSorter={orderDesc}
+            labelFormatter={tooltipLabelFormatter}
+            formatter={tooltipValueFormatter}
+          />
+          {renderLines()}
+        </ComposedChart>
+      </ResponsiveContainer>
+      <span className="sp-linechart-yLabel">{yLabel}</span>
```
bkemper, line 175: Does this need to be a global style?

PatrickSison:
Still fixing the zIndex with the tooltip

src/pages/reportBuilder/components/Charts.module.scss
```diff
@@ -0,0 +1,9 @@
+@import '~@sparkpost/design-tokens-hibana/tokens';
+
+.ChartsWrapper > div {
+  margin-bottom: spacing(400);
```
bkemper, line 4: This can be replaced with Stack.  For the y-axis label, set the position on the Box.

PR #1661, opened 2020-06-12
SA-1417 -  Inbox Placement list page crashes when user's session expires and logs in again

shrirupa:
SA-1417 -  Inbox Placement list page crashes when user's session expires and logs in again

### What Changed
 - getTrendsFilterValuesError should be a string; TypeAhead expects a string

### How To Test
 - Login and open the network tab to get the token sent by authenticate. Copy that token.
 - Navigate to inbox-placement page
 - Use postman and using POST /authenticate/logout invalidate the token
 - Now on inbox-placement page, change the date range filter to something else; this should trigger a logout. Watchout for two error alerts meaning message-filter api and inbox-placement api must have thrown 401.
- Now login again, you should have landed on inbox-placement page without it crashing.

### To Do
- [ ] Address feedback

### DIFF
--- a/src/reducers/inboxPlacement.js
+++ b/src/reducers/inboxPlacement.js
@@ -54,7 +54,11 @@ export default (state = initialState, { type, payload, meta }) => {
       };
     }
     case 'GET_INBOX_PLACEMENT_TRENDS_FILTER_VALUES_FAIL':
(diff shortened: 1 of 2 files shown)


bkemper:
Instead of using Postman, I opened a second tab and logged out.

PR #1662, opened 2020-06-15
FE-1095 - Update @testing-library dependencies

nicklemmon:
[FE-1095](https://jira.int.messagesystems.com/browse/FE-1095)

### What Changed
- Updates libraries outlined in the tickets
- Removes all `query*` queries from Cypress tests
- Removed a couple of RTL tests that were causing problems following the `@testing-library/user-event` upgrade - rather than going down a huge rabbit hole, I realized these tests were duplicated by Cypress

### How To Test
- Verify packages were updated to their latest versions
- Verify tests continue to pass

### To Do
- [ ] Incorporate feedback


config/jest/matchers/toHaveStyle.js
```diff
@@ -1,4 +1,4 @@
-import { toHaveStyle as testingLibraryToHaveStyle } from '@testing-library/jest-dom';
+import { toHaveStyle as testingLibraryToHaveStyle } from '@testing-library/jest-dom/dist/to-have-style';
```
bkemper, line 1: `toHaveStyle` is available at `@testing-library/jest-dom/matchers`.

— https://github.com/testing-library/jest-dom#usage

package.json
```diff
@@ -128,11 +128,11 @@
...
-    "@testing-library/cypress": "^5.3.1",
-    "@testing-library/jest-dom": "^4.2.4",
-    "@testing-library/react": "^9.3.2",
-    "@testing-library/react-hooks": "^3.2.1",
-    "@testing-library/user-event": "^7.1.2",
+    "@testing-library/cypress": "^6.0.0",
+    "@testing-library/jest-dom": "^5.10.1",
+    "@testing-library/react": "^10.2.1",
+    "@testing-library/react-hooks": "^3.3.0",
+    "@testing-library/user-event": "^11.4.2",
```
bkemper, line 135: It would be good to post some highlights in our public channel.

config/jest/matchers/toHaveStyle.js
nicklemmon: oooh let me update those - they were originally just available at the index, but that clearly changed.

PR #1664, opened 2020-06-16
FE-1101: Change whitelist to allowlist in 2web2ui

bkemper:
### What Changed
 - Find and replace on all uses of "whitelist" or "whitelabel" in repo

### How To Test
 - Check build to confirm Cypress tests don't break
 - Confirm `usePageFilters` hook is working as expected on /blacklist/incidents and /inbox-placement


kchu1337:
https://github.com/cypress-io/cypress/pull/7622

PR #1666, opened 2020-06-16
SA-1424: Add Bounce Reasons tab

kchu1337:
### What Changed 
 - Added bounce reasons tab to report builder

### How To Test
 - go to `reports/summary` with hibana enabled. Select a metric with bounce (should be selected by default)
 - Verify that the table appears.


src/pages/reportBuilder/components/tabs/Wrappers.module.scss
```diff
@@ -0,0 +1,5 @@
+@import '~@sparkpost/design-tokens-hibana/tokens';
```
shrirupa, line 1: We are not using this import here right?

src/pages/reportBuilder/components/tabs/Wrappers.js
```diff
@@ -0,0 +1,35 @@
...
+
+export const TableWrapper = props => (
+  <>
+    <Table>{props.children}</Table>
+    <hr className={styles.Line} />
+  </>
+);
+
+export const EmptyWrapper = ({ message }) => (
+  <Box height={'200px'} paddingTop={'70px'}>
```
shrirupa, line 26: `{ }` may be removed from height="200px".

src/pages/reportBuilder/components/tabs/BounceReason.js
```diff
@@ -0,0 +1,82 @@
...
+
+  useEffect(() => {
+    if (reportOptions.to && reportOptions.from) {
+      refreshBounceReport(reportOptions);
+    }
+  }, [refreshBounceReport, reportOptions]);
+
+  const getRowData = useCallback(
+    item => {
+      const { reason, domain, bounce_category_name, bounce_class_name, count_bounce } = item;
```
shrirupa, line 41: Just curious, that this calculation included 
```
if (bounce_category_name === 'Admin') {
      numerator = count_admin_bounce;
      denominator += aggregates.countAdminBounce;
    }
```
in the old BouncePage, why is it different here ?

kchu1337: The old bounce page included admin bounces as it's own separate category but this one has all the bounces lumped together.

PR #1667, opened 2020-06-16
FE-1098 - Incorporate UX feedback for the events feature

nicklemmon:
[FE-1098](https://jira.int.messagesystems.com/browse/FE-1098)

### What Changed
- Made updates to the events list and details pages, incorporating UX feedback listed in the ticket

### How To Test
- Sign in with the usual account!
- Go to `/reports/message-events` and verify the following changes were made:
  - [x] The "Add Filters" and "Share" buttons use the `"secondary"` variant
  - [x] The "Remove" and "Add Filter" icons within the "Add Filters" modal were updated
  - [x] The `<Select/>` components used in the "Add Filters" modal are wider, allowing content to be read
  - [x] The form filtering `<Panel/>` and the `<TableCollection/>` are now connected
- Go to `/reports/message-events/details/<eventID>` and verify the following changes were made:
  - [x] There is a section divider below each `<Panel/>` title (note, the way I did it was hacky - I opened a feature request for Matchbox to make this better: https://github.com/SparkPost/matchbox/issues/467)
  - [x] When an event is selected in the "Message History" table, the font weight is semibold and the color is `blue.700`
  - [x] Verify the labels paired with form fields in the "Event Details" `<Panel/>` are vertically centered
 
### To Do
- [ ] Incorporate feedback


src/pages/reports/messageEvents/components/EventDetails.js
```diff
@@ -15,7 +13,7 @@ class EventDetails extends Component {
     // Renders value in a read only textfield
     if (typeof value === 'object') {
       return (
-        <LabelledValue key={key} label={label}>
+        <LabelledValue key={key} label={label} style={{ alignItems: 'center' }}>
```
kchu1337, line 16: I'm not seeing any difference between this and what's in prd.

src/pages/reports/messageEvents/components/EventDetails.js
```diff
@@ -54,6 +52,8 @@ class EventDetails extends Component {
 
     return (
       <Panel title="Event Details">
+        {/* TODO: We could stand to have a component like <Panel.Header> from Matchbox to address this scenario */}
+        <Box paddingTop="500" borderBottom="400" />
```
kchu1337, line 56: Instead of using title, you could put it in it's own Panel.Section. I've also been using <hr> in places. But this seems to work.

kchu1337: `<hr>`

src/pages/reports/messageEvents/components/EventDetails.js
nicklemmon: The label is vertically centered doing this - otherwise, it aligns to the top of its parent. That's all!

src/pages/reports/messageEvents/components/EventDetails.js
nicklemmon: I did try that! But the end result caused the two Panels to have different heights to their containers. Doing it this way allowed me to use a similar method with the other `Panel` to keep the height consistent.

It's all pretty hacky. I filed a feature request with Matchbox:
https://github.com/SparkPost/matchbox/issues/467

PR #1676, opened 2020-06-19
SA-1481: Fixed bug of report initializing twice

PatrickSison:
### What Changed
 - Added check for fully initializing params of report options before making request
   - Change comes from the translation from Class to functional with lifecycle methods to hooks

### How To Test
 - Load Report builder with options different from default
 - Refresh page
 - Check for network requests. There should only be one initial time series instead of 2

### To Do
- [ ] See if there's a better pattern for initialization rather than a flag
   - Could go with the option of making it a string enum.


bkemper:
I think this PR is missing a commit.

src/pages/reportBuilder/ReportBuilder.js
```diff
@@ -8,33 +8,68 @@ import { Box, Grid, Inline } from 'src/components/matchbox';
...
+
+  /**
+   * Updating
+   * 1. Report options update
+   * 2a. Retrieve data
+   * 2b. Update query params as needed (inside ReportOptions)
+   */
+
+  const { location } = useRouter();
+  const [initialized, setInitialized] = useState(false);
```
PatrickSison, line 57: This doesn't feel right, but not sure on a better to avoid running the effect the first time every mount.

bkemper:
This pull request can be withdrawn.  The bug has been fixed in #1688

PR #1677, opened 2020-06-19
SA-1425: Add Rejection Reasons Tab

kchu1337:
### What Changed 
 - Added rejection reasons tab to report builder

### How To Test
- Go to `reports/summary` with hibana enabled. Check that all only these metrics show the rejection tab. 
    - Rejected
    - Generation Rejection
    - Generation Failure
    - Policy Rejection
 - Verify that the tab and table appears.
 - Also check that I changed any text correctly when I copy+pasted


src/pages/reportBuilder/components/tabs/RejectionReasonsTable.js
```diff
@@ -0,0 +1,77 @@
...
+  { label: 'Reason', width: '45%', sortKey: 'reason' },
+  { label: 'Domain', sortKey: 'domain' },
+];
+
+export function RejectionReasonsTable(props) {
+  const { loading, reasons, refreshRejectionReport, reportOptions } = props;
+
+  useEffect(() => {
+    if (reportOptions.to && reportOptions.from) {
+      refreshRejectionReport(reportOptions);
```
PatrickSison, line 35: Do we need to refresh the whole rejection report? This extends to the other reports tabs. I think the only content in the table shown is the data from fetchRejectionReasonsByDomain?

src/pages/reportBuilder/components/tabs/RejectionReasonsTable.js
```diff
@@ -0,0 +1,77 @@
...
+  TableWrapper,
+} from './Wrappers';
+
+const filterBoxConfig = {
+  show: true,
+  keyMap: { category: 'rejection_category_name' },
+  itemToStringKeys: ['rejection_category_name', 'domain', 'reason'],
+  exampleModifiers: ['domain', 'category'],
+  matchThreshold: 5,
+  wrapper: FilterBoxWrapper,
```
PatrickSison, line 20: We should probably change the label of the input field to be the matchbox label, and use the id between the label and the input field

src/pages/reportBuilder/components/tabs/RejectionReasonsTable.js
kchu1337: The other ones need aggregate data for % calculations. Yeah this one doesn't but `fetchRejectionReasonsByDomain` needs to pass in the metrics and we might need aggregate data for % later. It's kinda of weird to just have a whole number rather than % like in the other tables. Most likely UX just copied the current table, which shows the total rejections in a panel at the top.

PR #1678, opened 2020-06-22
SA-1431- Create Cypress Tests for Report Builder

nicklemmon:
[SA-1431](https://jira.int.messagesystems.com/browse/SA-1431)

### What Changed
- Added Cypress tests for 'Report Builder' ('Analytics Report'? Should the files be renamed?)
- Cypress-produced screenshots are now uploaded to CircleCI's storage when errors occur in integration tests
- Introduced a few style tweaks here and there where necessary
- Added a few `data-id` attributes for tests

### How To Test
- Run the tests locally and be sure they pass:
  + `npm run start-app-hibana` in one terminal window/tab
  + `npm run test-integration-hibana` in one terminal window/tab
- Go through tests and verify relevant scenarios are captured by the automated testing
- I am _sure_ there are scenarios I've missed here

### To Do
- [ ] Incorporate feedback


cypress/support/commands.js
```diff
@@ -110,6 +110,18 @@ Cypress.Commands.add('stubAuth', () => {
...
+    url: '/sockjs-node/**/*',
+    status: 200,
+    response: {},
+  }).as('stubbedSockNodePost');
+  cy.route({
+    method: 'GET',
+    url: '/sockjs-node/**/*',
+    status: 200,
+    response: {},
+  }).as('stubbedSockNodeGet');
```
nicklemmon, line 124: The dev server makes these requests periodically - this just stubs those to basically do nothing instead of erroring

src/pages/reportBuilder/components/DateTimeSection.js
```diff
@@ -40,8 +40,11 @@ const DateTimeSection = ({
   return (
     <Grid>
       <Grid.Column xs={12} md={6}>
-        <div className={styles.FieldWrapperMetricsRollup}>
-          <label>Date Range</label>
+        <Box marginBottom={[200, 200, 200, 0]}>
```
nicklemmon, line 43: I cleaned this up a bit - though may have introduced some unusual styles just handling the margin this way - let me know what you all think.

The label was rendering incorrectly - needed to use the `<Label/>` component.

cypress/tests/integration/signals-analytics/analytics-report/constants/index.js
```diff
@@ -0,0 +1,173 @@
+export const STABLE_UNIX_DATE = 1581087062000; // Stable unix timestamp (2/6/2020)
+export const PAGE_URL =
+  '/reports/summary?from=2020-06-22T19%3A00%3A00Z&to=2020-06-23T19%3A03%3A48Z&range=day&timezone=America%2FNew_York&precision=hour&metrics=count_targeted&metrics=count_rendered&metrics=count_accepted&metrics=count_bounce';
+export const METRICS = [
```
nicklemmon, line 4: Importing this from the config file was pretty messy - Cypress doesn't have access to our Webpack aliases - that being said, I could be persuaded that that would be a better way to go here. Not sure! That does end up tying our tests more closely to the actual code that way so may not be the right choice.

src/pages/reportBuilder/components/DateTimeSection.js
PatrickSison: I've actually swapped this out in the PR by adding in the label properly for the TextField, so it we can probably avoid the margin issue

cypress/tests/integration/signals-analytics/analytics-report/analyticsReportPage.spec.js
```diff
@@ -0,0 +1,496 @@
...
+
+      // Filtering form elements
+      cy.findByText('Date Range').should('be.visible');
+      // These labels are visually hidden though are associated with the date picker controls
+      cy.findByLabelText('Narrow Date Range').should('be.visible');
+      cy.findByLabelText('Broad Date Range').should('be.visible');
+      cy.findByLabelText('Time Zone').should('be.visible');
+      cy.findByLabelText('Precision').should('be.visible');
+
+      // Default selected metrics
```
PatrickSison, line 23: A few of these will need to change with the date picker updates, but I can do that in that PR

.circleci/config.yml
```diff
@@ -51,6 +51,9 @@ commands:
       - store_artifacts:
           path: /home/circleci/project/cypress/videos
           destination: cypress-videos
+      - store_artifacts:
```
PatrickSison, line 54: Do we screenshot anywhere at the moment? I don't have issue with this, just curious.

nicklemmon: Basically when errors occur, Cypress will take screenshots in addition to video. This just uploads those so we can see the point of failure in a little higher res than the video.

PR #1680, opened 2020-06-22
FE-1106 - Fix outline offset bug

nicklemmon:
[FE-1106](https://jira.int.messagesystems.com/browse/FE-1106)

### What Changed
- Added some custom CSS to address clipped focus outlines
- See: https://jira.int.messagesystems.com/browse/FE-1106 -> for some reason with the new Chrome outline customs styles, just applying `outline-offset` does not work. Oddly enough, all of their custom outline styles have to be overridden.

### How To Test
- Go to `/recipient-validation/list`, tab through the navigation and verify all sides of the focus outline are visible

### To Do
- [ ] Incorporate feedback


bkemper:
This is very odd.  I image we will be revisiting this again in the near future.

nicklemmon:
> This is very odd. I image we will be revisiting this again in the near future.

Jon is working on implementing custom focus styles so that we have consistency across browsers + slightly nicer appearance. Hopefully we can use some sort of mixin or design tokens to easily apply those same styles to other elements so we aren't at the mercy of the browser vendors.

PR #1681, opened 2020-06-22
FE-1111 - Sentry now renders a tag for isHibanaEnabled

nicklemmon:
[FE-1111](https://jira.int.messagesystems.com/browse/FE-1111)

### What Changed
- Added new tag to Sentry that passes on the user's `optons.ui.isHibanaEnabled` state

### How To Test
1. Sign in to the app and make sure Hibana is disabled
1. Go to `/error-boundary` and take a look at the network request made to Sentry - verify that `tags: { isHibanaEnabled: false }` is in the request body
1. Enable Hibana
1.Go to `/error-boundary` and verify the request body now includes `tags: { isHibanaEnabled: true }`
1. Log out and create a new account without the feature flag enabled
1. Go to `/error-boundary` and verify the request body contains `tags: { isHibanaEnabled: false }`

### To Do
- [ ] Incorporate feedback


src/helpers/errorTracker.js
```diff
@@ -98,12 +98,13 @@ function filterRequest(request) {
 export function getEnricherOrDieTryin(store, currentWindow) {
   return function enrich(data) {
     const { currentUser } = store.getState();
-    const user = _.pick(currentUser, ['access_level', 'customer', 'username']);
+    const user = _.pick(currentUser, ['access_level', 'customer', 'username', 'options.ui']);
```
bkemper, line 101: Should we pick the ui options or just use `user` later when defining `isHibanaEnabled`?

nicklemmon: @bkemper  - Not sure I follow. Were you thinking just `_.get` directly from `currentUser`?

bkemper: Just discussed.

nicklemmon:
@bkemper Ready for another pass!

src/helpers/errorTracker.js
```diff
@@ -125,6 +126,7 @@ export function getEnricherOrDieTryin(store, currentWindow) {
         // all tags can be easily searched and sent in Slack notifications
         ...data.tags,
         customer: _.get(user, 'customer'),
+        isHibanaEnabled: isHibanaEnabled ? isHibanaEnabled : false, // Default to `false` when no flag is present
```
bkemper, line 129: nit: This is equivalent to `Boolean(isHibanaEnabled)`.

PR #1683, opened 2020-06-23
AC-1474: (UI) Add the Delete SCIM token flow

shrirupa:
AC-1474: (UI) Add the Delete SCIM token flow

### What Changed
 - Clicking on the Delete token opens Delete scim token modal.  https://sparkpost.invisionapp.com/public/share/FR169HZ78E#/screens/478309246

### How To Test
 - Click on the Delete Token in SCIM token section, it opens the Delete SCIM token modal. 
 - Run the cypress test added. This should help in checking the flow.

### To Do
- [ ] Address Feedback
- [x] Need to call the Delete endpoint when user overrides a SCIM token

### DIFF
--- a/cypress/tests/integration/accounts/account-settings/accountSettingsPage.js
+++ b/cypress/tests/integration/accounts/account-settings/accountSettingsPage.js
@@ -125,6 +125,7 @@ describe('Account Settings Page', () => {
             cy.wait('@scimTokenDelete');
             cy.wait('@scimTokenGet');
           });
+          cy.findByText('SCIM token deleted').should('be.visible');
           cy.findByText('No token generated').should('be.visible');
           cy.findByText('Delete token').should('not.be.visible');
         });
--- a/cypress/tests/integration/accounts/account-settings/accountSettingsPage.js
+++ b/cypress/tests/integration/accounts/account-settings/accountSettingsPage.js
@@ -94,6 +94,41 @@ describe('Account Settings Page', () => {
           cy.findByText('123f••••••••').should('be.visible');
         });
       });
+      describe('Delete token flow: ', () => {
+        beforeEach(() => {
+          cy.stubRequest({
+            url: 'api/v1/api-keys?grant=scim/manage',
+            fixture: 'api-keys/200.get.scim-token.json',
+            requestAlias: 'scimTokenGet',
+          });
+          cy.visit(PAGE_URL);
+        });
+        it('Delete Token flow for scim token', () => {
+          cy.findByText('old1••••••••').should('be.visible');
+          cy.findByText('Delete Token').click();
+          cy.withinModal(() => {
+            cy.findAllByText('Delete SCIM Token').should('be.visible');
+            cy.stubRequest({
+              url: 'api/v1/api-keys/oldid',
+              method: 'DELETE',
+              fixture: 'api-keys/200.get.scim-token-notoken.json',
+              requestAlias: 'scimTokenDelete',
+            });
+            cy.stubRequest({
+              url: 'api/v1/api-keys?grant=scim/manage',
+              fixture: 'api-keys/200.get.scim-token-notoken.json',
+              requestAlias: 'scimTokenGet',
+            });
+            cy.get('button')
+              .contains('Delete SCIM Token')
+              .click();
+            cy.wait('@scimTokenDelete');
+            cy.wait('@scimTokenGet');
+          });
+          cy.findByText('No token generated').should('be.visible');
+          cy.findByText('Delete token').should('not.be.visible');
+        });
+      });
     });
   });
 });
--- a/cypress/tests/integration/accounts/account-settings/accountSettingsPage.js
+++ b/cypress/tests/integration/accounts/account-settings/accountSettingsPage.js
@@ -83,12 +83,17 @@ describe('Account Settings Page', () => {
             fixtureAlias: 'newScimTokenGet',
             requestAlias: 'newScimTokenGet',
           });
+          cy.stubRequest({
+            url: 'api/v1/api-keys/oldid',
+            method: 'DELETE',
+            fixture: 'api-keys/200.get.scim-token-notoken.json',
+            requestAlias: 'scimTokenDelete',
+          });
 
           cy.withinModal(() => {
             cy.findByText('Override Your Current Token?').should('be.visible');
             cy.findByText('Generate New Token').click();
-            cy.wait('@scimTokenCreate');
-            cy.wait('@newScimTokenGet');
+            cy.wait(['@scimTokenDelete', '@scimTokenCreate', '@newScimTokenGet']);
             cy.findByText('Continue').click();
           });
           cy.findByText('123f••••••••').should('be.visible');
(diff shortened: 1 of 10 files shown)


nicklemmon:
@shrirupa This is on my radar for today! Got delayed with some Report Builder stuff.

cypress/tests/integration/accounts/account-settings/accountSettingsPage.js
```diff
@@ -94,6 +94,42 @@ describe('Account Settings Page', () => {
...
       });
+      describe('Delete token flow: ', () => {
+        beforeEach(() => {
+          cy.stubRequest({
+            url: 'api/v1/api-keys?grant=scim/manage',
+            fixture: 'api-keys/200.get.scim-token.json',
+            requestAlias: 'scimTokenGet',
+          });
+          cy.visit(PAGE_URL);
+        });
```
nicklemmon, line 105: If this is the whole flow (i.e., one step) probably no need for the `describe()` and the `beforeEach()` - unless this will be fleshed out more later?

src/pages/account/components/SingleSignOnPanel/SCIMTokenSection.js
```diff
@@ -37,6 +48,18 @@ export default function SCIMTokenSection(props) {
...
+      listScimToken();
+      closeModal();
+      showAlert({ type: 'success', message: 'SCIM token deleted' });
+    });
+  };
+  useEffect(() => {
+    if (error) {
+      closeModal();
+    }
+  }, [error, closeModal]);
```
nicklemmon, line 62: Should there be a test for this scenario? Seems like it should be tested somewhere - if not Cypress, then a Jest DOM test. But maybe Cypress?

cypress/tests/integration/accounts/account-settings/accountSettingsPage.js
```diff
@@ -94,6 +94,42 @@ describe('Account Settings Page', () => {
...
+            cy.stubRequest({
+              url: 'api/v1/api-keys?grant=scim/manage',
+              fixture: 'api-keys/200.get.scim-token-notoken.json',
+              requestAlias: 'scimTokenGet',
+            });
+            cy.get('button')
+              .contains('Delete SCIM Token')
+              .click();
+            cy.wait('@scimTokenDelete');
+            cy.wait('@scimTokenGet');
```
nicklemmon, line 126: You can simplify this a bit by using this syntax:
```
cy.wait(['@scimTokenDelete', '@scimTokenGet']);
```

src/actions/scimToken.js
```diff
@@ -28,3 +29,15 @@ export function listScimToken() {
...
+export function deleteScimToken({ id, subaccount = null }) {
+  const headers = setSubaccountHeader(subaccount);
+  return sparkpostApiRequest({
+    type: 'DELETE_SCIM_TOKEN',
+    meta: {
+      method: 'DELETE',
+      url: `/v1/api-keys/${id}`,
+      headers,
+    },
+  });
```
nicklemmon, line 42: I would think some unit tests for this would be helpful?

src/reducers/scimToken.js
```diff
@@ -34,6 +37,32 @@ export default (state = initialState, { type, payload }) => {
     case 'LIST_SCIM_TOKEN_FAIL': {
       return { ...state, scimTokenListLoading: false, error: payload };
     }
+    //DELETE_SCIM_TOKEN
```
nicklemmon, line 40: Same here? I would think some unit tests could be helpful here?

src/pages/account/components/SingleSignOnPanel/SCIMTokenSection.js
```diff
@@ -75,6 +98,35 @@ export default function SCIMTokenSection(props) {
...
+              <ButtonWrapper>
+                <Button
+                  variant="destructive"
+                  onClick={() => {
+                    handleDeleteToken();
+                  }}
+                >
+                  Delete SCIM Token
+                </Button>
+                <Button variant="secondary" onClick={() => closeModal()}>
```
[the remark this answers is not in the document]
nicklemmon: (I can understand why you wouldn't want to use `<DeleteModal/>` - I prefer doing this in a composable way like you have)

cypress/tests/integration/accounts/account-settings/accountSettingsPage.js
```diff
@@ -94,6 +94,42 @@ describe('Account Settings Page', () => {
...
+              requestAlias: 'scimTokenDelete',
+            });
+            cy.stubRequest({
+              url: 'api/v1/api-keys?grant=scim/manage',
+              fixture: 'api-keys/200.get.scim-token-notoken.json',
+              requestAlias: 'scimTokenGet',
+            });
+            cy.get('button')
+              .contains('Delete SCIM Token')
+              .click();
```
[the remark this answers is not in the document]
shrirupa: So I had it this way earlier, but since even the heading if the Modal is same, it errors out saying two elements found.

src/pages/account/components/SingleSignOnPanel/SCIMTokenSection.js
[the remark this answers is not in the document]
shrirupa: Thanks! I updated the variant to monochrome-secondary.

shrirupa:
@nicklemmon Thanks for the feedback. This is ready for another pass whenever you have some time.

cypress/tests/integration/accounts/account-settings/accountSettingsPage.js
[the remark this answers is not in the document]
nicklemmon: fair enough!

src/pages/account/components/SingleSignOnPanel/SCIMTokenSection.js
```diff
@@ -32,11 +43,32 @@ export default function SCIMTokenSection(props) {
...
+          listScimToken();
+        });
+      });
+    } else {
+      generateScimToken().then(() => {
+        openModal({ name: 'Generate SCIM Token' });
+        listScimToken();
+      });
+    }
+  };
```
nicklemmon, line 59: This may not be a legit bug since the API isn't ready I suppose? But it *might* be an edge case even when the API is ready:

1. I clicked on 'Generate SCIM Token'
2. Got a 404
3. 'Delete Token' and 'Generate SCIM Token' buttons now do nothing

nicklemmon: I believe it's the `useEffect()` a little further down. Also, it's firing a whole lot in this scenario!

Screen Shot 2020-06-26 at 8 40 54 AM

src/reducers/tests/__snapshots__/scimToken.test.js.snap
```diff
@@ -0,0 +1,147 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
nicklemmon, line 1: 👍 It's the only time I really like snapshots!

src/pages/account/components/SingleSignOnPanel/SCIMTokenSection.js
shrirupa: Good catch! I will try and fix this. May be reset redux state for scimToken ? 🤷

shrirupa:
@nicklemmon Nice catch! This is ready for another pass.

PR #1689, opened 2020-06-25
FE-1107 - Incorporate UX feedback for the Webhooks feature

nicklemmon:
[FE-1107](https://jira.int.messagesystems.com/browse/FE-1107)

### What Changed
- Incorporate Tooltips in to the webook create form events
- Update grid gap within between checkboxes to reduce whitespace as a result
- Conditionally render the `<EventCheckboxes/>` component rather than using a `show` prop to ensure the `<Stack/>` component handles spacing appropriately

### How To Test
- Go to `/webhooks/create`
- Verify there is no extraneous space at the bottom of the "Events" radio button fieldset
- Select the "Select individual events" radio button and verify that help text no longer renders below checkboxes, instead, tooltips render
- Verify all of the above in both the OG and Hibana themes

### To Do
- [ ] Incorporate feedback


src/pages/webhooks/components/WebhookForm.js
```diff
@@ -24,25 +24,23 @@ import HibanaStyles from './WebhookFormHibana.module.scss';
...
-          key={key}
-          label={display_name}
-          type="checkbox"
-          name={name}
-          helpText={description}
-          component={CheckboxWrapper}
-          disabled={disabled}
-        />
+        <div key={key}>
+          <Tooltip dark content={description}>
```
PatrickSison, line 34: Might be a nitpick or question for UX, but the Tooltip could wrap the larger div above so that the tooltips are aligned.

nicklemmon: Spoke about this on Slack - decided to leave as-is.

PR #1691, opened 2020-06-25
FE-1110 - Incorporate UX feedback for the A/B Testing feature

nicklemmon:
[FE-1110](https://jira.int.messagesystems.com/browse/FE-1110)

### What Changed
- Details button was already fixed in a previous PR!
- Removed use of icons in A/B testing popover
- The "View Template" button now has a right chevron
- The "Remove Variant" button now has a remove icon (a11y requirement missed in the A/C)
- Fixed color of info icon to use blue color

### How To Test
- Sign in to the usual test account
- Verify the above on the following route `/ab-testing/ab_cn_dev_test/2`
- Verify the above on the following route `http://localhost:3100/ab-testing/cutover_test_e/1`

### To Do
- [ ] Incorporate feedback


src/pages/abTesting/components/StatusTag.js
```diff
@@ -13,10 +13,6 @@ const StatusTag = ({ status }) => {
     tagColor = 'yellow';
   }
 
-  if (status === 'completed') {
```
kchu1337, line 16: Isn't it supposed to be green rather than removing it?

src/pages/abTesting/DetailsPage.js
```diff
@@ -6,7 +6,6 @@ import { selectSubaccountIdFromQuery } from 'src/selectors/subaccounts';
 
 import { Loading, DeleteModal, ConfirmationModal } from 'src/components';
 import { PageLink } from 'src/components/links';
-import { Delete, Block } from '@sparkpost/matchbox-icons';
```
kchu1337, line 9: You missed removing the icon from `override with new draft` option in `src/pages/abTesting/ViewMode.js`

nicklemmon: Good catch

PR #1695, opened 2020-06-29
SA-1490: Remove X hours remaining on test from Inbox Placement

kchu1337:
### What Changed
 - Removed the X hours remaining on test from running tests.

### How To Test
 - Go to /inbox-placement. See that a test is still running but not longer displays 'X hours remaining on test'


src/pages/inboxPlacement/helpers/formatScheduleLine.js
```diff
@@ -1,15 +1,11 @@
 import React from 'react';
 import moment from 'moment';
-import { STATUS, DURATION_HOURS } from '../constants/test';
```
nicklemmon, line 3: Looks like you can delete this value from the constants file now.

PR #1699, opened 2020-07-01
FE-1120 - Incorporate UX feedback for the IP Pools feature

nicklemmon:
[FE-1120](https://jira.int.messagesystems.com/browse/FE-1120)

### What Changed
- Incorporated fixes to address UX feedback for the IP pools feature.
- Fixed Cypress test failures (for Hibana only) for the Report Builder feature stemming from updated UI

### How To Test

- [x] Go to `/account/ip-pools` and verify the "Number of IPs Assigned" is right-aligned
- [x] Go to `/account/ip-pools/edit/<id>` and verify the extra space below the "Pool Name" field does not render. Verify on `/account/ip-pools/create` that the full form was not negatively impacted by the changes
- [x] Go to `/account/ip-pools/edit/<id>` and verify the double border on the bottom panel no longer renders

### To Do

- [ ] Incorporate feedback


src/components/collection/TableHeader.js
```diff
@@ -27,6 +27,7 @@ export default class TableHeader extends Component {
           onClick={() => this.handleSorting(sortKey)}
           direction={sortKey === sortColumn && sortDirection}
           label={label}
+          align={align}
```
nicklemmon, line 30: This prop wasn't properly being exposed previously

src/pages/ipPools/EditPage.module.scss
```diff
@@ -0,0 +1,5 @@
+.Panel {
+  > div {
+    border-bottom: 0; // TODO: This should be removed when `borderBottom` is properly exposed on the Hibana <Panel/> component
```
bkemper, line 3: It definitely removes it, but it looks like this is a scenario that we would want it.

Screen Shot 2020-07-01 at 11 30 41 PM

PR #1700, opened 2020-07-02
FE-1115 - For Hibana only, replace Health Score bar charts with report builder-style line charts

nicklemmon:
[FE-1115](https://jira.int.messagesystems.com/browse/FE-1115)

**Note:** Making progress, though after following up with UX they want to do a quick review before moving forward.

### What Changed

- Introduces new, Hibana-only, global `LineChart` component

### To Do
- [ ] Fix health score data formatting inconsistency - could this be handled with a selector? on `HealthScorePage.js`
- [ ] Figure out a way to turn Hibana on in the storybook docs
- [ ] Document component

### DIFF
--- a/src/components/charts/LineChart.js
+++ b/src/components/charts/LineChart.js
@@ -34,7 +34,14 @@ function Cursor({ data, height, points: [{ x, y }], width: chartWidth }) {
   );
 }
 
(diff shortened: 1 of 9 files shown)


bkemper:
Backlog

PR #1706, opened 2020-07-07
SA-1480: UX feedback for reportbuilder

PatrickSison:
### What Changed
 - Fixed typeahead behavior in filter drawers
 - Changed default metrics
 - Changed metrics color order
 - Add empty state for chart if metrics are all removed
 - Changed spacing on filter drawer fields
 - Buttons in drawer are now full width

### How To Test
 - Check Report builder and verify that the visual changes were implemented

### To Do
 - [ ] Change alphabetical order to be vertical in the metrics drawer


src/pages/reportBuilder/ReportBuilder.js
```diff
@@ -105,57 +112,70 @@ export function ReportBuilder({
...
-                <LinksTable />
-              </Tabs.Item>
-            )}
-          </Tabs>
-        </div>
+        {isEmpty ? (
+          <Empty
+            message={
+              <>
+                <span>No Data</span>
```
kchu1337, line 119: Should it be `No Metrics Selected` instead?

src/config/default.js
```diff
@@ -137,7 +137,12 @@ const config = identifier => ({
   },
   splashPage: '/reports/summary',
   summaryChart: {
-    defaultMetrics: ['count_targeted', 'count_rendered', 'count_accepted', 'count_bounce'],
```
kchu1337, line 140: Can you add this change into the AC?

src/pages/reportBuilder/ReportBuilder.js
PatrickSison: This one was based on the mock Aubrey gave for the empty state for chart

PR #1709, opened 2020-07-08
SA-1381: Update routes for new report builder

PatrickSison:
### What Changed
 - Updated nav links and routes for new report builder

### How To Test
 - Check that old reports links now redirect to the report builder
 - Check that the preset reports are set actively

### To Do
- [ ] Address feedback


src/components/appRoutes/index.js
```diff
@@ -1,33 +1,41 @@
...
-  Redirect,
-  Route,
-  Switch
-} from 'react-router-dom';
+import { Redirect, Route, Switch } from 'react-router-dom';
 import { PublicRoute, ProtectedRoute } from 'src/components/auth';
-import routes from 'src/config/routes';
+import defaultRoutes, { hibanaRoutes } from 'src/config/routes';
+import { useHibana } from 'src/context/HibanaContext';
+import qs from 'query-string';
```
bkemper, line 6: I just labeled FE-629 as a "grooming-candidate".  We should use `qs` instead of `query-string`.

src/components/appRoutes/index.js
```diff
@@ -1,33 +1,41 @@
...
+        if (route.redirect) {
+          return (
+            <Route
+              key={route.path}
+              exact
+              path={route.path}
+              render={({ location }) => {
+                const search = qs.stringify({
+                  ...qs.parse(route.search),
+                  ...qs.parse(location.search),
```
bkemper, line 29: These should probably be reverse if `route.search` is required to make it to its destination.

cypress/tests/integration/signals-analytics/analytics-report/analyticsReportRedirect.spec.js
```diff
@@ -0,0 +1,50 @@
...
+
+const oldBouncePage = '/reports/bounce';
+const oldSummaryPage = '/reports/summary';
+const oldRejectionsPage = '/reports/rejections';
+const oldAcceptedPage = '/reports/accepted';
+const oldDelayedPage = '/reports/delayed';
+const oldEngagementPage = '/reports/engagement';
+const signalsAnalyticsPage = '/signals/analytics';
+if (Cypress.env('DEFAULT_TO_HIBANA') === true) {
+  beforeEach(() => {
```
bkemper, line 11: This should be inside the `describe`.

src/helpers/reports.js
```diff
@@ -72,5 +72,5 @@ export function parseSearch(search) {
   }
 
   // filters are used in pages to dispatch updates to Redux store
-  return { options, filters: filtersList };
+  return { options, filters: filtersList, report };
```
bkemper, line 75: Should report be persisted in the url?  Or just used for initial load?

src/pages/reportBuilder/components/ReportOptions.js
```diff
@@ -26,8 +27,16 @@ export function ReportOptions(props) {
     reportLoading,
     searchOptions,
   } = props;
+  const [selectedReport, setReport] = useState(undefined);
 
   const { location, updateRoute } = useRouter();
+  const handleReportChange = report => {
+    if (report) {
+      const { options, filters = [] } = parseSearch(location.search);
```
bkemper, line 35: Wouldn't we want to use `reportOptions` instead of `location.search`?

src/pages/reportBuilder/components/ReportOptions.js
```diff
@@ -40,9 +49,17 @@ export function ReportOptions(props) {
 
   //Initializes the report options with the search
   useEffect(() => {
-    const { options, filters = [] } = parseSearch(location.search);
-
-    refreshReportOptions({ ...options, filters });
+    const { options, filters = [], report } = parseSearch(location.search);
+    if (report) {
+      const reportObj = PRESET_REPORT_CONFIGS.find(({ key }) => key === report);
```
bkemper, line 54: To make this a little more resilient, can we move this out of the if block and use it to determine if we have a valid report or not?

```
//Initializes the report options with the search
useEffect(() => {
  const { options, filters = [], report: reportKey } = parseSearch(location.search);
  const report = PRESET_REPORT_CONFIGS.find(({ key }) => key === reportKey);

  if (report) {
    // ...
  }  else {
    // ...
  }
}, []);  // eslint-disable react-hooks/exhaustive-deps
```

src/pages/reportBuilder/components/ReportOptions.js
```diff
@@ -26,8 +27,16 @@ export function ReportOptions(props) {
     reportLoading,
     searchOptions,
   } = props;
+  const [selectedReport, setReport] = useState(undefined);
 
   const { location, updateRoute } = useRouter();
+  const handleReportChange = report => {
+    if (report) {
```
bkemper, line 34: Is this a valid case?  I have tried and I can't get report to equal undefined.

src/config/routes/signals.js
```diff
@@ -64,6 +64,68 @@ export default [
...
+    category: 'Signals Analytics',
+    subcategory: 'Analytics Report',
+    search: '?report=delayed',
+  },
+  {
+    path: '/reports/engagement',
+    redirect: '/signals/analytics',
+    category: 'Signals Analytics',
+    subcategory: 'Analytics Report',
+    search: '?report=engagement',
```
bkemper, line 115: I don't know how much we will or should use this `search` option, but another way to achieve this is functionality of redirecting with a query string param is with a custom component.

— https://github.com/SparkPost/2web2ui/blob/e1a28a322bcaa8a58df35481f521772793df3f77/src/config/SecretBillingPlanOrBillingSummaryPage.js#L8

src/pages/reportBuilder/components/ReportOptions.js
PatrickSison: I think it was with the typeahead clear button, but that' removed, so that might be why

src/pages/reportBuilder/components/ReportOptions.js
PatrickSison: No, since refreshReportOptions merges in reportOptions when it gets there. I actually need to swap it out though since it shouldn't be locaiton.search either. With current implementation, should be report.query_string

src/helpers/reports.js
PatrickSison: For right now, I don't think so. I don't think it's too much of a change if we need to transition to direct linking these reports, but we can always adjust.

src/pages/reportBuilder/components/ReportOptions.js
PatrickSison: 👍

src/config/routes/signals.js
PatrickSison: I thought about this, but I didn't want to pull logic even more separated away from the routing. I guess the benefits though is I don't have to keep reusing the hibana logic in other components to separate routes. I'll try that out.

bkemper:
@PatrickSison Unit tests are failing.

src/components/appRoutes/index.js
bkemper: I see.  This is using query-string to match the "parseSearch" helper.

src/pages/reportBuilder/components/ReportOptions.js
bkemper: I think we should remove it unless we know why it is there.

PR #1714, opened 2020-07-13
AC-1553 - Starter plan description should not include "First dedicated IP address is free"

shrirupa:
AC-1553 - Starter plan description should not include "First dedicated IP address is free"

### What Changed
 - includesIps is set to true when (currentBundle has 'dedicated_ip' on it) &&  (current plan is not deprecated), otherwise it is set to false
- Dedicated Ip Panel shows up on billing page irrespective of canPurchaseIps.

### How To Test
 - Create an account on a starter plan and confirm that 'First dedicated IP address is free' line doesn't show on /account/billing/plan . 

### To Do
- [ ] Address feedback
- [x] confirm if canPurchaseIps condition needs to be changed too ?

### DIFF
--- a/src/selectors/accessConditionState.js
+++ b/src/selectors/accessConditionState.js
@@ -24,9 +24,7 @@ export const getCurrentAccountPlan = createSelector(
       code: currentPlan.plan,
       includesIp:
         Boolean(currentPlan.status) &&
-        !_.isEmpty(_.find(currentBundle.products, { product: 'dedicated_ip' })) //second condition added for starter plans
-          ? true
-          : false,
+        !_.isEmpty(_.find(currentBundle.products, { product: 'dedicated_ip' })), //second condition added for starter plans,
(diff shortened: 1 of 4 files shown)


src/selectors/accessConditionState.js
```diff
@@ -11,15 +11,22 @@ const currentFreePlans = ['free500-1018', 'free15K-1018', 'free500-0419', 'free5
...
       products: subscription.products,
     };
     return {
       billingId: currentPlan.billing_id,
       code: currentPlan.plan,
-      includesIp: !currentPlan.status ? false : true,
+      includesIp:
+        Boolean(currentPlan.status) &&
+        !_.isEmpty(_.find(currentBundle.products, { product: 'dedicated_ip' })) //second condition added for starter plans
+          ? true
```
gfriedmansp, line 28: Don't think we need the ternary here, should already resolve to a boolean

PR #1720, opened 2020-07-15
FE-1123: Upgrade lodash to latest in 2web2ui

bkemper:
### What Changed
 - Upgrade lodash from v4.17.11 to v4.17.19
 - Since lodash-es is a sub dependency, upgrade redux-form from v7.4.2 to v7.4.3 and react-redux from v5.0.7 to v5.1.2
 - Since lodash.template is a sub dependency, upgraded postcss-preset-env from v6.5.0 to v6.7.0 and uninstalled [workbox-webpack-plugin](https://github.com/facebook/create-react-app/blob/9123aae7a3beaba8b245f73eb0180baa7b2f8411/docusaurus/docs/making-a-progressive-web-app.md)

### How To Test
- [x] Confirm template preview still works with `_.debounce`
- [x] Confirm any list page sorting works correctly with `_.orderBy`
- [x] Confirm SummaryTable on any singles page updates after page props change with `_.isEqual`
- [x] Confirm a suppression can be created (redux-form)
- [x] Confirm dashboard loads (react-redux)
- [ ] Confirm `npm run build` succeeds on CI and does not create a `build/service-worker.js` file (can test locally)

### Change Log Notes

https://github.com/lodash/lodash/wiki/Changelog

**v4.17.19 - v4.17.16**
_It looks like there were some issues publishing .16 which caused the next three attempts, see https://github.com/lodash/lodash/commits/4.17.19_
 - fix(zipObjectDeep): prototype pollution
   - n/a, we don't use this function
 - Sanitize sourceURL so it cannot affect evaled code
   - n/a, we don't use this functionality
 - Fix lodash.isEqual for circular references
   - need to confirm
 - Document matches* shorthands for over* methods
   - n/a
 - Ensure `orderBy` will accept iteratee path arrays
   - need to confirm, this didn't break our TableCollection ordering
 - Short circuit sortedIndexBy methods for empty arrays
   - n/a, we don't use this  

**v4.17.15**
 - Removed unintentionally published project helper scripts
   - n/a

**v4.17.14**
 - Reverted use of the "type" field in package.json of lodash-es
   - n/a

**v4.17.13**
 - Fixed missing dependency of _createRound
   - n/a

**v4.17.12**
 - Ensured _.ceil, _.floor, & _round handle Infinity consistently
   - n/a, we don't use any of these functions
 - Ensured _.debounce clears old timers before starting new ones
   - need to confirm, it was not affected
 - Ensured _.mergeWith always provides a stack to customizer
   - n/a, don't use 
 - Ensured additions to Object.prototype don’t break lodash initialization
   - n/a
 - Ensured map and set clones contain custom properties of source values
   - n/a
 - Fixed prototype pollution for _.defaultsDeep
   - n/a, fix security vulnerability, didn't change behavior
 - Fixed prototype pollution for _.template options
   - n/a, fix security vulnerability, didn't change behavior


bkemper:
> just to confirm, the reason for removing workbox-webpack-plugin was that we weren't using the offline/cache-first behavior, right ?

Correct.  We [unregistered](https://github.com/SparkPost/2web2ui/commit/7372c864c8397f4d577d4ba10e8d15b2952771f1)  it 3 years ago.  We have no plans to enable it.  So, it is better to remove it.

PR #1722, opened 2020-07-15
SA-1517: Precision sticks only when explicitly set or non date update

PatrickSison:
### What Changed
 - Precision only sticks if explicitly set and it's hibana. Otherwise, falls back on preexisting logic for precision

### How To Test
 - Go to non hibana summary report
 - Go to 24 hour range
   - Precision should be in hours
- Go to 7 days
  -Precision should be in days
 - Go to 30 days
  - Precision should be in days
 - Go to 7 days
  -Precision should be in days

### To Do
- [ ] Address feedback
- [ ] Double check if rounding is working correctly


src/actions/reportOptions.js
```diff
@@ -157,6 +157,10 @@ export function refreshReportOptions(payload) {
     }
 
     const rollupPrecision = useMetricsRollup && update.precision;
+    const dateUpdate = Boolean(
```
PatrickSison, line 160: Still weird behavior for updating precision. Will map out logic

src/actions/tests/reportOptions.test.js
```diff
@@ -235,4 +239,129 @@ describe('Action Creator: Report Options', () => {
...
   });
+
+  describe('refreshReportOptions - new flow', () => {
+    let currentFrom;
+    let currentTo;
+    let updatedFrom;
+
+    beforeEach(() => {
+      currentFrom = new Date('2018-02-10');
+      currentTo = new Date('2018-02-11');
```
bkemper, line 250: We shouldn't use `new Date` in tests.  It inherits the timezone of the machine it is run on which is not deterministic if the machine is running in a different timezone then us.

src/actions/reportOptions.js
```diff
@@ -153,31 +153,48 @@ export function refreshReportOptions(payload) {
...
+        const { from, to } = getRelativeDates(update.relativeRange, {
+          precision: updatePrecision,
+        });
+        updateFrom = from;
+        updateTo = to;
+      }
+      const precision =
+        getRollupPrecision({ from: updateFrom, to: updateTo, precision: updatePrecision }) ||
+        getRecommendedRollupPrecision(updateFrom, moment(updateTo));
+      update = { ...update, from: updateFrom, to: updateTo, precision };
```
bkemper, line 175: This code is a little tough to follow... Can't we simplify it a little?

```
// explicit update of range, new dates, try to preserve precision
if (update.relativeRange !== 'custom') {
  //regardless, update the dates
  const { from, to } = getRelativeDates(update.relativeRange, {
    precision: update.precision,
  });
  
  update.from = from;
  update.to = to;
}

update.precision = (
  getRollupPrecision({ from: updateFrom, to: updateTo, precision: updatePrecision }) ||
  getRecommendedRollupPrecision(updateFrom, moment(updateTo));
);
```

PR #1725, opened 2020-07-17
AC-1339 - UI verifies and consumes promo codes via Billing API

shrirupa:
AC-1339 - UI verifies and consumes promo codes via Billing API

### What Changed
 - replaced instances of GET and POST account/subscription/promo-codes/:code -> /billing/subscription/promo-codes/:code
 - updated tests - cypress and unit tests 
 - updated location of fixtures

### How To Test
 - Check onboarding page(/onboarding/plan) and plan change page(/account/billing/plan) for promocode field, if it works correctly.
~**- Note - the API ticket is in review so we might have to wait till it goes out to test this.**~

### To Do
- [ ] Address feedback


shrirupa:
> Looks good. Only endpoint tested was the one to check the promo code but that seems like it's working. Not sure how far out the other endpoint is or if there's other tickets blocking this

@PatrickSison Thanks for the review. found a promo code that we can use for testing in staging, going to wait on AC-1009 to be verified in eu(it is production now) and then may be we can deploy this.

PR #1732, opened 2020-07-20
SA-1525: Missing metric description popover in summary report

bkemper:
### What Changed
 - Removed a z-index override on Modal that was hiding the metric tooltips on summary report

### How To Test
1. Navigate to /reports/summary
1. Click "Select Metrics" button
1. Hover over any of the metric checkboxes


src/components/matchbox/Modal.module.scss
```diff
@@ -1,5 +0,0 @@
-@import '~@sparkpost/matchbox/src/styles/config.scss';
-
-.Modal {
-  z-index: z-index(overlay) + 1; // Needed to help with overlay bug, can be removed when OG theme is removed
```
shrirupa, line 4: Not really sure, which bug is being referred to in the comment here.

shrirupa: Nothing visibly wrong though.

bkemper: I don't know either.  I'll check-in with Nick tomorrow.

PR #1733, opened 2020-07-20
AC-1557 - (UI) App should enforce online support via billing subscription

shrirupa:
AC-1557 - (UI) App should enforce online support via billing subscription

### What Changed
 - Show online support tab on help form if "online_support" product is returned from GET /billing/subscription

### How To Test
 - Check if you see the options(eg Daily Sending Limit Increase) in the menu in "Submit A ticket" tab on Support Form on the basis of whether you have "online_support" in subscription or not.

### To Do
- [ ] Address feedback


src/selectors/accountBillingInfo.js
```diff
@@ -37,6 +40,10 @@ export const hasPhoneSupport = createSelector(
   [selectPhoneSupportOnSubscription],
   hasPhoneSupport => hasPhoneSupport,
 );
+export const hasOnlineSupport = createSelector(
```
bkemper, line 43: This is just duplicate effort.  `selectOnlineSupportOnSubscription` is already a selector and will provide the benefit of memoization.

PR #1735, opened 2020-07-21
SA-1492: Remove region formatting

kchu1337:

### What Changed
 - Removed region formatting for inbox placement.

### How To Test
 - Go to /inbox-placement and click on any test to go to details page.
 - Check that the region data in the breakdown table has not been formatted.


src/pages/inboxPlacement/helpers/formatFilterName.js
```diff
@@ -16,6 +15,4 @@ export const formatFilterName = (filterType, filterName) => {
 
 const formatMailboxProvider = name => name;
```
shrirupa, line 16: this can be removed too.

PR #1736, opened 2020-07-23
AC-1563: (UI) VWO Tracking for Video Test

shrirupa:
AC-1563: (UI) VWO Tracking for Video Test

### What Changed
- `Verifies Sending Domain` is being tracked in VWO
- set a user option on the basis of the request params(video=true or video=false) received on sign up and trigger different goals on that basis.

### How To Test
 - Signup with following variations
      - /join?video=false
      - /join?video=true
      - /join
- Check if different vwo goals get triggered based on these variations
- Check for different roles.
- vwo test - https://app.vwo.com/#/test/ab/24/summary

### To Do
- [ ] Address Feedback

### DIFF
--- a/src/pages/sendingDomains/components/SetupSending.js
+++ b/src/pages/sendingDomains/components/SetupSending.js
@@ -13,8 +13,9 @@ import SetupInstructionPanel from './SetupInstructionPanel';
 // actions
 import { showAlert } from 'src/actions/globalAlert';
 import { verifyDkim } from 'src/actions/sendingDomains';
-import { updateUserUIOptions } from 'src/actions/currentUser';
+import { updateUserUIOptions, get as getCurrentUser } from 'src/actions/currentUser';
 import { hasAutoVerifyEnabledSelector } from 'src/selectors/account';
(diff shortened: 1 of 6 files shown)


src/pages/sendingDomains/components/SetupSending.js
```diff
@@ -30,11 +32,29 @@ export class SetupSending extends Component {
...
+      switch (sawVideoWhileSignUp) {
+        case 'true':
+          trackCustomConversionGoal([204]);
+          this.props.updateUserUIOptions({ sawVideoWhileSignUp: 'vwo goal triggered' }).then(() => {
+            this.props.getCurrentUser();
+          });
+          break;
+        case 'false':
+          trackCustomConversionGoal([205]);
+          this.props.updateUserUIOptions({ sawVideoWhileSignUp: 'vwo goal triggered' }).then(() => {
```
gfriedmansp, line 50: Should this text be the same for each? `'vwo goal triggered'` What's it used for?

shrirupa: not really used for anything, it just there to make sure that the goals get triggered only once.

PR #1740, opened 2020-07-28
FE-1140 - [Dashboard UX Redesign] Set up page plumbing

nicklemmon:
[FE-1140](https://jira.int.messagesystems.com/browse/FE-1140)

### What Changed
- Adds new V2 Dashboard page with a container for handling Redux
- Adds supporting tests

### How To Test
- Enable Hibana and the account UI option `allow_dashboard_v2`
- Visit `/dashboardV2` and verify the page renders
- Disable Hibana and visit `/dashboardV2` and verify the 404 page renders
- Re-enable Hibana and remove the `allow_dashboard_v2` - visit `/dashboardV2` and verify the 404 page renders

### To Do
- [ ] Incorporate feedback


src/pages/dashboard/DashboardPageV2.js
```diff
@@ -0,0 +1,95 @@
+import React from 'react';
+import { Code, ChatBubble, LightbulbOutline } from '@sparkpost/matchbox-icons';
+import { Box, Layout, Page, Panel, ScreenReaderOnly, Stack, Text } from 'src/components/matchbox';
+import { Heading, TranslatableText } from 'src/components/text';
+import useDashboardContext from './hooks/useDashboardContext';
+import styles from './DashboardPageV2.module.scss';
```
nicklemmon, line 6: The PR for FE-1141 actually eliminates this styling, so no worries - we won't be incorporating new CSS modules here.

src/pages/dashboard/DashboardPageV2.js
```diff
@@ -0,0 +1,95 @@
...
+  const { currentUser } = useDashboardContext();
+
+  return (
+    <Page>
+      <ScreenReaderOnly>
+        <Heading as="h1">Dashboard</Heading>
+      </ScreenReaderOnly>
+
+      <Stack>
+        {currentUser?.first_name && (
```
nicklemmon, line 18: 🕺

src/pages/dashboard/index.js
```diff
@@ -1,3 +1,4 @@
 import DashboardPage from './DashboardPage.container';
```
nicklemmon, line 1: Question for the reviewer - should I just move these files in to a new folder instead of co-locating here? `src/pages/dashboardV2`?

src/pages/dashboard/DashboardPageV2.container.js
```diff
@@ -0,0 +1,29 @@
...
+
+const mapDispatchToProps = {};
+
+export default connect(
+  mapStateToProps,
+  mapDispatchToProps,
+)(props => {
+  const [{ isHibanaEnabled }] = useHibana();
+
+  if (!isHibanaEnabled) return <Redirect to="/404" />;
```
bkemper, line 22: `/404`?  Shouldn't we just handle this logic in the route condition?

src/pages/dashboard/DashboardPageV2.container.js
```diff
@@ -0,0 +1,29 @@
...
+
+  return (
+    <DashboardContextProvider value={props}>
```
bkemper, line 25: 👏 Page State!

src/pages/dashboard/DashboardPageV2.js
```diff
@@ -0,0 +1,95 @@
...
+            {currentUser.first_name}!
+          </DashboardHeading>
+        )}
+
+        <Layout>
+          <Layout.Section>
+            <div className={styles.PanelRow}>
+              <Panel>
+                <Panel.Section>
+                  <PanelTitle>
```
bkemper, line 30: Should we just add this to our shared matchbox component?

src/pages/dashboard/DashboardPageV2.js
```diff
@@ -0,0 +1,95 @@
...
+                    <PanelIcon as={Code} />
```
bkemper, line 31: Same here.

src/pages/dashboard/index.js
bkemper: I would vote for making a new folder to avoid shared files and a much easier removal of the old dashboard when ready.

src/components/matchbox/Banner.stories.mdx
```diff
@@ -1,34 +1,23 @@
 import { Meta, Story, Preview, Props } from '@storybook/addon-docs/blocks';
```
nicklemmon, line 1: I was looking to document the new Panel components, but they don't work at the moment as Hibana is not enabled for storybook. We can work on getting that going at a later time though.

While I was there, I noticed some outdated documentation for the `<Banner/>` component so wanted to go ahead and update before I forgot.

src/components/matchbox/tests/Panel.test.js
```diff
@@ -4,6 +4,7 @@ import { shallow } from 'enzyme';
 import { useHibana } from 'src/context/HibanaContext';
 jest.mock('src/context/HibanaContext');
 
+/* eslint-disable jest/expect-expect */
```
nicklemmon, line 7: There were warnings littered throughout due to a series of re-usable functions that were written to test things.

src/reducers/currentUser.js
```diff
@@ -17,7 +17,18 @@ export default (state = initialState, { type, payload, meta }) => {
...
-      return { ...state, ...payload, loading: false, cookie_consent: !!payload.cookie_consent };
+      return {
+        ...state,
+        ...payload,
+        loading: false,
+        cookie_consent: !!payload.cookie_consent,
+        // TODO: Remove when OG theme is removed
+        // Just defaults the app state to having Hibana enabled when using the env variable
+        // Run the app locally with `npm run start-app-hibana`
+        ...(process.env.REACT_APP_DEFAULT_TO_HIBANA
```
nicklemmon, line 28: Rather than manually overriding the context value, I refactored this to override the value in the Redux store. Needed to do it this way so that the condition helpers would work as on the route (`isUserUiOptionSet('isHibanaEnabled')`)

cypress/tests/integration/accounts/billing/billingPage.spec.js
```diff
@@ -500,7 +500,7 @@ describe('Billing Page', () => {
 
       it('closes the modal when clicking "Cancel"', () => {
         cy.findByLabelText('First Name').should('be.visible');
-        cy.findByRole('button', { name: 'Cancel' }).click();
+        cy.findByRole('button', { name: 'Cancel' }).click({ force: true });
```
nicklemmon, line 503: Seeing some flakiness when clicking these buttons, so trying to use the `force` option to make sure it happens.

nicklemmon:
@bkemper - incorporated your feedback. All good stuff!

- I moved everything to a new folder
- Removed the `<Redirect />` and added the condition to the route - and as a result, I had to re-work `npm run start-app-hibana`
- Though UX indicated the larger panel headings and icons were dashboard-specific, we all know that might not last forever, so I made new `<Panel.Headline />` and `<Panel.HeadlineIcon />` global components.

src/components/matchbox/Banner.stories.mdx
bkemper: Can we enable Hibana by default for the docs?

src/components/matchbox/Panel.js
```diff
@@ -41,9 +44,47 @@ const Footer = props => {
...
 Footer.displayName = 'Panel.Footer';
+Headline.displayName = 'Panel.Headline';
+HeadlineIcon.displayName = 'Panel.HeadlineIcon';
+
+Headline.propTypes = {
+  as: PropTypes.oneOf(['h1', 'h2', 'h3', 'h4', 'h5', 'h6']),
+};
+
+HeadlineIcon.propTypes = {
+  as: PropTypes.func,
```
bkemper, line 83: Can we use `React.element`?

src/reducers/currentUser.js
```diff
@@ -17,7 +17,18 @@ export default (state = initialState, { type, payload, meta }) => {
...
+          ? { options: { ...state.options, ui: { ...state.options.ui, isHibanaEnabled: true } } }
```
bkemper, line 29: Hmmm... won't this override options being provided from `payload`?

src/reducers/currentUser.js
bkemper: This is clever.  Wouldn't this work better as `initialState`?

nicklemmon: I think the issue there is we use this flag primarily for Cypress testing - and unless we want to have a fixture that somehow dynamically overrides the value when Hibana is enabled, then when the user is successfully retrieved it would override initial state.

But let me try it! I suppose if the option isn't present in the fixture, then it won't override anything.

src/components/matchbox/Panel.js
nicklemmon: For some reason that didn't work?

Do you mean:
```
HeadlineIcon.propTypes = {
  as: React.element,
}
```

That doesn't seem to do anything? Similarly `as: PropTypes.element` doesn't work - proptypes seems to register the icon component as a function.

src/reducers/currentUser.js
nicklemmon: Possibly - let me try the initial state thing and see how that goes.

src/reducers/currentUser.js
```diff
@@ -17,7 +17,26 @@ export default (state = initialState, { type, payload, meta }) => {
...
+        cookie_consent: !!payload.cookie_consent,
+        // TODO: Remove when OG theme is removed
+        // Just defaults the app state to having Hibana enabled when using the env variable
+        // Run the app locally with `npm run start-app-hibana`
+        options: {
+          ...state.options,
+          ...payload.options,
+          ui: {
+            ...state.options.ui,
+            ...payload.options.ui,
```
bkemper, line 33: I believe there is one case when `options` is not provided — subaccount reporting user.

PR #1723, opened 2020-07-16
FE-1135 - Incorporate Cypress tests in to code coverage calculation

nicklemmon:
[FE-1135](https://jira.int.messagesystems.com/browse/FE-1135)

### What Changed
- Updates Cypress to `4.11.0`
- Adds `@cypress/code-coverage` plugin to help calculate code coverage
- Adds `cypress-test-retries` to help with flaky tests - though this will be incorporated by the Cypress team as a core feature shortly - https://docs.cypress.io/guides/references/roadmap.html#Upcoming-features
- Adds common configuration options for Cypress to `cypress/plugins/index.js`
- Adds `cypress.dev.json` and `cypress.ci.json` configuration files that have some additional configuration for values found within `cypress/plugins/index.js`
- Revamps CircleCI pipeline to use a production build of the app vs. a development build
- Instruments code coverage within production builds when the env `INSTRUMENT_CODE=true` is passed - see `npm run build-ci`
- Uses `nyc` to merge code coverage reports from Jest (4 runners) and Cypress (10 runners) before uploading a single, merged report to Codecov

### How To Test
- For production builds, verify that _no code instrumentation is present_ - in the console, type `window.__coverage__` => should return `undefined`
- Verify tests continue to pass as expected
- Go to https://codecov.io/gh/SparkPost/2web2ui and verify this PR has a code coverage report

### To Do
- [x] Figure out a good way to instrument code for a production build
- [x] Re-incorporate `@cypress/code-coverage`
- [x] Re-incorporate Hibana Cypress tests
- [x] Promote build artifacts to next job so that build step does not need to run each time
- [x] Properly blacklist API requests for pseudo staging environment
- [x] is `nyc` a necessary dependency? - nope!
- [x] Verify that coverage is negatively impacted when Cypress tests are removed
- [x] Verify that coverage is negatively impacted when Jest tests are removed
- [x] Codecov upload is silently failing sometimes, yet exiting with code 0
- [ ] Incorporate feedback

PR #1741, opened 2020-07-28
FE-1143 - [Dashboard UX Redesign] Incorporate Hibana-only footer in the app

nicklemmon:
[FE-1143](https://jira.int.messagesystems.com/browse/FE-1143)

### What Changed
- Adds new Hibana-only `<Footer/>` component
- Adds supporting Cypress tests
- Used `styled-components` for the first time! We need to chat about this this afternoon.
- Updates design tokens package for Hibana only - while working on this feature, I noticed we were on an outdated version
- **Note:** Test failure due to flakiness, addressed in #1723 

### How To Test
- Log in to the app, turn on Hibana, and verify the links render as expected and open in new tabs.
- Resize the browser slowly and verify it remains in alignment with content above it horizontally
- Verify that the design roughly matches the mocks (**Note:** I did have to make the text color darker to meet a11y standards and use our `<ExternalLink/>` component here for consistent behavior across the app)

### To Do
- [x] Incorporate feedback
- [x] Wait on meeting on 7/28/20 before review/merging


src/components/layout/App.js
```diff
@@ -1,16 +1,16 @@
...
-import OGStyles from './Layout.module.scss';
-import hibanaStyles from './LayoutHibana.module.scss';
 import { BannerContext } from 'src/context/GlobalBanner';
 import withContext from 'src/context/withContext';
 import { useHibana } from 'src/context/HibanaContext';
 import useHibanaOverride from 'src/hooks/useHibanaOverride';
 import ScrollToTop from './components/ScrollToTop';
 import FeedbackButton from './components/FeedbackButton';
+import OGStyles from './Layout.module.scss';
+import hibanaStyles from './LayoutHibana.module.scss';
```
nicklemmon, line 13: Just wanted to move these down - seems to be where we tend to import stylesheets.

src/components/layout/components/FeedbackButton.module.scss
```diff
@@ -8,7 +8,7 @@
   box-shadow: box-shadow(300);
 
   @media (min-width: media-query(md)) {
-    bottom: spacing(500);
+    bottom: spacing(600);
```
nicklemmon, line 11: This helps with some vertical alignment issues with the new footer

cypress/tests/integration/accounts/billing/billingPage.spec.js
```diff
@@ -301,23 +301,22 @@ describe('Billing Page', () => {
         });
 
         cy.visit(PAGE_URL);
-        cy.findByText(/Update Payment Information */i).click();
+        cy.findByRole('button', { name: /Update Payment Information */i }).click();
```
bkemper, line 304: Isn't `/Update Payment Information/i` the same as `/Update Payment Information */i`?

Screen Shot 2020-07-30 at 4 21 19 PM

src/components/hibana/Footer.js
```diff
@@ -0,0 +1,54 @@
...
+    <StyledFooter aria-labelledby={headingId}>
+      <ScreenReaderOnly>
+        <Heading as="h2" id={headingId}>
+          Helpful Links
+        </Heading>
+      </ScreenReaderOnly>
+
+      <Box maxWidth="1400" margin="0 auto">
+        <Inline space="600">
+          <StyledLink as={ExternalLink} to={LINKS.SECURITY}>
```
bkemper, line 39: As discussed, remove composition here and use `ExternalLink` in `StyledLink` definition.

src/components/hibana/Footer.js
```diff
@@ -0,0 +1,54 @@
...
+            System Status
+          </StyledLink>
+
+          <StyledLink as={ExternalLink} to={LINKS.STATUS}>
+            Documentation
+          </StyledLink>
+
+          <StyledLink as={ExternalLink} to={LINKS.DOCS}>
```
bkemper, line 47: It looks like the url doesn't match the text.

src/components/layout/App.js
```diff
@@ -35,6 +35,8 @@ export const App = ({ children, bannerOpen }) => {
           </div>
         </main>
 
+        {isHibanaEnabled && <Footer />}
```
bkemper, line 38: Shouldn't we add this for all the layouts?

src/components/text/Heading.js
```diff
@@ -14,14 +14,15 @@ export function OGHeading(props) {
       aria-level={headingLevel}
       className={className}
       data-id={props['data-id']}
```
bkemper, line 16: Can we default the "data-id" to "id"?

cypress/tests/integration/accounts/billing/billingPage.spec.js
nicklemmon: Hmm, let me look in to that - I hadn't authored this pattern previously.

src/components/hibana/Footer.js
nicklemmon: WOW! Sloppy. My fault!

src/components/text/Heading.js
nicklemmon: In my mind, I think it's better to keep these things separate - and in general, we want the `data-id` to be a last resort kind of thing rather than something that renders by default (and then could be used as a crutch potentially when `findByRole` and `findByText` are more effective)

src/components/layout/App.js
nicklemmon: UX wanted it only for authenticated pages - though we could incorporate it to the full page layout?

nicklemmon: Actually, messing with that it doesn't make much sense there either. We're only using that layout for the full screen code editor (which obviously excludes the navigation). Introducing it kind of defeats the purpose of having a 'no distraction' interface.

I vote we move forward with use in the current layout and see if UX has feedback.

nicklemmon:
@bkemper Ready for another pass when you have time.

PR #1743, opened 2020-07-29
SA-1541: Pulls subaccounts on first mount

PatrickSison:
### What Changed
 - Subaccounts pulled down on mounting report builder table
 - Subaccounts no longer show "Deleted" if it can't find the name

### How To Test
 - Go to /signals/analytics (Hibana)
 - Verify subaccounts endpoint is hit
 - Use `Group By` select and choose subaccounts at bottom of table
 - Verify that the table no longer shows `Deleted` for subaccounts with missing names


src/pages/reportBuilder/components/ReportTable.js
```diff
@@ -59,7 +65,7 @@ export const ReportTable = props => {
     }
 
     const subaccount = _.find(typeaheadCache, { type: 'Subaccount', id: subaccountId });
```
kchu1337, line 67: So I guess the reason it worked before was because the component would find the subaccount from the typeahead cache which loads subaccount on mount. Since you are already getting all the subaccounts, you can just directly get subaccounts from state. And you can get rid of lodash.

Something like 
```js
    const subaccount = subaccountList.find(({id}) => id === subaccountId);
    const value = sub ? `${sub?.name} (ID ${sub?.id})` : `Subaccount ${subaccountId}`;
```

PR #1750, opened 2020-07-31
Sa-1528: Keep date range and filters when changing preset reports

kchu1337:
### What Changed
 - Added preset tag to preset reports in config file.
 - Updated report options so that changing from unsaved report -> preset report & preset report -> preset report does not change date range and filters.
 - Added cypress tests for saved reports typeahead.

### How To Test
 - Go to `/signals/analytics` with Hibana enabled.
 - Add a filter ( I usually use campaign - Christmas sale)
 - Change the time range, timezone, and precision.
 - In the report field, select a preset report.
 - Verify that time range, timezone, precision, and filter does not get reset.
 - Select another preset report.
 - Verify again that time range, timezone, precision, and filter does not get reset.


src/pages/reportBuilder/components/ReportOptions.js
```diff
@@ -34,7 +34,21 @@ export function ReportOptions(props) {
     //If user presses escape in typeahead, clears report. Keeps current active report options.
     if (report) {
       const { options, filters = [] } = parseSearch(report.query_string);
-      refreshReportOptions({ ...options, filters });
+      //Keep time range and filters when changing to preset report from another preset report.
+      if (report.type === 'preset' && (!selectedReport || selectedReport.type === 'preset')) {
```
PatrickSison, line 38: Hmmm, I can see this section getting pretty crazy, especially when custom reports start to come down and we consolidate that business logic. Not sure what the best way to approach this is.

PR #1751, opened 2020-08-03
FE-1147 - [Domains UX Redesign] Set up plumbing for the new feature

nicklemmon:
[FE-1147](https://jira.int.messagesystems.com/browse/FE-1147)

### What Changed
- Adds the following routes to the app behind the account UI option `allow_domains_v2`:
  - `/domains`
  - `/domains/create`
  - `/domains/:id`
  - `/domains/:id/verify-bounce`
  - `/domains/:id/verify-sending`
  - `/domains/:id/verify-tracking`
- Adds supporting Cypress tests
- Adds `DomainsLayout` component that will eventually provide access to the Redux store via a context-driven hook

### How To Test
- Enable Hibana and enable the account UI option `allow_domains_v2` and visit the aforementioned routes. Verify pages and titles render according to the mocks.
- Disable Hibana and try to visit the aforementioned routes and verify a redirect to the 404 page occurs
- Enable Hibana and disable the account UI option `allow_domains_V2` and visit the aforementioned routes. Verify you are redirected to the 404 page.

### To Do
- [ ] Incorporate feedback


src/pages/domains/components/DomainsLayout.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
```
shrirupa, line 1: Curious as to why to have a component to redirect instead of just having this route only in hibanaRoutes.

nicklemmon: You're right - let me re-work that!

shrirupa:
@nicklemmon Looks good to me!

PR #1752, opened 2020-08-04
AC-1581 - (UI) New usage report displays messaging, RV & other product usage

shrirupa:
AC-1581 - (UI) New usage report displays messaging, RV & other product usage

### What Changed
 - Added the route to UsagePage.
 - Added the Usage Page as in the mocks 
          - https://sparkpost.invisionapp.com/console/share/SK18I10OTX/479065391
          - https://sparkpost.invisionapp.com/public/share/CA1BJGJS4V
 - This page is only visible to admins.
 - Page should be visible only when hibana is enabled
 - Made a helper global

### How To Test
 - Login as admin and enable hibana and only then you should be able to see this page. Verify no other role can see this page.
 - Check if it matches mocks.
 - Check if the numbers being displayed look correct.

### To Do
- [ ] Address feedback


src/pages/usage/components/FeatureUsageSection.js
```diff
@@ -0,0 +1,55 @@
...
+                  )}
+                ></ProgressBar>
+              </>
+            )}
+            {getProduct('subaccounts') && (
+              <>
+                <Heading looksLike="h6" as="h3">
+                  Subaccounts
+                </Heading>
+                <SubduedText as="p">
```
nicklemmon, line 39: I think this defaults to `<p>`?

src/pages/usage/components/FeatureUsageSection.js
```diff
@@ -0,0 +1,55 @@
...
+                <SubduedText>
+                  {getProduct('dedicated_ip').quantity || 0} of {getProduct('dedicated_ip').limit}
+                </SubduedText>
+                <ProgressBar
+                  completed={getPercent(
+                    getProduct('dedicated_ip').quantity || 0,
+                    getProduct('dedicated_ip').limit,
+                  )}
+                ></ProgressBar>
+              </>
```
nicklemmon, line 32: Each of these could stand to be wrapped in another `<Stack/>` - probably with a little less than the default spacing. Maybe even `<Stack space="100"/>`

Mocks:
Screen Shot 2020-08-05 at 10 12 07 AM

The app:
Screen Shot 2020-08-05 at 10 12 11 AM

src/pages/usage/components/LabelAndKeyPair.js
```diff
@@ -0,0 +1,21 @@
...
+    <Definition>
+      <Definition.Label>
+        <Box color={tokens.color_gray_600} fontSize="200">
+          {label}
+        </Box>
+      </Definition.Label>
+      <Definition.Value>
+        <Box color={tokens.color_white} fontSize="400">
+          {value}
+        </Box>
```
nicklemmon, line 17: I think Report Builder is using this same UI? Could we have a `dark` prop on the `<Definition/>` component or something to help standardize this?

src/pages/usage/components/MessagingUsageSection.js
```diff
@@ -0,0 +1,122 @@
...
+  const isNearingMonthlyLimit =
+    usage && subscription && remaining > 0 && remaining / subscription.plan_volume < 0.1;
+  const isNearingDailyLimit =
+    usage && usage.day.used - usage.day.limit > 0 && usage.day.used / usage.day.limit > 0.9;
+  return (
+    <>
+      <Layout.Section annotated>
+        <Layout.SectionTitle as="h2">Messaging Usage</Layout.SectionTitle>
+        <Stack>
+          <>
```
nicklemmon, line 24: Are you adding fragments here so that <Stack />` will work well? Does this work? I thought it only worked with real DOM 🤔 ? I have just been adding `<div>` when I need to.

src/pages/usage/components/MessagingUsageSection.js
```diff
@@ -0,0 +1,122 @@
...
+            <SubduedText>
+              Each SparkPost account has a daily and monthly quota based on the current plan level
+            </SubduedText>
+            <ExternalLink to={LINKS.DAILY_MONTHLY_QUOTA_LIMIT_DOC}>Quota Limits</ExternalLink>
+          </>
+          {(isNearingMonthlyLimit || isNearingDailyLimit) && (
+            <>
+              <Text fontSize="200" fontWeight="bold">
+                Upgrade
+              </Text>
```
nicklemmon, line 41: Semantically, these two `<Text/>` components should probably be headings:

Maybe `<Heading as="h3" looksLike="h6" />`?

src/pages/usage/components/MessagingUsageSection.js
```diff
@@ -0,0 +1,122 @@
...
+                your plan in <PageLink to="account/billing/plan">Billing</PageLink>.
+              </SubduedText>
+            </>
+          )}
+        </Stack>
+      </Layout.Section>
+
+      <Layout.Section>
+        {usage && (
+          <Box padding="400" backgroundColor={tokens.color_gray_1000}>
```
nicklemmon, line 53: Looks like there is more vertical spacing here in the mocks?

Mocks:
Screen Shot 2020-08-05 at 10 18 25 AM

App:
Screen Shot 2020-08-05 at 10 18 21 AM

src/pages/usage/components/RVUsageSection.js
```diff
@@ -0,0 +1,62 @@
...
+import { tokens } from '@sparkpost/design-tokens-hibana';
+import { LabelAndKeyPair } from './LabelAndKeyPair';
+import _ from 'lodash';
+
+export const RVUsageSection = ({ rvUsage: { recipient_validation } = {} }) => {
+  const volumeUsed = _.get(recipient_validation, 'month.used', 0);
+  return (
+    <>
+      <Layout.Section annotated>
+        <Layout.SectionTitle as="h2">Recipient Validation Usage</Layout.SectionTitle>
```
nicklemmon, line 16: 👍 Nice, good heading structure.

src/pages/usage/tests/UsagePage.test.js
```diff
@@ -0,0 +1,23 @@
...
+  };
+
+  const instance = shallow(<UsagePage {...defaultProps} />);
+
+  it('renders correctly', () => {
+    expect(instance.find('MessagingUsageSection')).toHaveLength(1);
+    expect(instance.find('FeatureUsageSection')).toHaveLength(1);
+    expect(instance.find('RVUsageSection')).toHaveLength(1);
+  });
+});
```
nicklemmon, line 23: I'm a little concerned about the use of shallow/Jest tests for a feature like this - I feel like Cypress would be a better choice. At this point, if these components were ever refactored or renamed, for example, all of the tests would break even if the feature is still working correctly.

You did do a great job not using snapshots and writing tests that aren't grabbing child components (so they are less brittle than a lot of our tests).

What do you think?

src/pages/usage/components/RVUsageSection.js
```diff
@@ -0,0 +1,62 @@
...
+        <SubduedText>
+          Validate email addresses by going to{' '}
+          <PageLink to="/recipient-validation/list">Recipient Validation</PageLink>
+        </SubduedText>
+      </Layout.Section>
+      <Layout.Section>
+        <Box padding="400" backgroundColor={tokens.color_gray_1000}>
```
nicklemmon, line 23: you shouldn't need to import the tokens here to use the value - the `styled-system` props have access already (they should?).

`<Box padding="400" backgroundColor="gray.1000" />`

src/config/routes/index.js
```diff
@@ -730,6 +742,7 @@ const hibanaRoutes = [
   ...templatesRoutes,
   ...inboxPlacementRoutes,
   ...hibanaBlocklistRoutes,
+  ...usageRoute,
```
nicklemmon, line 745: Should this just be in `appRoutes`? It seems like the composition here is mainly for external imports?

src/helpers/tests/recipientValidation.test.js
```diff
@@ -1,4 +1,4 @@
-import totalRVCost, { calculateCost, calculateNewCost } from '../totalRecipientValidationCost';
+import totalRVCost, { calculateCost, calculateNewCost } from '../recipientValidation';
```
nicklemmon, line 1: 👍

src/pages/usage/UsagePage.js
```diff
@@ -0,0 +1,55 @@
...
+    getAccount({ include: 'usage' });
+  }, [getAccount]);
+
+  useEffect(() => {
+    getSubscription();
+  }, [getSubscription]);
+
+  useEffect(() => {
+    getUsage();
+  }, [getUsage]);
```
nicklemmon, line 29: This is interesting - usually I've just been grouping these all in to one `useEffect` though this might make more sense - is this to prevent unnecessary re-requests?

src/pages/usage/UsagePage.js
```diff
@@ -0,0 +1,55 @@
...
+
+  return (
```
nicklemmon, line 31: Is there a loading state?

src/pages/usage/components/FeatureUsageSection.js
```diff
@@ -0,0 +1,55 @@
...
+          <Stack>
+            {getProduct('dedicated_ip') && (
+              <>
+                <Heading looksLike="h6" as="h3">
+                  Dedicated IPs
+                </Heading>
+                <SubduedText>
+                  {getProduct('dedicated_ip').quantity || 0} of {getProduct('dedicated_ip').limit}
+                </SubduedText>
+                <ProgressBar
```
nicklemmon, line 26: Thanks for using the standard component!

src/pages/usage/components/LabelAndKeyPair.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { Definition } from 'src/components/text';
+import { Box } from 'src/components/matchbox';
+import { tokens } from '@sparkpost/design-tokens-hibana';
+
+export const LabelAndKeyPair = ({ label, value }) => {
+  return (
+    <Definition>
+      <Definition.Label>
+        <Box color={tokens.color_gray_600} fontSize="200">
```
nicklemmon, line 10: I mentioned this elsewhere, but no need to import tokens to access the color:
`<Box color="gray.600" />`

src/pages/usage/components/LabelAndKeyPair.js
shrirupa: yes, report builder uses the same component. you are right may be the `Box` can be a part of it with the `dark` prop.

shrirupa: `dark` prop is a good idea. I m confused about difference in fontsizes used here. It looks bigger in UsagePage mocks as compared to Report Builder. Shouldn't they ideally be same ?

src/pages/usage/components/MessagingUsageSection.js
shrirupa: yes, `<>` get Stack to work.

src/pages/usage/components/LabelAndKeyPair.js
nicklemmon: They definitely should be the same - might need to circle back with UX to help standardize which design tokens they are using.

shrirupa: Chelsey confirmed that we should go ahead and have it same way as in the Usage Page mocks (bigger font sizes)

src/pages/usage/components/RVUsageSection.js
shrirupa: you are right. I get confused because of the use of token instances in some other places in the app. 😅

src/pages/usage/UsagePage.js
shrirupa: I don't think, there would be unnecessary re-requests in this case. I just find it cleaner this way, to have separate useEffect for each unrelated "effect".

cypress/tests/integration/inbox-placement/allMessagesPage.spec.js
```diff
@@ -31,20 +31,20 @@ describe('AllMessagesPage', () => {
...
 
-  it('renders aggregates', () => {
-    cy.findByPanelTitle('Deliverability').within(() => {
-      cy.findByAriaLabelledByText('Sent').should('contain', 4);
-      cy.findByAriaLabelledByText('Inbox').should('contain', 2);
-      cy.findByAriaLabelledByText('Spam').should('contain', 1);
-      cy.findByAriaLabelledByText('Missing').should('contain', 1);
-    });
+  // it('renders aggregates', () => {
+  //   cy.findByPanelTitle('Deliverability').within(() => {
```
shrirupa, line 35: this test was failing locally too.

nicklemmon: It's not failing on `master`...only on Hibana and only on this branch...

src/components/text/Definition.js
```diff
@@ -1,28 +1,36 @@
...
     return null;
   });
 };
 
-const Label = ({ children, id }) => {
-  return <Bold id={id}>{children}</Bold>;
+const Label = ({ children, id, dark }) => {
+  return (
+    <Bold id={id}>
+      <Box color={dark && tokens.color_gray_600}>{children}</Box>
```
nicklemmon, line 26: I think you can just use a string here instead of importing tokens? `"gray.600"`?

cypress/tests/integration/inbox-placement/allMessagesPage.spec.js
nicklemmon: I see why! It had to do with updating the `<Definition />` component.

src/components/text/Definition.js
```diff
@@ -1,28 +1,36 @@
...
 
     return null;
   });
 };
 
-const Label = ({ children, id }) => {
-  return <Bold id={id}>{children}</Bold>;
+const Label = ({ children, id, dark }) => {
+  return (
+    <Bold id={id}>
```
nicklemmon, line 25: I found the cause of the inbox placement test failure! It's partially the custom `cy.findByAriaLabelledBy()` function - you didn't actually write a bug here.

_But_ we can get around the test failure by just redoing the `<Label/>` to be:
```
<Box fontWeight="700" color={dark && 'gray.600'} id={id}>
  {children}
</Box>
```

cypress/tests/integration/accounts/usage/usagePage.spec.js
```diff
@@ -0,0 +1,61 @@
...
+    it('renders with a relevant page title ', () => {
+      cy.stubRequest({
+        url: '/api/v1/usage',
+        fixture: 'usage/200.get.json',
+        requestAlias: 'getRVUsage',
+      });
+
+      cy.visit(PAGE_URL);
+      cy.title().should('include', 'Usage');
+      cy.findByRole('heading', { name: 'Usage' }).should('be.visible');
```
nicklemmon, line 32: Nice!

cypress/tests/integration/accounts/usage/usagePage.spec.js
```diff
@@ -0,0 +1,61 @@
...
+    });
+
+    it("doesn't renders recipient validation section when rvUsage is null", () => {
+      cy.stubRequest({
+        url: '/api/v1/usage',
+        fixture: 'usage/200.get.no-results.json',
+        requestAlias: 'getRVUsage',
+      });
+
+      cy.visit(PAGE_URL);
```
nicklemmon, line 55: In order to reduce flakiness, it would be worth adding a `cy.wait` to each of these before making any assertions. Since you already set up a request alias, you can do this and it should make things more stable:

```
cy.wait('@getRVUsage');
```

You can put that right after visiting the page URL but before making any assertions.

src/pages/reportBuilder/ReportBuilder.js
```diff
@@ -95,14 +95,10 @@ export function ReportBuilder({
 
   const renderAggregateMetric = useCallback(({ label, value, unit }) => {
     return (
-      <Definition>
-        <Definition.Label>
-          <Box color={tokens.color_gray_600}>{label}</Box>
-        </Definition.Label>
+      <Definition dark>
```
nicklemmon, line 98: Yaaaay

src/pages/usage/UsagePage.js
```diff
@@ -0,0 +1,58 @@
...
+  );
+}
+
+const mapStateToProps = state => {
+  return {
+    usage: state.account.usage,
+    rvUsage: state.account.rvUsage,
+    subscription: state.account.subscription,
+    billingSubscription: state.billing.subscription,
+    loading: state.account.loading && state.billing.loading && state.billing.usageLoading,
```
nicklemmon, line 54: Should this be `||` instead of `&&` since each request is distinct?

src/pages/usage/components/FeatureUsageSection.js
```diff
@@ -0,0 +1,58 @@
...
+    <>
+      <Layout.Section annotated>
+        <Layout.SectionTitle as="h2">Feature Usage</Layout.SectionTitle>
+        <SubduedText>Feature Limits are specific to account's plan.</SubduedText>
+      </Layout.Section>
+      <Layout.Section>
+        <Box border="400" borderColor="gray.400" padding="400">
+          <Stack>
+            {getProduct('dedicated_ip') && (
+              <Stack space="100">
```
nicklemmon, line 22: 👍 Much better.

src/pages/usage/components/MessagingUsageSection.js
nicklemmon: I had no idea!

src/pages/usage/components/MessagingUsageSection.js
```diff
@@ -0,0 +1,125 @@
...
+  const isNearingDailyLimit =
+    usage && usage.day.used - usage.day.limit > 0 && usage.day.used / usage.day.limit > 0.9;
+  return (
+    <>
+      <Layout.Section annotated>
+        <Layout.SectionTitle as="h2">Messaging Usage</Layout.SectionTitle>
+        <Stack>
+          <>
+            <SubduedText>All message injections count toward messaging usage.</SubduedText>
+            <ExternalLink to={LINKS.DAILY_USAGE}>Messaging Usage Definition</ExternalLink>
```
nicklemmon, line 25: It does look like we might need another `<Stack/>` in here. The mocks show some space between these elements:
Screen Shot 2020-08-07 at 9 45 55 AM

And just a heads up, this link is gray in the mocks - after chatting with UX it looks like there will be scenarios with gray links. As a result, Kevin is working on a new component to handle it in another PR - `SubduedLink` I believe. No need to worry about it now! We can incorporate it later.

src/pages/usage/components/MessagingUsageSection.js
```diff
@@ -0,0 +1,125 @@
...
+            </SubduedText>
+            <ExternalLink to={LINKS.DAILY_MONTHLY_QUOTA_LIMIT_DOC}>Quota Limits</ExternalLink>
+          </>
+          {(isNearingMonthlyLimit || isNearingDailyLimit) && (
+            <>
+              <Heading as="h3" looksLike="h6">
+                Upgrade
+              </Heading>
+              <SubduedText>
+                Your usage indicates that you are nearing you're daily or monthly limits. Upgrade
```
nicklemmon, line 42: Looks like a typo here - should be 'your' instead of 'you're'

src/pages/usage/components/tests/FeatureUsageSection.test.js
```diff
@@ -0,0 +1,57 @@
...
+          price: 0,
+        },
+      ],
+      type: 'active',
+    },
+  };
+  const instance = (props = {}) => shallow(<FeatureUsageSection {...props} />);
+  it('renders the correct title for section', () => {
+    expect(instance(defaultProps)).toHaveTextContent('Feature Usage');
+  });
```
nicklemmon, line 33: For each of your jest tests, you could probably remove this duplication - looks like Cypress is handling this for you.

src/pages/usage/components/LabelAndKeyPair.js
```diff
@@ -0,0 +1,16 @@
...
+import { Box } from 'src/components/matchbox';
+
+export const LabelAndKeyPair = ({ label, value }) => {
+  return (
+    <Definition dark>
+      <Definition.Label>
+        <Box fontSize="200">{label}</Box>
+      </Definition.Label>
+      <Definition.Value>
+        <Box fontSize="400">{value}</Box>
```
nicklemmon, line 12: UX wants different font sizes here than in report builder? That doesn't seem right. 🤔 

If that's the case, let me know and I'll have a chat with them.

shrirupa: UX has decided to change it for Report Builder too.Font sizes in the Usage mock is correct.

cypress/tests/integration/inbox-placement/allMessagesPage.spec.js
shrirupa: I found the issue!

src/components/text/Definition.js
shrirupa: you are right!

src/pages/usage/components/LabelAndKeyPair.js
nicklemmon: Oooh, in that case, let's just set those sizes directly in the `<Definition />` component and then we'll be good.

src/pages/usage/UsagePage.js
shrirupa: yes, I shouldn't code after 6 😄

src/pages/usage/components/MessagingUsageSection.js
shrirupa: yes, Chelsey pointed this out. ok may be I can cut a small ticket replace this once kevin ticket goes out.

src/pages/usage/components/MessagingUsageSection.js
shrirupa: good catch!

PR #1759, opened 2020-08-07
FE-1148 - [Domains UX Redesign] List page - Tabs Setup

nicklemmon:
[FE-1148](https://jira.int.messagesystems.com/browse/FE-1148)

### What Changed
- Added tabs to the new domains page that render the following routes when clicked:
  + `/domains/list/sending`
  + `/domains/list/bounce`
  + `/domains//list/tracking`
- Updates route config for domains details pages

### How To Test
1. Ensure the `allow_domains_v2` is enabled, ensure Hibana is enabled, and visit `/domains`
2. Visit `/domains` and ensure you are redirected to `/domains/list/sending`
3. Visit `/domains/list` and ensure you are redirected to `/domains/list/sending`
4. Click on the tabs and verify the correct content/routes render

### To Do
- [ ] Incorporate feedback


src/config/routes/index.js
```diff
@@ -717,10 +717,25 @@ const appRoutes = [
...
   {
     path: '/domains',
+    redirect: '/domains/list/sending',
+    condition: all(isAccountUiOptionSet('allow_domains_v2'), isUserUiOptionSet('isHibanaEnabled')),
+    category: 'Configuration',
+    subcategory: 'Domains',
+  },
+  {
+    path: '/domains/list',
+    redirect: '/domains/list/sending',
```
shrirupa, line 727: I was thinking may be, instead of multiple redirects, we can have /domain/* in the end of all domains pages to redirect to list/seeding if the url doesn't match anything above. What do you think ?

nicklemmon: That's a good idea - let me try that

nicklemmon:
@shrirupa - Ready for another pass

PR #1763, opened 2020-08-12
FE-1150 - [Domains UX Redesign] List Page - Tables

nicklemmon:
[FE-1150](https://sparkpost.atlassian.net/browse/FE-1150)

### What Changed
- Adds "Sending Domains", "Bounce Domains", and "Tracking Domains" tables to the `/domains/list` route
- Incorporates `@cypress/webpack-preprocessor` so that Cypress tests can leverage import aliases:

```javascript
import { A_VALUE } from 'cypress/constants'; // Importing Cypress-specific modules
import { doSomething } from 'cypress/helpers';
import { doAnotherThing } from 'src/helpers/string'; // Importing application modules
import { ANOTHER_VALUE } from 'src/helpers/constants';
```
- Adds a new `IS_HIBANA_ENABLED` global const available to Cypress tests to simplify checking for whether Hibana is enabled during a CI run
- Replaces the domains `<Layout />` component with a more appropriately named `<Domains.Container />` component for re-use throughout the feature

### How To Test
- Verify all three tables match the mocks attached to the ticket:
  + Visit `/domains/list/sending` to verify the Sending Domains table
  + Visit `/domains/list/bounce` to verify the Bounce Domains table
  + Visit `/domains/list/tracking` to verify the Tracking Domains table
- Ensure Cypress tests adequately verify page states
- **Note:** There are some questions for UX that will need to be followed up on before this feature is released. We can proceed with this ticket and address those items at a later time.

### To Do
- [x] Add "Sending Domains" table
  + [x] Replace flag icon with bookmark icon
  + [x] Incorporate tooltip for bookmark icon
- [x] Add "Sending Domains" supporting tests
- [x] Add "Bounce Domains" table
- [x] Add "Bounce Domains" supporting tests
- [x] Add "Tracking Domains" table
- [x] Add "Tracking Domains" table supporting tests
- [x] Remove use of state/dispatch model for `useDomains()` hook
- [x] Double check mocks for consistency
- [x] Follow-up with UX:
  + [x] Should we build the non-table `<Table />` according to the mocks? This could require some significant re-work of existing table behavior - **Yes, this will be a pattern going forward - could add a new prop to <TableCollection /> to hide table headers visually - incorporate in to 
  + [x] How can we tell if a domain is DMARC compliant?  - **No way to tell, do not implement**
  + [x ] Should we include the unverified domains banners that exist in the feature currently? - **No**
  + [x] How do we tell if a domain is pending verification? - **No way to tell, do not implement **


cypress/constants/index.js
```diff
@@ -0,0 +1 @@
+export const IS_HIBANA_ENABLED = Cypress.env('DEFAULT_TO_HIBANA') === true ? true : false;
```
bkemper, line 1: Do you need the ternary?

nicklemmon:
@bkemper Ready for another pass

PR #1764, opened 2020-08-12
FE-731 - Upgrade "react" and "react-dom" to latest versions

nicklemmon:
[FE-731](https://sparkpost.atlassian.net/browse/FE-731)

### What Changed
- Upgrades `react` and `react-dom` to `16.13.1`
- Removes hack to fix `act()` problems with React Testing Library 
- Removed use of `useTabs()` that was causing an infinite loop to freeze the browser (both locally and in CI)

### How To Test
- Verify the whole app!

### To Do
- [ ] Incorporate feedback


src/pages/domains/ListPage.js
```diff
@@ -30,8 +29,7 @@ export default function DomainsPage() {
       onClick: () => history.push(TRACKING_DOMAINS_URL),
     },
   ];
-  const tabIndex = TABS.findIndex(tab => tab['data-to'] === location.pathname);
-  const [selectedTabIndex, tabs] = useTabs(TABS, tabIndex);
```
nicklemmon, line 34: Caught this in another branch - 😬 somehow Cypress tests were passing, but there was an infinite loop here. With the latest version of React, that was causing the browser to freeze. Didn't cause the freeze with the old version of React.

PR #1765, opened 2020-08-13
FE-1149 - [Domains UX Redesign] List page - Filtering UI

nicklemmon:
[FE-1149](https://sparkpost.atlassian.net/browse/FE-1149)

### What Changed
- Adds filtering UI for use in FE-1154
- Adds new `<LabelSpacer />` component to help with spacing problems in rows of fields
- Stylelint now runs against `.js` files only - the new rules are incompatible with .css files, and prior to a recent change, were not being run against them anyways so this results in no regression
- Resets `overflow` style on the content of each page - Cypress caught a bug in which a popover could flow outside of the bounds of that element, causing it to render as hidden
- Removed use of `useTabs` on the `/domains/list` page that was causing an infinite loop
- Adds some additional tests to check for a bug found in production while I'm here See: https://sparkpost.atlassian.net/browse/FE-1159

### How To Test
1. Sign in to an account with the account option `allow_domains_v2`
1. Go to `/domains` and verify the filtering UI renders on each tab
1. Verify filtering UI matches mocks (note, the `<Select/>` to filter by date should not be present - we will use regular table sorting)

### To Do
- [ ] Incorporate feedback


package.json
```diff
@@ -41,7 +41,7 @@
       "eslint --fix",
       "git add"
     ],
-    "src/**/*.{js,css,scss}": [
+    "src/**/*.{js}": [
```
nicklemmon, line 44: CSS files were not being linted as it is, so just changed this here so only the Stylelint rules would be run. Otherwise, our setup for Stylelint would get much more complex.

src/pages/domains/ListPage.js
```diff
@@ -44,38 +56,86 @@ export default function DomainsPage() {
...
-              />
+                  <Popover
+                    id="domains-status-filter"
+                    trigger={<Button variant="secondary">Domain Status</Button>}
+                  >
+                    <Box padding="300">
+                      <Checkbox id="domains-status-select-all" label="Select All" />
+                    </Box>
+
+                    <Box as="hr" margin="0" />
```
nicklemmon, line 85: We could turn this in to a very simple `<Divider />` component. What do you think?

cypress/tests/integration/hibana/hibanaNavigation.spec.js
```diff
@@ -164,6 +168,48 @@ if (Cypress.env('DEFAULT_TO_HIBANA') === true) {
         cy.get(secondaryNavSelector).should('not.be.visible');
       });
 
+      it('renders the subnav and routes to the sending domains when "Configuration" is active', () => {
```
nicklemmon, line 171: I'm not sure why these tests were missing earlier - unfortunately they didn't catch a bug as a result.

cypress/tests/integration/hibana/hibanaNavigation.spec.js
```diff
@@ -11,9 +11,12 @@ if (Cypress.env('DEFAULT_TO_HIBANA') === true) {
       cy.stubRequest({
         url: '/api/v1/account*',
         fixture: 'account/200.get.has-hibana-theme-controls.json',
+        requestAlias: 'accountReq',
       });
 
       cy.visit('/');
+
+      cy.wait('@accountReq');
```
nicklemmon, line 19: This should help with some flakiness for this test.

cypress/fixtures/account/200.get.has-domains-v2.json
```diff
@@ -1,32 +1,12 @@
...
     "subscription": {
-      "code": "50K-starter-0519",
-      "name": "50K Starter",
-      "plan_volume": 50000,
-      "period": "month",
-      "recurring_charge": 20,
-      "overage": 1,
-      "self_serve": true,
-      "type": "default"
+      "code": "50K-starter-0519"
```
nicklemmon, line 4: Just removed a lot of the extra junk in this fixture that wasn't necessary for its use.

src/pages/domains/ListPage.js
```diff
@@ -31,7 +44,6 @@ export default function DomainsPage() {
     },
   ];
   const tabIndex = TABS.findIndex(tab => tab['data-to'] === location.pathname);
-  const [selectedTabIndex, tabs] = useTabs(TABS, tabIndex);
```
bkemper, line 34: 😄 

This was a good call to remove `useTabs`.  It wasn't built for this use case where state is stored in the url — see how it adds click handler to control state.

https://github.com/sparkpost/2web2ui/blob/FE-1149/src/hooks/useTabs/useTabs.js#L13

PR #1773, opened 2020-08-19
SA-1463 & SA-1464: Replace Blacklist with Blocklist alert

bkemper:
[SA-1463](https://sparkpost.atlassian.net/browse/SA-1463) and [SA-1464](https://sparkpost.atlassian.net/browse/SA-1464)

### What Changed
_Please note that the database work has already been deployed to make all current blacklist alerts ~> blocklist alerts._
 - Replaced "Blacklist" alert with "Blocklist" alert metric

### How To Test
 1. Open /alerts
 1. Click "Create an Alert"
 1. Confirm "Blocklist" is not an "Alert Metric" option (ignore "Blacklist")
 1. Manually enable "blocklist_monitors" account option
 1. Refresh page
 1. Confirm "Blocklist" is an "Alert Metric" option
 1. Confirm you can create a Blocklist alert


src/__testHelpers__/http-responses/default/v1/events/message/documentation/get.json
```diff
@@ -1 +1,2668 @@
...
\ No newline at end of file
+{
```
nicklemmon, line 1: Prettier formatting?

src/pages/alerts/DetailsPage.js
```diff
@@ -1,16 +1,23 @@
...
 import RedirectAndAlert from 'src/components/globalAlert/RedirectAndAlert';
 import { useHibana } from 'src/context/HibanaContext';
 import styles from './DetailsPage.module.scss';
-import _ from 'lodash';
+
+// note, this is one off that shouldn't be reused
+const FloatingButtons = styled.div`
+  float: right;
+  margin-right: ${props => props.theme.space['500']};
+`;
```
nicklemmon, line 20: I believe the real solution here is to use the updated `<Panel/>` API! That being said, this feature will probably get re-built before any update would happen here.

src/pages/alerts/DetailsPage.js
```diff
@@ -58,10 +67,11 @@ export class DetailsPageComponent extends Component {
   };
 
   DuplicateDeleteAction = () => {
-    const { id } = this.props;
+    const { id, isHibanaEnabled } = this.props;
+    const Wrapper = isHibanaEnabled ? FloatingButtons : Fragment;
 
     return (
-      <Box style={{ float: 'right' }} marginRight="500">
```
nicklemmon, line 64: I'm actually wondering if `float` should just be exposed on `<Box/>` at this point. Had another similar scenario somewhat recently. I'm making a note of it to follow up with Jon/Logan.

src/__testHelpers__/http-responses/default/v1/events/message/documentation/get.json
bkemper: Yep, this file was added as a single line blob.

src/pages/alerts/DetailsPage.js
bkemper: I thought about it, but I hope we never use float again. 😄

nicklemmon: There are legit use cases...though typically it's within the flow of a document these days rather than an app.

src/pages/alerts/DetailsPage.js
bkemper: Oh yeah, you are right, but it is a little tangled with the actions either being used in Page's primaryArea or SectionComponent.  I'm going to hold off on this one.

PR #1788, opened 2020-09-01
AC-1617 - (UI) Add segment tracking for account activity

shrirupa:
AC-1617 - (UI) Add segment tracking for account activity

### What Changed
 - Cancelling an account creates an "Account Cancelled" event
 - Creating an account creates an "Account Created" event 
 - Upgrading an account creates an "Account Upgraded" event
 - Downgrading an account creates an "Account Downgraded"
 - Sending an Invitation creates an "Invite Sent" event

Note - only added the attributes that are missing in the identify

### How To Test
 - all these events should be logged https://app.segment.com/sparkpost-daniel-chalef/sources/webui_dev/debugger

### To Do
- [ ] Address feedback
- [ ] Not sure about this - `After each of these events, re-identify the user with the new attributes.` [if this is required].


src/actions/account.js
```diff
@@ -84,13 +85,23 @@ export function emailRequest(data) {
...
+    const state = getState();
+    return dispatch(
+      sparkpostApiRequest({
+        type: 'CANCEL_ACCOUNT',
+        meta: {
+          method: 'POST',
+          url: '/v1/account/cancellation-request',
+          onSuccess: () => {
+            segmentTrack(SEGMENT_EVENTS.ACCOUNT_CANCELLED, {
+              customer_id: state.account.customer_id,
```
gfriedmansp, line 98: Isn't `customer_id` in the `identify` call?

src/helpers/segment.js
```diff
@@ -61,3 +61,9 @@ export const segmentPage = () => {
     window.analytics.page();
   }
 };
+
+export const segmentTrack = (eventType, traits) => {
```
gfriedmansp, line 65: 👍  Thanks for making this

src/pages/join/JoinPage.js
```diff
@@ -75,6 +76,9 @@ export class JoinPage extends Component {
         analytics.trackFormSuccess(ANALYTICS_CREATE_ACCOUNT, {
           form_type: ANALYTICS_CREATE_ACCOUNT,
         });
+        segmentTrack(SEGMENT_EVENTS.ACCOUNT_CREATED, {
+          company_name: values.company_name,
```
gfriedmansp, line 80: Think this should be in the `identify` call as well

src/pages/account/ImmediateCancelAccountPage.js
```diff
@@ -27,28 +28,39 @@ export class ImmediateCancelAccountPage extends Component {
...
-        });
-        history.replace(ACCOUNT_SETTINGS_ROUTE);
-      }, (error) => {
-        this.setState({ loading: LOAD_STATE.FAILURE, error });
-      });
-  }
+      .then(
+        () => {
+          fetchAccount();
+          segmentTrack(SEGMENT_EVENTS.ACCOUNT_CANCELLED);
```
shrirupa, line 34: only added this, rest is all prettier.

PR #1796, opened 2020-09-08
FE-1163 Upgrade Cypress to v5.1.0

gfriedmansp:
### What Changed
 - Removed cypress-plugin-retries because it is included in cypress ^5.0.0
   - Updated tests with the new format of causing retries (as a parameter of the test as opposed to static call in the test)
 - Added `libgbm-dev` as a dependency for our CI
 - Renamed `blacklistHosts` to `blockHosts`
 - Renamed `whitelist` to `preserve`

### How To Test
 - Ensure all cypress tests run and pass in CI


nicklemmon (approved):
Looks good!

PR #1797, opened 2020-09-09
FE-1157 - [Dashboard UX Redesign] Incorporate artwork from Matchbox

nicklemmon:
[FE-1157](https://sparkpost.atlassian.net/browse/FE-1157)

### What Changed

- Adds new `<Picture/>` and `<Picture.Img />` components along with supporting tests and documentation
- Adds `url-loader` to Storybook webpack config to support image loading
- Adds support for `.webp` images via the `url-loader` Webpack plugin
- Adds new UI to the dashboard V2 page consisting of an image and some links

### How To Test

- On an account with the `allow_dashboard_v2` account flag, go to `/dashboardV2` and verify the illustration and links design in the mocks matches the app - **Note:** UX would prefer that the image is not visible on smaller screen sizes. A bug with the Matchbox Column component was discovered as a result and an issue was filed - https://github.com/SparkPost/matchbox/issues/625 - this can be addressed at a later date and time

### To Do
- [ ] Incorporate feedback


cypress/tests/integration/dashboardV2/dashboardPageV2.spec.js
```diff
@@ -24,22 +24,34 @@ describe('Version 2 of the dashboard page', () => {
...
-      cy.verifyLink({
-        content: 'Generate an API Key',
-        href: '/account/api-keys/create',
-      });
-
+      cy.findAllByRole('link', { name: 'Generate an API Key' }).should(
+        'have.attr',
+        'href',
+        '/account/api-keys/create',
+      );
```
nicklemmon, line 33: For now, there are multiple links with the same content, so had to change this to `findByAll`

src/pages/dashboardV2/DashboardPageV2.js
```diff
@@ -48,70 +51,102 @@ export default function DashboardPageV2() {
...
                       <SupportTicketLink>Contact our Support Team</SupportTicketLink>
-                    </Panel.LEGACY.Section>
+                    </Panel.Section>
                   </Dashboard.Panel>
                 </Column>
               </Columns>
 
               <Dashboard.Panel>
-                <Panel.LEGACY.Section>
+                <Panel.Section>
```
nicklemmon, line 82: Just went ahead and removed the use of `LEGACY` components here since they are all Hibana-only pages.

src/pages/dashboardV2/DashboardPageV2.js
```diff
@@ -48,70 +51,102 @@ export default function DashboardPageV2() {
...
                   <Panel.Headline>
                     <Panel.HeadlineIcon as={LightbulbOutline} />
 
                     <TranslatableText>Helpful Shortcuts</TranslatableText>
                   </Panel.Headline>
 
                   <Columns collapseBelow="md">
-                    <Dashboard.Shortcut>
+                    <Dashboard.Tip>
```
nicklemmon, line 90: I thought this was a better name? Also, wanted to use it for the new UI 🤷

src/components/picture/Picture.js
```diff
@@ -0,0 +1,57 @@
...
+
+const StyledImg = styled.img`
+  width: 100%; /* Contains <img /> within parent */
+  height: 100%;
+  object-fit: cover;
+  mix-blend-mode: ${props =>
+    props.seeThrough === true ? 'multiply' : 'unset'}; /* applies a sort of faux transparency */
+`;
+
+function Picture({ children, role, className }) {
```
kchu1337, line 24: Should we allow className as a prop?

src/pages/dashboardV2/DashboardPageV2.js
```diff
@@ -10,6 +12,7 @@ import {
   Text,
 } from 'src/components/matchbox';
 import { Heading, TranslatableText } from 'src/components/text';
+import Picture from 'src/components/picture';
 import { ExternalLink, PageLink, SupportTicketLink } from 'src/components/links';
 import { Loading } from 'src/components/loading';
```
kchu1337, line 17: Could add Picture to components index and import both {Picture, Loading} from src/components.

src/pages/dashboardV2/DashboardPageV2.js
```diff
@@ -48,70 +51,102 @@ export default function DashboardPageV2() {
...
+              </Dashboard.Panel>
+
+              <Dashboard.Panel>
+                <ScreenReaderOnly>
+                  <Heading as="h3">Next Steps</Heading>
+                </ScreenReaderOnly>
+
+                <Columns space="0">
+                  <Column width={0.55}>
+                    <Box backgroundColor="gray.100" height="100%" borderRight="400">
```
kchu1337, line 127: Could add a `collapseBelow` prop but would then need to add a bottomBorder and ideally set up a media query that has the right border when size > md and bottom border when size < md.

kchu1337: Also, not sure if the box background color is working.

kchu1337 (changes requested):
Really minor stuff, mostly stylistic. If you don't want to address anything, just request another review and I can approve it.

src/pages/dashboardV2/DashboardPageV2.js
nicklemmon: > Could add a collapseBelow prop but would then need to add a bottomBorder and ideally set up a media query that has the right border when size > md and bottom border when size < md.

This would work, but UX actually wants it to not be visible at all on smaller screen sizes. I could handle that with `<Box/>` or `styled-components` but the `Column` wasn't supporting `className` out of the box. I cut a ticket and pushed a fix with Matchbox but just don't have it available yet 🤷‍♂️

src/components/picture/Picture.js
nicklemmon: This is required to make `styled-components` work! So if I wanted to do this, it would work:
```
const StyledPicture = styled(Picture)``
```

nicklemmon:
@kchu1337 Addressed your feedback - take another look when you get a minute!

kchu1337 (approved):

kchu1337 (approved):
Seems to work. Artwork is gone when screen size is smaller.

kchu1337 (approved):
LG again

PR #1799, opened 2020-09-09
FE-1166: Fix date selector for raw metrics with Report Builder

kchu1337:
### What Changed
 - Simplified a lot of stuff.
 - Raw metrics always shows your local timezone.
 - Raw metrics always uses the recommended precision

### How To Test
- Without metrics rollup on both Hibana and OG:
     - Select 7 day range and submit. Then select 24 hours. The precision should be 24 hours.
     - In the date picker itself, select a time range < 24 hours. The text boxes for time should be editable because precision is < day. Then change to ~3 day time range. The text boxes should not be editable because time range = day.
     - Verify that the timezone is always your local timezone when changing date ranges.
- With metrics rollup on both Hibana and OG:
     - Select 7 day range and submit. Then select 24 hours. The precision should be days. Then select hour. The precision should change to 5 minutes.
     - In the date picker itself, select a time range < 24 hours. The text boxes for time should be editable because precision is < day. Then change to ~10 day time range. The text boxes should not be editable because time range = day.
     - Verify that changing precision and timezone still works.
 

### To Do
- [ ] There are still some wonky stuff. Discuss with UX. One solution would to always show recommended precision even with rollup.


src/actions/reportOptions.js
```diff
@@ -136,9 +135,10 @@ export function refreshReportOptions(payload) {
     const { reportOptions } = getState();
     let update = { ...reportOptions, ...payload };
     const { useMetricsRollup } = selectFeatureFlaggedMetrics(getState());
+    const getPrecision = useMetricsRollup ? getRollupPrecision : getRawPrecision;
     const isHibanaEnabled = isUserUiOptionSet('isHibanaEnabled')(getState());
 
-    if (!update.timezone) {
+    if (!update.timezone || !useMetricsRollup) {
```
kchu1337, line 141: This is because raw metrics should always use local timezone.

src/actions/reportOptions.js
```diff
@@ -156,45 +156,24 @@ export function refreshReportOptions(payload) {
       update.relativeRange = isHibanaEnabled ? '7days' : 'day'; //Default relative range
     }
 
-    if (isHibanaEnabled) {
```
kchu1337, line 159: Hibana enabled vs disabled has not effect. This was a carryover from when we thought report builder will always have metrics-rollup. It will be bundled in the future but we should just use the   metrics rollup flag instead

src/actions/tests/reportOptions.test.js
```diff
@@ -190,178 +186,26 @@ describe('Action Creator: Report Options', () => {
...
-          precision: '15min',
-        }),
-      );
+      expect(getPrecision).not.toHaveBeenCalled();
+      expect(getRollupPrecision).toHaveBeenCalled();
       expect(dispatchMock).toHaveBeenCalledTimes(1);
     });
-  });
 
-  describe('refreshReportOptions - new flow', () => {
```
kchu1337, line 244: Deleted all of the hibana/report builder stuff

src/components/datePicker/ManualEntryForm.js
```diff
@@ -27,6 +27,8 @@ export class ManualEntryFormClassComponent extends Component {
     fromTime: '',
   };
 
+  getPrecision = this.props.defaultPrecision ? getRollupPrecision : getRawPrecision;
```
kchu1337, line 30: Meant to add this as a comment but the only time defaultPrecision has a value would be when rollup is enabled. I could pass through the flag but this acts as a proxy. But will be confusing in the future?

src/helpers/metrics.js
```diff
@@ -90,13 +90,8 @@ export function getDelimiter(filters = []) {
  * and returns the closest precision value
  *
  */
-export function getPrecision(from, to = moment(), precision) {
-  const precisionOptions = getPrecisionOptions(moment(from), moment(to)).map(({ value }) => value);
-  if (precisionOptions.includes(precision)) {
-    return precision;
-  }
```
kchu1337, line 97: I think this was the main problem of the bug ticket. Instead of always returning recommended precision, it returns the current precision if it's an available option.

src/helpers/metrics.js
```diff
@@ -107,7 +102,7 @@ export function getPrecision(from, to = moment(), precision) {
  */
 export function getRollupPrecision({ from, to = moment(), precision }) {
   if (!precision) {
-    return undefined;
+    return getRecommendedRollupPrecision(from, to);
```
kchu1337, line 105: If no precision is given, then just give the recommended precision for the date range.

src/helpers/metrics.js
```diff
@@ -90,13 +90,8 @@ export function getDelimiter(filters = []) {
...
  *
  */
-export function getPrecision(from, to = moment(), precision) {
-  const precisionOptions = getPrecisionOptions(moment(from), moment(to)).map(({ value }) => value);
-  if (precisionOptions.includes(precision)) {
-    return precision;
-  }
-
-  const diff = to.diff(from, 'minutes');
+export function getPrecision({ from, to = moment() }) {
```
kchu1337, line 93: Made this in an object so it would be the same as getRollupPrecision. I think there might be times where we want to pass `from` and `precision` but not `to`, so made the parameters an object.

src/pages/reportBuilder/components/DateTimeSection/DateTimeSection.js
```diff
@@ -35,9 +35,12 @@ export const DateTimeSection = ({
   );
 
   const isForcedUTC =
-    reportOptions.precision && isForcedUTCRollupPrecision(reportOptions.precision);
+    useMetricsRollup &&
```
kchu1337, line 38: raw metrics should use local timezone.

package.json
```diff
@@ -23,11 +23,11 @@
     "test-e2e": "npm run test-cy-env && cypress open --config integrationFolder=cypress/tests/e2e",
     "test-e2e-headless": "cypress run --config integrationFolder=cypress/tests/e2e",
     "test-e2e-ci": "npm run start-app & wait-on http://localhost:3100 && npm run test-e2e-headless",
-    "test-integration": "npm run test-cy-env && cypress open --config-file cypress.dev.json",
-    "test-integration-hibana": "npm run test-cy-env && cypress open --config-file cypress.dev.json --env DEFAULT_TO_HIBANA=true",
-    "test-integration-headless": "npm run test-cy-env && cypress run --config-file cypress.dev.json",
-    "test-integration-ci": "npm run serve-app & wait-on http://localhost:5000 && cypress run --config-file cypress.ci.json",
-    "test-integration-ci-hibana": "npm run serve-app & wait-on http://localhost:5000 && cypress run --config-file cypress.ci.json --env DEFAULT_TO_HIBANA=true",
+    "test-integration": "npm run test-cy-env && TZ=America/New_York cypress open --config-file cypress.dev.json",
```
gfriedmansp, line 26: Do we need this here _and_ in the config.yml?

src/components/datePicker/ManualEntryForm.js
gfriedmansp: I think the comment in the code would suffice

src/helpers/metrics.js
gfriedmansp: Good idea 👍

gfriedmansp (approved):
Looks good to me! Let me know when you'd like to get this in.

PR #1806, opened 2020-09-14
AC-1668 Add Hibana Toggled On and Hibana Toggled Off to Segment

gfriedmansp:
### What Changed
 - In addition to logging to pendo, log the hibana toggled event to Segment

### How To Test
 - Toggle Hibana on and off via the profile page
 - Ensure events are being logged via the debugger: https://app.segment.com/sparkpost-daniel-chalef/sources/webui_dev/debugger
 - Ensure the event has an attribute `ux`, which we will be able to filter by in mixpanel, etc.


shrirupa (approved):
Looks good to me!

PR #1807, opened 2020-09-15
SA-1606 - Subscription limit on reports

PatrickSison:
### What Changed
 -Tooltip shows if limit for reports reached
 - Button for creating new report disabled if limit for reports reached

### How To Test
 - Go to report builder
 - Check appteam
 - Create new report button should be working
 - Create new account
 - Enable user ui option `allow_saved_reports`
 - Button should be disabled
 - Tooltip should exist

### To Do
- [x] Check usage (since list isn't guaranteed to be the full list)

### Notes
 - Rebased off of SA-1588. Will mark appropriate file changes


cypress/fixtures/billing/subscription/200.get.reports-limited.json
```diff
@@ -0,0 +1,44 @@
+{
```
PatrickSison, line 1: File added for subscription limit

src/pages/reportBuilder/ReportBuilder.js
```diff
@@ -64,6 +69,19 @@ export function ReportBuilder({
     }
   }, [refreshSummaryReport, reportOptions, isEmpty]);
 
+  useEffect(() => {
```
PatrickSison, line 72: Retrieves subscription

src/pages/reportBuilder/ReportBuilder.js
```diff
@@ -109,27 +127,35 @@ export function ReportBuilder({
...
     <Page
       title="Analytics Report"
-      primaryAction={
-        isSavedReportsEnabled
-          ? {
-              content: 'Save New Report',
-              onClick: () => setShowSaveNewReportModal(true),
-            }
-          : undefined
+      primaryArea={
```
PatrickSison, line 130: Tooltip and button disabling

cypress/fixtures/billing/subscription/200.get.json
```diff
@@ -32,6 +32,11 @@
         "plan": "ip-0519",
         "product": "dedicated_ip",
         "limit": 4
+      },
```
PatrickSison, line 35: Added higher report limit for default subscription

src/pages/reportBuilder/ReportBuilder.js
```diff
@@ -64,6 +69,19 @@ export function ReportBuilder({
     }
   }, [refreshSummaryReport, reportOptions, isEmpty]);
 
+  useEffect(() => {
+    getSubscription();
+  }, [getSubscription]);
+
+  const reportsProduct = subscription?.products?.find(({ product }) => product === 'reports');
```
kchu1337, line 76: I think you can move this within the memoization.

kchu1337 (changes requested):
Can you also rebase from master so that only your 1606 commit shows up?

kchu1337 (commented):
LGTM, can you just rebase onto master? It might not be necessary since we are squashing commits, but the PR would be cleaner.

kchu1337 (approved):

PR #1809, opened 2020-09-17
FE-1167: Secure webui cookies

bkemper:
[FE-1167](https://sparkpost.atlassian.net/browse/FE-1167)

### What Changed
 - Set [secure](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#SameSite_attribute#Restrict_access_to_cookies) to "true" and [sameSite](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#SameSite_attribute) to "strict" attributes for `auth` and `website_auth` cookies
 - Had to modify the test configuration to not require secure cookies and fix blocklist spec

### How To Test

_**You will need to setup [charger](https://github.com/SparkPost/charger) to verify this locally.**_

 1. Read https://github.com/js-cookie/js-cookie and https://web.dev/samesite-cookies-explained
 1. Visit http://app.sparkpost.test/auth
 1. Open "Application" tab in devtools 
 1. Clear cookies
 1. Login
 1. Confirm both cookies are created with secure and sameSite attributes are set
 1. Open "http://www.sparkpost.test"
 1. Open "Application" tab in devtools 
 1. Confirm `website_auth` cookie is set


bkemper:
@gfriedmansp Shoot, I ran the integrations tests locally and they passed.  I need to take another look at this.

cypress.ci.json
```diff
@@ -1,4 +1,4 @@
 {
-  "baseUrl": "http://app-staging.sparkpost.com:5000",
+  "baseUrl": "http://localhost:5000",
```
bkemper, line 2: I didn't want to do this because I know it breaks our E2E tests, but they aren't running right now anyways.  So, we can revisit E2E setup later and hopefully, after our runtime configuration is simplified.

cypress/plugins/index.js
```diff
@@ -52,16 +52,19 @@ module.exports = (on, config) => {
...
+    'localhost',
     '*pendo.io',
-    'api.segment.io',
-    'cdn.segment.com',
-    '*siftscience.com',
-    '*googletagmanager.com',
+    '*sparkpost.com',
+    '*sparkpost.test',
+    '*segment.com',
+    '*segment.io',
```
bkemper, line 64: This list is a little silly to maintain until we start running true end to end tests with actual requests.  However, I updated it and organized it.

package.json
```diff
@@ -218,6 +218,7 @@
     "redux-mock-store": "^1.3.0",
     "sass-loader": "7.1.0",
     "semver": "^5.5.0",
+    "serve": "^11.3.2",
```
bkemper, line 221: This should be a defined dependency since our CI relies on it.

cypress/tests/integration/blocklists/incidents.spec.js
```diff
@@ -1,279 +1,279 @@
-import moment from 'moment';
-const utcFormatMatcher = /\d+-\d+-\d+T/g;
-const PAGE_BASE_URL = '/signals/blocklist/incidents';
-const dateFormat = 'MMM D YYYY, h:mma';
-
-let nov = moment('2019-11-20T17:14:57.899Z');
-let novTime = nov.format(dateFormat);
+import { IS_HIBANA_ENABLED } from 'cypress/constants';
```
bkemper, line 1: Why did I touch this spec?

The "changes the date range and makes the correct http call" test was unexpectedly failing.  I first broken them up in to separate test to debug.   Then I couldn't figure out why the dates were off, typically by one day.  I chalked it up to a timezone thing and instead of debugging more, I decided to remove the use of `moment` — comparing today on a server (with Moment) vs browser (whatever our app is using) can be problematic.  My solution was to freeze time.  The benefits are less complexity in the spec and alignment with fixtures.  The problem I had to work around were running time off for debounce and animations.

bkemper:
🎉

nicklemmon (commented):
@bkemper Had a question about updates to the `blockHosts` configuration - overall things look good and my local dev environment is working as expected.

nicklemmon (approved):
@bkemper This looks good, though I'm curious why `localHost` was added to Cypress' block list. Mind explaining?

cypress/plugins/index.js
```diff
@@ -48,16 +48,18 @@ module.exports = (on, config) => {
 
   // Configuration options that do not change between environments
   config.blockHosts = [
-    'api.sparkpost.test',
-    'api-staging.sparkpost.com',
+    '*algolia.net',
     '*google-analytics.com',
+    '*googletagmanager.com',
     '*.storage.googleapis.com',
+    'localhost',
```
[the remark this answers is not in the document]
bkemper: We really shouldn't need it, so I was just making it explicit.

PR #1812, opened 2020-09-18
SD-2124: Selecting overflow pool of "None" does not always trigger the correct PUT request

bkemper:
[SD-2124](https://sparkpost.atlassian.net/browse/SD-2124)

### What Changed
 - The bugs was caused by combination of redux-form and setting initialValues.  I tried using `defaultValue` which did not work.  The fix is to explicitly set `auto_warmup_overflow_pool ` in the form submit handler.

### How To Test

Use the steps to reproduce in the bug ticket and let me know if you have issues.  This needs to be tested on both OG and Hibana.  There are no other authorization considerations because the field is present for all pools except the "default" pool which can't be updated anyways.


nicklemmon (approved):
@bkemper Looks good! Works as expected. Was able to reproduce the bug on `master`, and reproduce the expected behavior on this branch.

PR #1820, opened 2020-09-24
SA-1665 - (UI) Comparators - Filters update

nicklemmon:
[SA-1665](https://sparkpost.atlassian.net/browse/SA-1665)

### What Changed

- Adds new `<AddFiltersSectionV2 />` component that renders behind the `"allow_report_filters_v2"` account UI flag - **should this be a user UI flag?**
- Adds new `<Uppercase />` and `<Emphasized />` re-usable text components
- Adds new `<RadioButtonGroup />` and `<RadioButtonGroup.Button />` components
- Adds new `useMultiEntry` hook - **should this be named** `useComboBoxTextField`**?**

### How To Test

- Go to `/signals/analytics` with the `"allow_report_filters_v2"` account flag set to `true`
- Click on "Add Filters"
- Verify the form behaves according to the mocks
- Verify submitted form data is reformatted in to the expected API structure

### To Do

- [x] Update handling of removing a filter according to: https://msys.slack.com/archives/G0195L0U5LH/p1601921523000600?thread_ts=1601664409.004000&cid=G0195L0U5LH
- [x] Focus management isn't working quite right - *could reviewers help me troubleshoot a bit?*
- [x] Add Cypress tests
- [x] Handle form submission/query param stuff on submit - *but this in Patrick's ticket handling JSON parsing?*
- [x] The "AND / OR" radio buttons should always render if _any_ filter has values within a grouping
- [x] Verify subaccounts search based on criteria from message w/ Greg:

> If a subaccount makes a request using the query_string parameter, the query will always be restricted to data for that subaccount. e.g. if subaccount 10  makes a request for subaccounts eq 12, 14, 16 , they will not in fact get data for subaccounts 12, 14, 16, only subaccount 10

- [x] Ensure subaccounts can only be compared by `"eq"` and `"not.eq"`
- [x] Fix re-mapping function on form submission
- [x] Overflow hidden vertically when the form gets too long
- [x] Fix initial selection of the filter "Type" field - placeholder is rendering funny
- [x] Work with API team on restructuring endpoint
- [x] Adding a grouping causes a rendering error
- [x] Conditionally rendering the "AND +" button between groupings
- [x] Add "AND/OR" switch and conditional "AND +" and "OR +" buttons
- [x] Break up `<AddFiltersSectionV2 />` in to configuration variables and components outside of the main file
- [x] Get initial state rendering
- [x] Add helpers to restructure data that is more React-friendly
- [x] Prepare form fields
- [x] Prepare layout according to mocks
- [x] Add `useMultiEntry` hook


src/components/multiEmailField/useMultiEmailField.js
```diff
@@ -7,11 +7,13 @@ const useMultiEmailField = (value = '', emailList = [], error = '') => {
   const [multiEmailError, setMultiEmailError] = useState(error);
   const [multiEmailList, setMultiEmailList] = useState(emailList);
```
nicklemmon, line 9: I was just messing around in here and saved the file - just Prettier changes.

src/hooks/index.js
```diff
@@ -1,4 +1,5 @@
 export { default as useABTest } from './useABTest';
+export { default as useMultiEntry } from './useMultiEntry';
```
nicklemmon, line 2: Should this have a different name? `useComboBoxTextField` or something?

src/pages/reportBuilder/components/AddFiltersSectionV2/AddFiltersSectionV2.js
```diff
@@ -0,0 +1,319 @@
...
+                                      onBlur={props.handleBlur}
+                                      onKeyDown={props.handleKeyDown}
+                                      onChange={props.handleChange}
+                                      value={props.value}
+                                      selectedItems={props.valueList}
+                                      itemToString={item => (item ? item : '')}
+                                    />
+                                  );
+                                }}
+                              />
```
nicklemmon, line 206: Ideally wouldn't have to do this, but I needed to use `useMultiEntry` with this component - and I can't do that inside a `.map` callback. Is there some better way to do this?

src/pages/reportBuilder/components/ReportOptions.js
```diff
@@ -147,21 +150,21 @@ export function ReportOptions(props) {
 
   return (
     <div data-id="report-options">
-      <Panel.LEGACY.Section>
+      <Panel.Section>
```
nicklemmon, line 153: No need to use the `LEGACY` components in a Hibana-only feature.

src/pages/reportBuilder/components/TypeaheadV2.js
```diff
@@ -0,0 +1,84 @@
+import React, { useCallback, useState, useMemo } from 'react';
```
nicklemmon, line 1: Mostly just a copy of the original `Typeahead` with some tweaks - there may be a better way to do this? But if the other `Typeahead` is ultimately going away, this is a little simpler.

src/pages/reportBuilder/components/TypeaheadV2.js
```diff
@@ -0,0 +1,84 @@
...
+import { ComboBoxTypeahead } from 'src/components/typeahead/AsyncComboBoxTypeahead';
+import { METRICS_API_LIMIT } from 'src/constants';
+import { useUniqueId } from 'src/hooks';
+import sortMatch from 'src/helpers/sortMatch';
+
+function getItemToStr({ filterType, item }) {
```
nicklemmon, line 7: The `itemToString` function has to be special for subaccount searches - is there a better place to put this?

cypress/tests/integration/signals-analytics/analytics-report/analyticsReportBreakdownTable.spec.js
```diff
@@ -35,7 +35,7 @@ if (IS_HIBANA_ENABLED) {
       cy.stubRequest({
         url: '/api/v1/subaccounts',
         fixture: 'subaccounts/200.get.json',
-        requestAliasList: 'getSubaccountList',
```
nicklemmon, line 38: Not sure why this was being passed in? Must have been a typo!

src/pages/reportBuilder/components/ReportOptions.js
```diff
@@ -197,30 +200,37 @@ export function ReportOptions(props) {
                 <MetricsDrawer selectedMetrics={processedMetrics} handleSubmit={handleSubmit} />
               </Tabs.Item>
               <Tabs.Item>
-                <AddFiltersSection handleSubmit={handleSubmit} reportOptions={reportOptions} />
+                <AccessControl condition={not(isAccountUiOptionSet('allow_report_filters_v2'))}>
+                  <AddFiltersSection handleSubmit={handleSubmit} reportOptions={reportOptions} />
+                </AccessControl>
+
+                <AccessControl condition={isAccountUiOptionSet('allow_report_filters_v2')}>
+                  <FiltersForm handleSubmit={handleSubmit} reportOptions={reportOptions} />
```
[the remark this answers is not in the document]
nicklemmon: Yeah we will need an effect somewhere to grab what's in the URL and render it out to the form.

And yeah, I think all we'll get is the IDs for subaccounts - you're saying we wouldn't be able to render the names if we do it that way without making another request?

src/pages/reportBuilder/components/FiltersForm/useFiltersForm.js
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    case 'CLEAR_FILTERS': {
+      return {
+        ...state,
+        groupings: remapGroupings(initialGroupingsState),
+      };
+    }
+
+    default:
```
PatrickSison, line 134: This is a good pattern I think. Never thought about it before, but definitely prevents catching on incorrect actions

PatrickSison (commented):
Everything else generally looks good. Longer term issue would be making sure we understand what our query param format is VS what's available for Typeahead. Bigger issue is with subaccounts, but extends to others as well. This is an issue that I can address in my PR, but we should agree on which format we want to work with. For example, `[{ type: 'Recipient Domain', value: 'whatever.com' }]` is how the typeahead is expecting, but not the end result

PatrickSison (approved):
New filter remove behavior looks good 👍

PR #1822, opened 2020-09-29
AC-1522 -  Removed feature flag - sso_scim_section and added cool new button loading states

shrirupa:
AC-1522 -  Removed feature flag - sso_scim_section and used cool new button loading states 

### What Changed
 - Removed feature flag - sso_scim_section
 - Used cool new button loading states 

### How To Test
 - SSO SCIM Section is visible to users without enabling the feature flag
 - Check the loading states for buttons

### To Do
- [ ] Address feedback

### DIFF
--- a/src/pages/account/components/SingleSignOnPanel/SCIMTokenSection.js
+++ b/src/pages/account/components/SingleSignOnPanel/SCIMTokenSection.js
@@ -13,6 +13,9 @@ export default function SCIMTokenSection(props) {
     deleteScimToken,
     error,
     showAlert,
+    generateScimTokenPending,
+    deleteScimTokenPending,
+    scimTokenListLoading,
     resetScimTokenErrors,
   } = props;
   const { closeModal, isModalOpen, openModal, meta: { name } = {} } = useModal();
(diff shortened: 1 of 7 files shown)


PatrickSison (approved):
Looks good 👍 I think failing unit test is some update with moment or something. Had same issue.

src/pages/account/components/SingleSignOnPanel/SCIMTokenSection.js
```diff
@@ -93,8 +100,10 @@ export default function SCIMTokenSection(props) {
               <ButtonWrapper>
                 <Button
                   variant="primary"
+                  loading={
```
PatrickSison, line 103: Actually, I wonder if this ends up being an issue for non-hibana. Maybe we need to add a mapping to disabled in the wrapper.

src/pages/account/components/SingleSignOnPanel/SCIMTokenSection.js
```diff
@@ -126,6 +135,7 @@ export default function SCIMTokenSection(props) {
                   onClick={() => {
                     handleDeleteToken();
                   }}
+                  loading={deleteScimTokenPending || scimTokenListLoading}
```
PatrickSison, line 138: Same thing with this one

src/pages/account/components/SingleSignOnPanel/SCIMTokenSection.js
PatrickSison: Wrapper referring to the matchbox Button wrapper

shrirupa: Good point! I will it in.

PatrickSison (approved):

PR #1825, opened 2020-09-29
SA:1664 - Move reportOptions into Context

kchu1337:
### What Changed
 - Moved ReportOptions redux state into Context with hooks.
 - Replaced action/reducer/selectors.
 - Replaced instances of reportOptions redux action/reducer with a useReducer.

### How To Test
 - Go to Report Builder and do everything you can think of.
 - Try changing dates, filters, metrics; and check that the UI changes are correct. Also check that the API calls are correct as well.

### To Do
- [ ] Change the patterns? Best Practices?
- [ ] Add more tests?


src/actions/summaryChart.js
```diff
@@ -39,6 +39,40 @@ export function refreshSummaryReport(
...
+  return (dispatch, getState) => {
+    const { summaryChart } = getState();
+
+    // if new metrics are included, convert them to their full representation from config
+    if (updates.metrics) {
+      updates = { ...updates, metrics: getMetricsFromKeys(updates.metrics, true) };
+    }
+
+    // merge together states
+    const merged = {
```
kchu1337, line 59: Mostly this change. I checked and all instances of the function pass on the entire reportOptions state from context so there should be no need to merge it with the reportOptions state from redux.

src/pages/reportBuilder/components/AddFiltersSection.js
```diff
@@ -12,7 +12,7 @@ import {
 } from 'src/actions/metrics';
 import { list as listSubaccounts } from 'src/actions/subaccounts';
 import { list as listSendingDomains } from 'src/actions/sendingDomains';
-import { initTypeaheadCache, setFilters } from 'src/actions/reportOptions';
+import { initTypeaheadCache } from 'src/actions/reportOptions';
```
kchu1337, line 15: Left it here since it doesn't actually any reportOptions reducers.

src/pages/reportBuilder/components/GroupByOption.js
```diff
@@ -1,26 +1,29 @@
...
+  const [reportOptions] = useReportBuilderContext();
+
   const selectId = useUniqueId('break-down-by');
 
   const [topDomainsOnly, setTopDomainsOnly] = useState(true);
 
   const handleGroupChange = e => {
     if (e.target.value !== 'placeholder') {
-      _getTableData({ groupBy: e.target.value });
+      _getTableData({ groupBy: e.target.value, reportOptions });
```
kchu1337, line 18: Needed to add this since we cannot get reportOptions state in the action anymore.

src/selectors/bounceReport.js
```diff
@@ -4,7 +4,8 @@ import { selectReportSearchOptions } from './reportSearchOptions';
 import _ from 'lodash';
```
kchu1337, line 4: These are all prettier changes. I changed some contents, and then undid my changes but prettier still activated.

src/pages/reportBuilder/context/ReportBuilderContext.js
```diff
@@ -0,0 +1,217 @@
...
+    },
+    [dispatch],
+  );
+
+  //Not currently used but I am leaving them here for now.
+  const clearFilters = useCallback(() => {
+    dispatch({
+      type: 'CLEAR_FILTERS',
+    });
+  }, [dispatch]);
```
kchu1337, line 200: Should I remove these instead?

src/pages/reportBuilder/context/ReportBuilderContext.js
```diff
@@ -0,0 +1,217 @@
...
+
+  const actions = { addFilters, clearFilters, setFilters, removeFilter, refreshReportOptions };
+  const selectors = useMemo(() => getSelectors(state), [state]);
+  return (
+    <ReportOptionsContext.Provider value={[state, actions, selectors]}>
```
kchu1337, line 205: Is this the best option to return action and selectors?

PatrickSison: Not sure what best option for this is. I'm cool with the [state, action] pattern, but throwing selectors in feels a bit weird. Also a bit much to condense all into one object. I'm fine with it for now, but this pattern worth checking with @nicklemmon as well

PatrickSison (commented):
Looks good for most part. I have one suggested change, and one I'm unsure on as well, but seems ok for most part.

src/actions/summaryChart.js
```diff
@@ -39,6 +39,40 @@ export function refreshSummaryReport(
   };
 }
 
+export function refreshSummaryReportReportBuilder(
```
PatrickSison, line 42: We might as well change this to `refreshReportBuilder` I think

PatrickSison (approved):
I think overall looks good. One minor change, but behavior seems consistent and the same.

PR #1828, opened 2020-10-05
SA-1712: Update billing UI for Saved Reports

kchu1337:
### What Changed
 - Added conditions to the features comparison.
 - Added saved reports to the feature comparison.
 - Added saved reports to the plan change check.
 - Added lots of tests.

### How To Test
 - Go to an account with saved reports enabled. Go to `/account/billing/plan`.
 - Click on `compare features` and verify that the modal opens and that you can see a row for saved reports.
 - If the account has >30 saved reports & is on premium (appteam stg works), change plans to starter. Check that there is an authorization message that prevents the user from switching. Then switch to free and verify that the message is not present then.
 - Go to an account without saved reports enabled. Go to `/account/billing/plan`.
 - Click on `compare features` and verify that the modal opens and there is no row for saved reports.
 - Try changing plans and verify that there is no message about saved reports.


src/pages/billing/context/FeatureChangeContext.js
```diff
@@ -63,16 +63,16 @@ export const FeatureChangeProvider = ({
                 description: (
                   <div>
                     <span>
-                      {`Your new plan doesn't include dedicated IPs. 
-                    Your current IP(s) will be removed at the end of your current billing cycle.`}
+                      Your new plan doesn't include dedicated IPs. Your current IP(s) will be
+                      removed at the end of your current billing cycle.
```
kchu1337, line 67: Prettier kept making changes. Should have no effect on what is displayed

src/pages/billing/context/FeatureChangeContext.js
```diff
@@ -129,6 +129,50 @@ export const FeatureChangeProvider = ({
               };
             }
             return resObject;
+          }
+          case 'reports': {
```
kchu1337, line 133: Most of it is copied from the subaccounts case

src/pages/billing/context/FeatureChangeContext.js
```diff
@@ -129,6 +129,50 @@ export const FeatureChangeProvider = ({
...
+              const description = (
+                <div>
+                  <>
+                    {limitOfNewPlan === 0
+                      ? "Your new plan doesn't include saved reports."
+                      : `Your new plan only allows for ${pluralString(
+                          limitOfNewPlan,
+                          'saved reports',
+                          'saved reports',
+                        )}.`}
```
gfriedmansp, line 147: I think this would have an extra period in the case of line 142

src/pages/billing/context/FeatureChangeContext.js
```diff
@@ -129,6 +129,50 @@ export const FeatureChangeProvider = ({
...
+                  )}
+                </div>
+              );
+
+              resObject.reports = {
+                label: 'Saved Reports',
+                description,
+                condition: !qtyExceedsLimit,
+                action: qtyExceedsLimit ? (
+                  <Button variant="destructive" external to="/signals/analytics">
```
gfriedmansp, line 166: Is this an `external` link? Not sure if it's just being used for style or something.

gfriedmansp:
Should we add a way to visibly see the additional report creation failed in the Modal? It isn't terribly obvious to me that it failed after saving with the "Something went wrong".

gfriedmansp:
When I get the message "Your account has reached its limit...", it tells me to delete a report. Is the only way to delete a report to go to the list of reports and delete it from there? Not sure if we should give more guidance.

gfriedmansp:
After deleting a report, it doesn't let me create a new one until I refresh. Is there an easy solution for this? (I think I heard about a ticket similar to this one in our grooming).

gfriedmansp (approved):
After in-person review, I think this all looks good. I had the one question regarding the button prop being used, but it should not be a show-stopper.

src/pages/billing/context/FeatureChangeContext.js
kchu1337: I think it's because the page wants the user to make a change, so it opens a new window/tab for the user to do so.

gfriedmansp (approved):
LGTM 👍

PR #1832, opened 2020-10-07
AC-1715 Replace occurrences of api.segment.io and cdn.segment.com with Sparkpost proxy

gfriedmansp:
### What Changed

- Changed the single instance of `cdn.segment.com` with our proxy domain `cdn.analytics.sparkpost.com`
  - When the CDN is loaded on the page, it will make network requests to `api.analytics.sparkpost.com` instead of `api.segment.io`
- Added both of the proxy URLs to our `blockHosts` list for cypress tests (with a comment for clarity of what it is)

### How To Test

1. Load any page in the app
2. Open network request tab
3. Search for "segment", ensure there are no calls to either cdn.segment.com or api.segment.io
4. Search for "analytics.sparkpost.com", ensure there is at least a call to `cdn.analytics.sparkpost.com` and `api.analytics.sparkpost.com`.

1. Turn on adblocker (ublock, for example)
2. Repeat all of the previous steps and ensure none of those requests are being blocked by default.

### Findings
- We were unsure how Segment was changing their backend to accommodate our use of `api.analytics.sparkpost.com` since we do not manually use their original `api.segment.io` domain in our code.
  - They used a script to update the CDN call _for each source_ that replaced it with our proxy domain.


src/components/segment/Segment.js
```diff
@@ -4,7 +4,7 @@ import getConfig from 'src/helpers/getConfig';
...
+const getScript = key => `!function(){var analytics=window.analytics=window.analytics||[];if(!analytics.initialize)if(analytics.invoked)window.console&&console.error&&console.error("Segment snippet included twice.");else{analytics.invoked=!0;analytics.methods=["trackSubmit","trackClick","trackLink","trackForm","pageview","identify","reset","group","track","ready","alias","debug","page","once","off","on","addSourceMiddleware","addIntegrationMiddleware"];analytics.factory=function(e){return function(){var t=Array.prototype.slice.call(arguments);t.unshift(e);analytics.push(t);return analytics}};for(var e=0;e<analytics.methods.length;e++){var t=analytics.methods[e];analytics[t]=analytics.factory(t)}analytics.load=function(e,t){var n=document.createElement("script");n.type="text/javascript";n.async=!0;n.src="https://cdn.analytics.sparkpost.com/analytics.js/v1/"+e+"/analytics.min.js";var a=document.getElementsByTagName("script")[0];a.parentNode.insertBefore(n,a);analytics._loadOptions=t};analytics.SNIPPET_VERSION="4.1.0";
```
gfriedmansp, line 7: The only thing that changed here is `cdn.segment.com` -> `cdn.analytics.sparkpost.com`

bkemper: Yikes, I wish this was parameterized.

bkemper (approved):

src/components/segment/Segment.js
gfriedmansp: Definitely valid. Previously wasn't worried about it because it was copied directly from segment, but now that we've changed it, it makes sense to make it a bit more readable.

gfriedmansp: I pulled `PROXY_CDN` into a variable and injected into the script. We could have pulled it into a config (like we have with the API key), but I don't think we are going to change that value between environments.

bkemper (approved):

PR #1835, opened 2020-10-07
FE-1198 - Delete Domain Confirmation Modal

nstanard:
### What Changed
 - Added a delete confirmation modal to the domain details delete domain button (http://localhost:3100/domains/details/fad-5415.io)

### How To Test
 - Run the cypress tests for domains/domainDetails.spec.js
 - Load the app the navigate to `http://localhost:3100/domains/details/fad-5415.io` logged in with appteam. Then click on Delete Domain.

### To Do
- [ ] Address any feedback/suggestions that the team finds important to this work.


shrirupa (commented):
I think you can just delete the failing unit test. you have already covered it in the cypress test

src/pages/domains/components/tests/DeleteDomainSection.test.js
```diff
@@ -1,12 +1,12 @@
 import React from 'react';
 import { shallow } from 'enzyme';
```
shrirupa, line 2: you can delete this whole file actually doesn't look relevant anymore.

src/pages/domains/components/DeleteDomainSection.js
```diff
@@ -47,9 +50,25 @@ export default function DeleteDomainSection({ domain, isTracking, id, history })
...
           <Panel.Section>
-            <Button variant="destructive" onClick={handleDeleteDomain}>
+            <Button variant="destructive" onClick={() => openModal()}>
               Delete Domain
             </Button>
+
+            <DeleteModal
+              title="Are you sure you want to delete this domain?"
+              open={isModalOpen}
+              content={
```
shrirupa, line 60: I am really sorry, I didn't notice this earlier. 

isPending should actually equal to  `deletePending`, and you can add this to Container.js as 
`deletePending: state.trackingDomain.deleting || state.sendingDomain.deleting` in mapStateToProps.
you would have to update this src/reducers/sendingDomains with

>      case 'DELETE_SENDING_DOMAIN_PENDING':
>       return { ...state, deleting: meta.domain, deleteError: null };
> 
>     case 'DELETE_SENDING_DOMAIN_FAIL':
>       return { ...state, deleting: false, deleteError: payload };
> 
>     case 'DELETE_SENDING_DOMAIN_SUCCESS':
> 

you can reference src/reducers/trackingDomain for the same.

src/reducers/sendingDomains.js
```diff
@@ -31,6 +31,15 @@ export default (state = initialState, { type, payload, meta }) => {
     case 'LIST_SENDING_DOMAINS_FAIL':
       return { ...state, listError: payload, listLoading: false };
 
+    case 'DELETE_SENDING_DOMAIN_PENDING':
+      return { ...state, deleting: meta.domain, deleteError: null };
```
shrirupa, line 35: meta.domain isn't being passed from the action, we should probably just set it `true` here.

shrirupa (approved):
Looks good to me! 🎉

PR #1837, opened 2020-10-08
SA-1610: Compare By infrastructure

kchu1337:
### What Changed
 - Added feature flagged compare by button and drawer

### How To Test
 - Go to an account w/o ui user flag `allow_compare_by` and go to `signals/analytics`. 
 - Verify that there is no compare button.
 - Enable the ui user flag `allow_compare_by` and verify that clicking the compare button opens the drawer. (appteam should be enabled already).


nicklemmon (changes requested):
@kchu1337 Appears to work as expected - two things:
1. Looks like a rebase is needed? There are merge conflicts.
2. We should probably have some automated testing in place? I know it's small, but it would still make sense to do for each ticket.

nicklemmon (approved):
@kchu1337 Looks good!

PR #1843, opened 2020-10-13
SA-1721: Add filter link works for comparators table

PatrickSison:
### What Changed
 - Group by table adds filters for new filter structure

### How To Test
 - On hibana account with feature flag `allow_report_filters_v2`, go to `/signals/analytics`
 - Go to bottom group by table and select a filter
 - Go to group by table
 - Click link to add filter
 - Verify that link adds filter to the table

### To Do
- [ ] Address feedback


src/pages/reportBuilder/components/AddFilterLink.js
```diff
@@ -26,7 +26,7 @@ export const AddFilterLink = ({ newFilter }) => {
 
   const handleClick = e => {
     if (!isNewTabKeyPressed) {
-      addFilters([newFilter]);
+      addFilters(newFilter);
```
nicklemmon, line 29: Much better

src/pages/reportBuilder/context/ReportBuilderContext.js
```diff
@@ -110,6 +111,15 @@ const reducer = (state, action) => {
...
   switch (action.type) {
+    case 'ADD_FILTERS': {
+      return {
+        ...state,
+        filters: [
+          ...state.filters,
+          { AND: { [FILTER_KEY_MAP[action.payload.type]]: { eq: [action.payload] } } },
+        ],
+      };
+    }
```
nicklemmon, line 122: For some reason adding a new filter from the group by table causes the table to refresh and then show an empty state?

Screen Shot 2020-10-14 at 3 42 14 PM

src/pages/reportBuilder/components/tests/AddFilterLink.test.js
```diff
@@ -40,7 +40,7 @@ describe('Add Filter Link', () => {
       'to',
       '/report?filters=Campaign%3A%20shinynewfilter&filters=Subaccount%3Asubmarine%3A10&metrics=count_something',
     );
-    expect(addFilters).toHaveBeenCalledWith([{ id: 10, type: 'Subaccount', value: 'submarine' }]);
+    expect(addFilters).toHaveBeenCalledWith({ id: 10, type: 'Subaccount', value: 'submarine' });
```
nicklemmon, line 43: It looks like the integration tests are testing this behavior, but only the old implementation. We should probably add some that test the new implementation as well.

nicklemmon (changes requested):
@PatrickSison Seeing some unusual behavior with the table rendering an empty state?

Also, I think we could stand to have more testing backing this one up.

src/pages/reportBuilder/context/ReportBuilderContext.js
PatrickSison: Not sure how this was working in the first place, but requests are coming from 2 separate places (group by table, and the main `refreshReportBuilder`. The group by table request is not properly attaching filters to the request and doesn't seem like it has been. Will update and add testing around this.

src/helpers/metrics.js
```diff
@@ -43,11 +43,7 @@ export function getQueryFromOptions(
     precision,
   };
   if (isComparatorsEnabled) {
-    //Formatted as JSON of { groupings: [filterObject]}
-    //TODO: Uncomment once it's ready
-    // options.query_filters = filters.length
-    //   ? encodeURI(JSON.stringify({ groupings: filters }))
-    //   : undefined;
+    options.query_filters = filters.length ? JSON.stringify({ groupings: filters }) : undefined;
```
PatrickSison, line 46: This adds the actual API request query param. It's real

src/actions/summaryChart.js
```diff
@@ -179,7 +180,10 @@ export function _getTableDataReportBuilder({ params, metrics, groupBy, reportOpt
 
     // Gets filters and metrics for params
     if (!params && reportOptions) {
-      params = getQueryFromOptions({ ...reportOptions, metrics: activeMetrics });
+      params = getQueryFromOptions(
+        { ...reportOptions, metrics: activeMetrics },
+        { isComparatorsEnabled },
```
PatrickSison, line 185: This was missing and causing weird empty state behavior

src/helpers/metrics.js
nicklemmon: Nice!

So I'm seeing some errors for some of the "Break Down By" options - I think the `value` for each option is slightly different than what the API expects? When I selected "Sending IP" from the dropdown:

Screen Shot 2020-10-15 at 11 21 28 AM

cypress/tests/integration/signals-analytics/analytics-report/analyticsReportFiltersForm.spec.js
```diff
@@ -329,6 +335,30 @@ if (IS_HIBANA_ENABLED) {
...
+      });
+
+      cy.wait('@nextGetDeliverability').then(xhr => {
+        cy.wrap(xhr.url).should(
+          'include',
+          'query_filters={"groupings":[{"AND":{"templates":{"notLike":["template"]}}},{"AND":{"sending_domains":{"like":["thisisasendingdomain"]}}}]}',
+        );
+      });
+    });
   });
```
nicklemmon, line 362: Nice. Really good test here.

src/helpers/metrics.js
nicklemmon: It looks like this happens after adding one "Break Down By" filter. So steps to reproduce:
1. Select "Break Down By" -> "Campaign" and choose one of the campaigns
2. Then try and select any other "Break Down By" option

To be honest, the UX for this whole table is very confusing to me. It wouldn't be clear to me on its face whether clicking another option in the table would add to or replace the existing filter 🤷‍♂️

nicklemmon (changes requested):
@PatrickSison I found an issue - though I'm a little unsure what the UX expectation is!

nicklemmon (approved):
@PatrickSison Looks good - let's make sure tests pass, then move forward.

PR #1844, opened 2020-10-14
TR-2709 - Reflect the new Delivery Confidence Score into the App

nicklemmon:
[TR-2709](https://sparkpost.atlassian.net/browse/TR-2709)

### What Changed
- Adds new `delivery_confidence_score` to the `/recipient-validation/single/<email>` page
- Adds supporting tests
- Updates styles slightly to accommodate wider text
- After encountering a warning while running Jest, updated the `ExternalLink` component to no longer define a styled component within another React component

### How To Test
- Go to `/recipient-validation/single` and validate an email address on an account with a credit card
- Verify the delivery confidence score renders both in the list and the raw API response section
- Verify tests adequately capture acceptance criteria
- Verify instances of `ExternalLink` still render correctly

### To Do
- [ ] Incorporate feedback


src/pages/recipientValidation/SingleResult.js
```diff
@@ -225,6 +243,11 @@ function ResultCodeBlock({ data }) {
       )}
       <TabCharacter />
       <TabCharacter />
+      "delivery_confidence":{' '}
```
kchu1337, line 246: Could replace the entire formatting block with 
```js
  const resultObj = {
    results: {
      delivery_confidence,
      result,
      valid,
      reason,
      is_role,
      is_disposable,
      is_free,
      did_you_mean
    }
  }
  return (
    <pre id="json" className={styles.CodeSnippet}>
      {JSON.stringify(resultObj, undefined, 2)}
    </pre>
  );
}
```
But it would also get rid of the data-ids so if you need them for tests, then it might not be worth refactoring.

kchu1337 (changes requested):
One minor thing. Up to you if you want to change.

src/pages/recipientValidation/SingleResult.js
nicklemmon: I think this was done to allow for FE control of the rendered output. It isn't the pretties thing in the world, though, I agree.

kchu1337 (approved):

PR #1848, opened 2020-10-14
[AC-1735] (UI) Change the format of billing cycle dates to use UTC format

shrirupa:
[AC-1735] (UI) Add the offset of +08:00 to billing cycle dates

Note - we had to change this because api started returning the utc date with time T00:00

### What Changed
 - Add the offset of +08:00 to billing cycle dates for DashboardPage, DashboardPageV2 and UsageReport

### How To Test
 - Check if the dates match the month.start and month.end returned by GET /account?include=usage

### To Do
- [ ] Address feedback

[AC-1735]: https://sparkpost.atlassian.net/browse/AC-1735

### DIFF
--- a/src/selectors/accountBillingInfo.js
+++ b/src/selectors/accountBillingInfo.js
@@ -222,6 +222,14 @@ export const selectMonthlyTransmissionsUsage = createSelector(getTransmissionsUs
   _.get(usage, 'month.used', 0),
 );
 
+export const selectStartOfBillingPeriod = createSelector(
+  [getTransmissionsUsage, getBillingPeriod],
+  (usage, billingPeriod) => {
+    // IMPORTANT CAVEAT: This will not accurately return the billing period for annual plans due to an API limitation
+    return _.get(usage, `${billingPeriod}.start`);
+  },
+);
+
 export const selectEndOfBillingPeriod = createSelector(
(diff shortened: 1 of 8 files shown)


PatrickSison (approved):
Seems good. Do we need to clarify that date/times are in UTC at all? If users are behind UTC time, then the cycle would terminate earlier than expected.

src/selectors/accountBillingInfo.js
```diff
@@ -226,15 +228,23 @@ export const selectStartOfBillingPeriod = createSelector(
   [getTransmissionsUsage, getBillingPeriod],
   (usage, billingPeriod) => {
     // IMPORTANT CAVEAT: This will not accurately return the billing period for annual plans due to an API limitation
-    return _.get(usage, `${billingPeriod}.start`);
+    return _.get(usage, `${billingPeriod}.start`)
+      ? moment(_.get(usage, `${billingPeriod}.start`))
+          .utc()
+          .format('YYYY-MM-DD') + 'T08:00:00.000Z' //added the T08:00:00.000Z to convert it actual time of when the bill run happens
```
PatrickSison, line 234: This works, but I think moment itself could convert it to a date and has helper functions to add the 8 hours if we needed it.

PatrickSison (approved):

PR #1849, opened 2020-10-15
SA-1747 - (UI) "is equal to" rendering within the ActiveFilters UI for some saved reports

nicklemmon:
[SA-1747](https://sparkpost.atlassian.net/browse/SA-1747)

### What Changed
- Adds additional check for filters in the initial state for the new comparator filters
- Adds new helper function that retrieves UI state for tags - distinct from form fields though using some of the same logic
- Renames report builder integration tests to match standard - they were a little more verbose than needed now that the feature has grown a good bit

### How To Test
- Verify tags still render as expected using the filters form and query params with the account flag `allow_report_filters_v2` set to `true`
- Verify that when report filters are in their initial state (`[{ AND: {} ]`) that the filters section is no longer rendered

### To Do
- [ ] Incorporate feedback


src/pages/reportBuilder/components/ReportOptions.js
```diff
@@ -42,6 +42,12 @@ export function ReportOptions(props) {
     return !Boolean(processedMetrics.length);
   }, [processedMetrics]);
 
+  // Render filters when metrics are not empty, when filters are not in their initial state, and when the filters array is not empty
+  const hasFilters =
+    !isEmpty &&
+    !_.isEqual(reportOptions.filters, [{ AND: {} }]) &&
```
PatrickSison, line 48: I wonder where this issue originates where reportOptions filters can get into the empty state. Form submission properly clears filters, so not sure how it's getting into the weird empty state

PatrickSison (commented):
Should the initial state be the `[{ AND: {} }]` blob or just an empty array? I think the form returns an empty array, and not sure how difficult it is long term to equate the initial state to being the empty state. I feel like we can miss that.

src/pages/reportBuilder/components/ReportOptions.js
nicklemmon: the form initializes with `[{ AND: {} }]` since it renders with those empty form elements that represent that state if that makes sense. But maybe it shouldn't work that way?

src/pages/reportBuilder/helpers/index.js
```diff
@@ -71,11 +71,13 @@ export function getInitialGroupState() {
 
 /**
  * Returns UI state to render multi-layer filter forms based on the user's entry
+ * Unit tests for this function were intentionally not written as they are tied closely to UI state
+ * and are well tested by integration tests.
  *
- * @param {array} remappedGroupings - array of iterable filter groupings, usually derived from `getIterableFormattedGroupings`
+ * @param {array} iterableGroupings - array of iterable filter groupings, usually derived from `getIterableFormattedGroupings`
```
nicklemmon, line 77: I renamed this function to make it a little more clear why it exists in the first place

PR #1851, opened 2020-10-15
FE-1208 - Upgrade Redux and Redux-related dependencies

nicklemmon:
[FE-1208](https://sparkpost.atlassian.net/browse/FE-1208)

### What Changed
- Upgrades the following Redux-related dependencies:
  + `redux`
  + `react-redux`
  + `redux-form`
  + `redux-thunk`
  + `raven-for-redux`
- Updates snapshots that reference `Connect(<Whatever>)` components
- Updates explicitly tests where use of Enzyme's `.find()` method were searching for `Connect(<Whatever>)`. Alternatively, the components were imported in to the tests. Ideally, this sort of testing strategy would be used less over time.
- Adds a few Cypress tests for the AB testing feature in order to reproduce bug
- Resolves crashing on AB testing details page by adding an `if` statement to a helper - this might not be the right solution 🤔 

### How To Test
- Verify tests pass
- Go through core flows in the app and verify nothing is broken - probably worth doing with the console open

### To Do
- [ ] Incorporate feedback

#### Bugs Found:
- [x] Creating a new AB test resulted in a page crash upon redirect - "Cannot read property 'from' of undefined" stemming from `src/helpers/date` - **verified that this does not happen on `master`** - our automated testing didn't catch this


src/helpers/date.js
```diff
@@ -135,7 +135,10 @@ export function getRelativeDates(
 }
 
 export function getDuration(dates, unit = 'hours') {
+  if (!dates) return;
```
nicklemmon, line 138: This prevents the AB test page crash - but the page crash doesn't happen with an older version of `redux` and `redux-form`. Could an unusual race condition have been introduced?

src/pages/abTesting/components/tests/AbTestCreateForm.test.js
```diff
@@ -10,17 +10,13 @@ describe('A/B Test Create Form Component', () => {
...
   };
 
   beforeEach(() => {
     wrapper = shallow(<AbTestCreateForm {...props} />);
   });
 
-  it('should render correctly', () => {
-    expect(wrapper).toMatchSnapshot();
-  });
-
```
nicklemmon, line 23: This was the only change - the rest is Prettier

src/components/auth/tests/__snapshots__/SuspensionAlerts.test.js.snap
```diff
@@ -13,11 +13,11 @@ Array [
       <div>
         For any questions or to request reactivation, please
          
-        <Connect(SupportTicketLink)
+        <Memo(Connect(SupportTicketLink))
```
gfriedmansp, line 16: Why are we able to leave `Connect` in the names of these components, but not the ones later in the PR? Is it because they're composed?

src/components/reduxFormWrappers/tests/__snapshots__/SubaccountTypeaheadWrapper.test.js.snap
```diff
@@ -1,7 +1,7 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`SubaccountTypeaheadWrapper renders subaccounts typeahead 1`] = `
-<Connect(SubaccountTypeahead)
+<Component
```
gfriedmansp, line 4: Kind of unfortunate we lose the specificity of these component names

src/components/support/tests/Support.test.js
```diff
@@ -85,7 +86,7 @@ describe('Support', () => {
 
   it('calls closeSupportPanel when support form is closed', () => {
     wrapper.setProps({ currentSupportView: 'ticket' });
-    wrapper.find('Connect(ReduxForm)').simulate('close');
+    wrapper.find(SupportForm).simulate('close');
```
gfriedmansp, line 89: Big fan of this > the find by string component name 👏

src/pages/inboxPlacement/tests/__snapshots__/TestDetailsPage.test.js.snap
```diff
@@ -1,16 +1,6 @@
...
-  alert={
-    Object {
-      "message": "You dun goofed",
-      "type": "error",
-    }
-  }
-  to="/inbox-placement"
-/>
-`;
+exports[`Page: Single Inbox Placement Test handles errors 1`] = `null`;
```
gfriedmansp, line 3: Should this actually return null?

src/pages/users/components/tests/__snapshots__/EditForm.test.js.snap
```diff
@@ -7,7 +7,15 @@ exports[`Component: EditForm should render 1`] = `
       <Field
         allowSubaccountAssignment={false}
         allowSuperUser={false}
-        component={[Function]}
+        component={
```
gfriedmansp, line 10: Interesting that it uses the Connect(RoleRadioGroup) format here for a nested component.

gfriedmansp:
Left a couple comments - I think overall it looks good. Surprised there was not more we needed to change in our code to upgrade these dependencies!

src/pages/inboxPlacement/tests/__snapshots__/TestDetailsPage.test.js.snap
nicklemmon: hmm, probably not.

src/components/support/tests/Support.test.js
nicklemmon: me too - it is a lot more change-proof.

src/components/reduxFormWrappers/tests/__snapshots__/SubaccountTypeaheadWrapper.test.js.snap
nicklemmon: Agreed. Seems like a step backwards.

src/components/auth/tests/__snapshots__/SuspensionAlerts.test.js.snap
nicklemmon: Well, this is weird - I ran the test locally and the snapshot fails - but in CI it passed 😬

nicklemmon: Neeevermind - obviously due to swapping branches and running the test before running `npm install`

nicklemmon: It could be because the `SupportTicketLink` is not a default export?

nicklemmon: Unfortunately I'm not really sure after looking in to it some more.

src/pages/inboxPlacement/tests/__snapshots__/TestDetailsPage.test.js.snap
nicklemmon: This was another situation where I needed to update the test to import the component rather than `.find()` using the display name. Pushing that up now.

nicklemmon:
@gfriedmansp I fixed one thing that was definitely a problem - take a look and let me know when you're comfortable moving forward.

Thanks again.

src/components/auth/tests/__snapshots__/SuspensionAlerts.test.js.snap
gfriedmansp: At least it is consistent and only in a snapshot hah.

gfriedmansp (approved):
Thanks for the updates. I think this should be ready to go. Do we want to set up a group smoke test since this change is pretty far reaching?

nicklemmon:
> Thanks for the updates. I think this should be ready to go. Do we want to set up a group smoke test since this change is pretty far reaching?

@gfriedmansp That's a great idea. Let me schedule something.

PR #1852, opened 2020-10-15
SA-1721: Fixed issue with group by in non comparators filters

PatrickSison:
### What Changed
 - Fixed group by behavior on non comparators

### How To Test
 - Go to report builder with `allow_report_filters_v2` as `false`
 - Go to group by, select any category
 - Click one filter to add it in
 - Only one row should be visible after
 - Choose another category and then go back to original category
 - Still should only have one role

### To Do
- [ ] Address feedback

### DIFF
--- a/src/pages/reportBuilder/components/tests/GroupByOption.test.js
+++ b/src/pages/reportBuilder/components/tests/GroupByOption.test.js
@@ -1,9 +1,14 @@
 import React from 'react';
 import { GroupByOption } from '../GroupByOption';
 import { shallow } from 'enzyme';
+import { dehydrateFilters } from '../../helpers';
 
 jest.mock('../../context/ReportBuilderContext', () => ({
   useReportBuilderContext: jest.fn(() => ({ state: { foo: 'bar', filters: [] } })),
+}));
+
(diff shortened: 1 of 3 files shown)


nicklemmon (approved):
@PatrickSison This does fix the issue you described!

src/pages/reportBuilder/components/tests/GroupByOption.test.js
```diff
@@ -90,4 +95,14 @@ describe('Group By Option', () => {
...
+
+  it('should dehydrate filters if on new comparators filters', () => {
+    wrapper.setProps({ isComparatorsEnabled: true });
+    expect(wrapper.find('Checkbox')).toBeChecked();
+    wrapper.find('Checkbox').simulate('change');
+    expect(wrapper.find('Checkbox')).not.toBeChecked();
+    wrapper.find('Checkbox').simulate('change');
+    expect(props._getTableData).toHaveBeenCalledTimes(2);
+    expect(dehydrateFilters).toHaveBeenCalled();
+  });
```
nicklemmon, line 107: Cool!

PR #1854, opened 2020-10-16
[FE-1203] UX Feedback for Domain Detail Pages Part - 1

shrirupa:
[FE-1203] UX Feedback for Domain Detail Pages Part - 1

### What Changed
- [x] Reset the state for domains when the Domain Details page unmounts.
- [x] Font size should be 200 in Left well MB.
- [x] Vertical spacing in left well title and info copy should be consistent ~ 12px
- [x] In Verify through email Modal - Textfield should not crash into domain
- [x] Toggle should be closer to text in case of "share with all subaccounts". We need to add max-width to this. 
- [x] Remove extra spaces in v=spf1 mx a ~all
- [x] Update text style for Optional - Grey -7 and size is 200
- [x] Wrong text spec in mini banner component MB. Removed the Text in the banner from `<SubduedText>`
- [x] For Sending and Bounce Domain Page - The “Recommended” tag beside the “Add SPF Record” header is too close, per design update it to be 20px distance - use 400 design token

### How To Test
 -  Verify all the above changes look good.

### To Do
- [ ] Address feedback

[FE-1203]: https://sparkpost.atlassian.net/browse/FE-1203

### DIFF
--- a/src/components/toggleBlock/ToggleBlock.js
+++ b/src/components/toggleBlock/ToggleBlock.js
@@ -25,12 +25,15 @@ function OGToggleBlock({ input, label, helpText, ...rest }) {
 }
 
 function HibanaToggleBlock(props) {
-  const { input, label, helpText, ...rest } = props;
+  /* variant prop helps adjusting the spacing between toggle and label text*/
+  const { input, label, helpText, variant, ...rest } = props;
+  const columns = variant === 'dense' ? 4 : 8;
+  const justifyContent = variant === 'dense' ? 'flex-start' : 'flex-end';
 
   return (
     <Box data-id="toggle-block">
       <Grid middle="xs">
-        <Grid.Column xs={8}>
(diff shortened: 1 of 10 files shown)


nstanard:
Thank you for pointing me in the right direction to find this - I saw the input field and played around with it a bit. What do you think about the following for the "Verify through Email" modal?

image
image


sidenote: "Verify through Email" title might be "Verify Through Email" instead.

nstanard:
image

I was trying to figure out how to get the toggle even closer but was struggling because it's using ToggleBlock.js and I'd have to dig in to see if we can do that without messing other spots up. But that toggle does seem kinda out in middle.

nstanard (approved):
I made a couple of comments but UX would be the people to determine if those things are important.

nstanard (approved):
still lgtm!

PR #1856, opened 2020-10-19
[FE-1197] UX Feedback for Domain Details Pages - Part 3

shrirupa:
[FE-1197] UX Feedback for Domain Details Pages - Part 3

### What Changed

- [x] For tracking Domain Detail Page - update the placeholder value to be spgo.io etc use the the same values that we have in the unverified tracking domains banner

- [x] Blocked Domain Banner sub-text should be updated from placeholder to “If your domain’s status is “Blocked”, it’s generally because your domain is already in use by another SparkPost account, your domain has been previously blocked for sending abusive traffic through our service or another provider, or because one or more of our requirements are not met.”

- [x] If set up for sending means they just did the email verification not the sending DKIM verification.. So on that detail page the version to copy and paste the Hostname and value should appear with the details about DKIM information

- [x] The back buttons from bounce domain pages should redirect to /domains/list/bounce

### How To Test
 - Verify if the above changes look good
### To Do
- [ ] Address feedback

[FE-1197]: https://sparkpost.atlassian.net/browse/FE-1197


nstanard (approved):
LGTM!

PR #1857, opened 2020-10-20
SA-1763: Subaccounts are called on mounting report builder

PatrickSison:
### What Changed
 - Subaccounts called on mounting report builder
 - Subaccounts are no longer feature flagged behind additional comparators
 - Subaccounts called once

### How To Test
 - On any account (with or without account ui feature flag `allow_report_filters_v2`), subaccounts API call should be made
 - Click on Group By selector (bottom of page)
 - Select subaaccounts
 - Subaccount names should populate (if available)

### To Do
- [ ] Address feedback

### DIFF
--- a/src/pages/reportBuilder/components/ReportTable.js
+++ b/src/pages/reportBuilder/components/ReportTable.js
@@ -128,16 +128,16 @@ export const ReportTable = props => {
 
   return (
     <>
-      <Panel.LEGACY marginBottom="-1px">
-        <Panel.LEGACY.Section>
+      <Panel marginBottom="-1px">
+        <Panel.Section>
           <GroupByOption
             _getTableData={_getTableData}
             groupBy={groupBy}
             hasSubaccounts={hasSubaccounts}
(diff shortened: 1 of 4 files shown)


src/pages/reportBuilder/components/ReportTable.js
```diff
@@ -16,7 +16,7 @@ import styles from './ReportTable.module.scss';
 
 const tableWrapper = props => {
   return (
-    <Panel.LEGACY borderTop="0">
+    <Panel.LEGACY>
```
PatrickSison, line 19: This is an old bug, but there's supposed to be a line in between the table and the top area (with the select)

nicklemmon: We could just go ahead and swap this out with `<Panel>`? No need to use `Panel.LEGACY` anymore I don't think

nicklemmon (approved):
@PatrickSison Looks good!

PR #1858, opened 2020-10-21
SA-1770: Fixed filters not properly being pulled from query params

PatrickSison:
### What Changed
 - Filter are properly attached when parsed from URL params

### How To Test
 - Go to report builder
 - Don't select a report
 - Add a filter
 - Copy and paste URL
 - Filter should persist through URL

### To Do
- [ ] Address feedback


nicklemmon (approved):
@PatrickSison Looks good to me - thanks for adding comments and renaming that variable. Much more clear now.

PR #1859, opened 2020-10-21
SA-1736: Create a new Scheduled Report

kchu1337:
### What Changed
 - Created new page for scheduled reports
 - Created react-hook-form wrapper for combobox typeahead
 - Added link to schedule a report in the reports modal.
 - Unit tests and Integration test

### How To Test
 - enable user UI option `allow_scheduled_reports`
 - go to `signals/analytics`. Click the `View all reports` button. Click on the actionlist in the modal and select `schedule`.
 - Click the submit button and check the validation is correct.
 - Enter in valid values and check that the API submission works. Also check that the cronjob translates corectly. 

### To Do
- [x] UX questions:
      - What happens after submitting?
      - No way to create a cron to send in a certain week.
- [x] Wait for backed to add timezone as an option. Apply UI changes after other backend changes
       ^ Backed ticket has been created, will fix in another ticket.
- [x] Maybe account UI feature flag instead (only used user UI because saved reports was user UI).


src/components/reactHookFormWrapper/ComboBoxTypeaheadWrapper.js
```diff
@@ -0,0 +1,17 @@
...
+import { ComboBoxTypeahead } from 'src/components/typeahead/ComboBoxTypeahead';
+
+export default function CheckboxWrapper(props) {
+  const { name, setValue, value = [], ...rest } = props;
+  const [componentValue, setComponentValue] = useState(value);
+  const handleChange = e => {
+    setComponentValue(e);
+  };
+  useEffect(() => {
+    if (value !== componentValue) {
```
kchu1337, line 11: It was breaking my unit test without this. Not sure why. 🤷

src/pages/reportBuilder/components/ScheduledReportForm.js
```diff
@@ -0,0 +1,244 @@
...
+  users,
+}) => {
+  const { control, handleSubmit, errors, register, setValue, watch } = useForm({
+    defaultValues: {
+      timing: 'daily',
+      recipients: [],
+    },
+  });
+  const [period, setPeriod] = useState('AM');
+  const [timezone, setTimezone] = useState(getLocalTimezone());
```
kchu1337, line 71: I just handle these 2 fields without react-hook-form since these types of input are less common and I didn't think it was worth the time/effort to make wrappers.

src/pages/reportBuilder/components/ScheduledReportForm.js
```diff
@@ -0,0 +1,244 @@
...
+  { label: 'Tuesday', value: 2 },
+  { label: 'Wednesday', value: 3 },
+  { label: 'Thursday', value: 4 },
+  { label: 'Friday', value: 5 },
+  { label: 'Saturday', value: 6 },
+];
+
+const hasAtLeastOneRecipient = recipientList => recipientList.length > 0;
+
+export const formatFormValues = formValues => {
```
kchu1337, line 35: I'm not using the first/second/ect week value since theres no way to make a single cronjob that runs on the 1st Friday of every month.
Not adding timezone right now, but I talked with backend team and they said they would implement timezone as parameter since it is easier on their end b/c of the cron-parser library accepting timezone as an option.

src/pages/reportBuilder/components/ScheduledReportForm.js
```diff
@@ -0,0 +1,244 @@
...
+  const { name, description = 'NA', subject, period, ...rest } = formValues;
```
kchu1337, line 36: They will remove description from the API soon. But right now, it's a required parameter.

src/pages/reportBuilder/components/ScheduledReportForm.js
```diff
@@ -0,0 +1,244 @@
...
+                  <Heading as="h3" looksLike="h5">
+                    Report
+                  </Heading>
+                  <Text as="p">{report.name}</Text>
+                </div>
+                <div>
+                  <Heading as="h3" looksLike="h5">
+                    From Address
+                  </Heading>
+                  <Text as="p">reports@sparkpost.com</Text>
```
nicklemmon, line 126: We have a component for this I believe - `Definition` in `src/components/text`

src/pages/reportBuilder/components/ScheduledReportForm.js
```diff
@@ -0,0 +1,244 @@
...
+                        onChange={() => {
+                          setPeriod('AM');
+                        }}
+                        value="AM"
+                      >
+                        <Uppercase>AM</Uppercase>
+                      </RadioButtonGroup.Button>
+                      <RadioButtonGroup.Button
+                        id="pm"
+                        name="pm"
```
nicklemmon, line 210: These should have the same `name` attribute

src/pages/reportBuilder/components/ScheduledReportForm.js
```diff
@@ -0,0 +1,244 @@
...
+                  <TimezoneTypeahead initialValue={timezone} onChange={setTimezone} />
+                </Box>
+              </Inline>
+            </Panel.Section>
+          </Panel>
+          <Inline>
+            <Button type="submit" variant="primary">
+              Schedule Report
+            </Button>
+            <Button variant="secondary">Cancel</Button>
```
nicklemmon, line 232: This doesn't do anything?

src/pages/reportBuilder/components/ScheduledReportForm.js
```diff
@@ -0,0 +1,244 @@
...
+  useEffect(() => {
+    listUsers();
+  }, [listUsers]);
+
+  const Typeahead = (
+    <ComboBoxTypeaheadWrapper
+      disabled={loading}
+      error={errors.recipients && 'At least 1 recipient must be selected'}
+      id="to-address"
+      itemToString={item => (item ? `Name: ${item.name} ---- Email: ${item.email}` : '')}
```
nicklemmon, line 82: Is this how it's supposed to be formatted? Looks funky. Doesn't match the mocks either:
Screen Shot 2020-10-27 at 1 37 46 PM

src/pages/reportBuilder/components/ScheduledReportForm.js
```diff
@@ -0,0 +1,244 @@
...
+                    },
+                  })}
+                  label="Time"
+                  name="time"
+                  id="time"
+                  error={errors.time?.message}
+                  maxWidth="12rem"
+                  placeholder="hh:mm"
+                  connectRight={
+                    <RadioButtonGroup id="period" label="Grouping Type">
```
nicklemmon, line 196: I'm glad that worked!

src/pages/reportBuilder/components/ScheduledReportForm.js
nicklemmon: I think you wouldn't need the `setPeriod` function as a result? Not sure though.

src/pages/reportBuilder/components/ScheduledReportForm.js
```diff
@@ -0,0 +1,244 @@
...
+                  name="day"
+                  options={DAY_OF_WEEK_OPTIONS}
+                  disabled={currentTiming === 'daily'}
+                />
+                <TextField
+                  ref={register({
+                    required: 'Required',
+                    pattern: {
+                      value: /^(1[0-2]|0?[1-9]):[0-5][0-9]$/,
+                      message: 'Invalid time format, should be hh:mm 12 hour format',
```
nicklemmon, line 186: Screen Shot 2020-10-27 at 1 41 32 PM

The validation works well, but the error rendering is pretty funky. I think this is due to the flex alignment of the `<Inline/>` component?

src/pages/reportBuilder/components/tests/ScheduledReportForm.test.js
```diff
@@ -0,0 +1,126 @@
...
+    expect(screen.getByText('Email Subject')).toBeVisible();
+    expect(screen.getByText('Send To')).toBeVisible();
+    expect(screen.getByText('Send Report')).toBeVisible();
+    expect(screen.getByText('Daily')).toBeVisible();
+    expect(screen.getByText('Weekly')).toBeVisible();
+    expect(screen.getByText('Monthly')).toBeVisible();
+    expect(screen.getByText('Week')).toBeVisible();
+    expect(screen.getByText('Day')).toBeVisible();
+    expect(screen.getByText('Time')).toBeVisible();
+    expect(screen.getByText('Time Zone')).toBeVisible();
```
nicklemmon, line 42: Yaaay no snapshots!

src/pages/reportBuilder/components/tests/ScheduledReportForm.test.js
```diff
@@ -0,0 +1,126 @@
...
+      recipients: [{ username: 'bob', email: 'bob@ross.com', extra: true }],
+      subject: 'meh',
+    };
+    const formattedValuesBase = {
+      name: 'test',
+      recipients: ['bob'],
+      subject: 'meh',
+    };
+
+    it('formats form values for daily schedule correctly', () => {
```
nicklemmon, line 57: These formatting tests are really great. Very clear.

src/pages/reportBuilder/components/ScheduledReportForm.js
```diff
@@ -0,0 +1,244 @@
...
+                <Box minWidth="19rem">
+                  <TimezoneTypeahead initialValue={timezone} onChange={setTimezone} />
+                </Box>
+              </Inline>
+            </Panel.Section>
+          </Panel>
+          <Inline>
+            <Button type="submit" variant="primary">
+              Schedule Report
+            </Button>
```
nicklemmon, line 231: I don't see any form control disabling when the form is submitting? That's definitely needed across the board.

src/components/reactHookFormWrapper/ComboBoxTypeaheadWrapper.js
```diff
@@ -0,0 +1,17 @@
+import React, { useEffect, useState } from 'react';
+import { ComboBoxTypeahead } from 'src/components/typeahead/ComboBoxTypeahead';
+
+export default function CheckboxWrapper(props) {
```
nicklemmon, line 4: I don't think this is supposed to be called `CheckboxWrapper`

src/pages/reportBuilder/components/ScheduledReportForm.js
nicklemmon: Would wrappers be needed? I guess I don't understand why the checkbox controls would need it.

nicklemmon (changes requested):
@kchu1337 Left you some notes - let me know what questions you have for me.

src/pages/reportBuilder/components/ScheduledReportForm.js
kchu1337: There's also a new matchbox component for it as well https://matchbox-storybook.netlify.app/?path=/story/layout-labelvalue--vertical-orientation

src/pages/reportBuilder/components/ScheduledReportForm.js
kchu1337: Oh yeah I got it working. It wasn't working before and when I tried at first but then I realized I set both values to AM due to bad copy+paste.

src/pages/reportBuilder/components/ScheduledReportForm.js
kchu1337: UX said they would prefer if email was available as well. The ideal situation would be something like this: 

image

With email replacing `appteam`. But I didn't want to also make changes to the comboboxTypeahead in this ticket.

src/pages/reportBuilder/ScheduledReportCreatePage.js
```diff
@@ -0,0 +1,43 @@
...
+import ScheduledReportForm from './components/ScheduledReportForm';
+import { Loading } from 'src/components/loading';
+
+export default function ScheduledReportCreatePage() {
+  const {
+    requestParams: { reportId },
+  } = useRouter();
+  const history = useHistory();
+  const { report, loading } = useSelector(state => state.reports);
+  const dispatch = useDispatch();
```
nicklemmon, line 17: So much better.

src/pages/reportBuilder/components/ScheduledReportForm.js
```diff
@@ -0,0 +1,276 @@
...
+                  <TimezoneTypeahead
+                    disabled={loading}
+                    initialValue={timezone}
+                    onChange={setTimezone}
+                  />
+                </Box>
+              </Inline>
+            </Panel.Section>
+          </Panel>
+          <ButtonWrapper>
```
nicklemmon, line 258: 👍

src/pages/reportBuilder/components/ScheduledReportForm.js
```diff
@@ -0,0 +1,276 @@
...
+                id="scheduled-report-name"
+                error={errors.name?.message}
+              />
+            </Panel.Section>
+            <Panel.Section>
+              <Inline space="800">
+                <div>
+                  <Definition>
+                    <Definition.Label>Report</Definition.Label>
+                    <Definition.Value>{report.name}</Definition.Value>
```
nicklemmon, line 120: Screen Shot 2020-11-02 at 11 15 03 AM

This isn't your doing, but the `Definition.Value` component almost definitely has the wrong font size design token.

nicklemmon: Yeah, inspecting the mocks it should be `18px` which maps to `props.theme.fontSizes[300]` (rather than `400`).

Would you mind updating the component? The text just looks huge right here. I'm sure it does elsewhere too.

nicklemmon: Screen Shot 2020-11-02 at 11 18 48 AM

nicklemmon (changes requested):
@kchu1337 Things are looking good - I had some minor feedback - looks like the `Definition` component has some slightly off CSS. Not your doing, but we may as well fix it now.

Nice work! The form works well.

nicklemmon (approved):
@kchu1337 and @PatrickSison Looks good!

PR #1864, opened 2020-10-26
FE-1218 - Add pagination to the domain list pages

nstanard:
### What Changed
 - _coming soon_

1. http://localhost:3100/domains/list/sending
2. http://localhost:3100/domains/list/tracking
3. http://localhost:3100/domains/list/bounce

### How To Test
 - Go to the domains list pages mentioned above and make sure the paging works.
 
### To Do
- [ ] Incorporate Feedback


src/reducers/summaryTables.js
```diff
@@ -1,30 +1,35 @@
+import { DEFAULT_PER_PAGE as PER_PAGE, DEFAULT_CURRENT_PAGE as CURRENT_PAGE } from 'src/constants';
 export const DEFAULT = '@@DEFAULT_TABLE';
 
 const initialState = {
   [DEFAULT]: {
-    currentPage: 1,
```
nstanard, line 5: note for self: be careful here because introducing a current page constant set to 0 would change the behavior since this was 1 based, not 0 based.

nstanard: `DEFAULT_CURRENT_PAGE` is 1 now so no worries here

src/hooks/useTable/useTable.js
```diff
@@ -28,23 +41,61 @@ function getSortedRows({ rows, sortBy, sortDirection }) {
 }
 
 function getSortDirection({ state, column }) {
+  log('state, column: ', state, column);
   if (state.sortBy === column) {
```
nstanard, line 45: This condition/return seems ... off but I need to dig in more

nstanard: Ohhhh it's reversing the sort when they sort on the same column, of course. This was throwing me off because it was reversing the default sort I passed through on page load.

nstanard:
Closing in favor of react-table (but this can be re-visited depending on the feedback from the other team members)

PR #1869, opened 2020-10-28
SA-1722: Release Saved Reports

kchu1337:
### What Changed
 - Removed feature flag condition for 'allow_saved_reports'
 - Reverted the feature comparison modal to conditionally show saved reports limits.

### How To Test
 - Login to account without 'allow_saved_reports' feature flag. 
 - Check that saved reports is usable.
 - Check that the feature comparison modal shows saved reports.
 - Check that the change billing plan still checks for saved reports
     - Do this by PUT request to `/api/v1/billing/subscription/control/product/reports/usage` with body 
     ```
    {
      "action": "add",
      "quantity": 100,
      "ids": []
     }
     ```


src/pages/billing/components/FeatureComparisonModal.js
```diff
@@ -100,13 +96,9 @@ export function ComparisonModal({ flags, open, handleClose }) {
             <Table key={groupName}>
               <tbody>
                 <GroupHeading groupName={groupName} colSpan={PLANS.length + 1} />
-                {_.map(featureObj, (featureValues, featureName) => {
```
kchu1337, line 103: Basically just removed all the conditional rendering

src/pages/reportBuilder/components/SavedReportsSection/SavedReportsSection.js
```diff
@@ -69,66 +67,65 @@ const SavedReportsSection = props => {
...
-            open={modalStatus === 'save'}
-            saveQuery
-            isOwner={props.currentUser.userName === focusedReport.creator}
-            report={focusedReport}
-            onCancel={() => {
-              setModalStatus('');
-              setFocusedReport({});
-            }}
-          />
+        <Button variant="tertiary" onClick={() => setModalStatus('view')}>
```
kchu1337, line 112: Moved button here.

src/pages/reportBuilder/components/SavedReportsSection/SavedReportsSection.js
```diff
@@ -69,66 +67,65 @@ const SavedReportsSection = props => {
...
-          </Button>
+      <Column width="content">
+        <Button
+          data-id="edit-report-details-button"
+          variant="tertiary"
+          onClick={() => {
+            setFocusedReport(props.selectedItem);
+            setModalStatus('edit');
+          }}
+          disabled={
```
PatrickSison, line 78: Late UX thing that's not in the scope of this ticket, but I wonder if these buttons should be hidden instead of disabled for the release of this ticket

PatrickSison (approved):
Very straight forward. One UX question, but that can be on a later feedback cycle. Everything else is removing the conditional and it seems like it's gone

PR #1872, opened 2020-10-29
SA-1616: Compare By stores in query params

PatrickSison:
### What Changed
 - Added submission for compare by to modify state
 - Updates in query params and preserves in query params

### How To Test
 - Enable `allow_compare_by` feature flag
 - Fill out form
 - Submit
 - Verify that the query params properly reflect the compare filters
 - Refresh page
 - Verify that the form properly populates with the appropriate state

### To Do
- [ ] Address feedback
- [ ] Additional testing


src/helpers/reports.js
```diff
@@ -126,6 +127,27 @@ export function parseSearchNew(search) {
     ret.filters = filtersList;
   }
 
+  if (compare) {
+    const compareFilters = (typeof filters === 'string' ? [compare] : compare).map(filter => {
+      const parts = filter.split(':');
```
bkemper, line 132: Is this structure being used to align with the saved report API?

src/helpers/reports.js
```diff
@@ -126,6 +127,27 @@ export function parseSearchNew(search) {
...
+
+      // Subaccount filters include 3 parts
+      // 'Subaccount:1234 (ID 554):554' -> { type: 'Subaccount', value: '1234 (ID 554)', id: '554' }
+      if (type === 'Subaccount') {
+        id = parts.pop();
+        value = parts.join(':');
+      } else {
+        value = parts.join(':');
+      }
+      return { value, type, id };
```
bkemper, line 145: Are we alright with accepting any value or should we validate them?

cypress/tests/integration/signals-analytics/analytics-report/reportCompareBy.spec.js
```diff
@@ -97,5 +97,68 @@ if (IS_HIBANA_ENABLED) {
         cy.findAllByLabelText('Subaccount').should('have.length', 2);
       });
     });
+
+    it('Properly submits the form and adds to reportOptions', () => {
+      cy.findByRole('button', { name: 'Add Comparison' }).click();
```
bkemper, line 102: Should we make a custom command for this?  `cy.findByButtonName`?

cypress/tests/integration/signals-analytics/analytics-report/reportCompareBy.spec.js
```diff
@@ -97,5 +97,68 @@ if (IS_HIBANA_ENABLED) {
...
+      cy.wait(['@getSubaccounts', '@getDeliverability', '@getTimeSeries']);
+
+      cy.findByRole('button', { name: 'Add Comparison' }).click();
+      cy.findByLabelText(TYPE_LABEL).should('have.value', 'subaccounts');
+      cy.findAllByLabelText('Subaccount')
+        .eq(0)
+        .should('have.value', 'Fake Subaccount 1 (ID 101)');
+      cy.findAllByLabelText('Subaccount')
+        .eq(1)
+        .should('have.value', 'Fake Subaccount 3 (ID 103)');
```
bkemper, line 161: Should we add a test for the error message?

src/pages/reportBuilder/components/CompareByForm/CompareByForm.js
```diff
@@ -66,16 +79,31 @@ function CompareByForm({
...
+    return { filterType: FILTER_KEY_MAP[compare[0].type], filters: compare };
+  };
+  const [state, dispatch] = useReducer(reducer, getInitialState(compare));
+  const { filters, filterType, error } = state;
 
-  function handleSubmit(e) {
+  function handleFormSubmit(e) {
     e.preventDefault();
 
     const cleanedFilters = filters.filter(filter => filter); //filters the filter filters
```
bkemper, line 100: I know this didn't change, but it would help to make it explicit what we are filtering out.  `filters.filter(filter => filter !== null);`

bkemper (commented):
I left some comments.

src/helpers/reports.js
```diff
@@ -126,6 +127,27 @@ export function parseSearchNew(search) {
     ret.filters = filtersList;
   }
 
+  if (compare) {
+    const compareFilters = (typeof filters === 'string' ? [compare] : compare).map(filter => {
```
bkemper, line 131: This should be `typeof compare === 'string`.

src/helpers/reports.js
PatrickSison: Saved reports doesn't have any structure. This would be what's interpreted on the webui side. Not sure what the best data structure for keeping the filter on UI is, but this keeps it flexible to compare filter types later on and this structure is based on the old structure for filters

bkemper: I agree with the flexibility.  I think joining and splitting on a colon or any arbitrary separator is error prone.  Query strings support nested structures.  Here are some comments from qs.


> You may specify an index as well:
```
var withIndexes = qs.parse('a[1]=c&a[0]=b');
assert.deepEqual(withIndexes, { a: ['b', 'c'] });
```

> You can also nest your objects, like 'foo[bar][baz]=foobarbaz':

```
assert.deepEqual(qs.parse('foo[bar][baz]=foobarbaz'), {
    foo: {
        bar: {
            baz: 'foobarbaz'
        }
    }
});
```

> — https://github.com/ljharb/qs


I suggest using these qs features to maintain the structure in the query string.

```
// example compare state
[
  {
    id: '554', 
    type: 'subaccount',
    value: 'my subaccount'
  },
  {
    id: '555', 
    type: 'subaccount',
    value: 'your subaccount'
  }
]

// what it looks like as a query string
?compare[0][id]=554&compare[0][type]=subaccount&compare[0][value]=my%20subaccount&compare[1][id]=555&compare[1][type]=subaccount&compare[1][value]=your%20subaccount
```

src/helpers/reports.js
bkemper: Just discussed with @PatrickSison and decided that it will be addressed in a follow-up ticket.

src/pages/reportBuilder/components/ReportOptions.js
```diff
@@ -55,7 +55,7 @@ export function ReportOptions(props) {
         update.filters = selectedFilters;
       }
 
-      updateRoute({ ...update, report: selectedReport?.id });
+      updateRoute({ ...update, report: selectedReport?.id }, { arrayFormat: 'indices' });
```
bkemper, line 58: Should we just make this the default?

src/helpers/reports.js
```diff
@@ -126,6 +127,10 @@ export function parseSearchNew(search) {
     ret.filters = filtersList;
   }
 
+  if (compare) {
```
bkemper, line 130: Can you file a tech debt ticket to do the same thing for `filters`?

src/pages/reportBuilder/components/ReportOptions.js
PatrickSison: I think we need to take a deeper dive into where query params are used before making the change everywhere.

PatrickSison: I'm a bit nervous about making a widespread change like that

bkemper (approved):
Looks like there is a flaky integration spec that is failing the build, but LGTM.

Two last things to concern, but don't need to changed:
 * The compare[][type] is using the option label and not the option value.
 * The compare[][value] should be the only thing needed to uniquely identify the resource.  Subaccount should not need a value and id.

PR #1875, opened 2020-10-30
SA-1808 - (UI) Advanced comparators values "stick" after changing comparator type

nicklemmon:
[SA-1808](https://sparkpost.atlassian.net/browse/SA-1808)

### What Changed
- Renames function in `useMultiEntry` file to match filename
- Adds test to reproduce the bug
- Adds `useEffect` hook within `useMultiEntry` to update the `valueList` state when the `initialState` of the hook changes

### How To Test
- Enable Hibana and the `allow_report_filters_v2` account UI flag
- Go to `/signals/analytics`
- Click on "Add Filters"
- Select a filter type, swap "Compare By" to "contains" and enter some filter values
- Swap "Compare By" to "does not contain" and verify that the filter values clear properly
- Verify the form continues to work as expected and that no regressions are introduced

### To Do
- [ ] Incorporate feedback


shrirupa (approved):
Looks good to me!

PR #1877, opened 2020-10-30
SA-1805 - (UI) Advanced comparators error with less than 3 character values

nicklemmon:
[SA-1805](https://sparkpost.atlassian.net/browse/SA-1805)

### What Changed
- Adds new `error` and `minLength` properties to the `useMultiEntry` hook state
- Adds `minLength` of `3` to the `MultiEntryController` used in the `FiltersForm`
- Adds supporting integration and unit tests

### How To Test
- With Hibana enabled and the `allow_report_filters_v2` feature flag enabled, go to `/signals/analytics`
- Click on "Add Filters" and change the compare by value to "contains" or "does not contain"
- type 1 or 2 characters and try adding the tag to the field with the space key, enter key, or by blurring the field
- An error should render!
- The error should be removed when: the user clears their entry entirely _or_ the user enters 3 or more characters

### To Do
- [ ] Incorporate feedback

### DIFF
--- a/cypress/tests/integration/signals-analytics/analytics-report/filtersForm.spec.js
+++ b/cypress/tests/integration/signals-analytics/analytics-report/filtersForm.spec.js
@@ -30,7 +30,7 @@ if (IS_HIBANA_ENABLED) {
       });
     });
 
-    it('clears current filter values when the user swaps between "compare by" values', () => {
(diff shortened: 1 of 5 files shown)


PatrickSison (approved):
UX on this field feels really good with it surfacing error and keeping focus on the field when it needs it

PR #1878, opened 2020-10-30
[AC-1775] Removed the tfa_required condition for starter plans

shrirupa:
[AC-1775] Removed the tfa_required condition for starter plans

### What Changed
 - Removed the hardcoded condition for starter plans
 - tfa_required toggle accurately represents the api value

**Another update**
 - Update the language of the plan change page to say Your new plan no longer allows for single sign-on.  instead of 'Your new plan no longer allows for single sign-on and account-wide requirement of two-factor authentication.'

### How To Test
 - Added a cypress test to check this change.
 - Trying to figure out an account on staging where this can be tested.

### To Do
- [ ] Address feedback

[AC-1775]: https://sparkpost.atlassian.net/browse/AC-1775


cypress/tests/integration/signals-analytics/analytics-report/savedReports.spec.js
```diff
@@ -184,7 +185,10 @@ if (IS_HIBANA_ENABLED) {
         });
 
         cy.findByLabelText('Report').should('have.value', 'My Bounce Report');
-        cy.findByLabelText('Time Zone').should('have.value', '(UTC-04:00) America/New York');
+        cy.findByLabelText('Time Zone').should(
```
shrirupa, line 188: this test was failing because of the day light saving.

gfriedmansp: Nice catch - might be worth a comment in the code in case someone else comes across this

shrirupa:
@gfriedmansp Made the following additions - 
 - Added the comment to explain the day light saving part in cypress test
 - Updated the language for change billing plan page when a user (downgrades from premier to starter) OR (upgrades from free to starter) to say `Your new plan no longer allows for single sign-on.` instead of `Your new plan no longer allows for single sign-on and account-wide requirement of two-factor authentication.` This change was made since the tfa_required will now be present on all the bundles.

gfriedmansp (approved):
All looks good to me now. Can we have the API return a `"tfa_required": true,` for an account in staging when we merge this?

PR #1879, opened 2020-11-02
FE-1218 - Pagination for domain list pages

nstanard:
### What Changed
 - Client side paging for the domains/sending, domains/bounce, and domains/tracking pages
 - react-table useTable instead of our own custom useTable
 - Default sort for trackingDomains wasn't working A - Z - fixed with the use of react-table useTable
 - Rows with dates that were undefined | null are now sorting to the end Newest - Oldest
 - Numbers now fall to the end of A - Z
 - Sorting, Filtering, and Paging interactions all work together
 - Landing the page with url params still applies them to the UI
 - Updating the UI filters still updates the URL

### How To Test
http://localhost:3100/domains/list/sending
http://localhost:3100/domains/list/tracking

1. Sorting, Filtering, Paging all work together
2. URL is synced with applied filters 
3. Filters apply on page load based on URL params

1a. Go to page 2 and filter -> paging should be reset to first page
2a. Default sort on page load should be correct (dates that are undefined/null fall to end of list & numbers fall to the end of the list for [a - z]) 
3. Domain status filter is applied as an && filter. 

### To Do
- [ ] Test
- [ ] Review / Get Team Feedback

### Thoughts for the future
- Decide on a deprecation strategy

- Debounce the domain name input so it's a little smoother (sometimes it lags a little but I admit that might be because of the multiple things happening in the event handler)


src/pages/reports/messageEvents/tests/EventPage.test.js
```diff
@@ -9,18 +9,15 @@ describe('Page: Event tests', () => {
     getSelectedEvent: jest.fn(),
```
nstanard, line 9: prettier

src/pages/sendingDomains/index.js
```diff
@@ -5,5 +5,5 @@ import CreatePage from './CreatePage';
 export default {
```
nstanard, line 5: prettier

src/reducers/summaryTables.js
```diff
@@ -1,30 +1,35 @@
+import { DEFAULT_PER_PAGE, DEFAULT_CURRENT_PAGE } from 'src/constants';
 export const DEFAULT = '@@DEFAULT_TABLE';
 
 const initialState = {
   [DEFAULT]: {
-    currentPage: 1,
+    currentPage: DEFAULT_CURRENT_PAGE,
```
nstanard, line 6: using const defaults (should be same value as was hardcoded) + prettier

src/constants/index.js
```diff
@@ -167,6 +167,9 @@ export const ALL_EVENTS_FILTERS = {
 };
 
 export const DEFAULT_PER_PAGE_BUTTONS = [10, 25, 50, 100];
+export const DEFAULT_PAGE_RANGE = 5;
```
nstanard, line 170: migrated all defaults to this location, and used them in the files I found

src/components/summaryTable/PerPageControl.js
```diff
@@ -1,10 +1,14 @@
+/***
+ * DEPRECATED!
```
nstanard, line 2: Need to talk about an actual deprecation strategy

src/components/collection/Pagination.js
```diff
@@ -1,12 +1,13 @@
 import React from 'react';
 import { Box, Pagination } from 'src/components/matchbox';
 import styles from './Pagination.module.scss';
-import { DEFAULT_PER_PAGE_BUTTONS } from 'src/constants';
+import { DEFAULT_PER_PAGE_BUTTONS, DEFAULT_PAGE_RANGE } from 'src/constants';
```
nstanard, line 4: We have to maintain the existing functionality where it was used, but also allow for these new locations to use the component with pageIndex initially set to zero.

src/components/summaryTable/PerPageControl.js
```diff
@@ -1,10 +1,14 @@
...
+ */
 import React from 'react';
 import classnames from 'classnames';
 import useHibanaToggle from 'src/hooks/useHibanaToggle';
 import { Box, Button } from 'src/components/matchbox';
 import styles from './SummaryTable.module.scss';
+import { DEFAULT_PER_PAGE_BUTTONS } from 'src/constants';
 
-const SIZES = [10, 25, 50, 100];
+const SIZES = DEFAULT_PER_PAGE_BUTTONS;
```
shrirupa, line 11: we can use this directly rather than it to SIZES or just renaming the DEFAULT_PER_PAGE_BUTTONS : SIZES

src/pages/domains/components/SendingDomainsTab.js
```diff
@@ -183,69 +242,125 @@ export default function SendingDomainsTab({ renderBounceOnly = false }) {
...
+
+                  return {
+                    id: i,
+                    value:
+                      filtersState.checkboxes[filtersState.checkboxes.map(i => i.name).indexOf(i)]
+                        .isChecked || false,
+                  };
+                });
+
+                setAllFilters(newFilters);
```
shrirupa, line 269: we should abstract this out into a separate function

src/pages/domains/components/SendingDomainsTab.js
```diff
@@ -183,69 +242,125 @@ export default function SendingDomainsTab({ renderBounceOnly = false }) {
...
+              }}
+            />
+
+            <TableFilters.StatusPopover
+              disabled={listPending}
+              checkboxes={filtersState.checkboxes}
+              onCheckboxChange={e => {
+                filtersDispatch({ type: 'TOGGLE', name: e.target.name });
+
+                const newFilters = Object.keys(filters).map(i => {
```
shrirupa, line 279: `Object.keys(filters)` what is filters here ?

nstanard: Filters here is an object of the checkbox values.

src/pages/domains/components/SendingDomainsTab.js
```diff
@@ -183,69 +230,102 @@ export default function SendingDomainsTab({ renderBounceOnly = false }) {
...
+
+        {isEmpty && <Empty message="There is no data to display" />}
+
+        {!listPending && !isEmpty && <SendingDomainsTable tableInstance={tableInstance} />}
+      </Panel>
+
+      <Pagination
+        data={rows}
+        pageBaseZero={true}
+        currentPage={state.pageIndex}
```
shrirupa, line 315: instead of making the pageBaseZero change to the Pagination component can we make the change here
`currentPage={state.pageIndex + 1}`

src/pages/domains/components/SendingDomainsTab.js
```diff
@@ -183,69 +230,102 @@ export default function SendingDomainsTab({ renderBounceOnly = false }) {
...
+        perPage={state.pageSize}
+        saveCsv={false}
+        onPageChange={page => {
+          // TODO: Try to see if this is firing because of a mistake I made.... otherwise leave condition in place
+          // Only adding this if condition because this keeps firing on load
+          if (state.pageIndex !== page) {
```
shrirupa, line 321: I think once you make the change above https://github.com/SparkPost/2web2ui/pull/1879/files#r517388344 you won't have to add in this condition.

src/pages/domains/components/SendingDomainsTab.js
```diff
@@ -183,69 +230,102 @@ export default function SendingDomainsTab({ renderBounceOnly = false }) {
...
+                  'data-sort-direction': 'desc',
+                },
+              ]}
+              onChange={e => {
+                const { target } = e;
+                const selectedOption = target.options[target.selectedIndex];
+                const selectedAttribute = selectedOption.getAttribute('data-sort-by');
+                const selectedDirection = selectedOption.getAttribute('data-sort-direction');
+                const desc = selectedDirection === 'desc' ? true : false;
+                toggleSortBy(selectedAttribute, desc);
```
shrirupa, line 299: can we abstract this into a separate function ?

src/pages/domains/components/SendingDomainsTab.js
```diff
@@ -183,69 +230,102 @@ export default function SendingDomainsTab({ renderBounceOnly = false }) {
...
+
+            <TableFilters.StatusPopover
+              disabled={listPending}
+              checkboxes={filtersState.checkboxes}
+              onCheckboxChange={e => {
+                filtersStateDispatch({ type: 'TOGGLE', name: e.target.name });
+                const reactTableFilters = getReactTableFilters(filters, filtersState, {
+                  domainName: filtersState.domainName,
+                  targetName: e.target.name,
+                  targetChecked: e.target.checked,
```
shrirupa, line 257: instead of setting this in the onChange function, we can have a useEffect that handles calling getReactTableFilters

src/pages/domains/components/SendingDomainsTab.js
```diff
@@ -183,69 +230,102 @@ export default function SendingDomainsTab({ renderBounceOnly = false }) {
...
+            <TableFilters.DomainField
+              disabled={listPending}
+              value={filtersState.domainName}
+              onChange={e => {
+                filtersStateDispatch({ type: 'DOMAIN_FILTER_CHANGE', value: e.target.value });
+                const reactTableFilters = getReactTableFilters(filters, filtersState, {
+                  domainName: e.target.value,
+                });
+                setAllFilters(reactTableFilters);
+              }}
```
shrirupa, line 246: same as https://github.com/SparkPost/2web2ui/pull/1879/files#r517413718

shrirupa:
created a PR to this with above suggested changes https://github.com/SparkPost/2web2ui/pull/1881/files

src/pages/domains/components/getReactTableFilters.js
```diff
@@ -0,0 +1,28 @@
+export default (
+  filters,
+  filtersState,
```
shrirupa, line 3: this won't be needed if we call setAllFilters in useEffect instead of onChange

src/pages/domains/components/TrackingDomainsTab.js
```diff
@@ -1,14 +1,19 @@
+import _ from 'lodash';
```
shrirupa, line 1: same suggestions as for `SendingDomainsTab.js`

src/pages/domains/components/TrackingDomainsTab.js
```diff
@@ -130,55 +166,76 @@ export default function TrackingDomainsTab() {
...
+      <Pagination
+        data={rows}
+        pageBaseZero={true}
+        currentPage={state.pageIndex}
+        perPage={state.pageSize}
+        saveCsv={false}
+        onPageChange={page => {
+          // TODO: Try to see if this is firing because of a mistake I made.... otherwise leave condition in place
+          // Only adding this if condition because this keeps firing on load
+          if (state.pageIndex !== page) {
```
shrirupa, line 231: I missed removing this condition here.

src/components/summaryTable/PerPageControl.js
```diff
@@ -1,21 +1,23 @@
+/***
+ * DEPRECATED! Use src/components/collection/Pagination.js instead (also update this when a deprecation plan is in place)
```
nicklemmon, line 2: You could even `console.warn()` when this component is used - I think that's what Matchbox does.

src/pages/domains/components/SendingDomainsTab.js
```diff
@@ -1,14 +1,18 @@
-import React, { useEffect, useReducer, useState } from 'react';
+import React, { useEffect, useReducer } from 'react';
+import { useFilters, usePagination, useSortBy, useTable } from 'react-table'; // https://react-table.tanstack.com/docs/api/overview
```
nicklemmon, line 2: Don't think linking to the docs is necessary here - ideally a dev would be able to look up the module themselves from the name?

src/pages/domains/helpers/react-table-filters.js
```diff
@@ -0,0 +1,14 @@
...
+      if (!x[1]) {
+        return null;
+      }
+
+      return {
+        id: x[0],
+        value: x[1],
+      };
+    })
+    .filter(Boolean);
```
nicklemmon, line 13: Not sure what this is doing...? Might be better to name the variables or add some JSDoc-like comments.

nicklemmon: Also, our convention for filenaming would make this file: `getReactTableFilters.js` instead of the hyphenation here

src/components/summaryTable/PerPageControl.js
nstanard: I was going to handle this in a follow on but.... you thinking this is enough?

image

src/pages/domains/components/SendingDomainsTab.js
nstanard: threw it out for the chat with everybody - removing

shrirupa (approved):
Looks good to me 🚀

PR #1885, opened 2020-11-05
SA-1820 - (UI) "Per precision" on unique metrics acting weird

nicklemmon:
[SA-1820](https://sparkpost.atlassian.net/browse/SA-1820)

### What Changed
- In the new report builder feature, renamed `Legend` -> `ActiveMetrics` to be more consistent
- Semantically rendered a heading instead of a div for that part of the page
- Renamed a Cypress test file to match convention
- Properly grabbing `reportOptions` from context rather than the Redux Store <-- This resolved the bug

### How To Test
- Go to `/signals/analytics` with HIbana enabled
- Click on "Add Metrics"
- Check "Unique Clicks" and "Unique Confirmed Opens" and apply those metrics
- Verify the precision content renders within the tags

### To Do
- [ ] Incorporate feedback


nstanard (approved):
LGTM!

Did we want to tag a jira ticket number on the todo comment?

nicklemmon:
> LGTM!
> 
> Did we want to tag a jira ticket number on the todo comment?

Good idea - I picked up the ticket that will remove it so I'll remember. If I weren't, I would definitely say we should call it out somewhere in Jira explicitly.

PR #1886, opened 2020-11-05
FE-1226 - Domain list page - status filter "or" logic

nstanard:
### What Changed
 - Reversed the UI logic for domain status filtering

1. **The page url** "http://localhost:3100/domains/list/sending" **automatically selects and updates the url on page load** - "http://localhost:3100/domains/list/sending?readyForSending=true&readyForDKIM=true&readyForBounce=true&validSPF=true&unverified=true&blocked=true"
image

**2. As the user un-checks the status options the URL is updated along with the table.**
image

image

**3. Refreshing the page causes the same filter state to load.**

image

(Sorting and paging are not persistent in the URL ... yet...?)

### How To Test
 - Just have to load up the domain list pages and try to break the filtering/sorting/paging in any way possible. 
http://localhost:3100/domains/list/sending
http://localhost:3100/domains/list/bounce
http://localhost:3100/domains/list/tracking

### To Do
- [x] Fix tests (unit/integration)
- [ ] Ask UX the question at the end of the description.
- [ ] Feedback / Review

Question for UX:
image

**Should the bounce tab have all the same filter options as the sending tab?**

### DIFF
--- a/src/pages/domains/components/SendingDomainsTab.js
+++ b/src/pages/domains/components/SendingDomainsTab.js
@@ -1,4 +1,5 @@
-import React, { useEffect, useReducer } from 'react';
+/* eslint-disable no-unused-vars */
+import React, { useEffect, useRef, useReducer } from 'react';
 import { useFilters, usePagination, useSortBy, useTable } from 'react-table';
 import { ApiErrorBanner, Empty, Loading } from 'src/components';
 import { Pagination } from 'src/components/collection';
@@ -9,7 +10,12 @@ import { API_ERROR_MESSAGE } from '../constants';
 import useDomains from '../hooks/useDomains';
 import SendingDomainsTable from './SendingDomainsTable';
 import TableFilters, { reducer as tableFiltersReducer } from './TableFilters';
-import getReactTableFilters from '../helpers/getReactTableFilters';
+import {
+  getReactTableFilters,
+  customDomainStatusFilter,
+  getActiveStatusFilters,
+  filterStateToParams,
+} from '../helpers';
 
 const filtersInitialState = {
   domainName: '',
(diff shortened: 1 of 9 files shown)


src/pages/domains/helpers/getReactTableFilters.js
```diff
@@ -10,11 +10,18 @@ export default params => {
     .map(keyValueArr => {
       // keyValueArr[0] = params key
       // keyValueArr[1] = params value
-      if (!keyValueArr[1]) {
+      if (keyValueArr[1] === true) {
```
bkemper, line 13: Instead of the comment, deconstruct the args.

```
.map(([id, value]) => {
```

src/pages/domains/components/SendingDomainsTab.js
```diff
@@ -104,23 +109,53 @@ export default function SendingDomainsTab({ renderBounceOnly = false }) {
 
   const domains = renderBounceOnly ? bounceDomains : sendingDomains;
 
+  const filter = React.useMemo(() => customDomainStatusFilter, []);
```
bkemper, line 112: This is unnecessary.

src/pages/domains/helpers/customDomainStatusFilter.js
```diff
@@ -0,0 +1,11 @@
+const customDomainStatusFilter = function(rows, columnIds, value) {
```
bkemper, line 1: I'm not a huge fan of helpers since we React build of reusable building blocks (i.e. components, hooks, etc.)

bkemper (approved):
LGTM!

src/pages/domains/components/SendingDomainsTab.js
nstanard: Mainly leaving since docs says it's required. (although it does function without)

src/pages/domains/helpers/customDomainStatusFilter.js
nstanard: I wanna keep talking about this and come up with a place to keep these that's closer to the reducer for this area.

PR #1888, opened 2020-11-06
FE-1209 - [Dashboard] Add sending domain onboarding

nstanard:
Epic: https://sparkpost.atlassian.net/browse/FE-1185
Task: https://sparkpost.atlassian.net/browse/FE-1209

### What Changed
 - The current dashboardV2 onboarding panel is being broken up into individual steps the user will walk through depending on the state of their account/data. 

**Before:**
image

**After:** 
image

Step one design:
https://sparkpost.invisionapp.com/console/Progressive-Onboarding-ckfe1rusj0cqn0185gnro5f8c/ckfe1s8be019p010e2zwj0wc6/play

Step four design:
https://sparkpost.invisionapp.com/console/Progressive-Onboarding-ckfe1rusj0cqn0185gnro5f8c/ckfe1s8bf019s010e68fy5lhe/play

### How To Test
 - Load up an admin or dev user (preferably both for good test coverage) and make sure it's a fresh account without a sending domain. Give the account the feature flag for `allow_domains_v2` and navigate to the dashboardV2 and you should see the first onboarding step for adding a sending domain. 

### To Do
- [x] Swap out the `<Picture />` component and use the correct image/media asset
- [x] Correct the conditions for the add sending domain boolean (onboarding step)
- [x] Add Cypress test
- [x] Split off work into individual PRs
- [x] Incorporate Feedback


src/pages/dashboardV2/DashboardPageV2.container.js
```diff
@@ -18,6 +18,21 @@ import DashboardPageV2 from './DashboardPageV2';
 
 function mapStateToProps(state) {
   const isPending = state.account.loading || state.account.usageLoading || state.alerts.listPending;
+  const isAdminOrDev = isAdmin(state) || hasRole(ROLES.DEVELOPER)(state);
```
nstanard, line 21: It was discussed that we should first check the `/api/v1/usage -> results.messaging.last_usage_date` value first before checking the number of sending domains they have configured because calling to get the sending domains list can be an expensive and long call to make.

src/pages/dashboardV2/DashboardPageV2.js
```diff
@@ -5,6 +5,7 @@ import ConfigurationImg from '@sparkpost/matchbox-media/images/Configuration@sma
 import { Loading, Picture } from 'src/components';
```
nstanard, line 5: This Picture component needs swapped out for the Matchbox one - but thats blocked until we finish https://sparkpost.atlassian.net/browse/FE-1224

src/pages/dashboardV2/DashboardPageV2.js
```diff
@@ -61,57 +116,91 @@ export default function DashboardPageV2() {
...
-                      </Picture>
-                    </Box>
-                  </Box>
+              {isAdminOrDev && (
+                <Dashboard.Panel>
+                  <ScreenReaderOnly>
+                    <Heading as="h3">Next Steps</Heading>
+                  </ScreenReaderOnly>
 
-                  <Column>
```
PatrickSison, line 86: Is this old design going away completely?

src/pages/dashboardV2/DashboardPageV2.js
```diff
@@ -61,57 +116,91 @@ export default function DashboardPageV2() {
...
-                      <Dashboard.Shortcut to="/signals/analytics">
-                        Analyze your Data
-                      </Dashboard.Shortcut>
-
-                      <Dashboard.Shortcut to="/alerts/create">Create an Alert</Dashboard.Shortcut>
-                    </Box>
-                  </Column>
-                </Columns>
-              </Dashboard.Panel>
+                  {addSendingDomainOnboarding && (
```
PatrickSison, line 125: Is the plan for there to only be 1 setup step at a time? If so, would we want this to render off of some enumerable state (derived or not)

src/pages/dashboardV2/DashboardPageV2.js
```diff
@@ -14,35 +17,87 @@ import {
...
+    listSendingDomains,
   } = useDashboardContext();
 
+  // TODO: useReducer instead
+  const [hasSetupDocumentationPanel, setHasSetupDocumentationPanel] = useState();
+  const [addSendingDomainOnboarding, setAddSendingDomainOnboarding] = useState();
+  const [lastUsageDate, setlastUsageDate] = useState(-1);
+
+  // TODO: useReducer instead
+  function setupState(lastUsageDate, sendingDomains) {
```
PatrickSison, line 51: I'm guessing the main issue is that we want to avoid touching the state until all the necessary data is ready. Besides changing to useReducer (which is already called out the comment) and swapping to an enumerable state, we could also use use a useEffect to check weather to run the init state.

```javascript
useEffect(() => {
  if (!dep1.isReady || !dep2.isReady) { //whatever dependency readiness check
     return;
  }
  setInitState(); //All variables ready, do initial state calculation
}, [dep1, dep2]);
```
This detaches it from the promise at least, just in case there end up being more dependencies to wait on before finding initial state, and lets them run in parallel before running it. Not a necessary change, but worth a thought

PatrickSison (commented):
Few comments. Looks really good though 👍

src/pages/dashboardV2/DashboardPageV2.js
nstanard: Yea, 1 setup setup step at a time is the direction we're going here. I was still trying to decide how I wanted to control the display conditions but I like where you're going with this.

src/pages/dashboardV2/DashboardPageV2.js
nstanard: I believe so - I wasn't instructed to a/b test or anything. Just swap out for this new set of states.

cypress/tests/integration/dashboardV2/dashboardPageV2.spec.js
```diff
@@ -24,27 +79,15 @@ describe('Version 2 of the dashboard page', () => {
...
-        content: 'Create an Alert',
-        href: '/alerts/create',
+        content: 'Getting Started Documentation',
+        href: LINKS.ONBOARDING_SENDING_EMAIL,
       });
     });
 
-    it('does not render certain links when grants are not available for a user', () => {
+    // TODO: Fix, something is wrong here
+    it.skip('does not render certain links when grants are not available for a user', () => {
```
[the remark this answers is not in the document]
nstanard: I'll keep this in mind. One main reason this test was failing is because we've gone from the one panel with the four links to the individual steps with graphics/verbiage. I'll have to revisit this test one one of the future tasks - I will tag this comment with the jira ticket number I expect to do that on.

src/pages/dashboardV2/DashboardPageV2.js
nstanard: Yea that's a good point. I'm so used to chaining promise events and initing with the response.

cypress/tests/integration/dashboardV2/dashboardPageV2.spec.js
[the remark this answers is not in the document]
nstanard: Fixed on: https://github.com/SparkPost/2web2ui/compare/FE-1211...FE-1212#diff-8c258619eb8697c507b6530c3d2b825a3243e7a5c9ad073fd289dcd141ba3129R39

src/pages/dashboardV2/DashboardPageV2.js
nstanard: Resolving on FE-1214

PatrickSison (approved):

PR #1892, opened 2020-11-09
SA-1773 - (UI) Release Advanced Comparators

nicklemmon:
[SA-1773](https://sparkpost.atlassian.net/browse/SA-1773)

### What Changed
- Removes references to the `"allow_report_filters_v2"` account UI option
- Removes the old state reducer for handling filters
- Removes the old UI for handling filters
- Updates Cypress tests, removing tests for old filter UI and removing references to fixture files with feature flags
- Centralized use of the `dehydrateFilters` helper function

### How To Test
- Verify adding/removing filters works as expected without any particular feature flag for report builder
- Verify filters can be removed using the tags rendered on the page
- Verify old filters applied via the `filters=` query param are still re-mapped to `query_filters=` 

### To Do
- [ ] Incorporate feedback


src/pages/reportBuilder/ReportBuilder.js
```diff
@@ -100,32 +93,42 @@ export function ReportBuilder({
...
     }
 
     // If report is found from ID, consolidates reportOptions from URL and report
     if (report) {
-      const { filters: reportFilters = [], ...reportOptions } = parseSearch(report.query_string);
-      setReport(report);
-      refreshReportOptions({ ...reportOptions, filters: [...reportFilters, ...urlFilters] });
+      const {
+        filters: reportFiltersV1 = [],
+        queryFilters: reportFiltersV2 = [],
```
nicklemmon, line 121: There has to be a cleaner way to do this!

src/pages/reportBuilder/components/ActiveFilters.js
```diff
@@ -43,7 +15,7 @@ export function ActiveFiltersV2({ filters, handleFilterRemove }) {
       <Inline space="200">
         {groupings.map((grouping, groupingIndex) => {
           return (
-            <div data-id="active-filter-group">
+            <div data-id="active-filter-group" key={`active-filter-tag-${groupingIndex}`}>
```
nicklemmon, line 18: Was missing a `key`! Must have missed the warning previously.

src/pages/reportBuilder/components/FiltersForm/FiltersForm.js
```diff
@@ -25,7 +25,8 @@ import {
 } from 'src/components/matchbox';
 import { Uppercase, Comparison } from 'src/components/text';
 import { getGroupingFields, getApiFormattedGroupings } from '../../helpers';
-import Typeahead from '../TypeaheadV2';
+import { useReportBuilderContext } from '../../context/ReportBuilderContext';
```
nicklemmon, line 28: No need to accept this state as a prop when it can be grabbed from context.

src/pages/reportBuilder/helpers/index.js
```diff
@@ -261,19 +261,24 @@ export function hydrateFilters(groupings, { subaccounts } = {}) {
...
+                filterRet[comparison] = filterObj[comparison].map(rawValue => {
+                  const value = typeof rawValue === 'object' ? rawValue.value : rawValue;
                   const subaccount =
                     subaccounts.find(subaccount => subaccount.id === Number.parseInt(value)) || {};
-
                   const { name, id } = subaccount;
+
                   return {
-                    value: name ? `${name} (ID ${id})` : `Subaccount ${value}`,
+                    value: name ? `${name} (ID ${id})` : value,
```
PatrickSison, line 271: Would we not want the subaccount name to fallback onto the format of `Subaccount (ID)`? Or is this an extension of remapping and having the ID stored in a different variable?

PatrickSison (commented):
Most important change is the removal of the extra dehydrateFilters. It also seems like the subaccount filters aren't properly dehydrating? I keep seeing the full subaccount being passed through instead of just the subaccount ID, which ends up messing up the requests

src/pages/reportBuilder/helpers/index.js
nicklemmon: yeah it appears to be structured correctly just as a raw value - if you look at the app it looks like it's still rendering the content as it was. Is that not what you are seeing?

nicklemmon: hmm, but now that structure is being passed to the actual request - which is obviously causing problems. 🤔

PatrickSison (approved):
Looks good now I think. We got some test coverage of old filters to new filters, and tried breaking as much as I could. Seems good 👍

cypress/tests/integration/accounts/account-settings/accountSettingsPage.spec.js
```diff
@@ -53,36 +53,41 @@ describe('Account Settings Page', () => {
...
+            fixture: 'api-keys/200.post.json',
+            requestAlias: 'scimTokenCreate',
+          });
+          cy.stubRequest({
+            url: 'api/v1/api-keys?grant=scim/manage',
+            fixture: 'api-keys/200.get.scim-token-newtoken.json',
+            requestAlias: 'newScimTokenGet',
+          });
+
+          cy.findByRole('button', { name: 'Generate SCIM Token' }).click();
```
nicklemmon, line 77: Mostly just spacing changes, but I also used a more specific query in case that helped things

cypress/tests/integration/accounts/account-settings/accountSettingsPage.spec.js
```diff
@@ -53,36 +53,41 @@ describe('Account Settings Page', () => {
...
-              cy.findByText('Make sure to copy your SCIM token now.').should('be.visible');
-              cy.findByText('Continue').click();
-            });
-            cy.findByText('123f••••••••').should('be.visible');
-            cy.findByText('Delete Token').should('be.visible');
-          },
-        );
+
+        it('opens Generate SCIM token Modal when a token is not present and clicking on Continue dismisses the Modal and new token can be found', () => {
+          cy.wait('@oldScimTokenGet');
```
nicklemmon, line 64: This was the actual problem here - surprised it wasn't flaky earlier

src/config/routes/signals.js
```diff
@@ -244,7 +244,7 @@ const signalsRoutes = [
     component: signals.IntegrationPage,
     condition: hasAccountOptionEnabled('allow_events_ingest'),
     layout: App,
-    title: 'Signals Integration | SparkPost',
+    title: 'Signals Integration',
```
nicklemmon, line 247: This was a bug I ran in to - it was rendering the `<title>` as `Signals Integration | SparkPost | SparkPost`

PR #1894, opened 2020-11-10
SA-1847: Update form and tests

kchu1337:
### What Changed
 - Added timezone and schedule_type as request params.
 - Changed logic for formatting form values due to new cron format.

### How To Test
 - Make sure `allow_scheduled_reports` account ui flag is enabled.
 - Go to the create scheduled page. EX: /signals/schedule/a601ccd1-49da-4c8f-bad1-e16e77302fbc. OR from the list report modal in report builder and select schedule in the action list. 
 - Check that timezone and schedule_type is added correctly.
 - Check that cronjob formatting works correctly.


src/pages/reportBuilder/components/ScheduledReportForm.js
```diff
@@ -21,36 +21,47 @@ import { listUsers } from 'src/actions/users';
...
 
 const hasAtLeastOneRecipient = recipientList => recipientList.length > 0;
 
 export const formatFormValues = formValues => {
-  const { name, description = 'NA', subject, period, ...rest } = formValues;
+  const { name, subject, period, timing, timezone, ...rest } = formValues;
   const recipients = rest.recipients.map(({ username }) => username);
   const [hour, minute] = rest.time.split(':');
-  const day_of_week = rest.day || '*';
+  const day_of_week = timing === 'monthly' ? `${rest.day}${rest.week}` : rest.day || '*';
```
kchu1337, line 48: Monthly timing will be formatted like fri#2 for second friday of the month. Or monl as the last monday of the month. day is set to * when the day field is disabled for daily scheduled reports.

src/pages/reportBuilder/components/ScheduledReportForm.js
```diff
@@ -21,36 +21,47 @@ import { listUsers } from 'src/actions/users';
...
   const schedule = {
     day_of_week,
     month: '*',
-    day_of_month: '*',
+    day_of_month: '?',
```
kchu1337, line 52: Cron changes needed ? instead of *.

src/pages/reportBuilder/components/ScheduledReportForm.js
```diff
@@ -21,36 +21,47 @@ import { listUsers } from 'src/actions/users';
...
+  { label: 'Monday', value: 'mon' },
+  { label: 'Tuesday', value: 'tue' },
+  { label: 'Wednesday', value: 'wed' },
+  { label: 'Thursday', value: 'thu' },
+  { label: 'Friday', value: 'fri' },
+  { label: 'Saturday', value: 'sun' },
+];
+
+const WEEK_OPTIONS = [
+  { label: `First`, value: '#1' },
```
nicklemmon, line 34: Don't think you meant to use backticks here?

src/pages/reportBuilder/components/ScheduledReportForm.js
nicklemmon: This comment clarifies things - would you mind adding it to the code rather than just here? I think it would come in handy for future maintainers.

nicklemmon (changes requested):
@kch Overall looks pretty good! I did notice an issue that seems to be unrelated to this ticket - when I'm searching in the "Send To" field and get no results, the "No results" message renders then quickly disappears.

Screen Shot 2020-11-12 at 1 46 50 PM

Other than that, the feedback I have is very minor.

src/pages/reportBuilder/components/ScheduledReportForm.js
nicklemmon: Very helpful, thanks for doing that.

src/components/typeahead/ComboBoxTypeahead.js
```diff
@@ -98,11 +98,9 @@ export const ComboBoxTypeahead = ({
           item,
         }),
       );
-    const isMenuOpen = isOpen && Boolean(items.length);
-
     const inputProps = getInputProps({
       disabled,
-      error: error && !isMenuOpen ? error : undefined,
+      error: error && !isOpen ? error : undefined,
```
nicklemmon, line 103: Screen Shot 2020-11-13 at 9 21 09 AM

Hmm, looks like this state shouldn't be possible - there should be some user entry required before the empty state renders?

nicklemmon: To reproduce:
1. Sign in with the usual account in prod
2. Go to `/alerts/edit/71570`
3. Focus on the "Subaccounts" field and "No Results" will render before typing anything

nicklemmon (commented):
@kchu1337 Take a look at my comments - I think I found a bug with the Typeahead changes?

nicklemmon (approved):
@kchu1337 Looks good to me. Thanks.

PR #1897, opened 2020-11-12
FE-1234 - Last few cleanup items on domains page

shrirupa:
FE-1234 - Last few cleanup items on domains page

### What Changed

- Visually hide the Apply button
- Added Select all checkbox for domain status filter
- Input placeholder text for domain name
- Sending domain tab  - Rename Sending -> Verified in Domain status filter
- Bounce Domain tab - Remove Bounce from the list
- Tracking domain tab  - Rename Tracking -> Verified in Domain status filter

### How To Test
 - testing if all the above changes look good 
mock - https://sparkpost.invisionapp.com/console/share/CF19LDYPB2/478910828

### To Do
- [ ] Addressing feedback

### DIFF
--- a/cypress/tests/integration/domains/listPage.spec.js
+++ b/cypress/tests/integration/domains/listPage.spec.js
@@ -480,6 +480,32 @@ describe('The domains list page', () => {
         });
       });
 
+      it('all domain status are checked if select all is checked and select all is not checked if even one of checkbox is unchecked', () => {
+        stubSendingDomains({ fixture: 'sending-domains/200.get.multiple-results.json' });
+        stubSubaccounts();
+        cy.visit(PAGE_URL);
+        cy.wait(['@sendingDomainsReq', '@subaccountsReq']);
+
+        cy.findByRole('button', { name: 'Domain Status' }).click();
+
+        cy.findByLabelText('Verified').uncheck({ force: true });
+        cy.findByLabelText('Unverified').uncheck({ force: true });
+        cy.findByLabelText('Blocked').uncheck({ force: true });
+        cy.findByLabelText('SPF Valid').uncheck({ force: true });
+        cy.findByLabelText('Bounce').uncheck({ force: true });
+        cy.findByLabelText('DKIM Signing').uncheck({ force: true });
+        cy.findByLabelText('Select All').check({ force: true });
+        cy.findByLabelText('Verified').should('be.checked');
+        cy.findByLabelText('Unverified').should('be.checked');
+        cy.findByLabelText('Blocked').should('be.checked');
+        cy.findByLabelText('SPF Valid').should('be.checked');
+        cy.findByLabelText('Bounce').should('be.checked');
+        cy.findByLabelText('DKIM Signing').should('be.checked');
+
+        cy.findByLabelText('DKIM Signing').uncheck({ force: true });
+        cy.findByLabelText('Select All').should('not.be.checked');
+      });
+
       it('filters by domain status', () => {
         stubSendingDomains({ fixture: 'sending-domains/200.get.multiple-results.json' });
         stubSubaccounts();
@@ -487,6 +513,13 @@ describe('The domains list page', () => {
         cy.wait(['@sendingDomainsReq', '@subaccountsReq']);
 
         cy.findByRole('button', { name: 'Domain Status' }).click();
+
+        cy.findByLabelText('Verified').should('be.visible');
+        cy.findByLabelText('Unverified').should('be.visible');
+        cy.findByLabelText('Blocked').should('be.visible');
+        cy.findByLabelText('SPF Valid').should('be.visible');
+        cy.findByLabelText('Bounce').should('be.visible');
+        cy.findByLabelText('DKIM Signing').should('be.visible');
 
         verifyTableRow({
           rowIndex: 1,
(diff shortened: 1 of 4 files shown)


src/pages/domains/components/TableFilters.js
```diff
@@ -145,28 +187,45 @@ function StatusPopover({ checkboxes, onCheckboxChange, disabled }) {
...
 
         <Divider />
 
-        <Box padding="300" display="flex" justifyContent="flex-end">
+        <VisuallyHiddenBox padding="300" display="flex" justifyContent="flex-end">
           <Button variant="primary" size="small" onClick={() => setIsPopoverOpen(false)}>
             Apply
           </Button>
-        </Box>
+        </VisuallyHiddenBox>
```
nicklemmon, line 228: Hmm, the purpose of the button is that it is actually there visually - screen reader users can always close the popover with the escape key. When it's wrapped in `<ScreenReaderOnly>` I believe it is removed from the tab order anyways - and even if it weren't, that would break the experience for users who are _not_ blind but use the keyboard to navigate pages. If UX wants this design, it would be best to remove the button and have some directions at the top of the checklist wrapped in `<ScreenReaderOnly>`. For example:

```
<ScreenReaderOnly as="p">
  Checkboxes filter the table. When checked, table elements are visible, when unchecked they are hidden from the table.
</ScreenReaderOnly>
```

I'm not sure if that's actually the behavior of the filtering?

This is actually why I advocated for just having the `<button/>` as it removes some of the complexity, and makes maintenance easier as you don't have to think about updating that content if the filter behavior changes.

src/pages/domains/components/TableFilters.js
```diff
@@ -145,28 +187,45 @@ function StatusPopover({ checkboxes, onCheckboxChange, disabled }) {
...
-                checked={filter.isChecked}
-              />
-            ))}
+            {checkboxes
+              .filter(filter => filter.name !== 'selectAll')
+              .filter(checkbox => {
+                if (domainType === 'bounce') return checkbox.name !== 'readyForBounce';
+                return true;
+              })
+              .map((filter, index) => (
```
nicklemmon, line 209: Probably want to have some testing checking the different types of checkboxes depending on the selected tab, right? Just seems like something that could be accidentally broken in the future.

src/pages/domains/components/TableFilters.js
```diff
@@ -145,28 +187,45 @@ function StatusPopover({ checkboxes, onCheckboxChange, disabled }) {
           </Button>
         }
       >
+        <Box padding="300">
+          <Checkbox
+            key="select-all"
```
nicklemmon, line 192: I don't think this needs a `key`? You may have a reason for having one here.

src/pages/domains/components/TableFilters.js
```diff
@@ -33,6 +52,19 @@ export function reducer(state, action) {
     case 'TOGGLE': {
       const isChecked = state.checkboxes.find(filter => filter.name === action.name).isChecked;
 
+      if (action.name === 'selectAll' && !isChecked) {
```
nicklemmon, line 55: Should have some testing around this as well - could maybe even add it to an existing test

src/pages/domains/components/TableFilters.js
```diff
@@ -21,6 +21,25 @@ const Chevron = styled(ChevronRight)`
...
+
+  white-space: nowrap;
+
+  clip: rect(0 0 0 0);
+  clip-path: inset(
+    100%
+  ); /* Added both since clip is deprecated now, but clip path isn't widely supported yet*/
+  overflow: hidden;
+`;
+
```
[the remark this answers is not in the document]
nicklemmon: Actually, read my later comments - I don't think this is necessary either way. 👇

nicklemmon (changes requested):
@shrirupa - it works well, though I have a few concerns:

1. I think we need some new testing for some of the changes here
2. Making the "Apply" button visually hidden actually defeats the purpose if it being there - I left some notes about a possible alternative, but I don't think it's as good as the existing design.

src/pages/domains/components/TableFilters.js
shrirupa: Thanks for explaining this! I will go ahead and add `ScreenReaderOnly` text then and remove the Apply button.

src/pages/domains/components/TableFilters.js
```diff
@@ -146,26 +169,38 @@ function StatusPopover({ checkboxes, onCheckboxChange, disabled }) {
...
-                name={filter.name}
-                onChange={onCheckboxChange}
-                checked={filter.isChecked}
-              />
-            ))}
-          </Checkbox.Group>
+          <ScreenReaderOnly as="p">
+            Checkboxes filter the table. When checked, table elements are visible, when unchecked
+            they are hidden from the table.
+          </ScreenReaderOnly>
```
nicklemmon, line 175: This is great! 👍

cypress/tests/integration/domains/listPage.spec.js
```diff
@@ -480,6 +480,32 @@ describe('The domains list page', () => {
...
+        cy.findByLabelText('Verified').should('be.checked');
+        cy.findByLabelText('Unverified').should('be.checked');
+        cy.findByLabelText('Blocked').should('be.checked');
+        cy.findByLabelText('SPF Valid').should('be.checked');
+        cy.findByLabelText('Bounce').should('be.checked');
+        cy.findByLabelText('DKIM Signing').should('be.checked');
+
+        cy.findByLabelText('DKIM Signing').uncheck({ force: true });
+        cy.findByLabelText('Select All').should('not.be.checked');
+      });
```
nicklemmon, line 507: 👍 Nice.

nicklemmon (approved):
@shrirupa Looks good to me. Let's get it out.

PR #1900, opened 2020-11-16
[FE-1237] - The domains page should be visible to user with either tracking_domains/manage or sending_domains/manage grant

shrirupa:
[FE-1237] - The domains page should be visible to user with either tracking_domains/manage or sending_domains/manage grant

### What Changed
 - Added the condition with grant to domain pages
 - Hide the domains page if the user doesn't have either tracking_domains/manage or sending_domains/manage
 - Hide the sending and bounce domains tabs from user who do not have sending_domains/manage grant
 - Hide the tracking domains tab from user who do not have tracking_domains/manage grant

### How To Test
 - Domains page shouldn't be visible to Templates, Reporting and Subaccount Reporting user.

### To Do
- [ ] Address feedback

[FE-1237]: https://sparkpost.atlassian.net/browse/FE-1237


cypress/tests/integration/domains/listPage.spec.js
```diff
@@ -42,19 +42,24 @@ describe('The domains list page', () => {
       cy.wait('@accountDomainsReq');
 
       // Going right to left since the first tab is already active!
-      cy.findByRole('tab', { name: 'Tracking Domains' })
-        .click({ force: true })
-        .should('have.attr', 'aria-selected', 'true');
+      cy.findByRole('tab', { name: 'Tracking Domains' }).click({ force: true });
```
shrirupa, line 45: for some reason chaining was causing the issue locally.

nicklemmon (changes requested):
@shrirupa I have a few comments I left for you.

One more thing - should the tabs on `src/pages/domains/ListPage` render conditionally as well? Right now _technically_ a user could see the tab for "Tracking Domains" but only have the grand for sending domains.

src/config/navItems/index.js
```diff
@@ -198,6 +199,7 @@ const subaccounts = {
 const domains = {
   label: 'Domains',
   to: '/domains',
+  condition: any(hasGrants('sending_domains/manage'), hasGrants('tracking_domains/manage')),
```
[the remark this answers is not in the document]
shrirupa: ok, as we discussed I am adding, `all(hasGrants('tracking_domains/manage'), hasGrants('sending_domains/manage')),` to routes for ListPage and CreatePage.

shrirupa:
@nicklemmon this is ready for another pass

src/config/routes/index.js
```diff
@@ -827,6 +748,87 @@ const sendingAndTrackingDomainsRoutes = [
...
+    condition: hasGrants('sending_domains/manage'),
+    layout: App,
+    title: 'Domain Details | Domains',
+    category: 'Configuration',
+    subcategory: 'Domains',
+  },
+  {
+    path: '/domains/details/tracking/:id',
+    component: domains.DetailsPage,
+    condition: hasGrants('tracking_domains/manage'),
```
nicklemmon, line 782: I think you would want to use the same condition for all these pages:

```js
all(hasGrants('tracking_domains/manage'), hasGrants('sending_domains/manage'))
```

I know that might seem weird on its face, but I'm imagining a user going to the list page, clicking on a tracking domain, and then getting an error.

cypress/tests/integration/hibana/hibanaNavigation.spec.js
```diff
@@ -496,6 +503,9 @@ if (IS_HIBANA_ENABLED) {
           cy.verifyLink({ content: 'API Docs', href: 'https://developers.sparkpost.com/api' });
           cy.findByText('Log Out').should('be.visible');
         });
+        cy.get(desktopNavSelector).within(() => {
+          cy.verifyLink({ content: 'Configuration', href: '/domains' });
+        });
```
nicklemmon, line 508: All of these assertions are actually nested within a `describe('account settings dropdown', () => {})` so this isn't the right place for them.

It actually looks like our Hibana navigation tests are missing role-based assertions, whereas our OG theme tests do have those assertions. I know it's expanding the scope of the ticket, but would you mind adding some? Should just basically involve creating users with those roles and then writing tests to assert against what you see when signed in as those users.

nicklemmon (changes requested):
@shrirupa I left a couple of comments - looks like you found a gap in our testing!

cypress/tests/integration/hibana/hibanaNavigation.spec.js
shrirupa: I meant to paste it inside the that }); 😄  . Thanks for catching that!

nicklemmon (approved):
@shrirupa Looks good to me!

PR #1901, opened 2020-11-18
SA-1617: Charts localize request

PatrickSison:
### What Changed
 - Charts will separate based on number of comparison filters
 - Charts and tabs panel separated from the report options panel

### How To Test
 - Enable account UI option `allow_compare_by`
 - Go to /signals/analytics
 - Click on `Add Comparison` and  fill out form w/ at least 2 options
 - Verify that charts render multiple
 - Reopen form and clear filters
 - Verify charts render w/o label

### To Do
- [ ] Address feedback

### Notes
 A lot of this branch is rebased on SA-1773. Will be smaller once merged


src/pages/reportBuilder/components/Charts.js
```diff
@@ -1,37 +1,122 @@
...
+
+export function Charts(props) {
+  const { reportOptions } = props;
+  const { compare, metrics } = reportOptions;
+
+  const formattedMetrics = useMemo(() => {
+    return getMetricsFromKeys(metrics, true);
+  }, [metrics]);
+  const formattedOptions = useMemo(() => {
+    return getQueryFromOptions({ ...reportOptions, metrics: formattedMetrics });
```
PatrickSison, line 76: Maybe `getMetricsFromKeys` can get baked into `getQueryFromOptions` to avoid having to pull it in twice

src/pages/reportBuilder/components/Charts.js
```diff
@@ -1,37 +1,122 @@
...
+            </Box>
+          </Panel.Section>
+        );
+      })}
+    </>
+  );
+}
+
+export function Charts(props) {
+  const { reportOptions } = props;
```
PatrickSison, line 69: Need to verify if we expect hydrated or dehydrated metrics. If we expect hydrated, might be harder to integrate with other pages.

src/pages/reportBuilder/components/Charts.js
```diff
@@ -1,37 +1,122 @@
...
-    metrics: metrics.filter(metric => metric.unit === unit),
+  const charts = getUniqueUnits(formattedMetrics).map(unit => ({
+    metrics: formattedMetrics.filter(metric => metric.unit === unit),
     ...METRICS_UNIT_CONFIG[unit],
   }));
-
   let height = 150;
-  switch (charts.length) {
+
+  switch (charts.length * (compare.length || 1)) {
```
PatrickSison, line 119: Chart calculation to account for comparison

src/pages/reportBuilder/components/Charts.js
```diff
@@ -1,37 +1,122 @@
...
-export default function Charts(props) {
-  const { chartData = [], metrics, chartLoading, precision, yScale, to } = props;
+export default function ChartContainer() {
+  const { state: reportOptions } = useReportBuilderContext();
+  return <ChartGroups reportOptions={reportOptions} />;
+}
+
+export function ChartGroups(props) {
+  const { reportOptions } = props;
+  const { compare } = reportOptions;
```
nicklemmon, line 32: This name confuses me a bit. Should it be `comparisons`?

src/pages/reportBuilder/components/Charts.js
```diff
@@ -1,37 +1,122 @@
...
+    <>
+      {compare.map((compareFilter, index) => {
+        const filterType = FILTER_KEY_MAP[compareFilter.type];
+        const newFilters = [
+          ...reportOptions.filters,
+          { AND: { [filterType]: { eq: [compareFilter] } } },
+        ];
+        return (
+          <Panel.Section>
+            <Box key={`chart_group_${index}`}>
```
nicklemmon, line 52: The `key` needs to be on the outermost element here. `Panel.Section` in this case. Probably no need for the `Box`?

src/pages/reportBuilder/components/Charts.js
```diff
@@ -1,37 +1,122 @@
...
+              <Heading data-id={`heading_${index}`} as="h2" looksLike="h5">
```
nicklemmon, line 53: This probably shouldn't be an `h2` - double check the heading structure of the page. At most it would be an `h3` but might even be an `h4`.

src/pages/reportBuilder/components/Charts.js
```diff
@@ -1,37 +1,122 @@
...
+  }
+
+  return (
+    <>
+      {compare.map((compareFilter, index) => {
+        const filterType = FILTER_KEY_MAP[compareFilter.type];
+        const newFilters = [
+          ...reportOptions.filters,
+          { AND: { [filterType]: { eq: [compareFilter] } } },
+        ];
```
nicklemmon, line 49: I don't quite understand what's happening here. Would you mind adding some comments to explain things?

src/pages/reportBuilder/components/Charts.js
nicklemmon: I would double check the font size token used in the mock - the mocks show:
1. A larger font size
2. Space below the heading. Probably worth using a `<Stack/>` here.

src/pages/reportBuilder/components/Charts.js
```diff
@@ -56,11 +141,10 @@ export default function Charts(props) {
...
               name: label,
-              stroke: chartLoading ? tokens.color_gray_100 : stroke,
+              stroke: chartStatus === 'loading' ? tokens.color_gray_100 : stroke,
             }))}
             {...formatters}
             yTickFormatter={chart.yAxisFormatter}
-            yScale={yScale}
             yLabel={chart.label}
             tooltipValueFormatter={chart.yAxisFormatter}
             showXAxis={i === charts.length - 1}
```
nicklemmon, line 150: The mocks don't appear to have the label on the lefthand side, or a dividing line between charts. Are these old mocks?

**Mocks:**
Screen Shot 2020-11-19 at 9 18 53 AM

**Actual:**
Screen Shot 2020-11-19 at 9 19 22 AM

nicklemmon (changes requested):
@PatrickSison A bit of feedback for you to start on - the review is a little confusing with the other PR here too - but hopefully we can get out SA-1773 today to help with that.

src/pages/reportBuilder/components/Charts.js
PatrickSison: Yeah, let me attach new mocks to the ticket

src/pages/reportBuilder/components/Charts.js
PatrickSison: Yeah, that's true. Will update

src/pages/reportBuilder/components/Charts.js
PatrickSison: I'm cool with changing the variable name. I didn't settle on a specific variable name. Will update this + query param

src/pages/reportBuilder/components/Charts.js
```diff
@@ -1,37 +1,122 @@
...
 }
 
-export default function Charts(props) {
-  const { chartData = [], metrics, chartLoading, precision, yScale, to } = props;
+export default function ChartContainer() {
+  const { state: reportOptions } = useReportBuilderContext();
+  return <ChartGroups reportOptions={reportOptions} />;
+}
+
+export function ChartGroups(props) {
```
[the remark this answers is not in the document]
PatrickSison: Yeah. Report options might not be coming from reportbuildercontext and may need to be sourced from somewhere else I think (e.g. Dashboard V2, or IP monitoring). We could make the reportBuilder context reusable so that other pages can use the same context provider, and then keep some of that business logic for the report options? What do you think?

src/pages/reportBuilder/components/Charts.js
```diff
@@ -1,37 +1,122 @@
...
+    return getQueryFromOptions({ ...reportOptions, metrics: formattedMetrics });
+  }, [reportOptions, formattedMetrics]);
+  const { precision, to } = formattedOptions;
+
+  const { data: rawChartData, status: chartStatus, refetch: refetchTimeSeries } = useSparkPostQuery(
+    () => {
+      return getTimeSeries(formattedOptions);
+    },
+    {
+      enabled: reportOptions.isReady,
```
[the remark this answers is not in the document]
PatrickSison: Yeah, didn't realize it wasn't rendering until then. Will remove.

src/pages/reportBuilder/components/Charts.js
PatrickSison: But, the mocks have always had the y axis. I'll double check how we want to go forward with the charts

PatrickSison:
@nicklemmon For easier diff comparison: https://github.com/SparkPost/2web2ui/compare/SA-1773...SA-1617

src/helpers/api/defaultQuery.js
```diff
@@ -23,12 +23,10 @@ export default function defaultQuery(key, { method, params, headers, auth }) {
     },
   }).then(response => {
     const results = _.get(response, 'data.results', response.data);
-    const links = _.get(response, 'data.links', {});
-    const total_count = _.get(response, 'data.total_count');
+    //TODO: Remove or find place with usePaginatedQuery();
+    // const links = _.get(response, 'data.links', {});
+    // const total_count = _.get(response, 'data.total_count');
```
nicklemmon, line 28: I think you can just delete this to be honest - may be awhile before we even need paginated queries.

src/helpers/reports.js
```diff
@@ -130,8 +130,8 @@ export function parseSearchNew(search) {
     ret.filters = mapFiltersToComparators(filtersList);
   }
 
-  if (compare) {
-    ret.compare = compare;
+  if (comparisons) {
+    ret.comparisons = comparisons;
```
nicklemmon, line 134: 👍 More consistent!

src/pages/reportBuilder/components/Charts.js
```diff
@@ -1,37 +1,127 @@
...
   // Keeps track of hovered chart for Tooltip
   const [activeChart, setActiveChart] = React.useState(null);
 
-  if (!chartData.length || !metrics) {
-    return null;
+  if (chartStatus === 'loading' || chartStatus === 'idle') {
+    return (
+      <Box height="10rem">
+        <Loading />
+      </Box>
```
nicklemmon, line 112: You shouldn't need a `Box` here - the `<Loading />` component has a `maxHeight` prop you can leverage.

Also, it looks like two loading states follow each other, and the size is dramatically different so it looks kind of janky. The size should either be consistent (controlled via a shared component or variable) or there shouldn't be multiple, different loading states:

**First loading state:**
Screen Shot 2020-11-20 at 2 09 09 PM

**Second loading state:**
Screen Shot 2020-11-20 at 2 09 28 PM

src/pages/reportBuilder/components/Charts.js
```diff
@@ -56,11 +146,10 @@ export default function Charts(props) {
               key: name,
               dataKey: name,
               name: label,
-              stroke: chartLoading ? tokens.color_gray_100 : stroke,
+              stroke: chartStatus === 'loading' ? tokens.color_gray_100 : stroke,
```
nicklemmon, line 149: If the `<Loading/>` just takes the place of the chart entirely, is this necessary?

src/pages/reportBuilder/ReportBuilder.js
```diff
@@ -214,17 +203,16 @@ export function ReportBuilder({
...
         ) : (
           <>
-            <hr className={styles.Line} />
             <div data-id="summary-chart">
               <Tabs defaultTabIndex={0} forceRender tabs={tabs}>
                 <Tabs.Item>
-                  <Panel.Section className={styles.ChartSection}>
-                    <Charts {...chart} metrics={processedMetrics} to={to} yScale="linear" />
-                  </Panel.Section>
+                  <Charts {...chart} metrics={processedMetrics} to={to} yScale="linear" />
```
nicklemmon, line 215: I tried this for the loading and it seemed to work well?
https://github.com/SparkPost/2web2ui/compare/master...SA-1619#diff-7ebbaa622d25702925ac215015c3dd0edbcfa747d72cd17088a3bd05a7a33abaR214

Might need to move the loading state further down though, so might need to just ignore this entirely

src/helpers/api/signals.js
```diff
@@ -0,0 +1,7 @@
+export function getTimeSeries(params) {
```
nicklemmon, line 1: Should this file be `metrics.js` to mirror the endpoint? I could see it either way.

src/pages/reportBuilder/ReportBuilder.js
```diff
@@ -23,7 +23,6 @@ import {
   LinksTable,
   RejectionReasonsTable,
 } from './components/tabs';
-import styles from './ReportBuilder.module.scss';
```
nicklemmon, line 26: Yaaay

src/pages/reportBuilder/components/Charts.js
nicklemmon: I'm usually a fan of doing it like you have it - keeping these sorts of things separate - each function only doing one thing.

src/pages/reportBuilder/components/Charts.js
```diff
@@ -1,37 +1,127 @@
...
+    {
+      refetchOnWindowFocus: false,
+    },
+  );
+
+  useEffect(() => {
+    if (reportOptions.isReady) {
+      refetchTimeSeries();
+    }
+  }, [reportOptions.isReady, refetchTimeSeries]);
```
nicklemmon, line 99: Is this even a possible state? It looks like `<Loading/>` is rendered on `ReportBuilder` if `!reportOptions.isReady`.

src/pages/reportBuilder/components/Charts.js
nicklemmon: This is more of a UX question, but shouldn't the things being compared have some sort of label? Like these:
Screen Shot 2020-11-20 at 2 31 46 PM

Would I necessarily know that `greg-hackathon` and `geoip-testing` are templates? They might be something else.

cypress/tests/integration/signals-analytics/analytics-report/reportCompareBy.spec.js
```diff
@@ -99,6 +99,16 @@ if (IS_HIBANA_ENABLED) {
...
     });
 
+    it('Properly submits the form and renders multiple charts', () => {
+      openCompareByModal();
+      fillOutForm();
+      cy.withinDrawer(() => {
+        cy.findByRole('button', { name: 'Compare' }).click();
+      });
+      cy.findByRole('heading', { name: 'Fake Subaccount 1 (ID 101)' }).should('exist');
+      cy.findByRole('heading', { name: 'Fake Subaccount 3 (ID 103)' }).should('exist');
```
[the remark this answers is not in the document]
nicklemmon: It looks like there are a few other spots you can add `cy.wait()` now that the network requests are going through.

nicklemmon (changes requested):
@PatrickSison Have some more feedback for you. Things seem to be working well.

Also, looks like query params aren't updating for this yet? Has that been wired up? I looked at Jira but couldn't tell if there was an additional ticket to address that.

src/helpers/api/signals.js
PatrickSison: Makes it more clear

src/pages/reportBuilder/ReportBuilder.js
nicklemmon: Looks much better now. Thank you for doing this.

src/pages/reportBuilder/components/Charts.js
```diff
@@ -1,37 +1,114 @@
...
 
-  // Keeps track of hovered chart for Tooltip
-  const [activeChart, setActiveChart] = React.useState(null);
+function ChartGroups(props) {
+  const { reportOptions } = props;
+  const { comparisons } = reportOptions;
 
-  if (!chartData.length || !metrics) {
-    return null;
+  if (!comparisons.length) {
```
nicklemmon, line 33: Should this be `!Boolean(comparisons.length)` to be a little safer?

src/pages/reportBuilder/components/LineChart.js
```diff
@@ -1,3 +1,177 @@
-import asyncComponent from 'src/components/asyncComponent/asyncComponent';
+/*
+  This component is meant to be loaded asynchronously, do not import directly.
+  See ../LineChart.js for async export
+*/
```
nicklemmon, line 4: You can remove this comment now

src/pages/reportBuilder/components/Charts.js
```diff
@@ -1,37 +1,114 @@
...
+      refetchOnWindowFocus: false,
+    },
+  );
+
+  const chartData = useMemo(() => {
+    return transformData(rawChartData, formattedMetrics);
+  }, [rawChartData, formattedMetrics]);
+
+  // Keeps track of hovered chart for Tooltip
+  const [activeChart, setActiveChart] = React.useState(null);
```
nicklemmon, line 101: Screen Shot 2020-11-24 at 10 18 08 AM

Not sure if this is the expectation: hovering the first chart renders the tooltip, then the second one renders both, but hovering the fourth renders them all.

nicklemmon (commented):
@PatrickSison I had a couple of questions - though depending on your answers we can move forward with this one.

src/pages/reportBuilder/components/Charts.js
PatrickSison: Hopefully fixed the behavior for the tooltip. This is a hack in the first place so it's a bit janky

nicklemmon (approved):
@PatrickSison Looks good to me!

PR #1909, opened 2020-11-20
SA-1825: 5 and 15 minute precision are no longer the default precisions

PatrickSison:
### What Changed
 - Rollups precision select 1 min when range is up to 8 hours
 - Rollups precision select hour when range is up to 10 days (unchanged)
 - Rollups no longer recommend 5/15 minute precision (but are still options)

### How To Test
 - On account with rollups enable, select date range below 8 hours
 - Recommended precision should be 1min
 - Select precision above 8 hours, below 10 days
 - Precision should be hour
 - Verify that 5/15 minute precision still selectable
 - Verify that behavior is correct in OG theme 😢 

### To Do
- [ ] Does this need to apply to non-rollup metrics?
- [ ] Verify behavior of precision in metrics


kchu1337 (commented):
Current changes look good. Just need to fix up the failing tests and it should be good to deploy.

src/components/datePicker/tests/DatePickerNew.test.js
```diff
@@ -64,7 +64,7 @@ describe('Component: DatePicker', () => {
     //From will be 12:00 due to recommended day precision rounding
```
kchu1337, line 64: Can problem delete this comment.

kchu1337 (approved):
One small change that's up to you

kchu1337 (approved):

PR #1911, opened 2020-11-20
FE-1257 - DashboardV2 Onboarding Grant Check

nstanard:
### What Changed
 - Added grant checks around the dynamic onboarding states. Set fallback as the else if the needed grants are not met.
 - Removed the extra <Abbr> element

Follow on to: https://github.com/SparkPost/2web2ui/pull/1899
(in that pr I added an extra abbr and didnt set the right grant checks - this fixes that)

**Note**: At some point I think it'd be good to move the onboarding logic/string to be on the main app state where any component/state could pick up on it and react to the users onboarding state. 

### How To Test
 - Load up a user of each type and check to see the onboarding states are correct.
admin and dev will see the individual onboarding steps based on sending domain and api key data. 
reporting and templates will _currently_ always see the analytics promo until a future ticket changes that. 

### To Do
- [x] Review Feedback

### DIFF
--- a/src/pages/dashboardV2/DashboardPageV2.container.js
+++ b/src/pages/dashboardV2/DashboardPageV2.container.js
@@ -27,8 +27,6 @@ function mapStateToProps(state) {
   // TODO: https://sparkpost.atlassian.net/browse/FE-1249 - rvUsage rename
   let lastUsageDate = state?.account?.rvUsage?.messaging?.last_usage_date;
 
-  // let onboarding = state?.onboarding;
-  // TODO: Move onboarding to a higher state/provider where it can be pulled into any area of the app
   const sendingDomains = state.sendingDomains.list;
   const verifiedDomains = selectVerifiedDomains(state);
   const apiKeysForSending = selectApiKeysForSending(state);
@@ -51,7 +49,7 @@ function mapStateToProps(state) {
 
     if (!addSendingDomainNeeded && !verifySendingNeeded && !createApiKeyNeeded)
       onboarding = 'startSending';
-  } else if (lastUsageDate === null) {
+  } else if (!canViewUsage || (canViewUsage && lastUsageDate === null)) {
     onboarding = 'fallback';
   } else {
     onboarding = 'done';
--- a/src/pages/dashboardV2/DashboardPageV2.container.js
+++ b/src/pages/dashboardV2/DashboardPageV2.container.js
@@ -36,7 +36,7 @@ function mapStateToProps(state) {
   const canManageApiKeys = hasGrants('api_keys/manage')(state);
   const canManageSendingDomains = hasGrants('sending_domains/manage')(state);
 
-  let onboarding = 'done';
+  let onboarding;
   if (canViewUsage && lastUsageDate === null) {
     if (isAnAdmin || isDev) {
       let addSendingDomainNeeded;
--- a/src/pages/dashboardV2/DashboardPageV2.container.js
+++ b/src/pages/dashboardV2/DashboardPageV2.container.js
@@ -36,7 +36,7 @@ function mapStateToProps(state) {
   const canManageApiKeys = hasGrants('api_keys/manage')(state);
   const canManageSendingDomains = hasGrants('sending_domains/manage')(state);
 
-  let onboarding;
+  let onboarding = 'done';
   if (canViewUsage && lastUsageDate === null) {
     if (isAnAdmin || isDev) {
       let addSendingDomainNeeded;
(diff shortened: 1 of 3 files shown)


src/pages/dashboardV2/DashboardPageV2.container.js
```diff
@@ -50,8 +49,12 @@ function mapStateToProps(state) {
 
     if (!addSendingDomainNeeded && !verifySendingNeeded && !createApiKeyNeeded)
       onboarding = 'startSending';
+  } else {
+    onboarding = 'fallback';
```
shrirupa, line 53: GET /usage api is forbidden only for reporting user. so what we can do is either always default to 'fallback' state or based on GET /account?include=usage messaging.used count display the analytics step (this condition is not flawless, since count goes down to 0 at the end of the month, but it will only be a problem for "reporting user" so I guess it should be fine.)

nstanard: https://github.com/SparkPost/2web2ui/pull/1911/commits/8c1c249ef43c5ee8e1d4cc8a01d338a211bce79a#diff-ebaa9f807ddc9bcd5bc926ca8ec1e78d19fa4e2538836a326fdb634b57ad0c3aR52

This is a commit I had put in place but then reverted (basically what you're describing here). I keep going back and forth on this. It'd be ideal if we could just get the last usage date without being blocked by grants. Then as long as we have a last_usage_date and not null then we know their account is through onboarding. Perhaps I don't have the right understanding of Reporting users but if we simply always were able to look at the last_usage_date I'd feel a lot better about flipping onboarding to "done"

shrirupa: I am not aware of the exact reason why usage/view is not granted to reporting and subaccount_reporting users, we can follow-up with accounts team on that but till that change is made we can use account?include=usage for this.

nstanard: That was another thing that I ran into when I was first going through the onboarding tasks - the last_usage_date wasn't coming back on the account object when passing `?include=usage`

image

I now realize I probably should have reached out directly to the account team and talked to them about that when I saw the two endpoints returning different data structures.

nstanard: Were you thinking, If a user doesn't have the `usage/view` grant, we can check `account.usage.month.used` is zero? and then display onboarding? I put checks back in for role types but I'm still on the fence about that.

nstanard:
Note: 
`/account/control?include=usage` has a number of options.ui properties relating to onboarding (this is a glance at the historical changes in relation to tracking onboarding using account ui options)

`account.options.ui.message_onboarding`
`account.options.ui.messaging_onboarding`
`account.options.ui.onboarding_active_step`

```
account.options.ui.onboarding: {
    "active_step": "Let's Code",
    "add_sending_domain_completed": true,
    "invite_collaborator_completed": true,
    "explore_analytics_completed": true,
    "isGuideAtBottom": true,
    "check_events_completed": true,
    "view_developer_docs_completed": true,
    "send_test_email_completed": true
},
```


**odd things:**
mis-spelled ones?
`account.options.ui.onbaording.isGuideAtBottom`

mis-spelled one sub-property duplicated without a reference to onboarding?
`account.options.ui.isGuideAtBottom`

image

src/pages/dashboardV2/DashboardPageV2.container.js
shrirupa: yes, I was thinking of using `account.usage.month.used` using that initially, but as we discussed this endpoint (account?include=usage may be deprecated soon. So we should not use this. We should rather just default to "fallback" for reporting users for now as @bkemper had suggested.

src/pages/dashboardV2/DashboardPageV2.container.js
```diff
@@ -21,37 +21,57 @@ import { selectVerifiedDomains } from 'src/selectors/sendingDomains';
...
 
-    if (!addSendingDomainNeeded && !verifySendingNeeded && !createApiKeyNeeded)
-      onboarding = 'startSending';
+      if (!addSendingDomainNeeded && !verifySendingNeeded && !createApiKeyNeeded)
+        onboarding = 'startSending';
+    }
+
+    if (isTemplatesUser || isReportingUser) {
+      onboarding = 'analyticsReportPromo';
+    }
```
shrirupa, line 68: can we move this to the else block below.

shrirupa:
looks like few integration tests on dashboardv2 are failing

bkemper (approved):

PR #1915, opened 2020-11-23
FE-1238 - Hitting the escape key with the PlanPicker in the open state renders the error boundary

nicklemmon:
[FE-1238](https://sparkpost.atlassian.net/browse/FE-1238)

### What Changed
- Updates `downshift` to latest version

### How To Test
- To go `/onboarding/plan` and click on the custom plan select to open it, then hit the escape key to close it - the app shouldn't crash!
- Test other instances of components using `downshift` to ensure they still work as expected

### To Do
- [ ] Incorporate feedback


bkemper (approved):
- [X] I confirmed esc no longer breaks the plan picker.  
- [X] I confirmed downshift was working for alerts, signals, etc.  
- [X] I reviewed the changelog.  The version bump was a big jump from v3.2.10 to v6.0.6 that primarily included new features, support, and functionality (i.e. hooks, Preact, treeshaking, etc.)

PR #1917, opened 2020-11-23
TR-2897: delivery_confidence example need to match other values

bkemper:
[TR-2897](https://sparkpost.atlassian.net/browse/TR-2897)

### What Changed
 - Set delivery_confidence example to zero on RV API Integration tab

### How To Test
 - Visit /recipient-validation/api
 - Confirm "delivery_confidence" example value equals zero


nicklemmon (approved):
@bkemper Looks good to me. For real this time.

PR #1919, opened 2020-11-24
Revert "FE-1167: Secure webui cookies"

bkemper:
While testing in staging, we found that pre-existing cookies were not being deleted.

Screen Shot 2020-11-24 at 10 35 52 AM


nicklemmon (approved):
@bkemper Looks good to me

PR #1920, opened 2020-11-24
Restore FE-1167: Secure webui cookies

bkemper:
There was a bug in #1809 that was setting the cookie domain option only for dev and test environments.  It would be worthless to set for all environments, especially considering the default is what we want.

> domain - A string representing the domain of the cookie. If omitted, the cookie becomes a host-only cookie.
> — https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/set


nicklemmon (approved):
@bkemper Looks good - after correcting for my own reviewer error.

PR #1923, opened 2020-11-30
SA-1619 - (UI) Compare By - Aggregate section update

nicklemmon:
[SA-1619](https://sparkpost.atlassian.net/browse/SA-1619)

### What Changed

- Adds API helpers to use with `useSparkPostQuery` that map to existing Redux actions for `metrics` endpoints
- Expose relevant query params for the `getQueryFromOptionsV2` helper
- Conditionally render `CompareByAggregatedMetrics` component depending on whether there are active comparisons present in feature's state
- Moves constants to a single index file within the report builder feature
- Upgrades Review Badger

### How To Test
- Click on "Add Comparison" and add at least two comparisons
- Verify that below the charts, aggregated data renders
- Look at the network requests in dev tools and verify the correct requests are being made
- Verify that the section updates when metrics are added and removed
- Verify that the section renders when the user refreshes the page or pastes the URL in a new tab or browser window

### To Do
- [ ] Incorporate feedback


src/pages/reportBuilder/components/CompareByForm/CompareByForm.js
```diff
@@ -69,6 +69,14 @@ const reducer = (state, action) => {
...
 };
 
+const getInitialState = comparisons => {
+  if (!comparisons || !comparisons.length) {
+    return initialState;
+  }
+
+  return { filterType: FILTER_KEY_MAP[comparisons[0].type], filters: comparisons };
+};
+
```
nicklemmon, line 79: I happened to be peeking in this file and realized this could be pulled out of the component pretty easily.

src/helpers/metrics.js
```diff
@@ -59,7 +61,7 @@ export function getQueryFromOptions({
   return options;
 }
 
-// TODO: Replace original once original theme removed.
+// TODO: Replace original once OG theme is removed
 export function getQueryFromOptionsV2({
```
nicklemmon, line 65: Should we add some unit tests to support this?

src/helpers/metrics.js
```diff
@@ -69,6 +71,13 @@ export function getQueryFromOptionsV2({
   filters = [],
   match = '',
   limit,
+  domains,
+  campaigns,
+  templates,
+  sending_ips,
+  ip_pools,
+  subaccounts,
+  sending_domains,
```
nicklemmon, line 80: Is there a cleaner way to do this that would be easier to maintain?

src/pages/reportBuilder/constants/index.js
```diff
@@ -47,3 +47,69 @@ export const PRESET_REPORT_CONFIGS = [
     type: 'preset',
   },
 ];
+
+export const GROUP_BY_CONFIG = {
```
nicklemmon, line 51: I moved these in to a central location rather than having multiple config files with one value each.

src/pages/reportBuilder/components/CompareByAggregatedMetrics.js
```diff
@@ -0,0 +1,107 @@
...
+  return (
+    <Stack>
+      <Inline>
+        <Definition dark>
+          <Definition.Label>{comparison.type}</Definition.Label>
+          <Definition.Value>{comparison.value}</Definition.Value>
+        </Definition>
+
+        {hasMetrics
+          ? aggregatedMetricsKeys.map((metricKey, metricKeyIndex) => {
```
kchu1337, line 88: Is there a reason you use the response object vs formattedMetrics? Then you wouldn't need `getMetricFromKey`.

src/pages/reportBuilder/components/CompareByAggregatedMetrics.js
```diff
@@ -0,0 +1,107 @@
...
+  if (status === 'loading' || status === 'error') return null;
+
+  const aggregatedMetricsObj = data[0] || {};
+  const aggregatedMetricsKeys = Object.keys(aggregatedMetricsObj);
+  const hasMetrics = Boolean(aggregatedMetricsKeys.length);
+
+  return (
+    <Stack>
+      <Inline>
+        <Definition dark>
```
kchu1337, line 82: Should this be LabelValue?

src/pages/reportBuilder/ReportBuilder.js
```diff
@@ -216,37 +209,59 @@ export function ReportBuilder({
...
+                                <Box marginRight="600" key={key}>
+                                  <LabelValue dark>
+                                    <LabelValue.Label>
+                                      <Box color="gray.600">{label}</Box>
+                                    </LabelValue.Label>
+
+                                    <LabelValue.Value>
+                                      <Box display="flex" alignItems="center" color="white">
+                                        {stroke && (
+                                          <LegendCircle marginRight="200" color={stroke} />
```
kchu1337, line 249: Should we add the color circle to the compare by? or remove it here?

src/pages/reportBuilder/ReportBuilder.js
```diff
@@ -216,37 +209,59 @@ export function ReportBuilder({
...
-                                  <Box display="flex" alignItems="center">
-                                    {stroke && <LegendCircle marginRight="200" color={stroke} />}
-                                    <Unit value={value} unit={unit} />
-                                  </Box>
-                                </MetricDefinition>
+
+                  {hasActiveComparisons ? (
+                    <CompareByAggregatedMetrics date={dateValue} />
+                  ) : (
+                    <Box padding="400" backgroundColor="gray.1000">
```
kchu1337, line 216: Could also move this into it's own component, so the size of report builder decreases. Or implement the ternary in another separate component that contains both the compare by and normal aggregate metrics.

src/helpers/metrics.js
kchu1337: I'm not sure what this does?

src/helpers/metrics.js
kchu1337: Yeah, but it's not within the scope of this ticket. Up to you.

kchu1337 (changes requested):

src/pages/reportBuilder/components/CompareByAggregatedMetrics.js
nicklemmon: good point - let me try that.

src/pages/reportBuilder/ReportBuilder.js
nicklemmon: It's not in the mocks, though it potentially should be. Let me check with UX.

src/pages/reportBuilder/ReportBuilder.js
nicklemmon: I like that idea

src/pages/reportBuilder/ReportBuilder.js
nicklemmon: UX confirmed they missed this. Good catch.

While I was there, I also incorporated the `<Unit/>` component for the value. Definitely would have caused problems otherwise.

nicklemmon:
@kchu1337 Ready for another pass.

kchu1337 (approved):

PR #1924, opened 2020-11-30
Upgrade to Cypress 6

nicklemmon:
**Note:** This will not be merged in to `master`, but rather a feature branch that requires it. Testing this independently to avoid introducing too much complexity when performing the upgrade.


nicklemmon:
Closing this, will be addressed by ticket FE-1275.

PR #1925, opened 2020-11-30
FE-1242 - Upgrades Matchbox to version 4.6.0

nicklemmon:
[FE-1242](https://sparkpost.atlassian.net/browse/FE-1242)

### What Changed
- Upgrades Matchbox to version 4.6.0
- For Jest DOM, adds the `window.matchMedia` property to avoid test failures introduced by Matchbox changes
- Adds component wrappers for:
  + `EmptyState` and `EmptyState.LEGACY`
  + `InlineCode`
  + `LabelValue`
- Removes usage of the `Definition` component within Hibana-only features - leaves it in place in features that leverage the OG theme
- Replaces instances of `EmptyState` with `EmptyState.LEGACY` for components that render in both the OG and Hibana themes
- **New:** Adds padding to Dashboard onboarding images - Matchbox removed baked-in padding within its image library

### How To Test
- Ensure the app continues to render as expected
- Ensure test scenarios adequately test changes

### To Do
- [ ] Incorporate feedback


src/components/errorBoundaries/ErrorBoundary.js
```diff
@@ -29,8 +35,8 @@ export default class ErrorBoundary extends Component {
 
     if (this.state.hasError) {
       return (
-        <div style={{ margin: '0 auto', maxWidth: 1080 }}>
```
nicklemmon, line 32: Remove inline styles to help resolve the ESLint error.

src/pages/reportBuilder/ReportBuilder.js
```diff
@@ -1,14 +1,21 @@
 import React, { useState, useEffect, useMemo } from 'react';
 import { connect } from 'react-redux';
 import moment from 'moment';
-import { tokens } from '@sparkpost/design-tokens-hibana';
```
nicklemmon, line 4: No need tor this since in most cases the theme variables can be accessed via a string

src/components/text/Definition.js
```diff
@@ -18,6 +18,11 @@ const StyledValue = styled.div`
 const Definition = ({ children, dark }) => {
   const id = useUniqueId('definition');
 
+  // eslint-disable-next-line
+  console.warn(
```
PatrickSison, line 22: Neat way to do this

src/pages/reportBuilder/ReportBuilder.js
PatrickSison: Should we block the import of this? I don't really see much of a case for this anymore either since the tokens are also accessible in the props of styled components (but not sure what's the preferred style for styled-components)

PatrickSison (approved):
Seems to work fine. Haven't seen any visual bugs so far

src/pages/dashboardV2/DashboardPageV2.js
```diff
@@ -19,16 +21,14 @@ import {
...
 import Dashboard from './components/Dashboard';
 import Sidebar from './components/Sidebar';
 import { LINKS } from 'src/constants';
-import styled from 'styled-components';
-import { ChartGroups } from 'src/pages/reportBuilder/components/Charts';
-import { usePinnedReport } from 'src/hooks';
-import _ from 'lodash';
 
-const OnboardingPicture = styled(Picture.Image)`
+const OnboardingImg = styled(Picture.Image)`
```
nicklemmon, line 31: This is a more accurate name - it isn't a `Picture`

PR #1927, opened 2020-12-01
FE-1192: Add the Change Report Modal to dashboard

shrirupa:
FE-1192: Add the Change Report Modal to dashboard

### What Changed

- Modified the existing ReportsListModal
      - moved in into src/components/reportBuilder
      - used the Tabs from src/components/matchbox as the other Tabs component was defaulting to the My Reports Tab on any re-render 
      - Added a `onDashboard` prop to this Modal to add support for https://sparkpost.invisionapp.com/console/Dashboard_PinnedReport-ckfydojkz00cr01vsbf9z1pt4/ckfydokg600cu01vs3qyndqlx/play

     - on Dashboard, on opening this Modal, the Submit button should be disabled when no checkbox is selected
     - on Dashboard, on submit of this modal, `pinned_report` user ui option should be set with correct report id.

### How To Test
- Use the Appteam account on staging to get to this onboarding step.
- On Clicking the “Change Report” button, this [modal](https://sparkpost.invisionapp.com/console/Dashboard_PinnedReport-ckfydojkz00cr01vsbf9z1pt4/ckfydokg600cu01vs3qyndqlx/play) Dashboard_PinnedReport  should open up
- On clicking “Change Report” save the selection as a User UI Option. - “pinned_report”
- Check if the ReportsList Modal works as earlier on the ReportBuilder Page.

### To Do

- [ ] Address feedback
- [x] Fix this issue - Modal is resetting when you search for a report name and select the checkbox.


src/components/reportBuilder/ReportsListModal.js
```diff
@@ -0,0 +1,130 @@
...
+    if (!onDashboard) return <>{children}</>;
+
+    return (
+      <Radio.Group label="reportList" labelHidden>
+        {children}
+      </Radio.Group>
+    );
+  };
+
+  const TABS = [
```
PatrickSison, line 58: I do like the idea of making the tabs reusable, but not sure how much I like passing in the onDashboard prop to basically create a separate version of everything. It might make things a bit tangled and confusing to pull apart if needed later on.

src/components/collection/Collection.js
```diff
@@ -80,6 +85,7 @@ export class Collection extends Component {
     }
 
     this.setState(update);
+    onChange(pattern);
```
PatrickSison, line 88: Could just change to
```javascript
if (onChange) {
  onChange(pattern);
}
```
instead of defaulting to an empty function. Feels a bit safer than possibly getting a working function, but passing in function becomes an issue

src/components/reportBuilder/ReportsListModalTabs.js
```diff
@@ -119,74 +181,55 @@ export function ReportsListModal(props) {
...
       ''
     );
+    const row = [
+      <div>{formatDateTime(modified)}</div>,
+      <div>{creator}</div>,
+      <Tag>
+        <Subaccount id={subaccount_id} master={subaccount_id === 0} shrinkLength={12} />
+      </Tag>,
+    ];
+    if (onDashboard)
```
PatrickSison, line 191: Partially why I'm a bit scared of passing in onDashboard. There's a lot of elements now being conditionally rendered (like columns and this radio) that are tied into the this modal and might be easier to maintain if they were separate components. At the same time, it's also a bit painful to rebuild components. Might be worth getting thoughts too from @nicklemmon  and @kchu1337  on this as well.

src/pages/dashboardV2/DashboardPageV2.js
```diff
@@ -90,6 +105,9 @@ export default function DashboardPageV2() {
                         <TranslatableText>Analyze Report</TranslatableText> <ShowChart size={25} />
```
PatrickSison, line 105: This has an underline as well. Not sure if that's what's intended

src/components/reportBuilder/ReportsListModal.js
```diff
@@ -0,0 +1,130 @@
...
+  const [tabIndex, setTabIndex] = useState(0);
+
+  const [selectedReportId, setSelectedReportId] = useState(null);
+
+  const [searchedText, setSearchedText] = useState('');
+
+  const handleRadioChange = id => setSelectedReportId(id);
+
+  const onSubmit = () => {
+    dispatch(updateUserUIOptions({ pinned_report: selectedReportId })).then(() => {
```
PatrickSison, line 36: Shouldn't this update the active report on the page?

PatrickSison (commented):
Hey! Left some preliminary comments. It doesn't seem like changing the report actually changes the report used for pinnedReport? Also would be worth getting some integration testing around this as this is a complex pattern I think

src/components/reportBuilder/ReportsListModal.js
shrirupa: we have a separate ticket to handle updating the Chart with the user ui option. In this ticket we only handled adding the Modal and updating the user ui option on Changing the report.

src/pages/dashboardV2/DashboardPageV2.js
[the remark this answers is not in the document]
shrirupa: did you mean like this - https://github.com/SparkPost/2web2ui/pull/1931/files#diff-0581037d04e85bf29360df4f0ce66f27c04c6d7aa77d7794a965b1f2128f530dR54 . (Added this on FE-1266)

src/components/reportBuilder/ReportsListModalTabs.js
nicklemmon: I tend to agree, Patrick.

I would say either build a new component or tweak the API a bit to make it more generic. We may end up using this modal somewhere else some day, so `onDashboard` could really be something more general like `hasRadioButtons` with a generic `onChange` handler that could be passed in.

But after looking it over I think just a new component entirely, even if there is some repetition, would be better here. If there is shared business logic, then extract that in to a custom hook.

Also, just a random tidbit while I'm here, that `Radio` still needs a `label` - even if it is only exposed to screen reader users. That will also help from a testing perspective.

src/pages/dashboardV2/DashboardPageV2.js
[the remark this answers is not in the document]
PatrickSison: 👍

src/components/reportBuilder/ReportsListModalTabs.js
shrirupa: Thanks @PatrickSison and @nicklemmon for the feedback! I will update the Tabs to be separate components and make ChangeReportModal also a separate component.

src/pages/dashboardV2/DashboardPageV2.js
shrirupa: Thanks! that wasn't intended. 
I have also updated the text to say "View Report" instead of "Analyze Report" (UX feedback)

shrirupa:
@PatrickSison Thanks for all the feedback. This PR is now ready for another pass.

src/components/reportBuilder/ChangeReportModal.js
```diff
@@ -0,0 +1,104 @@
+import React, { useState } from 'react';
```
PatrickSison, line 1: I think because this is only living in dashboard, it would probably be better if this component gets moved to the dashboard. Also the tabs/children components should probably have their own copies to separate out the business logic. Apologies if this ends up being extra!

src/pages/dashboardV2/DashboardPageV2.js
```diff
@@ -70,6 +76,15 @@ export default function DashboardPageV2() {
         <Heading as="h1">Dashboard</Heading>
       </ScreenReaderOnly>
 
+      {isModalOpen && (
+        <ChangeReportModal
+          onDashboard={true}
```
PatrickSison, line 81: This prop is no longer needed

PatrickSison: Also a lot cleaner w/ the props here 👏

src/components/reportBuilder/ChangeReportModal.js
```diff
@@ -0,0 +1,104 @@
...
+      selectedReportId={selectedReportId}
+      searchedText={searchedText}
+      setSearchedText={setSearchedText}
+    />,
+  ];
+
+  return (
+    <Modal open={open} onClose={onClose} showCloseButton maxWidth="1300">
+      <Modal.Header>Change Report</Modal.Header>
+      <Modal.Content>
```
PatrickSison, line 68: We should remove the padding of the modal content to match the mocks

src/components/reportBuilder/ChangeReportModal.js
```diff
@@ -0,0 +1,104 @@
...
+        <ModalContentContainer>
+          <Tabs
```
PatrickSison, line 70: Not related to just this component, but the reportBuilder one as well. It feels weird with the tabs getting scrolled away in a modal. Maybe we can talk with UX about figuring out if we can make the tabs sticky? Or attached to the header, and then the table is the only scrollable section? Outside of this ticket though

src/pages/dashboardV2/DashboardPageV2.js
```diff
@@ -85,14 +100,15 @@ export default function DashboardPageV2() {
                 <Dashboard.Panel>
                   <Panel.Header>
                     <Panel.Headline>{pinnedReport.name}</Panel.Headline>
-                    <Panel.Action>
-                      <PageLink to={pinnedReport.linkToReportBuilder}>
-                        <TranslatableText>Analyze Report</TranslatableText> <ShowChart size={25} />
-                      </PageLink>
+                    <Panel.Action to={pinnedReport.linkToReportBuilder}>
```
PatrickSison, line 103: Is there a way we can make this an in-app navigation instead of a full page navigation?

PatrickSison (commented):
A few small things (I hope) with mainly moving some components around. Let me know if you have any questions. Otherwise, it's looking good

src/components/reportBuilder/ChangeReportModal.js
shrirupa: yes, I agree! we should definitely follow-up with UX regarding this.

src/pages/dashboardV2/DashboardPageV2.js
shrirupa: Good catch! Looks like <Panel.Action to={} treats link as external links. I switched it to use onClick with history.push instead

PatrickSison (approved):
It's working correctly. Other misc comments are if the hitbox for the radio should extend past the radio (to maybe the whole row?). And there seems to be some weird lag with the All Reports Tab. It doesn't break functionality, and if it can be fixed later on, that's fine too

PR #1928, opened 2020-12-02
SA-1756: Add Deletion of Scheduled Reports

kchu1337:
### What Changed

- Added delete section in edit scheduled reports
- Added delete action in scheduled reports modal.

### How To Test
-  Go to '/signals/analytics' in hibana.
-  Select a saved report that already has scheduled reports. ( Sample Gmail Report! on staging works well.)
-  Click the `Schedule Report` button.
-  In the modal, click the `...` and select delete and then confirm in the new modal. Verify that the scheduled report has been deleted.
 - Click on the name of another scheduled report. This should redirect you to the edit page.
 - Scroll down to the delete panel and click delete and then confirm. Verify that the scheduled report has been deleted.


cypress/tests/integration/signals-analytics/analytics-report/scheduledReports.spec.js
```diff
@@ -276,5 +283,17 @@ if (IS_HIBANA_ENABLED) {
         'be.visible',
       );
     });
+
+    it('Deletes existing scheduled report', () => {
```
nicklemmon, line 287: Excellent test here - using the best methods and waiting at the right times. Well done.

src/pages/reportBuilder/ScheduledReportEditPage.js
```diff
@@ -59,20 +72,64 @@ export default function ScheduledReportEditPage() {
...
+                <Button variant="destructive" onClick={openModal}>
+                  Delete Item
+                </Button>
+              </Panel.Section>
+            </Panel>
+          </Layout.Section>
+        </Layout>
+      </Page>
+      <DeleteModal
+        title="Are you sure you want to Delete Scheduled Report"
```
nicklemmon, line 122: Should there be a question mark here? Not sure the capitalization is right in the sentence. Maybe titlecase?

> "Are You Sure You Want to Delete Scheduled Report?"

Don't see mocks for this, so I suppose it's a UX gap.

src/pages/reportBuilder/components/SavedReportsSection/ScheduledReportsModal.js
```diff
@@ -16,21 +29,61 @@ PageLinkAsButton.displayName = 'Button';
...
+          <Button variant="minimal" aria-controls={schedule_id} data-id={schedule_id}>
+            <Button.Icon as={MoreHoriz} />
+            <ScreenReaderOnly>Open Menu</ScreenReaderOnly>
+          </Button>
+        }
+      >
+        <ActionList>
+          <ActionList.Action
+            content="Delete"
+            onClick={() => openDeleteModal({ scheduledReport })}
```
nicklemmon, line 83: Are these supposed to render on top of one another like this or one at a time?

Screen Shot 2020-12-07 at 12 15 08 PM

src/pages/reportBuilder/components/SavedReportsSection/ScheduledReportsModal.js
```diff
@@ -16,21 +29,61 @@ PageLinkAsButton.displayName = 'Button';
...
 
     return [
       <PageLink to={`/signals/schedule/${report.id}/${schedule_id}`}>{name}</PageLink>,
       <Tag>Email ({recipients.length})</Tag>,
+      <Popover
+        left
+        top={isLast}
+        id={schedule_id}
+        trigger={
+          <Button variant="minimal" aria-controls={schedule_id} data-id={schedule_id}>
```
nicklemmon, line 74: I don't think you need `aria-controls` here, or the `data-id`

src/pages/reportBuilder/components/SavedReportsSection/tests/ScheduledReportsModal.test.js
```diff
@@ -71,4 +73,13 @@ describe('Scheduled Reports Modal', () => {
...
     expect(mockClose).toHaveBeenCalled();
   });
+
+  it('handles delete properly', () => {
+    subject();
+    fireEvent.click(screen.getByRole('button', { name: 'Open Menu' }));
+    fireEvent.click(screen.getByRole('button', { name: 'Delete' }));
+    expect(screen.queryByText('Are you sure you want to Delete Scheduled Report')).toBeVisible();
+    fireEvent.click(screen.getByRole('button', { name: 'Delete' }));
+    expect(deleteScheduledReport).toHaveBeenCalledWith('123', '12345');
```
nicklemmon, line 83: Should these be in Cypress? Might make for easier refactoring in the future, though these do do the job.

src/pages/reportBuilder/components/SavedReportsSection/ScheduledReportsModal.js
```diff
@@ -39,53 +92,68 @@ export function ScheduledReportsModal(props) {
...
+                  </Box>
+                  {collection}
+                  {pagination}
+                </>
+              )}
+            </TableCollection>
+          )}
+        </Modal.Content>
+        <Modal.Footer>
+          <PageLinkAsButton variant="primary" to={`/signals/schedule/${report.id}`}>
```
nicklemmon, line 137: This is a clever workaround.

I just filed this to change the component API:
https://github.com/SparkPost/matchbox/issues/710

nicklemmon (changes requested):
@kchu1337 Overall looks good - I had a few tiny bits of feedback though nothing major.

src/pages/reportBuilder/ScheduledReportEditPage.js
kchu1337: There's this https://msys.slack.com/archives/GLLAW6UAX/p1606165485165700?thread_ts=1606164866.161600&cid=GLLAW6UAX

src/pages/reportBuilder/components/SavedReportsSection/tests/ScheduledReportsModal.test.js
kchu1337: I like it here since it seems fairly self-contained. My logic is basically 
```
IF (Is it fairly self-contained on this component )
    IF(Is there already a unit test file)
       Put the test in the test file
    ELSE
       Add it to cypress test
ELSE
 Cypress
```

nicklemmon: That makes sense - I think the only downside is it does negatively impact our ability to refactor. I think I tend to prefer tests like this when the code is re-usable unit rather then feature-specific code. So anything in `src/components`, for example, should have lots of Jest testing, but then features are largely verified with integration testing in the browser.

That being said, helper functions and other small things can be helpful to test in Jest even if they are feature-specific.

I don't think it's worth holding the ticket up over, but just my two cents.

cypress/tests/integration/signals-analytics/analytics-report/savedReports.spec.js
```diff
@@ -324,6 +324,47 @@ if (IS_HIBANA_ENABLED) {
...
+          url: '/api/v1/reports/**/schedules/**',
+          fixture: 'blank.json',
+          requestAlias: 'deleteScheduledReport',
+        });
+
+        cy.findByRole('button', { name: 'Schedule Report' }).click();
+        cy.withinModal(() => {
+          cy.findByText('My Scheduled Report').should('be.visible');
+          //For some reason this command needed to be repeated to open the actionlist
+          cy.findByRole('button', { name: 'Open Menu' }).click({ force: true });
```
nicklemmon, line 351: This is some sort of a bug with using `TableCollection` inside of a modal - just using an arbitrary `cy.wait(500)` also fixes the issue - though I'm fine with your solution too.

Neal ran in to this as well. We did some digging and still unclear what's going on.

src/pages/reportBuilder/ScheduledReportEditPage.js
```diff
@@ -59,20 +72,65 @@ export default function ScheduledReportEditPage() {
...
+          users={users}
+        />
+        <Layout>
+          <Layout.Section annotated>
+            <Layout.SectionTitle>Delete Scheduled Report</Layout.SectionTitle>
+          </Layout.Section>
+          <Layout.Section>
+            <Panel accent="red">
+              <Panel.Section>
+                Make sure you absolutely want to delete this item before proceeding. This action
```
nicklemmon, line 109: This needs to be a `<p>` rather than just raw text here. Sorry, should have caught that earlier.

nicklemmon (changes requested):
@kchu1337 Everything looks good - I found one tiny HTML issue that I should have caught earlier. Let's fix that and then get this out the door.

nicklemmon (approved):
@kchu1337 Looks good to me.

PR #1931, opened 2020-12-03
FE-1266 - Add aggregate metrics section

shrirupa:
FE-1266 - Add aggregate metrics section

### What Changed

- Made a reusable component for Aggregate Metrics Section
- Removed the refreshReportBuilder from src/actions/summaryChart
- Localized the getAggregateData api call to the AggregateMetricsSection

### How To Test

- Verifying everything looks good on Report Builder page
- Add the metrics section as in the mocks - https://sparkpost.invisionapp.com/console/share/RU1SP41BG7/483505018
- if no report is pinned (user ui options "pinned_report"), these metrics are default metrics for Summary Report
- if user ui options has a "pinned_report", then these metrics are the one associated with pinned report.

### To Do

- [ ] Address feedback
- [x] Add related cypress tests
- [x] Clicking on reportbuilder link for reports with filters as subaccounts not working ?


src/pages/reportBuilder/ReportBuilder.js
```diff
@@ -73,13 +54,28 @@ export function ReportBuilder({
   }, [reportOptions.metrics]);
 
   useEffect(() => {
-    if (reportOptions.isReady && !isEmpty) {
-      refreshReportBuilder({
+    if (reportOptions.isReady && !isEmpty && chart.groupBy !== 'aggregate') {
+      let updates = {
         ...reportOptions,
         filters: reportOptions.filters,
```
shrirupa, line 60: looking for feedback on whether this useEffect can be moved to ReportTable.

src/components/reportBuilder/AggregateMetricsSection.js
```diff
@@ -0,0 +1,63 @@
...
+const MetricDefinition = ({ label, children }) => {
+  return (
+    <LabelValue>
+      <LabelValue.Label>
+        <Box color="gray.600">{label}</Box>
+      </LabelValue.Label>
+      <LabelValue.Value>
+        <Box color="white">{children}</Box>
+      </LabelValue.Value>
+    </LabelValue>
```
nicklemmon, line 18: @shrirupa  Just a heads up, I added a `dark` prop on `LabelValue` yesterday...and unfortunately I think we did almost the exact same thing - made a new component here

shrirupa: Thanks for the heads up! ohh, so AggregateMetricSection is already a separate component ?

shrirupa: found it, `AggregateMetrics`. 🤔  looks like this one depends on reportBuilderContext. I think we don't want dashboard to be wrapped in ReportBuilderContext, so may be we need a more generic component there. what do you think ?

nicklemmon: Yeah that's totally fine if you want to refactor it.

shrirupa:
closing for https://github.com/SparkPost/2web2ui/pull/1945

PR #1933, opened 2020-12-04
Adds Security Badger GitHub Action

nicklemmon:
Adds the Security Badger GitHub Action. Works in tandem with a Slack bot to post messages to `#front-end-guild` once a week with information about security vulnerabilities.


.github/workflows/security-badger.yml
```diff
@@ -0,0 +1,19 @@
+name: Security Badger
+
+on:
+  schedule:
+    # Run every Monday at 2 PM EST (6:00 PM UTC)
```
bkemper, line 5: 😄

.github/workflows/security-badger.yml
```diff
@@ -0,0 +1,19 @@
...
+    - cron: '0 18 * * 1'
+
+jobs:
+  security-vulnerabilities-reminder:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Security Badger
+        uses: nicklemmon/security-badger@v0.0.7-alpha
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
bkemper, line 15: Looks like the README calls for "GITHUB_API_TOKEN".

nicklemmon: 🤦 The README is wrong. Let me fix that.

.github/workflows/security-badger.yml
```diff
@@ -0,0 +1,19 @@
...
+  schedule:
+    # Run every Monday at 2 PM EST (6:00 PM UTC)
+    - cron: '0 18 * * 1'
+
+jobs:
+  security-vulnerabilities-reminder:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Security Badger
+        uses: nicklemmon/security-badger@v0.0.7-alpha
```
bkemper: What do you think about making Security Badger write an error to the Slack channel if it breaks?

nicklemmon: that's not a bad idea - write down it would fail in the GitHub Actions panel - but we wouldn't really know what was going wrong unless we looked there

bkemper: Also, `vulnerabilityAlerts` could provide duplicates.  I think we only want uniques or the latest for each package?

bkemper: > A wild Security Badger appeared!

😆

bkemper (approved):
LGTM.  I left some notes about SB itself.

.github/workflows/security-badger.yml
nicklemmon: Just opened: https://github.com/nicklemmon/security-badger/issues/1

PR #1934, opened 2020-12-07
SA-1868: useSparkPostQuery error handling

PatrickSison:
### What Changed

- useSparkPostQuery error defaults to an empty object instead of undefined if request fails due to network
- Error state added to charts

### How To Test

- Go to report builder
- Block network request for `time-series` in browser
 - Verify error state renders correctly
 - Verify error does not trigger error boundary page and keeps report builder rendered
 - Verify error message

### To Do

- [ ] Address feedback


src/pages/reportBuilder/components/Charts.js
```diff
@@ -129,6 +129,10 @@ export function Charts(props) {
     return <Loading minHeight="200px" />;
   }
 
+  if (chartStatus === 'error') {
+    return <Empty message="Unable to load report" description="Please try again" />;
```
PatrickSison, line 133: UX was good with the design and the padding. Only feedback was the weight of the state match the mocks

src/components/empty/Empty.js
```diff
@@ -28,11 +28,11 @@ function HibanaEmpty({ description, message }) {
       size="100%"
     >
       <Stack space="0">
-        <Text color="gray.900" fontSize="400" fontWeight="500">
+        <Text color="gray.900" fontSize="400" fontWeight="medium">
           {message}
         </Text>
         {description && (
-          <Text color="gray.900" fontSize="400" fontWeight="500">
+          <Text color="gray.900" fontSize="400" fontWeight="normal">
```
PatrickSison, line 35: Changed weight to match the mocks

src/pages/reportBuilder/components/Charts.js
nicklemmon: We should definitely have some Cypress testing to check for this state - should be fairly straight forward if you just stub the response as a failure (I usually do a `400` status code)

src/hooks/useSparkPostQuery/useSparkPostQuery.js
```diff
@@ -29,8 +29,9 @@ export default function useSparkPostQuery(queryFn, config = {}) {
 }
 
 function handleError({ error, method, queryCache, auth, dispatch }) {
-  const { response } = error;
+  const { response = {} } = error;
```
nicklemmon, line 32: Good call!

src/components/empty/Empty.js
nicklemmon: Nice.

src/pages/reportBuilder/components/Charts.js
nicklemmon, line 133: I don't think the `Empty` component is meant to render within a `Panel.Section` like this. There should be no white space around the gray area:

Screen Shot 2020-12-07 at 11 33 55 AM

nicklemmon (changes requested):
@PatrickSison Just a couple of things:
1. The `Empty` should not be within a `Panel.Section` - this unfortunately might be a little painful
2. We should have some testing around this state

src/pages/reportBuilder/components/Charts.js
PatrickSison: I asked UX about the padding, and they were cool with the padding. I'll double check again though if that's their preference

nicklemmon: Hmm, this was something they previously said was a bug - we even did a ticket or two to fix this problem in other places in the app as a result - so it would be inconsistent.

nicklemmon: Not to mention, in other features we don't use the empty component but the `ApiErrorBanner` when a request fails.

src/pages/reportBuilder/components/Charts.js
```diff
@@ -134,29 +134,31 @@ export function Charts(props) {
...
-            {...formatters}
-            yTickFormatter={chart.yAxisFormatter}
-            yLabel={chart.label}
-            tooltipValueFormatter={chart.yAxisFormatter}
-            showXAxis={index === charts.length - 1}
-          />
-        </Box>
-      ))}
-    </Stack>
+    <Box padding={[400, null, 500]}>
```
nicklemmon, line 137: Mind explaining this? Not really sure why the padding would go away sometimes...I'm sure  there is a good reason though.

nicklemmon (changes requested):
@PatrickSison Still seeing some padding issues and this test isn't passing for me locally - we spoke about it in Slack but just wanted to officially add that as review feedback for tracking purposes.

nicklemmon (approved):
@PatrickSison Looks good to me. As always, appreciate all the back and forth.

PR #1939, opened 2020-12-08
SA-1812: Send a Test Scheduled Report

kchu1337:
### What Changed

- Added the sent test scheduled report modal
- Added action/reducers
- Added cypress tests
- Added Nav selection that was missed in previous tickets.
- Added breadcrumb action that was missed in previous tickets.
- Changed how the recipients list if formatted.
- Added new AsyncActionModal

### How To Test
 - Go to a scheduled report page (go to signals/analytics -> select a report -> click on schedule report button -> either select an existing report or click add schedule.
 - Click on the send test button
 - In the modal, select a recipient and click send test.


src/config/routes/signals.js
```diff
@@ -128,13 +128,17 @@ const hibanaReportsRoutes = [
     component: reportBuilder.ScheduledReportCreatePage,
     condition: isAccountUiOptionSet('allow_scheduled_reports'),
     layout: App,
```
kchu1337, line 130: Added this since it was in the mocks but not previously implemented.

src/pages/reportBuilder/components/ScheduledReportForm.js
```diff
@@ -126,11 +182,13 @@ export const ScheduledReportTimingForm = ({
...
+  const { control, errors, formState, register, setValue } = formControls;
 
-  const timingFormValue = watch('timing');
-  const periodFormValue = watch('period');
-  const timezoneFormValue = watch('timezone');
+  const {
+    period: periodFormValue,
+    timezone: timezoneFormValue,
+    timing: timingFormValue,
+  } = useWatch({ control, name: ['period', 'timezone', 'timing'] });
```
kchu1337, line 191: Docs say useWatch has better performance but only re-renders at the components level. https://react-hook-form.com/api/#useWatch

src/pages/reportBuilder/components/ScheduledReportForm.js
```diff
@@ -28,8 +35,41 @@ export const ScheduledReportDetailsForm = ({
   users,
 }) => {
   const history = useHistory();
+  const [testSendRecipients, setTestSendRecipients] = useState([]);
```
kchu1337, line 38: Not sure if this is the best approach or if I should just use another form. IDK, this seems easier.

src/pages/reportBuilder/components/ScheduledReportForm.js
```diff
@@ -45,77 +85,102 @@ export const ScheduledReportDetailsForm = ({
...
+                  >
+                    Cancel
+                  </Button>
+                </ButtonWrapper>
+              </Panel.Section>
+            )}
+          </Panel>
+        </Layout.Section>
+      </Layout>
+      <ConfirmationModal
```
kchu1337, line 166: Should I use this? Technically, it's not what the component is meant for.

bkemper:
General comment, why is reportId passed as a path and query string param?

Screen Shot 2020-12-10 at 9 54 56 AM

src/actions/reports.js
```diff
@@ -93,3 +93,14 @@ export function updateScheduledReport({ reportId, scheduleId, data }) {
     },
   });
 }
+
+export function testScheduledReport(reportId, data) {
```
bkemper, line 97: These arguments don't help me understand this endpoint.  What data is being sent?  What is required?

src/pages/reportBuilder/components/ScheduledReportForm.js
```diff
@@ -45,77 +85,102 @@ export const ScheduledReportDetailsForm = ({
...
+            </Panel.Section>
+            {isUpdatingScheduledReport && (
+              <Panel.Section>
+                <ButtonWrapper>
+                  <Button type="submit" variant="primary" disabled={!formState.isDirty || disabled}>
+                    Update Details
+                  </Button>
+                  <Button
+                    variant="secondary"
+                    onClick={() => history.push('/signals/analytics')}
```
bkemper, line 155: Is `useHistory` needed?  Can't we use `<PageLink />`?

```
<PageLink as={Button} variant="secondary" to="/signals/analytics" disabled={disabled}> 
 Cancel
</PageLink>
```

src/pages/reportBuilder/components/ScheduledReportForm.js
```diff
@@ -14,11 +17,15 @@ import {
...
+import { showAlert } from 'src/actions/globalAlert';
+import { ButtonWrapper, ConfirmationModal, RadioButtonGroup } from 'src/components';
 import { hasAtLeastOneRecipient } from '../helpers/scheduledReports';
 import { DAY_OF_WEEK_OPTIONS, WEEK_OPTIONS } from '../constants/scheduledReports';
+import { ComboBoxTypeahead } from 'src/components/typeahead/ComboBoxTypeahead';
 import { ComboBoxTypeaheadWrapper } from 'src/components/reactHookFormWrapper';
 import { TimezoneTypeahead } from 'src/components/typeahead/TimezoneTypeahead';
+import useModal from 'src/hooks/useModal';
 
 export const ScheduledReportDetailsForm = ({
```
bkemper, line 30: What is the benefit of having `ScheduledReportDetailsForm` and `ScheduledReportTimingForm` in the same file?

bkemper:
Another general feedback item, the "Week", "Day", and "Timezone" fields should be top aligned. 

Screen Shot 2020-12-10 at 11 23 27 AM

src/pages/reportBuilder/components/ScheduledReportForm.js
bkemper: This approach persists the state if a user clicks cancel.  I would vote for a separate form.

src/pages/reportBuilder/components/ScheduledReportForm.js
```diff
@@ -45,77 +85,102 @@ export const ScheduledReportDetailsForm = ({
...
+                label="Email Subject"
+                name="subject"
+                helpText="Text which will appear as subject line in report email"
+                id="email-subject"
+                error={errors.subject?.message}
+              />
+            </Panel.Section>
+            <Panel.Section>
+              <Panel.Action>
+                <ButtonLink variant="tertiary" onClick={openModal} disabled={disabled}>
```
bkemper, line 135: The placement of this "Send Test" button is really confusing.  I see in the mocks that it is going to be with attachment type eventually.  In the meantime, I think we need to find a new home for it.  It is next to "Sent To", but doesn't use those values.

Screen Shot 2020-12-10 at 11 33 02 AM
Screen Shot 2020-12-10 at 11 33 07 AM

src/pages/reportBuilder/components/ScheduledReportForm.js
bkemper: Probably not, especially since it is using a bunch of legacy components.  We should use the new Modal provided by Matchbox.

src/pages/reportBuilder/components/ScheduledReportForm.js
```diff
@@ -45,77 +85,102 @@ export const ScheduledReportDetailsForm = ({
...
+                    Cancel
+                  </Button>
+                </ButtonWrapper>
+              </Panel.Section>
+            )}
+          </Panel>
+        </Layout.Section>
+      </Layout>
+      <ConfirmationModal
+        confirming={testSendRecipients.length < 1}
```
bkemper, line 167: We are missing the modal title.

Screen Shot 2020-12-10 at 11 46 07 AM

src/pages/reportBuilder/components/ScheduledReportForm.js
kchu1337: I think now we can. Was related to this bug: https://github.com/SparkPost/matchbox/pull/705

src/pages/reportBuilder/components/ScheduledReportForm.js
```diff
@@ -45,77 +85,102 @@ export const ScheduledReportDetailsForm = ({
...
+        open={isModalOpen}
+        confirmVerb="Send Test"
+        isPending={testScheduledReportStatus === 'loading'}
+        onConfirm={onConfirmTestSend}
+        onCancel={closeModal}
+      >
+        <ComboBoxTypeahead
+          id="test-send"
+          itemToString={item => (item ? `Name: ${item.name} ---- Email: ${item.email}` : '')}
```
bkemper, line 176: Do we need name and email?  Name is not a field on the invite user form, so I would bet that the majority of the time name is going to be username which is an ugly unique identifier.

If we must use name and email, we don't have to invent a format.  We should use the RFC format, see [Addresss specification](https://tools.ietf.org/html/rfc2822#section-3.4).

```
Name <my@email.com>
```

src/pages/reportBuilder/components/ScheduledReportForm.js
kchu1337: I think it was. UX ask that the test recipients be separate from the actual recipients 🤷

src/pages/reportBuilder/components/ScheduledReportForm.js
```diff
@@ -28,8 +35,41 @@ export const ScheduledReportDetailsForm = ({
...
+  const { closeModal, isModalOpen, openModal } = useModal();
+  const dispatch = useDispatch();
+
+  const onConfirmTestSend = () => {
+    const { name, subject } = formValues;
+    if (!name || !subject) {
+      return dispatch(
+        showAlert({
+          type: 'error',
+          message: `Please fill out Scheduled Report Name and Email Subject in the form`,
```
bkemper, line 50: Is this the desired experience?  I would have to close the modal, update the fields, re-open the modal, etc.

Screen Shot 2020-12-10 at 11 38 18 AM

src/pages/reportBuilder/components/ScheduledReportForm.js
kchu1337: I can certainly use that format. I had cut another ticket to add a custom render function but this looks better in the meantime. 
Also, it looks like there is some way to set name? 
image

src/config/routes/signals.js
```diff
@@ -128,13 +128,17 @@ const hibanaReportsRoutes = [
...
+    category: 'Signals Analytics',
+    subcategory: 'Analytics Report',
     title: 'Schedule Report',
   },
   {
     path: '/signals/schedule/:reportId/:scheduleId',
     component: reportBuilder.ScheduledReportEditPage,
     condition: isAccountUiOptionSet('allow_scheduled_reports'),
```
bkemper, line 138: Is there going to be authorization on this page or is it going to be available for all?

src/pages/reportBuilder/components/ScheduledReportForm.js
kchu1337: Although one reason I actually like persisting is that the confirm button is covered by the typeahead dropdown. Pressing escape or clicking outside the modal would close both. Personally, I've done that a couple of times.

src/config/routes/signals.js
kchu1337: Doesn't 
```    condition: isAccountUiOptionSet('allow_scheduled_reports'),
```

src/pages/reportBuilder/components/ScheduledReportForm.js
```diff
@@ -45,77 +85,102 @@ export const ScheduledReportDetailsForm = ({
...
+        open={isModalOpen}
+        confirmVerb="Send Test"
+        isPending={testScheduledReportStatus === 'loading'}
+        onConfirm={onConfirmTestSend}
+        onCancel={closeModal}
+      >
+        <ComboBoxTypeahead
+          id="test-send"
+          itemToString={item => (item ? `Name: ${item.name} ---- Email: ${item.email}` : '')}
+          label="Send To"
```
bkemper, line 177: This field label, "Send To", is misleading.  It gives me the impression that I can send to anyone, but in reality, I can only send to other users.

src/pages/reportBuilder/components/ScheduledReportForm.js
```diff
@@ -45,77 +85,102 @@ export const ScheduledReportDetailsForm = ({
...
+          onChange={setTestSendRecipients}
+          results={users}
```
bkemper, line 179: This design is problematic.  I'm assuming reporting users are going to want to create scheduled reports.  However, they don't have the user/manage grant to view the list of users in their account.

Screen Shot 2020-12-10 at 12 19 42 PM

bkemper (commented):
Left a bunch of comments.

src/pages/reportBuilder/components/ScheduledReportForm.js
kchu1337: It used to be one form. The edit page changes required it to be split into 2 forms. Just kept it in 1 file so that there was less refactoring needed. If it makes a big difference, I can move it into 2 files.

src/components/modals/AsyncActionModal.js
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
```
kchu1337, line 1: Kinda a copy of ConfirmationModal. Mostly the same but using the new modal components.

src/components/modals/tests/AsyncActionModal.test.js
```diff
@@ -0,0 +1,109 @@
+import React from 'react';
```
kchu1337, line 1: Pretty much a copy+paste version of confirmation modal tests

src/pages/reportBuilder/components/ScheduledReportForm.js
```diff
@@ -168,7 +246,7 @@ export const ScheduledReportTimingForm = ({
             </Radio.Group>
           </Panel.Section>
           <Panel.Section>
-            <Inline>
+            <Inline alignY="top">
```
kchu1337, line 249: Right now, this does nothing. But in a future matchbox update, this will be included as a prop.

shrirupa (commented):
Few of the things - 
- Extra space between the combobox and the Modal Footer
- Modal is persisting data if the Modal is closed without sending
- Send test is not underlined in the mock [Panel.Action is also a button internally and I think accepts all the props of a Button so we can user that instead of button link]

shrirupa (approved):
Looks good to me! 

As we discussed on the call, the availability of this page for different roles is to be tested in the release ticket and there is no existing solution for the space issue of comboxbox and Modal footer

PR #1941, opened 2020-12-08
SA-1825: Fixed precision type for charts

PatrickSison:
### What Changed

Fixed precision behavior (15 and 5 minutes change for charts)


src/helpers/metrics.js
```diff
@@ -181,7 +180,13 @@ export function getMomentPrecisionByPrecision(precision) {
 }
 
 export function getPrecisionType(precision) {
-  return indexedPrecisions[precision].time <= 60 * 24 * 2 ? 'hour' : precision;
+  switch (precision) {
```
PatrickSison, line 183: Removed the back and forth mapping

kchu1337 (approved):

PR #1945, opened 2020-12-09
FE-1266 - Add aggregate metrics section

shrirupa:
FE-1266 - Add aggregate metrics section

### What Changed

- Used AggregateMetrics and CompareByAggregateMetrics  for displaying metrics on dashboard
- Removed useReportBuilderContext from AggregateMetrics and CompareByAggregateMetrics and used props instead
- updated usePinnedReport to get the correct reportOption when report is changed.

### How To Test
- Verify with the report builder page that metrics for all the reports look correct on dashboard too.
- Verify with the report builder page that chart for all the reports look correct on dashboard too.
- Add the metrics section as in the mocks - https://sparkpost.invisionapp.com/console/share/RU1SP41BG7/483505018
- if no report is pinned (user ui options "pinned_report"), these metrics are default metrics for Summary Report
- if user ui options has a "pinned_report", then these metrics are the one associated with pinned report.

### To Do

- [ ] Address feedback
- [x] Move AggregateMetrics and CompareByAggregateMetrics into src/components/reportBuilder


src/pages/dashboardV2/DashboardPageV2.js
```diff
@@ -105,9 +130,22 @@ export default function DashboardPageV2() {
                       <TranslatableText>Change Report</TranslatableText> <Sync size={25} />
                     </Panel.Action>
                   </Panel.Header>
-                  <Panel.Section>
-                    <ChartGroups reportOptions={pinnedReport.options} p="0" />
-                  </Panel.Section>
+                  {!pinnedReport.loading && (
+                    <Panel.Section>
+                      <ChartGroups reportOptions={pinnedReport.options} p="0" />
+                    </Panel.Section>
```
nicklemmon, line 136: Screen Shot 2020-12-11 at 11 03 58 AM

I don't think the `Panel.Section` should be used here. On the `/signals/analytics` route, it renders without the extra padding:
Screen Shot 2020-12-11 at 11 05 45 AM

src/pages/dashboardV2/DashboardPageV2.js
```diff
@@ -105,9 +130,22 @@ export default function DashboardPageV2() {
...
+                  )}
+                  {pinnedReport.options.comparisons.length > 0 ? (
+                    <CompareByAggregatedMetrics
+                      date={dateValue}
+                      reportOptions={pinnedReport.options}
+                    />
+                  ) : (
+                    <AggregatedMetrics
+                      date={dateValue}
+                      processedMetrics={getMetricsFromKeys(pinnedReport?.options?.metrics)}
```
nicklemmon, line 146: Is there a risk these would be undefined?

src/pages/dashboardV2/DashboardPageV2.js
```diff
@@ -67,10 +73,29 @@ export default function DashboardPageV2() {
...
+        _getAggregateDataReportBuilder({
+          ...pinnedReport.options,
+          filters: pinnedReport.options.filters,
+        }),
+      );
+  }, [dispatch, pinnedReport.loading, pinnedReport.options, prevReportOptions]);
+
+  const dateValue = `${moment(pinnedReport?.options?.from).format('MMM Do')} - ${moment(
+    pinnedReport?.options?.to,
+  ).format('MMM Do, YYYY')}`;
```
nicklemmon, line 97: Could this logic be moved in to `useDashboardContext()`? I think that was my intention originally - to try and keep the business logic distinct from the UI.

src/pages/dashboardV2/DashboardPageV2.js
```diff
@@ -30,6 +31,11 @@ import Sidebar from './components/Sidebar';
 import { LINKS } from 'src/constants';
 import { useModal } from 'src/hooks';
 import ChangeReportModal from './components/ChangeReportModal';
```
nicklemmon, line 33: Unrelated to this ticket, but I noticed this modal isn't animating when opened. I think it's because it's being conditionally rendered?
```
{isModalOpen && (
  <ChangeReportModal open={isModalOpen} onClose={closeModal} reports={allReports} />
)}
```

You _should_ just be able to do this:
```
<ChangeReportModal open={isModalOpen} onClose={closeModal} reports={allReports} />
```

src/pages/dashboardV2/DashboardPageV2.js
```diff
@@ -30,6 +31,11 @@ import Sidebar from './components/Sidebar';
...
+import moment from 'moment';
```
nicklemmon, line 34: Can we use `date-fns` instead? I believe we are going to make an effort to remove `moment` in 2021 so it would be helpful to avoid using it.

src/pages/dashboardV2/DashboardPageV2.js
```diff
@@ -105,9 +130,22 @@ export default function DashboardPageV2() {
                       <TranslatableText>Change Report</TranslatableText> <Sync size={25} />
                     </Panel.Action>
                   </Panel.Header>
-                  <Panel.Section>
-                    <ChartGroups reportOptions={pinnedReport.options} p="0" />
-                  </Panel.Section>
+                  {!pinnedReport.loading && (
+                    <Panel.Section>
+                      <ChartGroups reportOptions={pinnedReport.options} p="0" />
```
nicklemmon, line 135: Since this is a "global" component now, you could move it to co-exist alongside the aggregated metrics components

src/pages/dashboardV2/DashboardPageV2.js
```diff
@@ -105,9 +130,22 @@ export default function DashboardPageV2() {
...
+                    </Panel.Section>
+                  )}
+                  {pinnedReport.options.comparisons.length > 0 ? (
+                    <CompareByAggregatedMetrics
+                      date={dateValue}
+                      reportOptions={pinnedReport.options}
+                    />
+                  ) : (
+                    <AggregatedMetrics
```
nicklemmon, line 144: Screen Shot 2020-12-11 at 11 18 51 AM

It looks like the colors aren't quite aligning here - these appear to be the OG colors for some reason?

src/pages/dashboardV2/DashboardPageV2.js
```diff
@@ -105,9 +130,22 @@ export default function DashboardPageV2() {
...
-                  <Panel.Section>
-                    <ChartGroups reportOptions={pinnedReport.options} p="0" />
-                  </Panel.Section>
+                  {!pinnedReport.loading && (
+                    <Panel.Section>
+                      <ChartGroups reportOptions={pinnedReport.options} p="0" />
+                    </Panel.Section>
+                  )}
+                  {pinnedReport.options.comparisons.length > 0 ? (
+                    <CompareByAggregatedMetrics
```
nicklemmon, line 139: It looks like UX wants the layout to be slightly different on the dashboard. Could potentially be a prop that controls it? Like `isStacked` or `stacked` or `collapsed` or ??

Screen Shot 2020-12-11 at 11 22 49 AM

src/hooks/usePinnedRport/usePinnedReport.js
```diff
@@ -57,9 +79,8 @@ export default function usePinnedReport(onboarding) {
...
       filters: hydrateFilters(summaryReportOptions.filters, { subaccounts }),
     });
-    pinnedReport.linkToReportBuilder = getLinktoReportBuilder(
-      _.omitBy(pinnedReport.options, (_value, key) => excludeOptionsFromLink.includes(key)),
-    );
+    pinnedReport.linkToReportBuilder = `/signals/analytics?${summaryReportQueryString}`;
   }
+
   return { pinnedReport };
 }
```
nicklemmon, line 86: Glad to see this custom hook here.

When I link to a saved report it isn't populated in the `<select />` as I would expect? Not sure if that's on your end here or on our end really:
Screen Shot 2020-12-11 at 11 30 30 AM

src/hooks/usePinnedRport/usePinnedReport.js
```diff
@@ -1,35 +1,39 @@
...
+  const { list: subaccounts, listLoading: subaccountsLoading } = useSelector(
+    state => state.subaccounts,
+  );
+  const pinnedReportId = useSelector(state =>
+    selectCondition(isUserUiOptionSet('pinned_report_id'))(state),
+  );
+
   useEffect(() => {
     if (onboarding === 'analytics') {
+      isFirstRender.current = false;
```
nicklemmon, line 31: Mind explaining this a bit? Might be helpful to just add some comments describing the requirement you're addressing by doing this.

nicklemmon (changes requested):
@shrirupa First pass is done - let me know what questions you have. Overall things largely work as expected, though I did find a few things worth looking at.

src/pages/dashboardV2/DashboardPageV2.js
shrirupa: have we finalized upon date-fns ?

src/hooks/usePinnedRport/usePinnedReport.js
shrirupa: I will add the comment, reason being loadingState for api is false on first render and hence it triggers the getAggregateTable action without the correct pinnedReport.options

src/pages/dashboardV2/DashboardPageV2.js
shrirupa: ahh, good catch! looks like getMetricsFromKeys needed true as second argument to get right colors.

src/pages/dashboardV2/DashboardPageV2.js
shrirupa: I think they want the layout to this only different when View filters button shows up otherwise it is same as Report Builder. That "View Filters"button(added in another PR) is rendered conditionally only when pinned report has associated filters to it.

src/hooks/usePinnedRport/usePinnedReport.js
shrirupa: fixed it! report parameter was being added twice

shrirupa: some saved reports seem to have report param in query_string and some don't 🤷

src/pages/dashboardV2/DashboardPageV2.js
nicklemmon: Well, I know are for sure moving away from `moment` and don't have any plans to move away from `date-fns`.

src/pages/dashboardV2/DashboardPageV2.js
nicklemmon: Really? 🤔  That seems off to me.

src/actions/summaryChart.js
```diff
@@ -107,6 +107,60 @@ export function _getAggregateData({ params, metrics, isCurrentGroupingAggregates
...
+    const metrics = merged.metrics;
+
+    let params = getQueryFromOptionsV2(merged);
+
+    // Get selected metrics
+    const activeMetrics = metrics || state.summaryChart.metrics;
+
+    // Gets filters and metrics for params
+    if (!params) {
+      params = getQueryFromOptions({ ...state.reportOptions, metrics: activeMetrics });
```
nicklemmon, line 137: Should we be mixing and matching these two helpers? I thought V2 was for use with the newer/Hibana chart UI.

src/hooks/usePinnedRport/usePinnedReport.js
```diff
@@ -42,11 +46,32 @@ export default function usePinnedReport(onboarding) {
...
+      precision: 'hour',
+      ...options,
+      isReady: true,
+      filters: hydrateFilters(options.filters, { subaccounts }),
+    });
+
+    pinnedReport.linkToReportBuilder = report.query_string.includes('report=')
+      ? `/signals/analytics?${report.query_string}`
+      : `/signals/analytics?${report.query_string}&report=${pinnedReportId}`;
+  } else {
```
nicklemmon, line 74: I don't see tests supporting a lot of these changes? It would be worth adding more to ensure this doesn't break in the future.

src/pages/dashboardV2/DashboardPageV2.js
```diff
@@ -67,10 +76,29 @@ export default function DashboardPageV2() {
...
+    if (!pinnedReport.loading && !_.isEqual(prevReportOptions, pinnedReport.options))
+      dispatch(
+        _getAggregateDataReportBuilder({
+          ...pinnedReport.options,
+          filters: pinnedReport.options.filters,
+        }),
+      );
+  }, [dispatch, pinnedReport.loading, pinnedReport.options, prevReportOptions]);
+
+  const dateValue = `${moment(pinnedReport?.options?.from).format('MMM Do')} - ${moment(
```
nicklemmon, line 98: Thinking it over again, I think this could probably be extracted as a helper function, actually - since it's now in use in two places that are quite far apart from one another.

nicklemmon (changes requested):
@shrirupa Have some additional feedback - things seem to work well overall and you addressed most of my initial concerns.

I'm a little concerned that some of the logic in the feature is not covered by automated testing. There is a lot going on here, and since we plan to do a fairly significant refactor of report builder charts in the near future, I think we would feel safer making those sorts of changes with more testing in place.

Nice work so far!

src/actions/summaryChart.js
shrirupa: good point, I don't think we will need getQueryFromOptions here.

nicklemmon (approved):
@shrirupa This looks good to me. Let me know when you would like to get it out.

PR #1948, opened 2020-12-10
SA-1849 - (UI) Compare By - Bounce Report

nicklemmon:
[SA-1849](https://sparkpost.atlassian.net/browse/SA-1849)

### What Changed

- Adds new logic to handle dynamic tab rendering when there are active comparisons
- Adds new custom `usePrepareQuery` to prepare deliverability requests based on current report builder context - updates `CompareByAggregatedMetrics` to use this hook.
- Adds new `getComparisonArguments` helper
- Renames table components within `src/pages/reportBuilder/components/tabs`

### How To Test

- Go to report builder with the compare by UI flag enabled
- Add several comparisons without changing metrics
- Verify the default "Bounce Reason" tab does not render
- Verify that a tab for each comparison renders and makes the correct requests - _I could use some help with this - I think I did this right?_

### Questions

- **Should the tabs logic be extracted in to a custom hook?** This should be feasible, except for the unusual `setShowTable` side effect baked-in. This could also be baked in to report builder context to centralize things? Or we just don't worry about it until we do our refactoring in Q1.
- **Should `usePrepareQuery` have a different/more-specific name?**

### To Do

- [x] Handle API errors by adding relevant content and retries
- [x] Add supporting tests for API errors
- [x] Rename API helpers in `src/helpers/api` and remove imports in to the root index file
- [x] Rename existing components within `src/pages/reportBuilder/tabs` to be consistent with `BounceReportTab`
- [X] Ensure "Count %" is rendering correctly in the table - doesn't appear to be at the moment.


src/selectors/bounceReport.js
```diff
@@ -3,51 +3,46 @@ import { getBandTypes, reshapeCategories, formatAggregates } from 'src/helpers/b
...
+const selectClassifications = state => _.get(state, 'bounceReport.classifications', []);
+export const selectReasons = ({ bounceReport }) =>
+  bounceReport.reasons.filter(reason => reason.count_bounce > 0);
+const selectAdminReasons = ({ bounceReport }) =>
+  bounceReport.adminReasons.filter(reason => reason.count_admin_bounce > 0);
 
-const selectFormattedAggregates = createSelector(
-  [selectAggregates],
-  (aggregates) => {
+export const selectFormattedAggregates = createSelector([selectAggregates], aggregates => {
```
nicklemmon, line 21: Exported two of these, the rest of Prettier formatting.

src/pages/reportBuilder/ReportBuilder.js
```diff
@@ -196,49 +251,70 @@ export function ReportBuilder({
         {isEmpty ? (
           <Empty message="No Data" description="Must select at least one metric." />
         ) : (
-          <>
-            <div data-id="summary-chart">
-              <Tabs defaultTabIndex={0} forceRender tabs={tabs}>
+          <div data-id="summary-chart">
+            <Tabs defaultTabIndex={0} forceRender tabs={tabs}>
```
nstanard, line 255: Seeing inconsistent behavior with the tab list. At first when I added 8+ comparisons the tabs listed out and the page layout/scrolling was broken. Then after resizing the window to mobile and back to desktop the tabs showed in the `...` dropdown (like I'd expect) but the layout/scrolling was still broken. 

image

I'm trying to think of how we want to handle the UI when the tab count grows beyond the available horizontal space.  

image

overflow scrolling the tab list container could be a cheap and easy solution. But even better (IMO) would be if the tabs component recognized the tabs go out of the container width, and adjust to have the `...` dropdown just like the mobile view has. 

image

src/pages/reportBuilder/ReportBuilder.js
```diff
@@ -122,30 +123,84 @@ export function ReportBuilder({
...
-      [
-        { content: 'Report', onClick: () => setShowTable(true) },
-        hasBounceTab && { content: 'Bounce Reason', onClick: () => setShowTable(false) },
-        hasRejectionTab && { content: 'Rejection Reason', onClick: () => setShowTable(false) },
-        hasDelayTab && { content: 'Delay Reason', onClick: () => setShowTable(false) },
-        hasLinksTab && { content: 'Links', onClick: () => setShowTable(false) },
-      ].filter(Boolean),
-    [hasBounceTab, hasRejectionTab, hasDelayTab, hasLinksTab],
-  );
+  const tabs = useMemo(() => {
```
nstanard, line 143: Perhaps outside the scope of this work but I think we should consider putting a limit on the number of things the user can compare. (maybe this is already a consideration and I'm just late to the party)

nstanard: One reason I'm thinking we should consider this is some of the interactions on the page visibly slow down a little. But especially when having multiple comparison. I'm going to do a production build locally to see if that cuts down on the click handler event time.

src/pages/reportBuilder/ReportBuilder.js
nstanard: image

I was a little confused by the UX/Interactions of the tabs at first. The way the "Report" tab shows the graphs and each individual tab shows the table data. This wasn't confusing for me when the tabs were listed out and visible in a row. But when the tab list grows and the UI changes to show them in the dropdown I think we might want to keep the "Report" Tab fixed outside the list, and only the comparisons are listed in the dropdown (if the UI is displaying with the dropdown list)

src/pages/reportBuilder/ReportBuilder.js
```diff
@@ -196,49 +251,70 @@ export function ReportBuilder({
...
+                  <LinksTab />
+                </Tabs.Item>
+              )}
+
+              {hasActiveComparisons
+                ? reportOptions.comparisons.map((comparison, comparisonIndex) => {
+                    return (
+                      <Tabs.Item key={`tab-${comparison.value}-${comparisonIndex}`}>
+                        {hasBounceMetrics ? (
+                          <BounceReasonComparisonTab comparison={comparison} />
```
nstanard, line 295: I'm seeing our (Table) Collection component re-render a lot without me actually doing anything on the page. I can have the page open and click on a different browser window and back... and our collection component re-renders multiple times. This was something I started suspecting was a problem while investigating on [FE-1284](https://sparkpost.atlassian.net/browse/FE-1284). I could absolutely be wrong but something seems off here. Could be part of the problem with the issue I wrote up for FE-1284.

image

nstanard (commented):
First pass, hitting submit so I dont lose this info. Picking this back up on Monday.

src/pages/reportBuilder/ReportBuilder.js
nicklemmon: I believe in this case that's due to `react-query` refetching from the API in the background on window focus?

src/pages/reportBuilder/ReportBuilder.js
nicklemmon: There is a ticket for that, fortunately! Though only recently. You're totally right:
https://sparkpost.atlassian.net/browse/SA-1893

src/pages/reportBuilder/ReportBuilder.js
nicklemmon: Unfortunately this isn't something we have control over - entirely a Matchbox thing. I think you're right that the UX is a tad confusing.

In my mind, maybe even moving *everything* to a dropdown would make more sense than just "all of the tabs except the one I'm viewing".

nicklemmon: I spoke with Jon and he is going to adjust the `Tabs` to handle overflowing content without relying on a window resize occurring, so he will address that, but your (valid) UX concern will remain.

src/pages/reportBuilder/hooks/usePrepareQuery.js
```diff
@@ -0,0 +1,25 @@
...
+import {
+  getMetricsFromKeys,
+  getQueryFromOptionsV2 as getQueryFromOptions,
+} from 'src/helpers/metrics';
+import { useReportBuilderContext } from '../context/ReportBuilderContext';
+
+/**
+ * Prepares options for requests based on the current state of the page
+ */
+export default function usePrepareComparisonRequest() {
```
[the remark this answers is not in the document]
nicklemmon: Good catch, I meant to keep that consistent. This is the danger of default exports.

src/pages/reportBuilder/ReportBuilder.js
nstanard: Did we want to put a flag on this work until that `Tabs` component gets fixed? The scrolling on the page was broken in both situations when I had it loaded up locally.

src/pages/reportBuilder/ReportBuilder.js
nstanard: Awesome, I won't worry about it then.

src/pages/reportBuilder/ReportBuilder.js
nstanard: I wanted/needed to dig into `react-query` anyway (to get familiar) so I'll double check this in the morning tomorrow.

src/pages/reportBuilder/ReportBuilder.js
nstanard: If you want to rebase this I can do another pass tomorrow. We can probably just fix the page scrolling issue locally instead of waiting on the matchbox fix. Whatever you feel is right.

nicklemmon: it is behind a feature flag - not to worry. I would rather wait on Matchbox - not to mention we have 5-6 more tickets before this epic wraps anyways so it will be a little while longer before a customer sees it.

src/components/reportBuilder/CompareByAggregatedMetrics.js
```diff
@@ -45,26 +41,14 @@ export default function CompareByAggregatedMetrics({ date, reportOptions }) {
...
-      ...reportOptions,
-      metrics: formattedMetrics,
-    });
-  }, [reportOptions, formattedMetrics]);
-  const requestOptions = {
-    ...formattedOptions,
-    [comparisonObj.value]: comparisonObj.value === 'subaccounts' ? comparison.id : comparison.value, // Subaccount formatting means different data must be passed to the request
+function ComparisonRow({ comparison, reportOptions, hasDivider }) {
+  const sharedArguments = usePrepareReportBuilderQuery(reportOptions);
+  const comparisonArguments = getComparisonArguments(comparison);
```
PatrickSison, line 46: I don't think this ends up being a valid way to use comparison right? If `query_filters` exists, it overwrites the filters (where in this case, this is using the old filters to the API request). I think if the combination of filters + comparisons exist, it appends the filter on top of the other filters before the request (as an independent grouping)

src/hooks/usePrepareReportBuilderQuery/usePrepareReportBuilderQuery.js
```diff
@@ -0,0 +1,23 @@
+import { useMemo } from 'react';
+import {
+  getMetricsFromKeys,
+  getQueryFromOptionsV2 as getQueryFromOptions,
+} from 'src/helpers/metrics';
+
+/**
+ * Prepares options for requests based on the current state of the page
+ */
+export default function usePrepareReportBuilderQuery(reportOptions) {
```
PatrickSison, line 10: 👏 At least in the mean time. Hopefully we can clean up some of these formatters, but big 👍

src/pages/reportBuilder/ReportBuilder.js
```diff
@@ -121,30 +122,84 @@ export function ReportBuilder({
     return quantity >= limit;
   }, [subscription]);
 
-  const hasBounceTab = processedMetrics.some(({ key }) => {
+  const hasBounceMetrics = processedMetrics.some(({ key }) => {
```
PatrickSison, line 125: Maybe we should move some of these functions later on. This is growing to be a really big file

PatrickSison: I suppose not for just the sake of a line count though. But it is becoming a monolith of a file for a top level component

nicklemmon: Definitely - I think most of this could live in `useReportBuilderContext()`

src/pages/reportBuilder/ReportBuilder.js
```diff
@@ -195,52 +249,73 @@ export function ReportBuilder({
...
+              )}
+
+              {hasLinksTab && (
+                <Tabs.Item>
+                  <LinksTab />
+                </Tabs.Item>
+              )}
+
+              {hasActiveComparisons
+                ? reportOptions.comparisons.map((comparison, comparisonIndex) => {
```
PatrickSison, line 292: Won't this not render tabs for each item? It has 4 tab contents in a tab item?

src/pages/reportBuilder/components/tabs/BounceReasonComparisonTab.js
```diff
@@ -0,0 +1,75 @@
...
+import { Panel } from 'src/components/matchbox';
+import { getMetricsFromKeys, getFilterByComparison } from 'src/helpers/metrics';
+import { getBounceReasonByDomain, getDeliverability } from 'src/helpers/api/metrics';
+import { selectReasons, selectFormattedAggregates } from 'src/selectors/bounceReport';
+import { BounceReasonTable } from '../tables';
+import { usePrepareReportBuilderQuery } from 'src/hooks';
+import { useReportBuilderContext } from '../../context/ReportBuilderContext';
+
+export default function BounceReasonComparisonTab({ comparison }) {
+  const [bounceReasonParams, aggregatesParams] = useRequestArguments(comparison);
```
PatrickSison, line 13: Is the order of these params incorrect? I think the function below has the aggregates as the first return in the tuple

src/pages/reportBuilder/components/tabs/BounceReasonComparisonTab.js
```diff
@@ -0,0 +1,75 @@
...
+    filters: [...existingFilters, comparisonFilter],
+    metrics: deliverabilityMetrics,
+  });
+  const bounceReasonParams = usePrepareReportBuilderQuery({
+    ...reportOptions,
+    filters: [...existingFilters, comparisonFilter],
+    metrics: bounceReasonMetrics,
+  });
+
+  return [aggregatesParams, bounceReasonParams];
```
PatrickSison, line 74: aggregates is the first return in the tuple

src/pages/reportBuilder/components/tabs/BounceReasonComparisonTab.js
nicklemmon: Whoops. That's the risk with this kind of destructuring.

src/pages/reportBuilder/ReportBuilder.js
nicklemmon: AH! I see what you mean.

nicklemmon:
@PatrickSison Ready for another pass.

src/pages/reportBuilder/components/tabs/BounceReasonComparisonTab.js
```diff
@@ -0,0 +1,75 @@
...
+
+  if (isError) {
+    return (
+      <Panel.Section>
+        <ApiErrorBanner reload={handleReload} status="muted" />
+      </Panel.Section>
+    );
+  }
+
+  // This re-structuring using the `bounceReport` key is a holdover from Redux - could we refactor these selectors to be less opinionated?
```
PatrickSison, line 36: Would be nice eventually. Or even with the selectors in react-query v3

cypress/tests/integration/signals-analytics/analytics-report/reportTabs.spec.js
```diff
@@ -246,5 +235,346 @@ if (IS_HIBANA_ENABLED) {
         cy.findByText('No delay reasons to report').should('be.visible');
       });
     });
+
+    describe('the links table', () => {
```
PatrickSison, line 239: Good catch

cypress/tests/integration/signals-analytics/analytics-report/reportTabs.spec.js
```diff
@@ -246,5 +235,346 @@ if (IS_HIBANA_ENABLED) {
...
+          fixture: 'metrics/deliverability/bounce-reason/domain/200.get.json',
+          requestAlias: 'getBounceReasons',
+        });
+        cy.findByRole('tab', { name: 'Bounce Reason Fake Subaccount 1 (ID 101)' }).click();
+
+        cy.wait(['@getBounceReasons', '@getDeliverabilityAgain']).then(xhrs => {
+          const [deliverabilityReq, bounceReasonsReq] = xhrs;
+
+          // Verify the subaccount filters that were already present are in the request
+          cy.wrap(deliverabilityReq.url).should('include', '104');
```
PatrickSison, line 462: Neat. Didn't know you could do this with multiple requests

src/helpers/metrics.js
```diff
@@ -373,3 +373,27 @@ export function average(item, keys = []) {
 export function rate(item, keys = []) {
   return safeRate(item[keys[0]], item[keys[1]]);
 }
+
+/**
+ * Prepares request options/params based on the current state of the page and the passed in comparison object.
+ *
+ * @param {Object} comparison - passed in comparison when the user selects comparisons via "compare by"
+ */
+export function getFilterByComparison(comparison) {
```
PatrickSison, line 382: We could probably extend on top of this and do something like pass in the whole report options, and it prepare an array of report options, each with the comparison converted to the filters. Might be overabstracting though and we fall into the rut of of obscuring logic

PatrickSison (approved):
Looks good

PR #1950, opened 2020-12-11
SA-1620: Group by table comparison filters

PatrickSison:
### What Changed

- Group By table swaps for CompareByGroupByTable (if there are comparison filters)
- Column added to comparison filter
- Added new useSparkPostQueries

### How To Test

- ON report builder with compare by enabled, add at least 2 comparison filters
 - Go to group by table and verify table works as expected
 - Verify that clicking filter links still adds filters to table
 - Verify that table is still sortable by column
 - Verify that default sort is by group by column (no longer first column)
- Block one of the comparison filters request from going through (and verify error state)
 - Block all of the comparison filters requests from going through (and verify error state)

### To Do

- [ ] Add testing
 - [ ] Disable option for group by if select column is used (or maybe just disable 2nd row if it's the same?)


src/helpers/api/metrics.js
```diff
@@ -33,10 +33,10 @@ export function getTemplatesMetrics(params) {
   };
 }
 
-export function getDeliverabilityMetrics(params) {
+export function getDeliverabilityMetrics(params, path) {
   return {
     method: 'GET',
-    url: DELIVERABILITY_BASE_URL,
+    url: `${DELIVERABILITY_BASE_URL}${path ? `/${path}` : ''}`,
```
nicklemmon, line 39: Doesn't the `params` argument handle this rather than interpolating here?

src/pages/reportBuilder/ReportBuilder.js
```diff
@@ -21,8 +21,10 @@ import {
   AggregatedMetrics,
   CompareByAggregatedMetrics,
   ReportOptions,
-  ReportTable,
+  GroupByTable,
+  // CompareByGroupByTable,
```
nicklemmon, line 25: Can this be removed?

src/pages/reportBuilder/components/GroupByTable/CompareByGroupByTable.js
```diff
@@ -0,0 +1,148 @@
...
+  const {
+    selectors: { selectSummaryMetricsProcessed: metrics },
+  } = useReportBuilderContext();
+  const hasSubaccounts = useSelector(hasSubaccountsSelector);
+  const subaccounts = useSelector(state => state.subaccounts.list);
+  const group = GROUP_BY_CONFIG[groupBy];
+
+  const tableData = _.flatten(data);
+
+  const getColumnHeaders = () => {
```
nicklemmon, line 38: I'm not seeing the default sort within the table?
Screen Shot 2020-12-11 at 1 43 02 PM

src/pages/reportBuilder/components/GroupByTable/CompareByGroupByTable.js
```diff
@@ -0,0 +1,148 @@
...
+      return null;
+    }
+
+    if (status === 'loading') {
+      return <PanelLoading minHeight="250px" />;
+    }
+
+    if (!tableData.length) {
+      return (
+        <Panel.LEGACY>
```
nicklemmon, line 108: No need to use `Panel.LEGACY` within this feature.

src/pages/reportBuilder/components/GroupByTable/GroupByTable.js
```diff
@@ -23,7 +23,7 @@ const tableWrapper = props => {
   );
 };
 
-export const ReportTable = () => {
+export const GroupByTable = () => {
```
nicklemmon, line 26: 👏  Much clearer

src/pages/reportBuilder/components/GroupByTable/CompareByGroupByTable.js
```diff
@@ -0,0 +1,148 @@
+import React from 'react';
```
nicklemmon, line 1: We should have some Cypress testing around all of this? Probably useful to do testing around the multiple requests and their arguments as well as the user's interaction with the table when comparisons are enabled.

src/pages/reportBuilder/components/GroupByTable/useGroupByTable.js
```diff
@@ -0,0 +1,85 @@
+import { useState, useMemo } from 'react';
```
nicklemmon, line 1: This could all be tested via the feature itself rather than Jest unit testing, though unit testing might be a little easier to test independently (as well as edge cases).

We have this installed:
https://github.com/testing-library/react-hooks-testing-library

Up to you how it should be handled.

nicklemmon (changes requested):
@PatrickSison Seems to work well overall. I have a few comments I've left, plus I don't see any testing to support these changes so that's a concern.

src/pages/reportBuilder/components/GroupByTable/CompareByGroupByTable.js
PatrickSison: I think default sort is the sort is uses if no sort is selected. Active sorting works otherwise

src/helpers/api/metrics.js
PatrickSison: Was on the fence with how to handle this, but also it uses the paths `/deliverability/${groupByKey}` to actually route to the appropriate break down table. It feels separate from the rest of the params, but we could pass it in through params and deconstruct just this variable, and send the rest down to the API

nicklemmon: Yeah I think that would be better in my mind - make this as generic/simple as possible and keep complexity elsewhere.

src/pages/reportBuilder/components/GroupByTable/CompareByGroupByTable.js
```diff
@@ -0,0 +1,162 @@
...
+import { useReportBuilderContext } from '../../context/ReportBuilderContext';
+import AddFilterLink from '../AddFilterLink';
+import { ApiErrorBanner } from 'src/components';
+
+import styles from './ReportTable.module.scss';
+import useGroupByTable from './useGroupByTable';
+
+const tableWrapper = props => {
+  return (
+    <Panel.LEGACY>
```
nicklemmon, line 22: Same here, no need to use `Panel.LEGACY`

src/hooks/useSparkPostQuery/useSparkPostQuery.js
```diff
@@ -5,6 +5,33 @@ import { refresh, logout } from 'src/actions/auth';
...
+
+export function useSparkPostQueries(queries, config = {}, queryKey) {
+  const queryCache = useQueryCache();
+  const auth = useSelector(state => state.auth);
+  const dispatch = useDispatch();
+
+  const queryRequests = () =>
+    queries.map(queryFn => {
+      const { url, method, params, headers } = queryFn();
+      return defaultQuery(url, { method, params, headers, auth });
```
nicklemmon, line 18: You shouldn't have to pass things in to the `defaultQuery` manually? From what I understand, this is how it works:
- A `queryKey` can contain lots of request-related information (method, URL, params) which is passed to each instance of `useQuery`
- `useQuery` takes that `queryKey` and passes it to our axios instance, `sparkPostRequest` - our custom `defaultQuery` destructures the `queryKey` and does that for us
- We pass the `defaultQuery` to the `QueryCache` component so that it is always there, consuming/handling passed-in query keys without us having to manually reference it

src/pages/reportBuilder/components/GroupByTable/CompareByGroupByTable.js
```diff
@@ -0,0 +1,162 @@
...
+      return <PanelLoading minHeight="250px" />;
+    }
+
+    if (!tableData.length) {
+      return (
+        <Panel>
+          <Empty message="There is no data to display" />
+        </Panel>
+      );
+    }
```
nicklemmon, line 126: It would be worth adding some testing for the error and empty state scenarios.

src/pages/reportBuilder/components/GroupByTable/CompareByGroupByTable.js
```diff
@@ -0,0 +1,162 @@
...
+
+    const subaccount = subaccounts.find(({ id }) => {
+      return id === subaccountId;
+    });
+
+    const value = subaccount
+      ? `${subaccount?.name} (ID ${subaccount?.id})`
+      : `Subaccount ${subaccountId}`;
+    return { type: 'Subaccount', value, id: subaccountId };
+  };
```
nicklemmon, line 78: Screen Shot 2020-12-14 at 1 12 30 PM

Appear to be some issues with rendering subaccounts.

nicklemmon: When clicking on these links that does appear to cause some invalid queries to be passed to the API. Fortunately error scenarios handle this well.

nicklemmon (changes requested):
@PatrickSison taking another pass and I think I caught one bug. Other feedback is minor, though I do think some additional testing of edge cases is warranted.

src/hooks/useSparkPostQuery/useSparkPostQuery.js
PatrickSison: I think this is a limit of `useQuery` solved by `useQueries` in `react-query`. For this instance, we don't necessarily know the number of requests that could come in because of the comparisons. Regardless of the number of requests, these have to be joined into one promise as if one fails, the rest fails because the data can't be loaded altogether.

nicklemmon: That makes sense to me.

And in the not too distant future, we'll migrate to React Query 3, and then a lot of this should go away - though not all of it.

cypress/tests/integration/signals-analytics/analytics-report/reportBreakdownTable.spec.js
```diff
@@ -459,27 +459,94 @@ if (IS_HIBANA_ENABLED) {
...
 
-      cy.get('table').within(() => {
-        cy.findAllByText('Fake Subaccount 1 (ID 101)').should('have.length', 3);
-        cy.findAllByText('Fake Subaccount 3 (ID 103)').should('have.length', 3);
+        cy.wait('@getWatchedDomains');
+        cy.wait('@getWatchedDomains');
+        cy.findAllByText('There is no data to display').should('be.visible');
+      });
+
+      it('will render an empty state the requests fail', () => {
```
nicklemmon, line 507: This is an error state, no? Not an empty state?

cypress/tests/integration/signals-analytics/analytics-report/reportBreakdownTable.spec.js
```diff
@@ -459,27 +459,94 @@ if (IS_HIBANA_ENABLED) {
...
+
+        // Hmmmmmm
+        cy.wait('@getWatchedDomains');
+        cy.wait('@getWatchedDomains');
+        cy.wait('@getWatchedDomains');
+        cy.wait('@getWatchedDomains');
+        cy.wait('@getWatchedDomains');
+        cy.wait('@getWatchedDomains');
+        cy.wait('@getWatchedDomains');
+        cy.wait('@getWatchedDomains');
```
nicklemmon, line 531: Seems like a bug!

cypress/tests/integration/signals-analytics/analytics-report/reportBreakdownTable.spec.js
```diff
@@ -459,27 +459,94 @@ if (IS_HIBANA_ENABLED) {
...
+
+        cy.stubRequest({
+          url: '/api/v1/metrics/deliverability/watched-domain**/*',
+          fixture: 'metrics/deliverability/watched-domain/200.get.json',
+          requestAlias: 'getWatchedDomains',
+        });
+
+        cy.findByRole('button', { name: 'Try Again' }).click();
+        cy.wait('@getWatchedDomains');
+        cy.wait('@getWatchedDomains');
```
nicklemmon, line 544: Since they occur simultaneously, probably use this:
```
cy.wait(['@getWatchedDomains', '@getWatchedDomains']);
```

Just helps things be a little bit more self documenting.

src/pages/reportBuilder/components/GroupByTable/CompareByGroupByTable.js
```diff
@@ -0,0 +1,162 @@
...
+      group.label === 'Subaccount'
+        ? getSubaccountFilter(filterKey)
+        : { type: group.label, value: filterKey };
+
+    const primaryCol = <AddFilterLink newFilter={newFilter} />;
+    const comparisonCol = <Tag>{row[comparisonType]}</Tag>;
+    const metricCols = metrics.map(({ key, unit }) => (
+      <Box textAlign="right" key={key}>
+        <Unit value={row[key] || '- - -'} unit={unit} />
+      </Box>
```
nicklemmon, line 92: This component should be implemented here. If it's global and re-usable, we can also use it in lots of spots for consistency's sake:
```jsx
function EmptyCell() {
  return (
    <>
      {/* hidden from screen reader users */}
      <span aria-hidden="true">---</span>
      {/* hidden from everyone else */}
      <ScreenReaderOnly>No Data</ScreenReaderOnly>
    </>
  );
}
```

src/pages/reportBuilder/components/GroupByTable/useGroupByTable.js
```diff
@@ -0,0 +1,97 @@
...
+    return [reportOptions];
+  }
+
+  return comparisons.map(compareFilter => {
+    const filterType = REPORT_BUILDER_FILTER_KEY_MAP[compareFilter.type];
+    const comparedFilters = [
+      ...reportOptions.filters,
+      { AND: { [filterType]: { eq: [compareFilter] } } },
+    ];
+    return { ...reportOptions, filters: comparedFilters };
```
nicklemmon, line 25: This could be tough to fix, but looks like it's possible to add the same filter multiple times through this table, resulting in an API error. It's a weird edge case, so I'm not sure how that should be handled:
Screen Shot 2020-12-16 at 10 13 31 AM

src/pages/reportBuilder/components/GroupByTable/CompareByGroupByTable.js
```diff
@@ -0,0 +1,162 @@
...
+  };
+
+  const TableContent = () => {
+    if (!group || metrics.length === 0) {
+      return null;
+    }
+
+    if (status === 'error') {
+      return (
+        <Panel>
```
nicklemmon, line 105: We would want to put this inside a `Panel.Section` for consistency and to avoid having a double border when we upgrade Matchbox (white line is there to show the alignment inconsistency):

Screen Shot 2020-12-16 at 10 15 18 AM

jsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+  "compilerOptions": {
+    "allowSyntheticDefaultImports": false,
+    "baseUrl": "./"
+  },
+  "exclude": ["node_modules", "dist"]
```
nicklemmon, line 6: Our build folder is `build` rather than `dist` looks like.

src/hooks/useSparkPostQuery/useSparkPostQuery.js
```diff
@@ -5,8 +5,35 @@ import { refresh, logout } from 'src/actions/auth';
 import { useRefreshToken } from 'src/helpers/http';
 import { showAlert } from 'src/actions/globalAlert';
 import { fetch as fetchAccount } from 'src/actions/account';
+import { defaultQuery } from 'src/helpers/api';
 
-export default function useSparkPostQuery(queryFn, config = {}) {
+export function useSparkPostQueries(queries, config = {}, queryKey) {
```
nicklemmon, line 10: Would you mind adding some tests to `useSparkPostQuery.spec.js` and some docs to `useSparkPostQuery.stories.mdx`?

I would think just a tiny bit of both would go a long way.

src/pages/reportBuilder/components/GroupByTable/useGroupByTable.js
```diff
@@ -0,0 +1,97 @@
...
+    // Array of objects (unique group-by key)
+    const objectKeys = _.uniqBy(_.flatten(rawData), groupKey);
+
+    return objectKeys.reduce((acc, object) => {
+      const key = object[groupKey];
+      const newRows = comparisons.map((compareFilter, index) => {
+        const currentDataCompare = keyedData[index][key] || {};
+        return {
+          ...currentDataCompare,
+          [groupKey]: key, //Readd key (in case it's empty)
```
nicklemmon, line 77: Typo here

cypress/tests/integration/signals-analytics/analytics-report/reportBreakdownTable.spec.js
nicklemmon: Nevermind! This is right. It looks like there are two requests that are invoked via `useSparkPostQueries` each time it attempts the request, so that should total 8. This works too:

```js
cy.wait(['@getWatchedDomains', '@getWatchedDomains']);
cy.wait(['@getWatchedDomains', '@getWatchedDomains']);
cy.wait(['@getWatchedDomains', '@getWatchedDomains']);
cy.wait(['@getWatchedDomains', '@getWatchedDomains']);
```

nicklemmon (changes requested):
@PatrickSison Left a few comments, though not a whole lot - I think this is almost ready.

cypress/tests/integration/signals-analytics/analytics-report/reportBreakdownTable.spec.js
PatrickSison: That's a lot nicer

nicklemmon (approved):
@PatrickSison 👍 Looks good to me.

PR #1952, opened 2020-12-14
FE-1275 - Upgrade Cypress to version 6.1.0

nicklemmon:
[FE-1275](https://sparkpost.atlassian.net/browse/FE-1275)

### What Changed

- Upgrades Cypress to version 6.1.0
- Replaces uses of `should('not.be.visible')` with `should('not.exist')` in the vast majority of test cases -> [see the relevant portion of the migration guide](https://docs.cypress.io/guides/references/migration-guide.html#Non-existent-element-assertions)
- Swaps CI browser to Chrome instead of electron due to [a bug with newer versions of Cypress](https://github.com/cypress-io/cypress/issues/9208)
- As a result of ☝️ , uses `cypress/browsers` Docker image instead of generic Node image - this also allows us to remove some of the setup steps in our CI config that are handled by default with this image and helps us avoid manually installing Chrome as part of our CI pipeline
- Upgrades Review Badger to the latest release
- Changes path to Cypress artifacts when targeted for upload

### How To Test

- Verify Cypress tests continue to pass with new version of Cypress
- Verify the new version of Cypress runs locally after running `npm install`
- Verify in CircleCI that build artifacts match artifacts that are present on recent runs on the `master` branch. You can do so by comparing these two jobs:
  + https://app.circleci.com/pipelines/github/SparkPost/2web2ui/3496/workflows/338d9f2f-1f79-47b7-bb9f-a55634d021c1
  + https://app.circleci.com/pipelines/github/SparkPost/2web2ui/3443/workflows/2bfa4425-eaec-486e-8ac4-439e16deaaf2

### To Do

- [ ] Incorporate feedback


cypress/tests/integration/signals-analytics/summaryReport.spec.js
```diff
@@ -112,7 +112,7 @@ describe('Summary Report page', () => {
       });
 
       cy.findByDataId('summary-chart').within(() => {
```
nstanard, line 114: Had a failure locally - was able to change it to pass like so:

image

nicklemmon: Cool, that's super simple. Let me incorporate that.

nicklemmon:
@nstanard Ready for another pass when you have a minute. Thanks.

package.json
```diff
@@ -18,12 +18,11 @@
     "analyze": "source-map-explorer build/static/js/main.*",
     "test": "node scripts/test.js --env=jsdom --config=./jest.config.js",
     "test-ci": "CI=true npm test -- --coverage --colors --maxWorkers=2",
-    "test-cy-run": "cypress run",
```
nicklemmon, line 21: I just removed this - we weren't using it anywhere.

nstanard:
didn't actually get a chance to do another pass, but let the test suite run in the bg and found that I'm getting A LOT of local test failures that aren't happening on circleci. I'm going to pick this up first thing in the morning and make sure this wasn't a mistake on my part. It most likely was.

nicklemmon:
> didn't actually get a chance to do another pass, but let the test suite run in the bg and found that I'm getting A LOT of local test failures that aren't happening on circleci. I'm going to pick this up first thing in the morning and make sure this wasn't a mistake on my part. It most likely was.

I just ran all of the tests locally and only had two failures!

**accountSettingsPage.spec.js**
Screen Shot 2020-12-16 at 9 33 51 PM

**events/listPage.spec.js**
Screen Shot 2020-12-16 at 9 33 40 PM

Both of which pass when run in isolation. I'm thinking this is some unusual flakiness that is only encountered locally, but also only encountered when running all tests locally, possibly due to performance slowdown with Cypress itself as it runs so many tests in sequence.

Fortunately, there isn't really a great reason to run _all_ tests locally so I don't think this is really worth solving for. Sounds like you're seeing other problems than these two?

nstanard:
I have a feeling that a production build and the minification process plays a role in Cypress test results on circle vs locally. Picking this back up now that pin to dashboard is getting further to merge ready.

nstanard (approved):
Alright! My local testing is consistent with what Nick got. This all LGTM!

PR #1956, opened 2020-12-15
FE-1247 - Add API Keys Hibana Empty State

jonambas:
### What Changed

- Adds new empty state for API Keys
- Note
  - Doesn't match the mocks, since copy was changed
  - Doesn't include the muted banner. UX determined we shouldn't include it: https://msys.slack.com/archives/GLLAW6UAX/p1608130726001400

### How To Test

- Run the app
- Enable account ui flag: "allow_empty_states"
- With no api keys and hibana is enabled, verify you see the Empty state

### To Do

- [x] Rebase with Shrirupa's branch / master when FE-1246 is merged


src/pages/api-keys/ListPage.js
```diff
@@ -127,7 +109,7 @@ export class ListPage extends Component {
         primaryAction={primaryAction}
         title="API Keys"
         empty={{
-          show: keys.length === 0,
+          show: !error && keys.length === 0,
```
jonambas, line 112: Caught a small bug – api errors just reverted to the empty state instead of showing the banner

shrirupa (approved):
Looks good to me!

cypress/fixtures/api-keys/200.get.empty.json
```diff
@@ -0,0 +1,3 @@
+{
+  "results": []
```
shrirupa, line 2: I think we are switching to using the common fixture `fixtures/200.get.no-results.json` and `fixtures/400.json` instead of adding separate ones.

PR #1959, opened 2020-12-17
SA-1896: Release Scheduled Reports

kchu1337:
### What Changed

- Removed the flag 'allow_scheduled_reports'

### How To Test
 - Use an account without the feature flag.
 - Check that schedule report is an option in the '...' in the view all reports modal. 
 - Check that you can see the scheduled report modal from the /signals/analytics
 - Check that you can edit a report from the  scheduled report modal.
 - Check that you can create a report from the  scheduled report modal.

### To Do

- [ ] Remove feature flag in confluence doc.


src/pages/reportBuilder/helpers/scheduledReports.js
```diff
@@ -53,9 +53,11 @@ export const getDefaultValues = (scheduledReport, users) => {
...
   const formattedMinute = formatMinuteString(minute);
   const time = `${formattedHour}:${formattedMinute}`;
-  const fullRecipients = recipients.map(recipient => {
-    return users.find(({ name }) => name === recipient);
-  });
+  const fullRecipients = recipients
+    .map(recipient => {
+      return users.find(({ name }) => name === recipient);
+    })
+    .filter(Boolean);
```
kchu1337, line 60: Removes users who cannot be found.

nicklemmon (approved):
@kchu1337 This looks good to me - test failure appears to be a flaky test.

nicklemmon (changes requested):
@kchu1337 I sent you a Slack message about it, but the test failure here is actually something that needs to be addressed before we merge. It's technically a Matchbox problem, but we don't want to introduce test failures if we can help it.

nicklemmon (approved):
@kchu1337 Looks good to me - thanks for fixing that test.

PR #1963, opened 2020-12-18
SA-1894: Add active comparisons to save report modal

PatrickSison:
### What Changed

- Active comparisons displayed on save report changes modal

### How To Test

- Go to report builder
- Save a custom report
- Click Add Comparisons and add at least 2 comparison filters (need feature flag `allow_compare_by`)
- Click `Save Changes` to get to save report modal
- Verify that active comparisons are displayed

### To Do

- [ ] Address feedback

### DIFF
--- a/cypress/tests/integration/signals-analytics/analytics-report/savedReports.spec.js
+++ b/cypress/tests/integration/signals-analytics/analytics-report/savedReports.spec.js
@@ -208,10 +208,7 @@ if (IS_HIBANA_ENABLED) {
         cy.findByLabelText('Report').focus(); // open typeahead
 
         cy.findListBoxByLabelText('Report').within(() => {
-          cy.findAllByRole('option')
-            .eq(0)
-            .should('have.contain', 'My Bounce Report')
-            .click();
+          cy.findByRole('option', { name: 'My Bounce Report mockuser' }).click();
         });
 
         cy.findByLabelText('Report').should('have.value', 'My Bounce Report');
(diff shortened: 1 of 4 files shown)


src/pages/reportBuilder/components/SavedReportsSection/SaveReportModal.js
```diff
@@ -131,6 +133,14 @@ export function SaveReportModal(props) {
...
               ) : null}
 
+              {hasComparisons ? (
+                <Box>
+                  <Heading as="h6">Comparisons</Heading>
+
+                  <ActiveComparisons comparisons={reportOptions.comparisons} />
+                </Box>
+              ) : null}
+
```
nicklemmon, line 143: Should we update these to use the new `LabelValue` component from Matchbox? I don't think that was ready when we started this modal.

nicklemmon (changes requested):
@PatrickSison Just a little bit of feedback. Shouldn't take too long to explore further.

cypress/tests/integration/signals-analytics/analytics-report/savedReports.spec.js
```diff
@@ -255,18 +255,20 @@ if (IS_HIBANA_ENABLED) {
 
         cy.findListBoxByLabelText('Report').within(() => {
           cy.findAllByRole('option')
-            .eq(0)
-            .should('have.contain', 'My Bounce Report')
+            .eq(2)
```
nicklemmon, line 258: I would remove the use of the index here - it will be a pretty brittle test otherwise. Try this:
```js
`cy.findByRole('option', { name: 'Comparison Report' })`.click()
```

nicklemmon: In fact, it might be worth removing the use of `.eq` throughout this spec file while you're here.

nicklemmon (changes requested):
@PatrickSison I think some tiny testing changes would help protect us from brittle tests in the future. Otherwise, things look good. The Use of `LabelValue` is nice here.

nicklemmon (approved):
@PatrickSison Looks good. Thanks for making those testing changes.

PR #1965, opened 2020-12-21
Sa 1896-2

kchu1337:
### What Changed

- Added error state handling

### How To Test

- Go to /signals/schedule/foo and /signals/schedule/foo/bar. Verify that you are redirected back to report builder.


nicklemmon (changes requested):
@kchu1337 One tiny bit of feedback - and obviously the test failures I described in the other branch are happening here as well, so with a little rebase I'm sure things will be alright.

nicklemmon (approved):

PR #1967, opened 2020-12-21
FE-1286 - Show Pinned Report to Reporting and templates user when their last_usage_date!==null

shrirupa:
[FE-1286] Show Pinned Report to Reporting and templates user when their last_usage_date!==null

### What Changed

- Updated the condition to show Analytics Promo step for templates and reporting user when last usage date is null
- And show Pinned Report for templates and reporting user when last usage date is not null
- Updated a condition on usePinnedReport which was causing the _skipped_ cypress test to fail
- Updated the usage/view grant in the fixture for reporting users.

### How To Test

- Check /dashboardv2 for reporting and templates user on appteam, it should show pinned report.

### To Do

- [ ] Address feedback

[FE-1286]: https://sparkpost.atlassian.net/browse/FE-1286

### DIFF
--- a/cypress/fixtures/authenticate/grants/200.get.reporting.json
+++ b/cypress/fixtures/authenticate/grants/200.get.reporting.json
@@ -44,6 +44,9 @@
     },
     {
       "key": "support/manage"
+    },
+    {
+      "key": "usage/view"
     }
   ]
 }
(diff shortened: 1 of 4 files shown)


bkemper (approved):
LGTM.  I was able to verify that it is working as expected.

Screen Shot 2020-12-24 at 12 01 25 PM

PR #1971, opened 2020-12-21
TR-2938 - (UI) When publishing a template, immediately duplicating it before saving prevents duplication

nicklemmon:
[TR-2938](https://sparkpost.atlassian.net/browse/TR-2938)

### What Changed

- Initially added failing Cypress test that reproduced the bug - it passes now
- Cleaned up the templates create page Cypress tests to use more accurate queries and to appropriately wait for network requests to reduce the risk of flakiness
- Replaces use of `usePageFilters` for the template `id`, `version`, and `navKey` values with the [useParams](https://reactrouter.com/native/api/Hooks/useparams) hook from `react-router-dom`. It looks like this bug was introduced when we removed the custom `useRouter` hook from the app.
- Updates unit tests with appropriate mocks accordingly
- Moved domains tests within the `/configuration` directory
- Moved templates tests within the `/content` directory

### How To Test

- Follow the bug reproduction steps in the Jira ticket and verify the bug is resolved
- Test various redirects in the templates feature ensuring no breaking changes were introduced

### To Do

- [ ] Incorporate feedback


src/pages/templates/hooks/tests/useEditorNavigation.test.js
```diff
@@ -1,15 +1,18 @@
 import React from 'react';
 import { act } from 'react-dom/test-utils';
 import { mount } from 'enzyme';
+import { useParams } from 'react-router';
```
shrirupa, line 4: ahh, I always tend to import these hooks from reacct-router-dom. https://reactrouter.com/native/api/Hooks/useparams

nicklemmon:
@shrirupa Updated that import - it worked like a charm. I'm not sure why but I had a comment indicating importing from `react-router-dom` wouldn't work - must have been another issue I eventually resolved.

src/pages/templates/hooks/tests/useEditorNavigation.test.js
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import { act } from 'react-dom/test-utils';
 import { mount } from 'enzyme';
+import { useParams } from 'react-router-dom';
```
shrirupa, line 4: similar import fix here. sorry, should have found this one earlier review.

src/pages/templates/context/tests/EditorContext.test.js
```diff
@@ -1,20 +1,28 @@
 import React from 'react';
 import { mount, shallow } from 'enzyme';
+import { useParams } from 'react-router';
```
shrirupa, line 3: similar import fix here. sorry, should have found this one earlier review.

shrirupa (approved):
just that last import fix, everything else looks good to me!

PR #1972, opened 2020-12-22
SA-1892: Added segment to report builder comparison form

PatrickSison:
### What Changed

- Added segment tracking event for compare by in report builder

### How To Test
- Go to Comparison form in report builder
- Apply form with no comparisons
- Check that request does not go through with applying
- Apply form with at least 2 comparison
- Check that request goes through with the appropriate data

### To Do

- [ ] Verify data on segment

### DIFF
--- a/src/pages/reportBuilder/components/CompareByForm/CompareByForm.js
+++ b/src/pages/reportBuilder/components/CompareByForm/CompareByForm.js
@@ -26,6 +26,8 @@ import { useReportBuilderContext } from '../../context/ReportBuilderContext';
 import Typeahead from './Typeahead';
 import styled from 'styled-components';
 
+import { segmentTrack, SEGMENT_EVENTS } from 'src/helpers/segment';
+
 const initialState = {
   filters: [null, null],
   filterType: undefined,
@@ -118,6 +120,13 @@ function CompareByForm({
       return dispatch({
         type: 'SET_MIN_COMPARE_ERROR',
         error: 'Select more than one item to compare',
+      });
+    }
+
+    if (formattedFilters.length >= 2) {
+      segmentTrack(SEGMENT_EVENTS.REPORT_BUILDER_COMPARISON_ADDED, {
+        comparisonType: state.filterType,
+        numberOfComparisons: formattedFilters.length,
       });
     }
 
--- a/src/pages/reportBuilder/components/CompareByForm/CompareByForm.js
+++ b/src/pages/reportBuilder/components/CompareByForm/CompareByForm.js
@@ -54,6 +54,7 @@ const reducer = (state, action) => {
       return {
         ...state,
         filters: [...state.filters, null],
+        formChanged: true,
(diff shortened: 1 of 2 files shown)


src/pages/reportBuilder/components/CompareByForm/CompareByForm.js
```diff
@@ -103,6 +105,13 @@ function CompareByForm({
       return dispatch({ type: 'SET_ERROR', error: 'Select more than one item to compare' });
     }
 
+    if (formattedFilters.length >= 2) {
+      segmentTrack(SEGMENT_EVENTS.REPORT_BUILDER_COMPARISON_ADDED, {
```
bkemper, line 109: Shouldn't this be called after `handleSubmit` is successful?

PatrickSison: handleSubmit is a synchronous function at least, and all it does is update state in the context. We could do it in the context provider and see if state has changed?

PatrickSison:
Wonder if it's more valuable to see regardless of if the form is empty if the form is submitted. Not sure if we want to see users who look at the form and apply nothing

src/pages/reportBuilder/components/CompareByForm/CompareByForm.js
kchu1337: Ideally, it should just check that the number of filters changes. For example, ```if(_.isEqual(comparisons, formattedFIlters)```. But I think there is a bug where it's updating without needed to click the submit button.

kchu1337: If you keep it like this, if can send a segment event even if they kept their comparisons the same. And if they remove all comparisons, this does not send an event.

kchu1337 (changes requested):

src/pages/reportBuilder/components/CompareByForm/CompareByForm.js
PatrickSison: @kchu1337 Is your suggestion then that we send any time comparison filters exclusively when it changes (including from any filters down to no filters)? I do agree though it should probably not send when the filters have been unchanged

src/pages/reportBuilder/components/CompareByForm/CompareByForm.js
```diff
@@ -103,6 +111,12 @@ function CompareByForm({
       return dispatch({ type: 'SET_ERROR', error: 'Select more than one item to compare' });
     }
 
+    segmentTrack(SEGMENT_EVENTS.REPORT_BUILDER_COMPARISON_ADDED, {
+      formChanged: state.formChanged,
```
kchu1337, line 115: I would either do `state.formChanged || false` or set it's initial state to false rather than undefined. I think Segment does some weird caching so if you don't explicitly define it, it keeps the previous value.

kchu1337 (changes requested):

kchu1337 (approved):

PR #1979, opened 2020-12-28
FE-1294 - Domain Configuration Tabs - Cleanup and Feedback

nstanard:
### What Changed
- Domain Status filter label is now ["All" and "None"](https://github.com/SparkPost/2web2ui/pull/1979/files#diff-19a75f0b8b84ed3a98be828d6783c9683427cea0c0554d0d5ce5c5d4381bb141R180) respectively. 
- Select All checkbox state is now maintained always through the reducer actions [TOGGLE](https://github.com/SparkPost/2web2ui/pull/1979/files#diff-19a75f0b8b84ed3a98be828d6783c9683427cea0c0554d0d5ce5c5d4381bb141R59) and [LOAD](https://github.com/SparkPost/2web2ui/pull/1979/files#diff-19a75f0b8b84ed3a98be828d6783c9683427cea0c0554d0d5ce5c5d4381bb141R79). (it wasn't being checked when re-checking all the options)
- [Select All](https://github.com/SparkPost/2web2ui/pull/1979/files#diff-19a75f0b8b84ed3a98be828d6783c9683427cea0c0554d0d5ce5c5d4381bb141R202) is now disabled if already all selected.
- Styled components being used for Panel, Panel.Section on the [Domain Configuration tabs](https://github.com/SparkPost/2web2ui/pull/1979/files#diff-fbf9141f2dc6901493f388c3b03cdc82835174d5e6b883ee03e0215cddc518bdR321) to prevent misalignment between screen resolution 350px -> 450px (need matchbox update to revert later). 
- Styled components used for domain status filter [button](https://github.com/SparkPost/2web2ui/pull/1979/files#diff-19a75f0b8b84ed3a98be828d6783c9683427cea0c0554d0d5ce5c5d4381bb141R169) and [icon](https://github.com/SparkPost/2web2ui/pull/1979/files#diff-19a75f0b8b84ed3a98be828d6783c9683427cea0c0554d0d5ce5c5d4381bb141R186).
- I've mapped the checkbox values to a [joined string](https://github.com/SparkPost/2web2ui/pull/1979/files#diff-19a75f0b8b84ed3a98be828d6783c9683427cea0c0554d0d5ce5c5d4381bb141R157) instead of mapping with a key. This allows for getting the text ellipsis to work as expected.
- [Rearranged the event cycles](https://github.com/SparkPost/2web2ui/pull/1979/files#diff-152b7d3698608074b564de26bc4f19dad494e2c8ba4a4356afe64cacf95724b7R169) for page load and filter updates.
- [Preventing page load init](https://github.com/SparkPost/2web2ui/pull/1979/files#diff-152b7d3698608074b564de26bc4f19dad494e2c8ba4a4356afe64cacf95724b7R170) unless there is data and its not loading.
- Thanks to Shrirupa we've also corrected the [custom domain status filter](https://github.com/SparkPost/2web2ui/pull/1979/files#diff-5f7b1c7cf9c9d8d716a9132e7166c018e58d964f893a1470b00ab8c8181ebec1R6) logic to be an "or" filter.
- Cypress test added for [resetting state between sending and bounce tab switching](https://github.com/SparkPost/2web2ui/pull/1979/files#diff-3ed6bd6d605b0858cb45f274b73211b2b414faadb474defbffcbef2018f58110R757)
- Cypress tests added for [landing the page with the different query parameters](https://github.com/SparkPost/2web2ui/pull/1979/files#diff-3ed6bd6d605b0858cb45f274b73211b2b414faadb474defbffcbef2018f58110R633) that that set the filters/table state.

## Before:
Chrome:

FireFox:

## After:
All:
image
None:
image
In between:
image
Mobile:
image
Disable select all: 
image

### How To Test
- Load up the app locally and navigate to the domain configuration `http://localhost:3100/domains/list/sending`
- Load up the same page on staging or production - compare the filters at all resolutions/screen widths.

### To Do
- [x] Self review changes
- [x] Discuss possible changes for Matchbox components. (talk with Jon & Logan)
- [x] Incorporate any review feedback 

### Notes during development

1. Buttons don't have text-align left by default... but buttons within ListBox do.
1b. Edit: I now see Icon.Button is what's being used just not fixed right at all widths.
1b. Edit: We have IconButton - But I'd have to style it like a button.

2. We could position absolute right the Button.Icon so it's always aligned right but text gets aligned left - see 1.
2b. List box also has it like this - I think button should behave the same by default.

3. Margin left on Panel and Panel Section flex child (div) was messing with the mobile padding/spacing.

4. Padding left/right on mobile for Panel/Section could be reduced. I used 0.75rem and then set 1rem in a media query.

5. Text vertical alignment for the filters (buttons) wasn't aligned.

6. Margin left override until matchbox fixes panel styling for 350 -> 450 screen resolutions


src/pages/domains/components/TableFilters.js
```diff
@@ -162,33 +182,29 @@ function StatusPopover({ checkboxes, onCheckboxChange, disabled, domainType }) {
...
-              {hasCheckedCheckboxes ? (
-                checkedCheckboxes
-                  .filter(checkbox => checkbox.name !== 'selectAll')
-                  .map((checkbox, index) => (
-                    <span key={`${checkbox.name}-${index}`}>{checkbox.label}&nbsp;</span>
-                  ))
-              ) : (
-                <span>Domain Status</span>
-              )}
+              {!hasCheckedCheckboxes && 'None'}
```
shrirupa, line 185: on visiting http://localhost:3100/domains/list/sending?readyForSending=true&readyForDKIM=true&readyForBounce=true&validSPF=true&unverified=true&blocked=true, all checkboxes except "Select All" is selected.

src/pages/domains/components/styles.js
```diff
@@ -1,34 +1,69 @@
...
   }
 `;
+
+export const AlignedTextButton = styled(Button)`
+  text-align: left;
+  line-height: 20px; /* Note: Fixes text and button alignment so they're all the same vertical alignment (see styles/note on span children) */
+
+  > span,
+  > span > span {
+    vertical-align: middle; /* Note: Fixes text and button alignment so they're all the same vertical alignment (v align the children of the button) */
```
shrirupa, line 36: 😱  we shouldn't have to add css like this. My guess is, this should be handled by matchbox.

src/pages/domains/components/styles.js
```diff
@@ -1,34 +1,69 @@
...
+  transform: rotate(90deg);
+`;
+
+export const DomainTabPanelSection = styled(Panel.Section)`
+  padding: 1rem 0.75rem;
+
+  > div > div,
+  > div > div > div {
+    margin-left: 0;
+  }
```
shrirupa, line 64: why do we need this ?

nstanard: I went too far, the issue was only happening under 350px width and that's not going to happen - thanks for asking about it!

src/pages/domains/components/TableFilters.js
nstanard: on it

src/pages/domains/components/styles.js
nstanard: https://msys.slack.com/archives/GLLAW6UAX/p1609181666078500 - i've got a conversation lined up to get this work integrated from the Matchbox side. I'll take responsibility for cleaning this up later.

src/pages/domains/components/styles.js
```diff
@@ -1,34 +1,55 @@
...
+
+  @media (min-width: ${props => props.theme.breakpoints[2]}) {
+    width: 250px;
+  }
+`;
+
+export const AlignedButtonIcon = styled(Button.Icon)`
+  position: absolute;
+  right: 0.75rem; /* note: same as Matchbox listbox svg icon */
```
shrirupa, line 49: I think the design token for this is spacing_300, ${props => props.theme.space['300']}

src/pages/domains/components/SendingDomainsTab.js
```diff
@@ -250,35 +251,44 @@ export default function SendingDomainsTab({ renderBounceOnly = false }) {
     }
   }, [hasSubaccounts, listSubaccounts, subaccounts]);
 
+  // useEffect to synce query params on page load to react state
   const firstLoad = useRef(true);
   useEffect(() => {
-    if (!listPending) {
-      const allStatusFilterNames = Object.keys(filters).filter(i => i !== 'domainName');
+    if (firstLoad.current) {
```
shrirupa, line 257: I think the change here introduced a bug - unselect all the status filters and then select "select all" no domains show up

shrirupa: I know you are still working on it, but just noticed Filter Domains filed doesn't seem to be working too.

nstanard: Table still isn't updating, but still working on it. Only thing left should be getting the url params on page load to correctly apply to the `setAllFilters` dispatch.

nstanard: Awesome, so as long as I can fix: https://github.com/SparkPost/2web2ui/pull/1979/files#diff-fbf9141f2dc6901493f388c3b03cdc82835174d5e6b883ee03e0215cddc518bdR290

Then I can add the cypress tests I spec'd out and this should all be good to go.

src/pages/domains/components/SendingDomainsTab.js
```diff
@@ -309,7 +358,32 @@ export default function SendingDomainsTab({ renderBounceOnly = false }) {
               checkboxes={filtersState.checkboxes}
               domainType={renderBounceOnly ? 'bounce' : 'sending'}
               onCheckboxChange={e => {
-                filtersStateDispatch({ type: 'TOGGLE', name: e.target.name });
+                const newCheckboxes = filtersState.checkboxes.map(checkbox => {
```
nstanard, line 361: both tab files have the same logic - can boil down - https://github.com/SparkPost/2web2ui/pull/1979/files#diff-152b7d3698608074b564de26bc4f19dad494e2c8ba4a4356afe64cacf95724b7R270

src/pages/domains/components/SendingDomainsTab.js
```diff
@@ -143,8 +144,9 @@ export default function SendingDomainsTab({ renderBounceOnly = false }) {
     listSubaccounts,
   } = useDomains();
 
-  const [filtersState, filtersStateDispatch] = useReducer(tableFiltersReducer, filtersInitialState);
+  // eslint-disable-next-line no-unused-vars
```
nstanard, line 147: temporary

src/pages/domains/components/SendingDomainsTab.js
```diff
@@ -234,8 +236,11 @@ export default function SendingDomainsTab({ renderBounceOnly = false }) {
 
   // resets state when tabs tabs switched from Sending -> Bounce or Bounce -> Sending
   useEffect(() => {
-    filtersStateDispatch({ type: 'RESET', state: filtersInitialState });
-    resetFilters();
+    // TODO: Figure out a way to do this without clearing out url/state on page load....
```
nstanard, line 239: this was causing page load issues for url syncing - we'll wanna figure out a different way to do this

src/pages/domains/components/SendingDomainsTab.js
nstanard: K, all the event handling and url syncing should work without any issues, we just need to call the reset differently.

src/pages/domains/components/SendingDomainsTab.js
```diff
@@ -250,35 +255,74 @@ export default function SendingDomainsTab({ renderBounceOnly = false }) {
...
+      }); // NOTE: Sets the filters display
+      updateFilters(filterStateToParams(newFiltersState)); // NOTE: Updates the URL query params, sets url if it's not set at all...
 
-        filtersStateDispatch({
-          type: 'LOAD',
-          names: statusFiltersToApply,
-          domainName: domainNameFilter,
-        });
+      // Question - should we strip off selectAll so it doesn't get dispatched to the table?
+      setAllFilters(getReactTableFilters(filterStateToParams(newFiltersState))); // NOTE: Updates the state/table filtering
```
shrirupa, line 292: Regarding our discussion on the or logic, rearranging the filters and accessor structure for domains got it working https://github.com/SparkPost/2web2ui/pull/1982/files. it would be great if you could add it in here.

nstanard: Thanks amazing - great work!

src/pages/domains/components/TrackingDomainStatusCell.js
```diff
@@ -5,7 +5,7 @@ import { TranslatableText } from 'src/components/text';
 import useUniqueId from 'src/hooks/useUniqueId';
 
 export default function TrackingDomainStatusCell({ row }) {
-  const { defaultTrackingDomain, unverified, blocked } = row;
+  const { defaultTrackingDomain, unverified, blocked } = row.DomainStatus;
```
shrirupa, line 8: tracking domain details page seems to be crashing because of this. we can remove this change here and instead pass the prop differently

shrirupa: in src/pages/domains/components/TrackingDomainsTable we can change <TrackingDomainStatusCell row={row.values.DomainStatus} />

shrirupa: in src/pages/domains/components/TrackingDomainsTable we can change `<TrackingDomainStatusCell row={row.values.DomainStatus} />`

nstanard: I'm not getting page crashes with this - can you try pulling and see if it's all up to date?

nstanard: OKKK reading is hard - domain details page fixed!

cypress/tests/integration/domains/listPage.spec.js
```diff
@@ -614,6 +614,170 @@ describe('The domains list page', () => {
...
+
+        cy.findByRole('button', { name: 'Domain Status' }).click();
+        cy.findByLabelText('Verified').should('be.checked');
+        cy.findByLabelText('DKIM Signing').should('not.be.checked');
+        cy.findByLabelText('Bounce').should('not.be.checked');
+        cy.findByLabelText('SPF Valid').should('not.be.checked');
+        cy.findByLabelText('Unverified').should('not.be.checked');
+        cy.findByLabelText('Blocked').should('not.be.checked');
+
+        // THIS ROW SHOULDN'T EXIST.... IMO
```
shrirupa, line 645: I think we should see all rows that are sending, if sending checkbox is selected

shrirupa: I remember we had it like the way you are suggesting earlier when we had "and" logic.

shrirupa: right, verified with AC of https://sparkpost.atlassian.net/browse/FE-1226,  so if we have 

> Unverified + Blocked should display all Domains that are either unverified or blocked.

shrirupa: so in case of when "verified" is selected, we should display all domains that are verified

src/pages/domains/helpers/customDomainStatusFilter.js
```diff
@@ -1,11 +1,22 @@
...
-  }
-
   const column = columnIds[0];
-  const mappedRows = rows.map(row => (row.values[column] === value ? row : false)).filter(Boolean);
-  return mappedRows;
+  return rows
+    .map(row => {
+      let atleastOneTrue = 0;
+      Object.keys(value).forEach(status => {
+        // QUESTION: Do we really need to call out these fields? or can we keep it generic?
```
shrirupa, line 7: this check is necessary since, in case of blocked domains which were previously verified or dkim signed aren't unset, but are no longer displayed as " Sending" or "DKIM Signed" as in UI.

src/pages/domains/helpers/getAllSelectedForCheckboxes.js
```diff
@@ -0,0 +1,13 @@
+const getAllSelectedForCheckboxes = function(checkboxes) {
```
shrirupa, line 1: instead of getAllSeleted should we call this areAllCheckboxesSeleted since it returns true or false ?

src/pages/domains/components/styles.js
```diff
@@ -1,34 +1,75 @@
...
+  position: absolute;
+  right: ${props => props.theme.space['300']}; /* Note: same as Matchbox listbox svg icon */
+`;
+
+export const Chevron = styled(ChevronRight)`
+  color: ${props => props.theme.colors.blue['700']};
+  transform: rotate(90deg);
+`;
+
+export const DomainsListPanel = styled(Panel)``;
```
shrirupa, line 63: looks like we can remove this, since there are no styles

src/pages/domains/components/SendingDomainsTab.js
```diff
@@ -143,70 +149,61 @@ export default function SendingDomainsTab({ renderBounceOnly = false }) {
     listSubaccounts,
   } = useDomains();
 
-  const [filtersState, filtersStateDispatch] = useReducer(tableFiltersReducer, filtersInitialState);
+  // eslint-disable-next-line no-unused-vars
   const { filters, updateFilters, resetFilters } = usePageFilters(initFiltersForSending);
+  const [filtersState, filtersStateDispatch] = useReducer(tableFiltersReducer, filtersInitialState);
+
+  if (previousRenderBounceOnly !== undefined && renderBounceOnly !== previousRenderBounceOnly) {
```
nstanard, line 156: Personal note here - we might be able to (or simply might be better to) use "useRef" for this - but this is working well and I put a cypress test in place.

src/pages/domains/helpers/getAllSelectedForCheckboxes.js
nstanard: Yea, that sounds great

src/pages/domains/components/styles.js
nstanard: I did add this back, and i'll remove it later once we talk with Jon &&|| Logan

cypress/tests/integration/domains/listPage.spec.js
nstanard: All sorted out - was totally thinking about this wrong

src/pages/domains/components/styles.js
nstanard: Going to leave in place so it's more descriptive and matches the `DomainsListPanelSection` - and we only import Panel here and not both places.

src/pages/domains/components/SendingDomainsTab.js
nstanard: Yep, was better to have useRef

shrirupa (approved):
Looks good to me!

PR #1980, opened 2020-12-29
AC-1816 Add tfa_enabled and sso_enabled traits to Segment Identify

gfriedmansp:
### What Changed

- Added tagging for when 2FA authentication is enabled or disabled and when SSO is enabled or disabled
- Updated tests to use the full redux state instead of just testing the manually passed traits

### How To Test

- Log into the app, navigate to any page
- Check segment debugger: https://app.segment.com/sparkpost-daniel-chalef/sources/webui_dev/debugger for events
- The events should include `sso_enabled` and `tfa_enabled` traits, depending on what the account is set to
- Repeat with different accounts (or update your current account) to test sso_enabled: true, sso_enabled: false, tfa_enabled: true, and tfa_enabled: false


src/components/segment/SegmentIdentify.js
```diff
@@ -30,6 +30,8 @@ const mapStateToProps = state => ({
   [SEGMENT_TRAITS.COMPANY]: state.account.company_name,
   [SEGMENT_TRAITS.FIRST_NAME]: state.currentUser.first_name,
   [SEGMENT_TRAITS.LAST_NAME]: state.currentUser.last_name,
+  [SEGMENT_TRAITS.SSO_ENABLED]: Boolean(state.currentUser.is_sso),
+  [SEGMENT_TRAITS.TFA_ENABLED]: Boolean(state.tfa.enabled),
```
kchu1337, line 34: there is also a tfa_enabled prop in current user. Not sure which one is the best to use.

gfriedmansp: I went ahead and switched to use this for consistency's sake. Also at some point we will likely switch to just spreading the `currentUser` object into segment.

kchu1337 (approved):

PR #1983, opened 2020-12-31
SA-1893: Compare by filters limited to 10 filters

PatrickSison:
### What Changed

- Added comparison limit of 10 filters

### How To Test

- On Hibana, with `allow_compare_by` feature flag enabled, go to Report Builder page
- Click on `Add Comparison`
- Select comparison type
- Fill out form, adding up to 10 filters
- Verify that warning appears at top of form

### To Do

- [ ] Address feedback

### Notes
- Just realized that user wouldn't immediately see the warning message. Might be worth thinking about how the message is displayed.

### DIFF
--- a/cypress/tests/integration/signals-analytics/analytics-report/reportCompareBy.spec.js
+++ b/cypress/tests/integration/signals-analytics/analytics-report/reportCompareBy.spec.js
@@ -202,6 +202,64 @@ if (IS_HIBANA_ENABLED) {
       });
     });
 
+    it('Shows form warning if 10 comparison filters used', () => {
+      cy.stubRequest({
+        url: '/api/v1/subaccounts',
+        fixture: 'subaccounts/200.get.many.json',
+        requestAlias: 'getManySubaccounts',
+      });
+
+      cy.visit(PAGE_URL);
+      cy.wait(['@getManySubaccounts', '@getDeliverability', '@getTimeSeries']);
+
+      openCompareByModal();
+      cy.withinDrawer(() => {
+        cy.findByLabelText(TYPE_LABEL).select('Subaccount');
+
+        cy.findAllByLabelText('Subaccount')
+          .eq(0)
+          .type('Fake Subaccount');
+        cy.findByText('Fake Subaccount 1 (ID 101)')
+          .should('be.visible')
+          .click();
+        cy.findAllByLabelText('Subaccount')
+          .eq(1)
+          .type('Fake Subaccount');
+        cy.findByText('Fake Subaccount 2 (ID 102)')
+          .should('be.visible')
+          .click();
+
+        Array(7)
+          .fill()
+          .forEach((_item, index) => {
+            cy.findByRole('button', { name: 'Add Subaccount' }).click();
+            cy.findAllByLabelText('Subaccount')
+              .eq(index + 2)
+              .type('Fake Subaccount');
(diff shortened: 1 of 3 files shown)


cypress/tests/integration/signals-analytics/analytics-report/reportCompareBy.spec.js
```diff
@@ -202,6 +202,64 @@ if (IS_HIBANA_ENABLED) {
...
+
+        cy.findByRole('button', { name: 'Add Subaccount' }).click();
+        cy.findAllByLabelText('Subaccount')
+          .eq(9)
+          .type('Fake Subaccount');
+        cy.findByText(`Fake Subaccount 10 (ID 110)`)
+          .should('be.visible')
+          .click();
+
+        cy.findByText('Limit on number of comparisons reached').should('exist');
```
nicklemmon, line 252: Yeah, per your point in your PR, this is really a bug - the banner `should.be.visible` rather than just exist. Probably worth working with UX on this one, but there are a few ways to handle it:
1. Move focus to the banner when it renders - though this would be kind of odd when the user hasn't finished entering their last comparison
2. Render a snackbar showing the message in addition to the banner at the top of the form

nicklemmon, line 252: it would be worth asserting that the "Add <whatever>" button is no longer rendered as well.

nicklemmon (changes requested):
@PatrickSison Your code is relatively straightforward and seems good to me, however, the UX leaves a bit to be desired. I think you already know that though

cypress/tests/integration/signals-analytics/analytics-report/reportCompareBy.spec.js
PatrickSison: I will follow up... as soon as slack comes back lol. But agreed

src/pages/reportBuilder/components/CompareByForm/CompareByForm.js
```diff
@@ -48,22 +49,39 @@ const RemoveButton = ({ onClick }) => (
...
       const newFilters = state.filters;
       newFilters[action.index] = action.value;
+      if (state.filters.length >= 10) {
+        return {
+          ...state,
+          error: 'Limit on number of comparisons reached',
+          errorStatus: 'warning',
+          filters: newFilters,
+        };
+      }
```
nicklemmon, line 77: I'm not sure why this would be necessary? How could the user have more than the number of fields that end up being rendered in the form?

src/pages/reportBuilder/components/CompareByForm/CompareByForm.js
```diff
@@ -224,6 +240,11 @@ function CompareByForm({
               </Button>
             </Box>
           )}
+          {error && (
+            <Banner size="small" status={errorStatus}>
+              {error}
+            </Banner>
+          )}
```
nicklemmon, line 247: The only problem with this generic error banner that is used for two sets of UI - is UX OK with the other error being at the bottom of the form as well?

It makes sense either way, just wanted to check.

nicklemmon (commented):
@PatrickSison Left a couple of comments - I could be missing something when it comes to the state reducer. I'll ping you about it.

nicklemmon (approved):
@PatrickSison Code is much cleaner and works as expected. Nice job. 👍

PR #1984, opened 2020-12-31
SA-1898: Checks subscription on delete and saving reports

PatrickSison:
### What Changed

- Subscription checked on delete/saving reports

### How To Test

- Go to report builder
- Check that subscription request made on saving report
- Check that subscription request made on deleting report

### To Do

- [ ] Address feedback


src/pages/reportBuilder/components/SavedReportsSection/SaveReportModal.js
```diff
@@ -95,6 +97,8 @@ export function SaveReportModal(props) {
     return saveAction({ ...data, id: report?.id }).then(response => {
       showAlert({ type: 'success', message: `You have successfully saved ${data.name}` });
       onCancel();
+      getSubscription();
```
nicklemmon, line 100: I feel like in this instance a state machine would be super helpful. Right now, this request is being made in this component but it really impacts one that's sort of unrelated.

nicklemmon (commented):
@PatrickSison I'm a little worried about this implementation - let's chat about it on the side and think some things through.

nicklemmon (commented):
@PatrickSison This looks good - hoping you could add a comment to describe a dependency, then we should be good to go.

PatrickSison:
Cancelled workflow. Greenlit before. Just added comment

nicklemmon (approved):
@PatrickSison Looks good to me 👍

PR #1987, opened 2021-01-04
SA-1897: Strip report ID when saving a report from an existing report

PatrickSison:
### What Changed

- Report params generated again instead of pulling straight from URL
- Report ID no longer attached to saving report (to avoid having a report point to another report

### How To Test

- Go to report builder
- Select a report
- Save a new report
- Save under a new name
- Verify that report id is not attached

### To Do

- [ ] Address feedback

### DIFF
--- a/src/pages/reportBuilder/components/SavedReportsSection/SaveReportModal.js
+++ b/src/pages/reportBuilder/components/SavedReportsSection/SaveReportModal.js
@@ -88,10 +88,10 @@ export function SaveReportModal(props) {
   }, [report, reset]);
 
   const onSubmit = data => {
-    const { filters: selectedFilters, ...update } = selectSummaryChartSearchOptions;
-
-    if (Boolean(selectedFilters.length)) {
-      update.query_filters = encodeURI(JSON.stringify(dehydrateFilters(selectedFilters)));
+    const { filters: _selectedFilters, ...update } = selectSummaryChartSearchOptions;
+    const { filters } = reportOptions;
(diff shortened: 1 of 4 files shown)


src/pages/reportBuilder/components/SavedReportsSection/SaveReportModal.js
```diff
@@ -85,8 +88,13 @@ export function SaveReportModal(props) {
   }, [report, reset]);
 
   const onSubmit = data => {
-    const query_string = search.charAt(0) === '?' ? search.substring(1) : search;
+    const { filters: selectedFilters, ...update } = selectSummaryChartSearchOptions;
+
+    if (Boolean(selectedFilters.length)) {
+      update.query_filters = encodeURI(JSON.stringify(dehydrateFilters(selectedFilters)));
```
kchu1337, line 94: I'm not sure if this actually works. It looks like the way we are getting filters is not working.
I tried to console.log filters for 1 filter and it returned `["undefined:undefined"]`

kchu1337 (changes requested):

src/pages/reportBuilder/components/SavedReportsSection/SaveReportModal.js
kchu1337: Reproduce an error by 
1. Add a filter ( i used campaign: Black Friday)
2. Save the page as a new report
3. Try to access that report. Should give an error.

kchu1337 (approved):

PR #1988, opened 2021-01-04
SD-2300 - IP Graph shows incorrect date time for the included data (UI) 

nicklemmon:
[SD-2300](https://sparkpost.atlassian.net/browse/SD-2300)

### What Changed

- Removes unused `colors` dependency that was previously leveraged when generating a Cypress env file for local testing
- Properly Hibana-fy the chart on the IP pool details page by conditionally passing in `stroke` colors to the chart components depending on whether Hibana is enabled
- The user's timezone is now passed with the request to get time series data for a particular IP pool on the IP pool detail page

### How To Test
- Visit the IP pool details page - `/account/ip-pools/edit/default/<ip-pool-id>` (I believe there is only one IP address in staging for our usual test account)
- Verify in the console that the request made to `/api/v1/metrics/deliverability/time-series` includes the `timezone` parameter in the request matching your local timezone
- Turn Hibana on and off and verify the chart colors render as expected within both themes

### To Do

- [ ] Incorporate feedback


PatrickSison (commented):
Everything in code is working as intended. But didn't realize this issue when filing the bug: ranges longer than 1 day are defaulted to UTC time in metrics endpoint. No matter what, it will end up as UTC. Unfortunately, since this is a static range of 10 days, we might have to change this to display that it's in UTC

PatrickSison (approved):
Looks good. Properly attaching timezone to the request

PR #1994, opened 2021-01-05
FE-1269 - Create Empty State for Analytics Report 

shrirupa:
FE-1269 - Create Empty State for Analytics Report 

### What Changed

-  Empty state should show up on Analytics Report when user has no sending domain.
-  Added the three Tabs  as in the mock (with placeholder components)

mocks - https://sparkpost.invisionapp.com/console/Empty-States-ckcno05yk00tf011ij29ebpmx/ckevla6xu04f301yh7ymqfocx/play#project_console
### How To Test

- Enable "allow_empty_states" feature flag and check the Analytics page when you have no sending domain, you should see the Empty state as in the mock, with three placeholder tabs
- When "allow_empty_states" feature flag is not enabled the Analytics Page should appear as it is now.

### To Do

- [ ] Address feedback

### DIFF
--- a/src/pages/reportBuilder/ReportBuilder.js
+++ b/src/pages/reportBuilder/ReportBuilder.js
@@ -24,6 +24,7 @@ import {
 import { parseSearchNew as parseSearch } from 'src/helpers/reports';
 import { getFormattedDateRangeForAggregateData } from 'src/helpers/date';
 import { selectVerifiedDomains } from 'src/selectors/sendingDomains';
+import { isAccountUiOptionSet } from 'src/helpers/conditions/account';
 import {
   Charts,
   ReportOptions,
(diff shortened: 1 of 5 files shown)


src/pages/reportBuilder/ReportBuilder.js
```diff
@@ -251,6 +262,10 @@ export function ReportBuilder({
           </Button>
         </Box>
       }
+      empty={{
+        show: sendingDomains.length === 0 && isEmptyStateEnabled,
```
kchu1337, line 266: Do you need to check `isEmptyStateEnabled` since the Page component already is?

kchu1337 (changes requested):
Having some trouble with my openresty settings. I'll take another look at content in another pass once I get it fixed.

src/pages/reportBuilder/ReportBuilder.js
shrirupa: yes, actually we need this because show: sendingDomains.length === 0 will trigger matchbox to show the default empty state which doesn't exist for this Report Builder Page. you can see this problem, on any account where empty states is not enabled.

kchu1337 (approved):

PR #1996, opened 2021-01-06
FE-1239, FE-1240 - Sending and Bounce empty states

nstanard:
### What Changed
 - Sending domain Empty State shows when no sending domains
 - Bounce domain Empty State shows when there are no sending and bounce domains
 - Sending domain Info Banner shows, but only until the user dismisses it
 - Bounce domain Info Banner shows, but only until the user dismisses it
 - Create domains page accepts param of: sending, bounce, and tracking
 - Segment tracks each tab through the new page prop, `props.empty?.trackingOnly`
 - Cypress test coverage added
 - ESLint rule to enforce page tracking calls are being done correctly - so the call to segment doesn't happen more than once on page load. 

### How To Test
 - Cypress tests in the listPage.spec.js should express the needed test scenarios, find the "empty state" tests and go through them to understand the states that the app can handle.
 - Load up the http://localhost:3100/domains in hibana with empty states turned off, then on.
 
 - Sending Domain Empty State shows when no sending domains **(with correct mock content)**
 - Bounce domain Empty State shows when there are no sending and bounce domains  **(with correct mock content)**
 - Sending Domain Info Banner shows, but only until the user dismisses it  **(with correct mock content)**
 - Bounce Domain Info Banner shows, but only until the user dismisses it  **(with correct mock content)**
 - Routing to the /domains/create state with the type param works and selects the type of domain to create
 - Cypress and unit tests pass and there is enough coverage
 - Page tracking calls segment, and only once for each tab/state

### To Do
- [x] Talk to UX about bounce empty state & banner (see comment)
- [x] Make UX updates provided by Tim
- [x] Add Cypress Test Coverage
- [ ] Feedback


src/pages/domains/ListPage.js
```diff
@@ -31,39 +36,77 @@ export default function DomainsPage() {
...
     if (matchesSendingTab) {
+      if (sendingDomains.length === 0) {
+        return <SendingDomainsEmptyState />;
+      }
+
       return <Domains.SendingDomainsTab />;
     }
+
     if (matchesBounceTab) {
+      if (bounceDomains.length === 0 && sendingDomains.length === 0) {
```
nstanard, line 58: TODO: Talk with UX about how bounce empty state shows after adding an unverified "bounce" domain. (but sending empty goes away, cause it has an unverified domain in it's list)

src/pages/domains/ListPage.js
```diff
@@ -31,39 +38,92 @@ export default function DomainsPage() {
...
+    listSendingDomains();
+    // listTrackingDomains();
+  }, [listSendingDomains, listTrackingDomains]);
+
   const matchesSendingTab = useRouteMatch(SENDING_DOMAINS_URL);
   const matchesBounceTab = useRouteMatch(BOUNCE_DOMAINS_URL);
   const matchesTrackingTab = useRouteMatch(TRACKING_DOMAINS_URL);
 
+  const renderInfoBanner = () => {
+    // TODO: Add extra conditionals
```
nstanard, line 51: Need to fix the loading/error state on page load, empty states flash right now while things are loading

src/pages/domains/ListPage.js
```diff
@@ -31,39 +38,92 @@ export default function DomainsPage() {
...
+    <Page
+      title="Domains"
+      primaryAction={{
+        to: `/domains/create?type=${getTabType()}`,
+        content: 'Add a Domain',
+        component: PageLink,
+      }}
+      empty={{
+        trackingOnly: true,
+      }}
```
nstanard, line 110: might need to pass in loading prop, for segment call to work right

src/pages/domains/ListPage.container.js
```diff
@@ -0,0 +1,22 @@
+import { connect } from 'react-redux';
+import { showAlert } from 'src/actions/globalAlert';
+import { isAccountUiOptionSet } from 'src/helpers/conditions/account';
+import { ListPage } from './ListPage';
+import React from 'react';
+import { useHibana } from 'src/context/HibanaContext';
+
+function ListPageContainer(props) {
+  const [{ isHibanaEnabled }] = useHibana();
```
nstanard, line 9: I'm not actually 100% sure we need to check isHibanaEnabled but we do need the isEmptyStateEnabled later in this file. Will do more testing and clean up.

src/pages/domains/ListPage.js
nstanard: Also, bounce banner only shows when sendingDomains > 0, not bounceDomains > 0

.eslintplugin/tests/require-is-first-render-empty-state-loading.test.js
```diff
@@ -0,0 +1,11 @@
+import runner from '../utils/eslint-test-runner';
+import rule from '../require-is-first-render-empty-state-loading';
+
+runner('require-is-first-render-empty-state-loading', rule, {
```
nstanard, line 4: I think this is the last thing I could do... 

There might also be some situations that aren't covered by the cypress tests yet.... I'll check those and think about adding tests here.

package.json
```diff
@@ -8,6 +8,7 @@
   },
   "engineStrict": true,
   "scripts": {
+    "lint": "npx eslint src/",
```
nstanard, line 11: note, this command is super slow compared to the staging/commit run

.eslintplugin/require-is-first-render-empty-state-loading.js
```diff
@@ -0,0 +1,82 @@
+const fs = require('fs');
+
+/**
+ * ESLint Docs:
+ *  - https://eslint.org/docs/developer-guide/working-with-plugins#create-a-plugin
+ *  - https://eslint.org/docs/developer-guide/working-with-rules
+ */
+
+const requireIsFirstRenderEmptyStateLoading = {
```
nstanard, line 9: Main issues with this rule right now

VSCode not finding rule definition
image

We might also change the logic to look for more than just the empty prop.

.eslintplugin/require-is-first-render-empty-state-loading.js
```diff
@@ -0,0 +1,82 @@
...
+
+          const loadingAttr = node.openingElement.attributes.find(attr => {
+            if (attr && attr.name && attr.name.name && attr.name.name === 'loading') {
+              return attr;
+            }
+          });
+
+          if (emptyAttr) {
+            if (!loadingAttr) {
+              report({
```
nstanard, line 48: Not sure if we really wanna report here ... talk with team about it

package.json
nstanard: and produces a ton of output, since our codebase has a number of rules that produce long lists of warnings and errors.

.eslintplugin/require-is-first-render-empty-state-loading.js
nstanard: Checking for `HibanaEmptyStateComponent` only now, then checking the loading attr.

src/pages/domains/ListPage.js
```diff
@@ -31,39 +50,117 @@ export default function DomainsPage() {
...
+
+  return (
+    <Page
+      title="Domains"
+      primaryAction={{
+        to: `/domains/create?type=${getTabType()}`,
+        content: 'Add a Domain',
+        component: PageLink,
+      }}
+      empty={{
```
jonambas, line 144: A heads up, hopefully this doesn't throw a wrench into anything, but `empty` is deprecated and we were planning on removing it soon

nstanard: Yea, that might be a bit of an issue for the decisions we've made here.

src/pages/domains/ListPage.js
nstanard: https://sparkpost.invisionapp.com/console/Empty-States-ckcno05yk00tf011ij29ebpmx/ckevl9t7f062r011f6camaz74/play

Tim has provided an updated mock for bounce domain verbiage - to account for the sending/bounce domain relationship.

nstanard: Updated spec doc: https://docs.google.com/document/d/1_95itrqZyFQ-UHoaVyxvkRlHNPPgb6RAOhJqGgc92fk/edit#heading=h.8f2kqzymhgcy

(provided by Tim)

src/pages/domains/ListPage.js
nstanard: Or, as long as it's ok we leave that prop for the `Page` component, then it shouldn't hurt.

src/pages/domains/ListPage.js
```diff
@@ -31,39 +50,117 @@ export default function DomainsPage() {
...
+    isEmptyStateEnabled &&
+    sendingDomains.length === 0 &&
+    !sendingDomainsListError;
+
+  const showBounceDomainsEmptyState =
+    !listPending &&
+    matchesBounceTab &&
+    isEmptyStateEnabled &&
+    bounceDomains.length === 0 &&
+    sendingDomains.length === 0 &&
```
shrirupa, line 82: we can skip checking `sendingDomains.length === 0`

src/pages/domains/ListPage.js
```diff
@@ -31,39 +50,117 @@ export default function DomainsPage() {
...
+  return (
+    <Page
+      title="Domains"
+      primaryAction={{
+        to: `/domains/create?type=${getTabType()}`,
+        content: 'Add a Domain',
+        component: PageLink,
+      }}
+      empty={{
+        trackingOnly: showSendingDomainsEmptyState || showBounceDomainsEmptyState,
```
shrirupa, line 145: we can use `show` props here and then in hibanaEmptyStateComponent={getEmptyState()}

const getEmptyState = () => {
if (matchesSendingTab) {
      return SendingDomainsEmptyState;
    }

    if (matchesBounceTab) {
      return BounceDomainsEmptyState;
    }

    if (matchesTrackingTab) {
      return TrackingDomain;
    }

}

src/pages/domains/ListPage.js
shrirupa: noticed that when no verified domains are present but a unverified sending domain is present the Empty state for bounce domain is not showing up.

nstanard: I had added that intentionally at first, but I think you're right we should probably still show the bounce empty state when sending domains exist.

src/pages/domains/ListPage.js
nstanard: I had tried it that way but then the tabs don't show because the Page component returns early with the empty state.  `<Tabs selected={tabIndex} tabs={TABS} />`

I decided to go this route to ensure the tracking happens in the Page component, but we are still able to render the tabs with the empty states under them.

shrirupa: ok, upto you. I just thought we could avoid adding one off props in Page wrapper and wrap the hibanaEmptyStateComponent in Tabs.

jonambas (commented):
Working through verifying:
- [x] Verify sending domain empty state shows
- [x] Verify bounce domain empty state shows
- [x] Verify clicking on the primary empty state action selects the domain type by default on the create page
- [x] Verify table is shown with domains after creation for sending
- [ ] Verify table is shown with domains after creation for bounce
- [x] Verify the sending banner is visible
- [x] Verify the sending banner can be dismissed and does not reappear
- [ ] Verify the bounce banner is visible
- [ ] Verify the bounce banner can be dismissed and does not reappear

src/pages/domains/components/BounceDomainsEmptyState.js
```diff
@@ -0,0 +1,59 @@
...
+          </TranslatableText>
+          <Bold>Only verified domains&nbsp;</Bold>
+          <TranslatableText>
+            can be used for bounce domains. Unverified bounce domains will appear under Sending
+            Domains.
+          </TranslatableText>
+        </p>
+        <EmptyState.List>
+          <li>
+            <TranslatableText>Add a new bounce domain.</TranslatableText>
```
jonambas, line 36: Nitpick – mocks don't have periods

jonambas (approved):
LGTM. Was able to verify everything is working correctly. Left some comments, but I don't see any reason not to merge this

src/pages/domains/components/BounceDomainsEmptyState.js
nstanard: I've been adding periods to all of them 😬  (i can write that up to fix it though if UX wants them gone)

src/pages/domains/ListPage.js
```diff
@@ -31,39 +50,125 @@ export default function DomainsPage() {
...
+
+  return (
+    <Page
+      title="Domains"
+      primaryAction={{
+        to: `/domains/create?type=${getTabType()}`,
+        content: 'Add a Domain',
+        component: PageLink,
+      }}
+      empty={{
```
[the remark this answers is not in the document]
nstanard: Yea, that's correct. I put together the custom eslint rule to describe that a little, and make sure `isFirstRender` is set on the loading attribute.

src/pages/domains/CreatePage.js
```diff
@@ -1,10 +1,29 @@
...
 import { PageLink } from 'src/components/links';
 import { BASE_URL } from './constants';
 import Domains from './components';
 
+const initFilters = {
+  type: { defaultValue: 'sending' },
+};
+
 export default function CreatePage() {
+  const { filters } = usePageFilters(initFilters);
```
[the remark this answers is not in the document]
nstanard: Good thoughts, I was thinking about linking from outside the app. I have nothing lined up that requires that, but seemed good to make it work that way just in case. I'll explore moving the reference to the form component.

cypress/fixtures/users/200.get.sending-domain-banner-dismissed.json
```diff
@@ -0,0 +1,39 @@
...
+    "tou_auto_accept": false,
+    "is_sso": false,
+    "email_verified": true,
+    "access_level": "admin",
+    "options": {
+      "ui": {
+        "isHibanaBannerVisible": false,
+        "isHibanaEnabled": true,
+        "onboardingV2": {
+          "sendingDomainBannerDismissed": true
```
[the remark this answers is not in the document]
nstanard: I'm not sure to be honest. I like that idea but I also think we need to build a factory to reduce all the .json files and instead use javascript objects that we are able to re-use and compose into what we need. There might also be a desire to make tests that check one tab is dismissed, but the other isn't. Kinda on the fence on this one.

src/pages/domains/ListPage.js
nstanard: Yea, I like the readability of the render/children of the <Page /> and having one instance of <Tabs />.

src/pages/domains/CreatePage.js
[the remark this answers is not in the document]
nstanard: Moved that down lower 👍

PR #1997, opened 2021-01-06
FE-1292: `usage` can sometimes be undefined, resulting in a rendering error/crash when users visit the usage page

shrirupa:
FE-1292: `usage` can sometimes be undefined, resulting in a rendering error/crash when users visit the usage page

### What Changed

- Added tests for usage page for enterprise customers and for customers on annual messaging plan
- Removed ?include=usage from Usage Page
- Added a separate reducer for usage endpoint which we can later reuse when we remove include=usage endpoint from other places in app
- Hid Billing cycle for Annually billed customer since currently api doesn't return this info for users on annual plan
- Hid Plan Volume for customers who have unlimited volume (plan_volume === -1)

### How To Test

- Cypress tests for enterprise and annually billed customer can help these two scenarios.

### To Do

- [ ] Address feedback


cypress/fixtures/usage/200.get.no-results.json
```diff
@@ -1,3 +1,18 @@
 {
-  "results": {}
+  "results": {
```
nicklemmon, line 2: So the API will *always* return this data structure? I would think the `"start"` and `"end"` values wouldn't be specific dates if so.

cypress/tests/integration/accounts/usage/usagePage.spec.js
```diff
@@ -135,9 +160,9 @@ describe('The usage page', () => {
 
       // Usage request occurs 3 times due to retries
       cy.wait(['@usageReq', '@usageHistoryReq', '@subscriptionReq', '@accountReq']);
-      cy.wait('@usageReq');
-      cy.wait('@usageReq');
-      cy.wait('@usageReq');
+      cy.wait(['@usageReq']);
+      cy.wait(['@usageReq']);
```
nicklemmon, line 164: An array shouldn't be needed here - but maybe you had a reason for introducing it?

src/pages/usage/components/MessagingUsageChart.js
```diff
@@ -21,7 +21,9 @@ function MessagingUsageChart(props) {
   });
 
   const thresholdPercentage =
-    thresholdIndex === -1 ? 100 : (thresholdIndex / cumulativeData.length) * 100;
+    thresholdIndex === -1 || planVolume === -1
```
nicklemmon, line 24: Is there any way we could store this `-1` value in a way that makes it more self-documenting?

Like a constant maybe?

```
// And here is a comment linking to the API docs explaining this value
const UNLIMITED_VOLUME = -1
```

Then later:
```
planVolume === UNLIMITED_VOLUME
```

src/reducers/usage.js
```diff
@@ -0,0 +1,30 @@
+const initialState = {
+  messaging: [],
+  receipient_validation: [],
+  loading: false,
+  error: null,
+};
+
+export default (state = initialState, { type, payload }) => {
```
nicklemmon, line 8: So this isn't in use yet, you just wanted to set it up so people in the future use it safely?

src/selectors/accountBillingInfo.js
```diff
@@ -34,8 +34,10 @@ export const hasPhoneSupport = selectCondition(hasProductOnSubscription('phone_s
 const selectBillingSubscription = state => state.billing.subscription || {};
 const currentFreePlans = ['free500-1018', 'free15K-1018', 'free500-0419', 'free500-SPCEU-0419'];
 export const isManuallyBilled = state => _.get(state, 'billing.subscription.type') === 'manual';
-const getRecipientValidationUsage = state => _.get(state, 'account.rvUsage.recipient_validation');
-const getTransmissionsUsage = state => _.get(state, 'account.usage');
+const getRecipientValidationUsageFromAccountApi = state =>
+  _.get(state, 'account.rvUsage.recipient_validation');
+const getTransmissionsUsageFromAccountApi = state => _.get(state, 'account.usage');
+const getTransmissionsUsageFromUsageApi = state => _.get(state, 'usage.messaging');
```
nicklemmon, line 40: Is there any way you could add some notes explaining why you might want to get this data from the usage API vs. the account API? Or maybe just links to docs or something.

nicklemmon (changes requested):
@shrirupa I left some comments for you - nothing major that I can see though it is difficult to test locally obviously.

cypress/fixtures/usage/200.get.no-results.json
shrirupa: yes, so api will always return that data structure. start and end dates being almost always billing cycle dates.

src/pages/usage/components/MessagingUsageChart.js
shrirupa: good point!

src/reducers/usage.js
shrirupa: no actually, its in use now here `getTransmissionsUsageFromUsageApi = state => _.get(state, 'usage.messaging');`

src/selectors/accountBillingInfo.js
shrirupa: usage is in weird place in app actually, we have messaging usage being returned from GET account/include=usage but not roles and tenants have access to this and we also have usage from GET /usage and more tenants and user have access to this. for specific roles that have access, its best to consult access library. https://github.com/SparkPost/access/blob/master/lib/token-access.js

shrirupa: Added the same comment above these selectors.

cypress/tests/integration/accounts/usage/usagePage.spec.js
```diff
@@ -12,7 +12,7 @@ describe('The usage page', () => {
       requestAlias: 'subscriptionReq',
     });
     cy.stubRequest({
-      url: 'api/v1/account?include=usage',
+      url: 'api/v1/account',
       fixture: 'account/200.get.include-usage.json',
       requestAlias: 'accountReq',
     });
```
nicklemmon, line 18: Should have caught this earlier, but looks like this can be removed entirely - except in the tests that are explicitly stubbing a different account response (like the enterprise and annual plan tests).

So that means waiting for `'@accountReq'` in nearly every test except for those can be removed.

src/selectors/accountBillingInfo.js
```diff
@@ -34,8 +34,15 @@ export const hasPhoneSupport = selectCondition(hasProductOnSubscription('phone_s
 const selectBillingSubscription = state => state.billing.subscription || {};
 const currentFreePlans = ['free500-1018', 'free15K-1018', 'free500-0419', 'free500-SPCEU-0419'];
 export const isManuallyBilled = state => _.get(state, 'billing.subscription.type') === 'manual';
-const getRecipientValidationUsage = state => _.get(state, 'account.rvUsage.recipient_validation');
-const getTransmissionsUsage = state => _.get(state, 'account.usage');
+const getRecipientValidationUsageFromAccountApi = state =>
+  _.get(state, 'account.rvUsage.recipient_validation');
+//usage is in weird place in app, we have messaging usage being returned from
+//GET account/include=usage but not roles and tenants have access to this and
```
nicklemmon, line 40: Typo here "but not roles and tenants" should be "but not all roles and tenants"

nicklemmon (changes requested):
@shrirupa I found a couple of issues that I missed earlier - sorry about that.

After these we should be able to move forward.

shrirupa:
@nicklemmon  Thanks for the feedback! This is ready for last pass.

nicklemmon:
@shrirupa Looks like that error state test might be a little flaky. Mind taking a look? Is there a missing `cy.wait()` maybe?

shrirupa:
> @shrirupa Looks like that error state test might be a little flaky. Mind taking a look? Is there a missing `cy.wait()` maybe?

Looks like I missed removing one accountReq. They should all pass now.

nicklemmon (approved):

PR #2000, opened 2021-01-07
SA-1885 - (UI) Address comparator filters w/ same key are dropped by adding form validation

nicklemmon:
[SA-1885](https://sparkpost.atlassian.net/browse/SA-1885)

### What Changed

- Renamed spec files within `cypress/tests/integration/signals-analytics/analytics-report` to improve consistency
- Refactored `useFiltersForm` to use a `status` enum to help manage state
- Removes `filterByCollectionValues` helper that is no longer used in the app (I just happened to run in to this while looking at our existing array helpers)
- Adds a new error banner that renders at the top of each group that has a duplicate violation (when the user adds filters that have duplicate "Type" and "Compare By" values within the same grouping, i.e. "Campaign" "contains" and then another filter for "Campaign" "contains")

### How To Test

1. Go to `/signals/analytics` and click on "Add Filters"
1. Select "Type" of "Campaign" and in the `<select />` that appears next to it, select "contains"
1. Enter any value within the "Campaign" field and hit enter
1. Click the "AND +" button _within that same grouping_
1. _Within that same grouping_ select the exact same values for "Type" and "Compare By" ("Campaign" and "contains")
1. Try to submit the form
1. Verify the error banner renders at the top of the relevant filter grouping
1. Verify that focus is moved to the error (this can be verified via the console via `document.activeElement`)
1. Afterwards, verify the error clears when:
    -  Swapping the filter "Type" for either of the offending filters
    -  Swapping the filter "Compare By" for either of the offending filters
    -  Clearing the form via the "Clear Filters" button
1. Add an additional filter grouping (By clicking "AND +" outside of the Panel but within the drawer) and introduce another duplication error
1. Submit the form and verify focus moves to the _first error that is still present_
1. Verify that this error clears independently of the first and vice versa
1. It would be worth testing more complex scenarios as well - the error will render (but only once!) when the user has duplicates within _any_ grouping in the form - whoever picks up the review, please try and break this thing.

### To Do

- [x] Run design/copy by UX - [relevant Slack thread](https://msys.slack.com/archives/GLLAW6UAX/p1610044824129000)
- [x] Add automated testing
- [ ] Incorporate feedback


cypress/tests/integration/signals-analytics/analytics-report/filtersForm.spec.js
```diff
@@ -17,14 +17,6 @@ if (IS_HIBANA_ENABLED) {
       commonBeforeSteps();
     });
 
-    it('allows the user to add a list of grouped filters and apply them by submitting the form', () => {
```
nicklemmon, line 20: This test wasn't doing a whole lot that others weren't already handling.

src/pages/reportBuilder/components/FiltersForm/useFiltersForm.js
```diff
@@ -116,9 +131,18 @@ function reducer(state, action) {
       return {
         ...state,
         groupings: getIterableFormattedGroupings(initialGroupingsState),
+        status: 'idle',
```
nicklemmon, line 134: GitHub is collapsing this in an unusual way. This change is within the `'CLEAR_FILTERS'` case.

src/pages/reportBuilder/helpers/index.js
```diff
@@ -99,13 +99,17 @@ export function getGroupingFields(iterableGroupings) {
...
       ...grouping,
 
-      // Renders below all groups *except* the last one in the list
-      hasAndBetweenGroups: groupingIndex + 1 < groupings.length,
+      // Renders an "AND" below all groups *except* the last one in the list
+      hasAndBetweenGroups: Boolean(groupingIndex + 1 < groupings.length),
 
       // Renders below the last group in the list when one of the filters has values
-      hasAndButton:
+      hasAndButton: Boolean(
```
nicklemmon, line 106: This was returning as `undefined` when it should have been `false`

src/pages/reportBuilder/helpers/index.js
```diff
@@ -99,13 +99,17 @@ export function getGroupingFields(iterableGroupings) {
     return {
       ...grouping,
 
-      // Renders below all groups *except* the last one in the list
-      hasAndBetweenGroups: groupingIndex + 1 < groupings.length,
+      // Renders an "AND" below all groups *except* the last one in the list
```
nicklemmon, line 102: Just clarifying a comment here.

src/pages/reportBuilder/helpers/index.js
```diff
@@ -99,13 +99,17 @@ export function getGroupingFields(iterableGroupings) {
...
+      hasAndBetweenGroups: Boolean(groupingIndex + 1 < groupings.length),
```
nicklemmon, line 103: This was returning as undefined when it should have been false

src/helpers/array.js
```diff
@@ -8,47 +8,3 @@ export const toSentence = (array, conjuctionTerm = 'and') => {
     array.slice(-2).join(conjunction)
   ); //joins the last 2 items using oxford comma for >=3 items and append to end of sentence.
 };
-
-/**
- * @param {*} arr - Array to filter
- * @param {*} filters - Set of filters applied to the passed in array.
- * @example // TODO: Flesh this out
- */
-export const filterByCollectionValues = (arr, { filters }) => {
```
nicklemmon, line 17: This function was no longer in use after the introduction of `react-table`

src/pages/reportBuilder/components/FiltersForm/FiltersForm.js
```diff
@@ -28,11 +28,12 @@ import { getGroupingFields, getApiFormattedGroupings } from '../../helpers';
...
 import Typeahead from '../Typeahead';
 import {
-  TypeSelect,
+  AddButton,
   CompareBySelect,
+  DuplicateErrorBanner,
   MultiEntryController,
-  AddButton,
   RemoveButton,
+  TypeSelect,
```
nicklemmon, line 36: Just re-ordered some things and introduced `DuplicateErrorBanner` here

src/pages/reportBuilder/components/FiltersForm/FiltersForm.js
```diff
@@ -28,14 +28,17 @@ import { getGroupingFields, getApiFormattedGroupings } from '../../helpers';
...
 } from './components';
 import useFiltersForm from './useFiltersForm';
 
+const FILTER_VALUE_PLACEHOLDER_TEXT = 'e.g. resource_01 or resource_02';
```
nicklemmon, line 40: UX requested this change to make the interaction more clear

src/pages/reportBuilder/components/FiltersForm/useFiltersForm.js
```diff
@@ -18,33 +20,53 @@ function reducer(state, action) {
...
     }
 
     case 'SET_FILTER_COMPARE_BY': {
-      const { groupings } = state;
+      const { groupings, status: prevStatus } = state;
       const targetFilter = groupings[action.groupingIndex].filters[action.filterIndex];
       targetFilter.compareBy = action.compareBy;
       targetFilter.values = [];
+      // Only show the error if coming from the error state post submission
+      const nextStatus =
```
PatrickSison, line 51: Wouldn't this logic just need to be if it's an error previously, just swap the state to editing? Or is this a just to capture the case of the duplicate grouping error?

PatrickSison (commented):
I think most things look good. I'm still a little concerned with the UX since the error banner pops up in the first filter section (although it's in the right grouping). There's not a necessarily clean way to calculate that logic off the top of my head though.

Example use case would be
[goodFilter, goodFilter, badFilter, goodFilter, badFilter]. That means the box pops up at the top, and doesn't exactly identify the offending filter. There's also a lot of UI elements, so I feel like the user might get lost. I'm also willing to defer for now since this is already a better experience than before

src/pages/reportBuilder/components/FiltersForm/FiltersForm.js
```diff
@@ -141,6 +162,14 @@ function FiltersForm({
                           position="relative"
                           key={`filters-${groupingIndex}-${filterIndex}`}
                         >
+                          {status === 'error' &&
```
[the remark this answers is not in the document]
nicklemmon: I had it coded that way originally but UX requested we change it to on submit only instead so that it moves in to focus when the error occurs.

src/pages/reportBuilder/components/FiltersForm/useFiltersForm.js
nicklemmon: Right, this is in the event they don't actually fix the error within a group. Say, for example, they just set the filter compare by for a filter that isn't contributing to the error.

PatrickSison:
I think just needs a rebase and this should be good to go. This is working as intended and is better than the previous situation

PatrickSison (approved):
Looks good still

PR #2001, opened 2021-01-07
FE-1285 - Add new ESLint rule warning when the user imports moment

nicklemmon:
[FE-1285](https://sparkpost.atlassian.net/browse/FE-1285)

### What Changed

- Adds new rule that warns when importing `moment` with a custom warning message

### How To Test

- Open a JS module that is importing `moment`, verify that an ESLint warning occurs
- Does the warning message make sense to you?

### To Do

- [ ] Incorporate feedback


gfriedmansp (approved):
Cool! Wonder if there's any other implied rules we're following like this that could be added. Works as expected.

nicklemmon:
> Cool! Wonder if there's any other implied rules we're following like this that could be added. Works as expected.

Another good one would probably be warning when someone imports `*.module.scss` since we are trying to move away from CSS modules.

PR #2003, opened 2021-01-08
SA-1984: use usernames for recipients in Scheduled Reports

kchu1337:

### What Changed

- Pass username into scheduled reports api instead of name.

### How To Test

- Create a report here `/signals/schedule/bounce`
- Add recipients with custom names. IE, for {name}<{email}>, their name does not follow the `first-last-####`.
- Verify that the API receives username, not name as the values for recipients.
- Also use `Send Test` button to verify that the email is being sent.

### To Do


src/pages/reportBuilder/helpers/tests/scheduledReports.test.js
```diff
@@ -1,20 +1,20 @@
...
-  { name: 'bob2', email: 'bob2@theBuilder.com' },
+  { name: 'bob', email: 'bob@theBuilder.com', username: 'bob-123' },
+  { name: 'bob2', email: 'bob2@theBuilder.com', username: 'bob2-456' },
 ];
 
 describe('getDefaultValues and ', () => {
   const formValuesBase = {
     name: 'test',
-    recipients: [{ name: 'bob', email: 'bob@theBuilder.com' }],
+    recipients: [{ name: 'bob', email: 'bob@theBuilder.com', username: 'bob-123' }],
```
nicklemmon, line 11: Does the `name` property need to be here anymore?

nicklemmon (changes requested):
@kchu1337 Looks good - I just see one tiny tweak to a test should be made and then I'll gladly approve.

src/pages/reportBuilder/helpers/tests/scheduledReports.test.js
kchu1337: It doesn't have to be, but i kept it there since name is still a key of the user object. I just simplified the user object to only what i need.

kchu1337: It also acts as sort of a regression test to make sure that username, and not name is the correct value passed.

nicklemmon: That does make sense to me. Let's move forward.

nicklemmon (approved):
@kchu1337 Looks good!

PR #2007, opened 2021-01-12
FE-1334: Add the functionality on new domain pages to verify domain through url

shrirupa:
FE-1334: Add the functionality on new domain pages to verify domain through url

### What Changed

- Added VerifyToken component on Domains list page
- Whenever domain/list/sending?domain={{domain}}&token={{token}}&mailbox={{mailbox}} is loaded, verify action will trigger on page load
      - if the domain is verified then it would redirect to domain details page with a success alert
      - if domain is not verified then it would stay on the list page with unsuccessful alert
      
      
### How To Test

- Load http://localhost:3100/domains/list/sending?domain=unverified.com&mailbox=abuse&token=akaka url, you should see a unsuccessful alert
- From appteam staging send yourself a email from here 
Screen Shot 2021-01-13 at 5 01 50 PM
     - use the token and mailboax in the link and navigate to http://localhost:3100/domains/list/sending?domain={{domain}}&mailbox={{mailbox}}&token={{token}}

### To Do

- [ ] Address feedback

### DIFF
--- a/cypress/tests/integration/configuration/domains/listPage.spec.js
+++ b/cypress/tests/integration/configuration/domains/listPage.spec.js
@@ -68,22 +68,20 @@ describe('The domains list page', () => {
       stubSendingDomains({ fixture: 'sending-domains/200.get.paginated-results.json' });
       stubSubaccounts();
       cy.stubRequest({
-        url: '/api/v1/sending-domains/failed-verification.com/verify',
+        url: '/api/v1/sending-domains/fake1.domain.com/verify',
         method: 'POST',
         fixture: 'sending-domains/verify/200.post-abusetoken.json',
         requestAlias: 'sendingDomainVerifyReq',
       });
       cy.stubRequest({
-        url: '/api/v1/sending-domains/failed-verification.com',
+        url: '/api/v1/sending-domains/fake1.domain.com',
         fixture: 'sending-domains/200.get.all-verified.json',
         requestAlias: 'sendingDomainsReq',
       });
-      cy.visit(
-        `${PAGE_URL}/list/sending?token=faketoken&domain=failed-verification.com&mailbox=abuse`,
-      );
+      cy.visit(`${PAGE_URL}/list/sending?token=faketoken&domain=fake1.domain.com&mailbox=abuse`);
 
-      cy.wait(['@sendingDomainsReq', '@subaccountsReq', '@sendingDomainsReq']);
-      cy.findByText('failed-verification.com has been verified').should('be.visible');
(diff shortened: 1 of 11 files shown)


cypress/tests/integration/configuration/domains/listPage.spec.js
```diff
@@ -64,6 +64,46 @@ describe('The domains list page', () => {
       cy.url().should('include', `${PAGE_URL}/list/sending`);
     });
 
+    it('successfully verifies the sending/bounce domain when url has token, mailbox and domain as query parameters and redirects to domain details page', () => {
```
nstanard, line 67: nit: The domain name used here makes for an odd success message. 😅 
image

src/pages/sendingDomains/components/VerifyToken.js
```diff
@@ -2,7 +2,11 @@ import React, { Component } from 'react'; // eslint-disable-line no-unused-vars
 import { connect } from 'react-redux';
```
nstanard, line 2: Note: I believe this is only prettier changes.

src/reducers/sendingDomains.js
```diff
@@ -104,7 +104,7 @@ export default (state = initialState, { type, payload, meta }) => {
     case 'VERIFY_TOKEN_PENDING':
       return {
         ...state,
-        verifyTokenLoading: false,
```
nstanard, line 107: How might I test to make sure this won't cause any issues in the old sending/bounce domain page and loading state?

src/pages/sendingDomains/components/VerifyToken.js
shrirupa: yep, just prettier changes.

src/reducers/sendingDomains.js
shrirupa: just did a search for it verifyTokenLoading in vscode, we weren't using it anywhere in the old sending domain pages

src/pages/domains/ListPage.js
```diff
@@ -154,6 +157,13 @@ function DomainTabPages() {
       }}
       loading={listPending || isFirstRender}
```
nstanard, line 158: I wonder if we can/should pass verifyTokenLoading into here too (and set this to && instead of || probably?)

shrirupa:
👀  looking at the failing tests

nstanard (approved):
I was able to verify a domain via email and follow the link back to the app, where it loaded the details page with a success message. LGTM!

PR #2012, opened 2021-01-14
SA-2003: Added fix for filters not properly being unset

PatrickSison:
### What Changed

- Removing filters properly unsets from query param

### How To Test

- Go to report builder
- Click Add Filters and add at least 1 filter
- Apply
- Verify `query_filters` param exists in URL
- Go to main report builder page
- Remove all filters via tags
- Verify `query_filters` doesn't exist in URL
- Refresh page
- Verify that filters don't repopulate from URL

### To Do
- [ ] Addressed feedback


src/pages/reportBuilder/components/ReportOptions.js
```diff
@@ -48,6 +48,7 @@ export function ReportOptions(props) {
       const { filters: selectedFilters, ...update } = selectSummaryChartSearchOptions;
       const { filters } = reportOptions;
 
+      update.query_filters = filters;
       if (Boolean(filters.length)) {
         update.query_filters = JSON.stringify(dehydrateFilters(filters));
       }
```
nicklemmon, line 54: I don't understand 🤔  - setting `query_filters` twice seems odd? Wouldn't they get set twice in this scenario?

If that's not the case, adding some clarifying comments explaining why these are both necessary would be helpful.

cypress/tests/integration/signals-analytics/analytics-report/reportPage.spec.js
```diff
@@ -336,6 +337,8 @@ if (IS_HIBANA_ENABLED) {
 
       // No tags render when no filters are applied
       getFilterTags().should('not.exist');
+      cy.url().should('not.contain', 'query_filters');
```
nicklemmon, line 340: Nice.

nicklemmon (commented):
@PatrickSison This fix works - though I'm a little confused _how_ it's working. Mind clarifying a bit?

nicklemmon (approved):

PR #2013, opened 2021-01-14
FE-1344 - track onboarding in segment

nstanard:
### What Changed
- Tracking onboarding steps in segment as long as it's not set to "done"

### How To Test
- Create a new account and go to the dashboard. 
- Go through each onboarding step manually calling the API to get past the sending domain verification process.
- Make sure calls are going out to segment for each step.

### To Do
- [x] Test
- [x] Feedback

### DIFF
--- a/src/helpers/segment.js
+++ b/src/helpers/segment.js
@@ -19,6 +19,7 @@ export const SEGMENT_EVENTS = {
   SENDING_DOMAIN_ADDED: 'Sending Domain Added',
   SENDING_DOMAIN_VERIFIED: 'Sending Domain Verified',
   EMPTY_STATE_LOADED: 'Empty state loaded',
+  ONBOARDING: 'Onboarding',
   REPORT_BUILDER_COMPARISON_ADDED: 'Report Builder Comparison Added',
   SCHEDULED_REPORT_CREATED: 'Created Scheduled Report',
 };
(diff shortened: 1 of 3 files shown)


src/helpers/segment.js
```diff
@@ -19,6 +19,7 @@ export const SEGMENT_EVENTS = {
   SENDING_DOMAIN_ADDED: 'Sending Domain Added',
   SENDING_DOMAIN_VERIFIED: 'Sending Domain Verified',
   EMPTY_STATE_LOADED: 'Empty state loaded',
+  ONBOARDING: 'Onboarding',
```
nstanard, line 22: Not sure if this is what we want, maybe DashboardV2 Onboarding?

jonambas: I would use a verb for the event so its consistent with the others, like "Dashboard Onboarding Loaded" or something

nstanard: I like that. Updated!

src/pages/dashboardV2/DashboardPageV2.container.js
```diff
@@ -66,6 +67,12 @@ function mapStateToProps(state) {
     onboarding = 'analyticsReportPromo';
   }
 
+  if (onboarding !== 'done') {
+    segmentTrack(SEGMENT_EVENTS.DASHBOARD_ONBOARDING, {
```
bkemper, line 71: The desired result is to only call `segmentTrack` when `onboarding` has changed and doesn't each "done".  `mapStateToProps` on the other hand is going to be called whenever state changes which is more than we want.  I recommend moving this to a `useEffect` in the component to control number of times it is called.

nstanard: testing now to confirm it's what we want

jonambas (approved):
This looks good to me – verified the track event is called once on load with the step

PR #2014, opened 2021-01-15
FE-1338 - Report Builder Info Banner

nstanard:
Branched from [FE-1270](github.com/SparkPost/2web2ui/pull/2005)

### What Changed
- Info banner for report builder doc and seedlist buttons was added at the top of the page

### How To Test
- On an account that hasn't dismissed the banner yet, go to the report builder page with at least one verified sending domain.
- The info banner should show at the top of the page until the user dismisses it.
- 

### Questions
- What's the link for the view seedlist button?
- When should that view seedlist button show? Always?

### To Do
- [x] Cypress test coverage
- [x] Feedback

### DIFF
--- a/src/pages/reportBuilder/components/InfoBanner.js
+++ b/src/pages/reportBuilder/components/InfoBanner.js
@@ -1,0 +1,50 @@
+import React, { useState } from 'react';
+import { useDispatch, useSelector } from 'react-redux';
+import { Banner, Picture } from 'src/components/matchbox';
+import AnalyticsJpg from '@sparkpost/matchbox-media/images/Analytics.jpg';
+import AnalyticsWebp from '@sparkpost/matchbox-media/images/Analytics.webp';
+import { updateUserUIOptions } from 'src/actions/currentUser';
+import { isUserUiOptionSet } from 'src/helpers/conditions/user';
+import { LINKS } from 'src/constants';
+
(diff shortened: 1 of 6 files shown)


src/pages/reportBuilder/components/InfoBanner.js
```diff
@@ -0,0 +1,50 @@
...
+      title="All the Metrics in One Place"
+      mb="600"
+    >
+      <p>
+        Build and save custom reports with SparkPost's easy to use dashboard. Apply unlimited
+        metrics across delivery and deliverability data. To learn how to unlock the full potential
+        of SparkPost's Analytics Report, visit the documentation below.
+      </p>
+      {/* TODO: VIEW SEEDLIST TO LINK */}
+      <Banner.Action color="blue" to="/inbox-placement/seedlist">
```
nstanard, line 35: This link doesn't actually work yet - need to find out where it should go.

Also need to find out when this should show. There are two mocks and one doesn't have this button.

tokonoma: Neal the mockup you're referring to, with the link to the seedlist, is for when signals deliverability is live and users have a route to that list. Until then, that empty state shouldn't be presented (and presumably can't be built completely until that route is available)

shrirupa (approved):
Looks good to me!

PR #2025, opened 2021-01-20
FE-1329 - Remove OG vs. Hibana-specific scripts and CI jobs

nicklemmon:
### What Changed

- Removes OG and Hibana-specific jobs from the CircleCI config
- Increased parallelism on integration tests from 5 to 6
- Removes Hibana-specific scripts from NPM scripts

### How To Test

- CircleCI tests should continue to run as expected with Codecov calculation and artifact storage

### To Do

- [ ] Incorporate feedback


.circleci/config.yml
```diff
@@ -35,16 +35,12 @@ commands:
     steps:
       - checkout
       - restore_npm_cache
-      - run: npm ci
+      - run:
+          name: 'Installing NPM dependencies'
+          command: npm ci
```
shrirupa, line 40: Looks like this step is the second most time consuming step(~ 1.5 min each time), so if we cut down runner may be we can save some time, not sure though.

shrirupa (approved):
Looks good to me! Thanks for working on this @nicklemmon  🎉

PR #2030, opened 2021-01-21
FE-1327 - Upgrades react-query to version 3.5+

nicklemmon:
[FE-1327](https://sparkpost.atlassian.net/browse/FE-1327)

### What Changed

- Upgrades `react-query` to version 3.5+
- Upgrades `enzyme`-related dependencies that were causing test failures
- Installs `nock` (https://github.com/nock/nock) for stubbing API requests in a Jest DOM environment
- Adds new `stubRequest` test helper that wraps `nock.intercept()`. Very similar API to `cy.stubRequest()`
- Refactors `useSparkPostQuery` by breaking up logic in to re-usable helpers
- Adds Jest unit testing for said helper functions as well as `useSparkPostQuery` and `useSparkPostQueries`
- Refactors `useSparkPostQueries` to more closely align with the newly released [useQueries](https://react-query.tanstack.com/reference/useQueries#_top)

### How To Test

- Verify all features that leverage `useSparkPostQuery` and `useSparkPostQueries` continue to function as expected
- Verify relevant steps from the [React Query v3 migration guide](https://react-query.tanstack.com/guides/migrating-to-react-query-3) were taken
- Does changing the configuration value for `apiBase` in `test-config.js` introduce any breaking changes? Is this just used for our Jest environment or is this used elsewhere?

### To Do

- [x] Incorporate feedback
- [x] Cut a follow-up ticket to add `useSparkPostMutation` hook wrapping [useMutation](https://react-query.tanstack.com/reference/useMutation#_top)


config/jest/setup.js
```diff
@@ -52,7 +52,6 @@ Object.defineProperty(global.window, 'matchMedia', {
   }),
 });
 Object.defineProperty(global.window, 'scrollTo', { value: jest.fn(), configurable: true });
-Object.defineProperty(global.window.location, 'assign', { value: jest.fn(), configurable: true });
```
nicklemmon, line 55: Newer versions of enzyme-related dependencies have made this unnecessary

jest.config.js
```diff
@@ -24,7 +24,7 @@ module.exports = {
     '<rootDir>/(eslint-rules|src|scripts)/**/*.test.js?(x)',
   ],
   testEnvironment: 'node',
-  testURL: 'http://phoenix.test',
+  testURL: 'http://mock.sparkpost.test',
```
nicklemmon, line 27: Makes things a little less confusing, I think

package.json
```diff
@@ -138,7 +138,7 @@
     "@testing-library/cypress": "^6.0.0",
     "@testing-library/jest-dom": "^5.10.1",
     "@testing-library/react": "^10.2.1",
-    "@testing-library/react-hooks": "^3.3.0",
+    "@testing-library/react-hooks": "^5.0.0",
```
nicklemmon, line 141: We were waaay behind here

src/pages/reportBuilder/components/CompareByForm/Typeahead.js
```diff
@@ -228,7 +228,7 @@ function Typeahead({ id, onChange, lookaheadRequest, lookaheadOptions, selector,
         match: inputValue,
         limit: METRICS_API_LIMIT,
       }),
-    { enabled: inputValue && inputValue.length >= 3, refetchOnWindowFocus: false },
+    { enabled: Boolean(inputValue && inputValue.length >= 3), refetchOnWindowFocus: false },
```
PatrickSison, line 231: I know it's outside the scope of this, but could you add in a `staleTime` to allow this hook to cache the requests it makes? Would be greatly beneficial to reduce load on API

src/hooks/api/tests/useSparkPostQueries.test.js
```diff
@@ -0,0 +1,58 @@
...
+  cacheTime: 0,
+};
+
+const wrapper = ({ children }) => (
+  <TestApp store={{ auth: { token: FAKE_AUTH_TOKEN, loggedIn: true } }}>{children}</TestApp>
+);
+
+describe('useSparkPostQueries', () => {
+  it('makes multiple network requests based on the passed in array of query functions', async () => {
+    stubRequest({
```
PatrickSison, line 37: Nice. Good way to test the hooks are working all the way through

PatrickSison (commented):
I think only comment is in regards to updates with our cache. Everything else looks good. Was going to say initially we could implement helper functions to useSparkPostQueries (like having a combined status field), but I doubt it's too useful right now since this is a rare use case hook

src/pages/reportBuilder/components/CompareByForm/Typeahead.js
nicklemmon: https://msys.slack.com/archives/GLLAW6UAX/p1612464846131600

So looks like:

1. Set the stale time to something fairly long like 2 minutes (is that long? I think it's long...)
2. Explore a way to use `queryClient.invalidateQueries()` (https://react-query.tanstack.com/reference/QueryClient#queryclientinvalidatequeries) to invalidate the cache for those particular requests when the form mounts - this might be tough to do though not totally sure yet

I'll give it a shot.

nicklemmon:
@PatrickSison Ready for another pass when you have time.

PatrickSison (approved):
This is working really well. 
Weird behavior with the typeahead caching. It appears to be caching? But then also still has an XHR going out? Not sure what's going on there, but don't want it to block the release of this ticket at this point.

PR #2031, opened 2021-01-21
FE-1296 - Resolves events page bug in which invalid ranges caused an application crash

nicklemmon:
[FE-1296](https://sparkpost.atlassian.net/browse/FE-1296)

### What Changed

- Adds check for valid range before updated returned date options in the `parseSearch` message events helper
- Adds supporting test case

### How To Test

1. Go to `/reports/message-events` and update the URL to `/reports/message-events?from=2021-01-21T17%3A27%3A38Z&range=29days&to=2021-01-21T18%3A27%3A38Z`
2. Verify that the `range` in the URL changes to `hour` and the page doesn't crash
3. Set different, valid ranges through the UI and through the URL and verify they still work as expected (i.e., range is still passed correctly to the relevant network requests)

### To Do

- [ ] Incorporate feedback


src/helpers/tests/messageEvents.test.js
```diff
@@ -53,104 +69,111 @@ describe('messageEvents helpers', () => {
...
+          to: new Date('2018-03-23T04:02:32.000Z'),
+        },
       },
-    'parses correctly when adding in unsupported filters by removing them':
-      {
-        searchText: `?${defaultDateURL}&filter3=foo`,
-        expected: {
-          dateOptions: mockRelativeRangeToDate
-        }}
+    },
```
nicklemmon, line 147: This is all that was added - the rest is Prettier.

src/helpers/messageEvents.js
```diff
@@ -25,14 +26,18 @@ export function formatDocumentation(data) {
...
+      ...fieldDescriptions,
     };
   });
+  /* eslint-enable lodash/preferred-alias */
 
   return events;
 }
 export function parseSearch(search) {
   const { from, to, range, ...rest } = qs.parse(search);
+  // If the passed in range is present in configuration, then a valid range has been set
```
nicklemmon, line 38: Changes were only made here...

src/helpers/messageEvents.js
```diff
@@ -43,20 +48,23 @@ export function parseSearch(search) {
     dateOptions.to = new Date(to);
   }
 
-  if (range) {
+  if (hasValidRange) {
```
nicklemmon, line 51: ...and here. The rest is Prettier.

shrirupa (approved):
Looks good to me!

PR #2041, opened 2021-01-26
AC-1851 - Merge account usage into main usage

nstanard:
### What Changed
- Merged the account usage response into the main usage response to make sure limit data is correctly displaying.

image

### How To Test
- Go to the `/usage` page and make sure daily and monthly limit data is displaying.
(this data will only show if on accounts that have limits & not on enterprise)

### To Do
- [ ] Feedback

### Relates to
- https://sparkpost.atlassian.net/browse/FE-1249

### DIFF
--- a/src/pages/usage/UsagePage.js
+++ b/src/pages/usage/UsagePage.js
@@ -56,7 +56,7 @@ export default function UsagePage() {
   // Merging data so existing selectors can work together to grab from a common object
   const data = { account, subscription, usage, usageHistory };
 
-  // NOTE: API usage data discrepancy
+  // NOTE: AC-1851 - API usage data discrepancy
   usage.messaging.day = {
     ...account.usage.day,
     ...usage.messaging.day, // Usage second so if the API fixes the response, it'll win out over the account call
@@ -65,7 +65,7 @@ export default function UsagePage() {
     ...account.usage.month,
     ...usage.messaging.month, // Usage second so if the API fixes the response, it'll win out over the account call
   };
-  // NOTE: API usage data discrepancy
+  // NOTE: AC-1851 - API usage data discrepancy
 
   const endOfBillingPeriod = selectEndOfMonthlyUsage(data);
   const startOfBillingPeriod = selectStartOfMonthlyUsage(data);
--- a/src/pages/usage/UsagePage.js
+++ b/src/pages/usage/UsagePage.js
@@ -12,8 +12,8 @@ import { MessagingUsageSection, FeatureUsageSection, RVUsageSection } from './co
 
 export default function UsagePage() {
   // API Requests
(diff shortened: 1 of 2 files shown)


src/pages/usage/UsagePage.js
```diff
@@ -12,8 +12,8 @@ import { MessagingUsageSection, FeatureUsageSection, RVUsageSection } from './co
 
 export default function UsagePage() {
   // API Requests
-  const { data: account, status: accountStatus, refetch: accountRefetch } = useSparkPostQuery(
-    getAccount,
+  const { data: account, status: accountStatus, refetch: accountRefetch } = useSparkPostQuery(() =>
+    getAccount({ include: 'usage' }),
```
bkemper, line 16: Should we use /account?  I thought the push was away from it and to use /billing/subscription and /usage.

nstanard: (we chatted on a call but for historical preservation)

I wasn't able to find the daily and monthly limit data through any other API call. For example, here's the billing subscription response for one of my users. I wouldn't mind having a call with an API engineer to discuss the /usage and /account?include=usage calls and how we are using them. 

image

Two tickets have been created as a result of the data discrepancy coming from the API. 

(the current ticket/pr) as well as https://sparkpost.atlassian.net/browse/FE-1249

src/pages/usage/UsagePage.js
```diff
@@ -55,6 +55,18 @@ export default function UsagePage() {
 
   // Merging data so existing selectors can work together to grab from a common object
   const data = { account, subscription, usage, usageHistory };
+
+  // NOTE: AC-1851 - API usage data discrepancy
+  usage.messaging.day = {
+    ...account.usage.day,
```
shrirupa, line 61: test in usagePage.spec.js need to be updated to use 

> cy.stubRequest({
>       url: 'api/v1/account*',
>     });

since you added back account?include=usage

shrirupa: I think after doing this you will notice the test for enterprise customers fail.

shrirupa: I think little update to fix this would just be 
``` 
usage.messaging.day = {
    ...account?.usage?.day,
    ...usage.messaging.day, 
  };
  usage.messaging.month = {
    ...account?.usage?.month,
    ...usage.messaging.month,
  };```

shrirupa: Sorry for the long comment, just wanted to note down how I went about testing this.

nstanard: Did it for both account and usage just to be safe. Good callout!

shrirupa (approved):
Looks good to me!

PR #2043, opened 2021-01-26
FE-1337 - Fix custom relativeRange reports on dashboard v2

nstanard:
### What Changed
- Custom relativeRange reports were calling into `getRelativeDates` and that function wasn't returning from and to information. That was resulting in the response from `reportOptionsWithDates` to not have from and to, and as a result the pinnedReport `pinnedReport` object used in DashboardPageV2.js had no from and to information. 
- This PR ensures that custom relativeRange reports have their from and to properties set for displaying on the dashboard. 

### How To Test
- Go to the /dashboard page and change reports to and from custom and non custom date range reports. Use the summary report too - and check to make sure the to and from dates used/showing are correct for the custom saved range. 

### To Do
- [x] Test (need to make sure the exact change I made is the best one I could have made, I have a feeling `getRelativeDates` could also be updated instead, to return the to and from datapoints, instead of setting a condition like I have here.)
- [x] Feedback

### DIFF
--- a/src/hooks/usePinnedRport/usePinnedReport.js
+++ b/src/hooks/usePinnedRport/usePinnedReport.js
@@ -35,10 +35,11 @@ export default function usePinnedReport(onboarding) {
   const reportOptionsWithDates = reportOptions => {
     const { relativeRange, precision } = reportOptions;
     const { from, to } = getRelativeDates(relativeRange, { precision });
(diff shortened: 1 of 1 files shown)


PatrickSison (approved):
Looks good. Hopefully we can clean up getRelativeDates in the future to make it easier to work with dates all around

PR #2044, opened 2021-01-26
SA-2053: Fix signals health score for mailbox provider

kchu1337:
### What Changed

- Fixed Mailbox Provider facet for Health Score due to new formatted names.
- Removed formatter for mb health score calls in API
- Encoded the url for mb provider to account for strange formatted names.

### How To Test

- Go to `/signals/health-score/`
- In the table, set facet filter to mailbox provider.
- Click on a mailbox provider (Do multiple with strange names such as Hotmail / Outlook & Verizon Media Group(.
- Verify that the heath score page works.


src/actions/signals.js
```diff
@@ -39,8 +40,9 @@ const signalsActionCreator = ({ dimension, type }) => ({
     filter = '';
   }
 
-  if (facet === 'mb_provider' && filter) {
-    filter = filter.toLowerCase().replace(' ', '_');
+  //Block spam trap request for mb provider
+  if (facet === 'mb_provider' && dimension === 'spam-hits') {
+    return () => undefined;
```
kchu1337, line 45: Replaced previous functions with this. It actually does a separate thing. I stopped trying to format the facet and added a check to stop the page from hitting the spam-hits endpoint when faceting by mailbox provider.

src/selectors/signals.js
```diff
@@ -30,6 +30,11 @@ export const getComplaintsByCohortData = state => _.get(state, 'signals.complain
 export const getHealthScoreData = state => _.get(state, 'signals.healthScore', {});
 export const getCurrentHealthScoreData = state => _.get(state, 'signals.currentHealthScore', {});
 
+export const getFacetIdFromParams = createSelector(
```
nicklemmon, line 33: You added this selector, but it isn't being used anywhere except the test?

nicklemmon (commented):
@kchu1337 The core bug appears to be fixed, though I'm noticing the filters are being applied differently when searching:

**This is production:**
Screen Shot 2021-01-28 at 10 32 19 AM

**This is your branch:**
Screen Shot 2021-01-28 at 10 32 24 AM

When I search for "Other Providers" I get no results in either environment despite that being a valid mailbox provider. This is all on the 108 account.

src/selectors/signals.js
kchu1337: I renamed the original to `getFacetIdFromParamsRaw` and made this new one that decodes the URI for mailbox provider.

nicklemmon (approved):
@kchu1337 👍  Looks good.

PR #2048, opened 2021-01-29
SA-2055: Performance improvements on typeahead

PatrickSison:
### What Changed

- Typeahead now passes to/from dates
- Typeahead now uses react-query to help with caching
- Typeahead now has debounce of 500ms

### How To Test

- Go to report builder
- Verify that typeaheads pass in proper params in filters form
- Verify that typeahesds pass in proper params in compare by
- Verify that typeaheads have longer debounce time

### To Do

- [ ] Address feedback

### DIFF
--- a/cypress/tests/integration/signals-analytics/analytics-report/compareBy.spec.js
+++ b/cypress/tests/integration/signals-analytics/analytics-report/compareBy.spec.js
@@ -231,17 +231,21 @@ describe('Analytics Report - Compare By', () => {
             .click();
           cy.findAllByLabelText('Subaccount')
             .eq(index + 2)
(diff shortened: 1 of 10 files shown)


PatrickSison:
Not sure if the caching is working... Seems to make request still. Wonder if it's how we have the react-query key setup?

nstanard (approved):
Asked a bunch of questions, pulled it down locally and ran the tests, reverted changes and added them back and compared against staging and production while using the filter and compare typeaheads in the slideout drawer of the analytics report page: 

image

This PR LGTM, let me know when you want to ship this and I can help get it moved along.

PR #2053, opened 2021-02-01
FE-1298 - Recommend using a subdomain when creating

nstanard:
### What Changed
- Added a mini banner in the domain and assignment to draw attention to using a subdomain and not a root domain.

### How To Test
- Go to the domain create form/page.
- Look at the left sidebar for the "domain and assignment" section.

### To Do
- [x] Talk with UX about no dismiss action. (no dismiss is fine)
- [x] Talk with UX about pronoun usage. (got updated verbiage from UX after asking)
- [x] Ask UX if the mini banner should always show or only after they select bounce domain to create. (always show banner)
- [ ] Feedback


shrirupa (approved):
Looks good to me! may be we can add a line to test this in cypress.

PR #2055, opened 2021-02-02
SA-2055: Hotfix for sending-ips and ip pools

PatrickSison:


nstanard (approved):
that's better:
image

lgtm

PR #2056, opened 2021-02-02
SA-2057 - Fixes compare by aggregated metrics to render calculated metrics

nicklemmon:
[SA-2057](https://sparkpost.atlassian.net/browse/SA-2057)

### What Changed

- Leverages `transformData` helper from `src/helpers/metrics` to properly render calculated metrics (metrics that has the `'computeKey'` key
- Updates integration and unit tests accordingly

### How To Test

1. Go to `/signals/analytics` and a calculated metric like "Accepted Rate"
2. Click on "Add Comparison" and compare two subaccounts
3. Verify that "Accepted Rate" renders below the charts as a percentage
4. Save this report and pin it to the dashboard.
5. Verify the data renders correctly on the dashboard as well

### To Do

- [ ] Incorporate feedback


PatrickSison (approved):
Nice! Good fix and testing around it

PR #2061, opened 2021-02-04
FE-1353 - dont show dashboard onboarding for on prem (allow_events_ingest)

nstanard:
### What Changed
- Skip onboarding on the dashboard for on prem users.

### How To Test
- Load up an account that has allow_events_ingest and go to the dashboard. (shouldn't see onboarding)
- Load up an account without allow_events_ingest and go to the dashboard. (should see onboarding)

### To Do
- [ ] Feedback


src/config/routes/signals.js
```diff
@@ -251,6 +251,7 @@ const signalsRoutes = [
   {
     path: '/signals/integration',
     component: signals.IntegrationPage,
+
     condition: hasAccountOptionEnabled('allow_events_ingest'),
```
nstanard, line 255: Is this the correct condition? Shouldn't it be `getAccountUiOptionValue` with state passed in? Or maybe state gets passed in later where these objects get used. But shouldn't it be `getAccountUiOptionValue` not `hasAccountOptionEnabled` @PatrickSison @bkemper ?

PatrickSison: Account option are tied to the account, and UI options are an nested object inside options. The account object would be similar to something like this

```javascript
{
  options: {
    allow_events_ingest: true,
    ...,
    ui: {
      custom_option: true,
      ...
    }
  }
}

```

PatrickSison: The UI object is freeform but the other options are a defined list. Those 2 functions are different accessors. 1 for main options on the top level, the second would be for just the UI options

nstanard: Sorry, I should have been more specific - the reason I ask is because the fixtures have the following object structure:

```
"options": {
      "ui": {
        ...
        "allow_events_ingest": true,
        ...
      }
}
```

nstanard: Meanwhile.... our actual redux store is both structures.... 🤦 

image

bkemper: It is better to use `hasAccountOptionEnabled` vs `getAccountUiOptionValue` when the expectation is a boolean value.  `has` will guarantee a boolean value where `getAccountUiOptionValue` may provide you either a boolean or undefined.  The use case for `getAccountUiOptionValue` was on the dashboard when we wanted to save the current step you were on as a string.

Regarding usage, this is a bad example because route configuration doesn't have access to state.  This condition is eventually handed off to `<AccessControl />`, see https://github.com/SparkPost/2web2ui/blob/master/src/components/auth/ProtectedRoute.js#L15

cypress/fixtures/account/200.get.json
```diff
@@ -22,6 +22,7 @@
       "type": "default"
     },
     "options": {
+      "allow_events_ingest": true,
```
nstanard, line 25: this change causes at least the following tests to fail:

```
it.only('all nav links renders correctly for templates', () => {
it.only('all nav links renders correctly for reporting user', () => {
```

cypress/fixtures/account/200.get.on-prem.json
```diff
@@ -0,0 +1,92 @@
+{
+  "results": {
```
shrirupa, line 2: I think account/200.get.has-integration-page.json is for the same purpose, we should probably keep only one of them.

shrirupa (changes requested):
~when i created a new account, even then I didn't see the other on-boarding steps ~

shrirupa (approved):

PR #2062, opened 2021-02-04
FE-1340 - No domain status filters selected by default

nstanard:
### What Changed
- All domain status filters are unchecked by default, then preserved using the url query params between page loads and using the browser back button.
- Clears filter state when switching from sending -> bounce tab. (request from UX)
- Leaves filter state when switching from bounce -> sending tab. (request from UX)

### How To Test
- go to /domains/list/sending and test the domain status filter on all three tabs
- test navigating away from the page and using the browser back button
- Test landing the page, updating the filters, switching between the tabs and refreshing the page (note the filter state reset logic mentioned in "what changed")

### To Do
- [ ] Feedback

### DIFF
--- a/src/pages/domains/components/SendingDomainsTab.js
+++ b/src/pages/domains/components/SendingDomainsTab.js
@@ -1,9 +1,9 @@
-import React, { useCallback, useReducer } from 'react';
+import React, { useEffect, useCallback, useRef, useReducer } from 'react';
 import { useFilters, usePagination, useSortBy, useTable } from 'react-table';
 import { ApiErrorBanner, Empty, Loading } from 'src/components';
 import { Pagination } from 'src/components/collection';
 import { DEFAULT_CURRENT_PAGE, DEFAULT_PER_PAGE } from 'src/constants';
-import { usePageFilters } from 'src/hooks';
+import { usePageFilters, usePrevious } from 'src/hooks';
 import { API_ERROR_MESSAGE } from '../constants';
 import useDomains from '../hooks/useDomains';
 import SendingDomainsTable from './SendingDomainsTable';
@@ -11,6 +11,7 @@ import TableFilters, { reducer as tableFiltersReducer } from './TableFilters';
 import {
   getReactTableFilters,
   customDomainStatusFilter,
+  getActiveStatusFilters,
   setCheckboxIsChecked,
   filterStateToParams,
 } from '../helpers';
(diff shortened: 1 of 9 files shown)


src/pages/domains/components/SendingDomainsTab.js
```diff
@@ -174,12 +161,12 @@ export default function SendingDomainsTab({ renderBounceOnly = false }) {
       { Header: 'SharedWithSubaccounts', accessor: 'sharedWithSubaccounts', canFilter: false },
       { Header: 'SubaccountId', accessor: 'subaccountId', canFilter: false },
       { Header: 'SubaccountName', accessor: 'subaccountName', canFilter: false },
+      { Header: 'DefaultBounceDomain', accessor: 'defaultBounceDomain', canFilter: false },
       {
         Header: 'DomainStatus',
         accessor: row => ({
           readyForBounce: row.readyForBounce,
           blocked: row.blocked,
-          defaultBounceDomain: row.defaultBounceDomain,
```
shrirupa, line 182: according to the grouping on Domains list page, isn't defaultBounceDomain displayed in the DomainStatus column, what was the reason to separate this ?

shrirupa (commented):
these changes look good to me! there is one thing that I observed though(this bug might have existed before 🤷  ) 
Steps to reproduce - 
- Apply some filter on the Sending Domain List page, eg select unverified
- Navigate to details page
- Use browser back button to navigate back
- the filters in url and in Domains status filters are correct but the rows aren't actually filtered.

Let me know, I should cut a separate ticket for this.

src/pages/domains/components/SendingDomainsTab.js
```diff
@@ -224,28 +211,62 @@ export default function SendingDomainsTab({ renderBounceOnly = false }) {
...
+    } else {
+      const flattenedFilters = filterStateToParams(filtersState);
+      const domainStatusValues = {
+        blocked: flattenedFilters['blocked'],
+        readyForDKIM: flattenedFilters['readyForDKIM'],
+        readyForSending: flattenedFilters['readyForSending'],
+        unverified: flattenedFilters['unverified'],
+        validSPF: flattenedFilters['validSPF'],
+      };
+      if (!renderBounceOnly) {
```
shrirupa, line 240: this condition is not necessary since in this else block !renderBounceOnly will always be true.

src/pages/domains/components/SendingDomainsTab.js
```diff
@@ -224,28 +211,62 @@ export default function SendingDomainsTab({ renderBounceOnly = false }) {
...
+  const previousFirstLoad = usePrevious(firstLoad.current);
+  // Because we need tab changes from sending and bounce to behave differently
+  useEffect(() => {
+    if (firstLoad.current || (!firstLoad.current && previousFirstLoad)) {
+      return;
+    }
+
+    // NOTE: Handles tab changes, ignores page load
+
+    if (renderBounceOnly === true) {
```
shrirupa, line 224: may be if(Boolean(renderBounceOnly)), upto you though.

nstanard: sure

shrirupa (commented):
found a small thing - 
- When switching tab from bounce -> sending domain filters are not reset, however when switching from sending -> bounce filters clear correctly.

src/pages/domains/components/SendingDomainsTab.js
```diff
@@ -149,18 +149,6 @@ export default function SendingDomainsTab({ renderBounceOnly = false }) {
...
 
-  const previousRenderBounceOnly = useRef(renderBounceOnly);
-  if (renderBounceOnly !== previousRenderBounceOnly.current) {
-    // filtersInitialState -> selectAll was being set to false after first reset... huh...
-    filtersInitialState.checkboxes.map(checkbox => {
-      checkbox.isChecked = true;
-      return checkbox;
-    }); // force selectAll true again...
-    filtersStateDispatch({ type: 'RESET', state: filtersInitialState });
-    resetFilters();
```
shrirupa, line 160: related to the change here  - https://github.com/SparkPost/2web2ui/pull/2062#pullrequestreview-587012305

nstanard: Updated ticket and PR description to mention this was a request from UX.

shrirupa (approved):
looks good to me!

PR #2064, opened 2021-02-05
FE-1368 - Migrates tests cases from lengthy Cypress test suites to individual spec files, fixes test reporting to split tests by timing

nicklemmon:
### What Changed

- Fixes both Jest and Cypress test reporters so that Cypress tests can split between parallel runners based on their duration
- Test reports are now uploaded as artifacts to help with troubleshooting in the future
- Properly introduces the `.eslintplugin` directory in to the unit test step
- Upgrades `jest-junit` test reporter so that we can have access to the `"addFileAttribute"` configuration option.
- Uninstalls `cypress-multi-reporters` and `mocha-junit-reporter` in favor of `cypress-circleci-reporter` in order to have access to filenames
- Splits up tests within `/signals-analytics/analytics-report/tabs.spec.js` in to distinct specs for each type of table rendered within the tabs
- Splits up tests within `/domains/listPage.spec.js` in to distinct specs for each type of table rendered on that page
- Updates handling of cookies between tests to use `Cypress.Cookies.preserveOnce()` instead of the `preserve` option - allowing the login specs to properly clear cookies prior to running
- Removes video compression step for Cypress tests, saving even more time

### How To Test

- Verify that test reporter XML files are available as artifacts
- Verify that CircleCI is now picking up on test timing data - you can verify this by viewing the runner logs within CircleCI itself - let me know if you need help finding this
- Verify the pipeline runs significantly faster when compared with the default branch

### To Do

- [ ] Incorporate feedback


cypress/tests/integration/recipients/recipient-lists/listPage.spec.js
```diff
@@ -1,4 +1,3 @@
-import { IS_HIBANA_ENABLED } from 'cypress/constants';
```
nicklemmon, line 1: Just noticed this was still here - removed it

cypress/.eslintrc
```diff
@@ -1,6 +1,5 @@
 {
   "rules": {
-    "no-undef": "off",
```
nicklemmon, line 3: This shouldn't have been turned off - especially now that we are using more JS modules within tests.

package.json
```diff
@@ -25,7 +24,7 @@
     }
   },
   "lint-staged": {
-    "{cypress,src,.eslintplugin}/**/*.js": [
+    "{config,cypress,src,.eslintplugin}/**/*[.spec][.test].js": [
```
nicklemmon, line 27: Linting wasn't probably running within Cypress tests - though it was working within VSCode locally.

package.json
```diff
@@ -8,7 +8,6 @@
   },
   "engineStrict": true,
   "scripts": {
-    "lint": "npx eslint src/",
```
nicklemmon, line 11: We never use this script

package.json
```diff
@@ -55,7 +54,8 @@
     "url": "https://github.com/SparkPost/2web2ui/issues"
   },
   "jest-junit": {
-    "outputDirectory": "test_results/junit"
+    "outputDirectory": "test_results/junit",
+    "addFileAttribute": "true"
```
nicklemmon, line 58: CircleCI uses the file name associated with each test to split them up by their duration.

package.json
```diff
@@ -145,8 +145,9 @@
     "css-loader": "^2.1.0",
     "cssnano": "^4.0.3",
     "cypress": "^6.3.0",
+    "cypress-circleci-reporter": "^0.2.0",
```
nicklemmon, line 148: Using this in favor of `mocha-junit-reporter` as it properly attaches filenames to each test to allow for splitting by timing

cypress/.eslintrc
shrirupa: Good call!

cypress/support/commands.js
```diff
@@ -28,6 +28,9 @@ Cypress.Commands.add('login', (options = {}) => {
         cy.wait('@loginPost');
         cy.wait('@twoFactorGet');
       }
+
+      // Preserves cookies once per spec to avoid repeating login flow for each test
+      Cypress.Cookies.preserveOnce('website_auth', '__ssid', 'auth');
```
shrirupa, line 33: 🎉

cypress/tests/integration/configuration/domains/listPage.spec.js
```diff
@@ -1,74 +1,29 @@
-import { USERNAME } from 'cypress/constants';
-import { LINKS } from 'src/constants';
+import { stubSendingDomains, stubSubaccounts } from './helpers';
```
shrirupa, line 1: I think we recently updated these tests that's why there is a merge conflict

cypress/tests/integration/configuration/domains/listPage/bounceDomainsTable.spec.js
```diff
@@ -0,0 +1,273 @@
+import { LINKS } from 'src/constants';
```
shrirupa, line 1: do you think we should limit to one describe block per spec ?

.circleci/config.yml
```diff
@@ -44,7 +44,11 @@ commands:
     steps:
       - store_test_results:
           name: 'Upload test report'
-          path: /root/project/cypress/results
+          path: test_results
+      - store_artifacts:
+          name: 'Upload test results'
+          path: /root/project/test_results/cypress
```
shrirupa, line 50: what is this step doing ?

cypress/tests/integration/configuration/domains/listPage.spec.js
nicklemmon: Yeah I have rebased this branch many times at this point. Let me see if I can do it one more time.

cypress/tests/integration/configuration/domains/listPage/bounceDomainsTable.spec.js
nicklemmon: that might be a good approach when it comes to splitting things up!

.circleci/config.yml
nicklemmon: This is storing the test report XML files as artifacts so they can be downloaded if we want - right alongside the videos and screenshots that get produced.

nicklemmon:
@shrirupa I rebased this one (it was easier than I thought it would be). Mind taking another look?

shrirupa (approved):
looks good to me!

PR #2066, opened 2021-02-08
FE-1366 - Alert is displayed only once

shrirupa:
FE-1366 - Alert is displayed only once

### What Changed

- Alert on verification of domain through token will display the success/error alert only once.

### How To Test

Steps to reproduce the bug:

- Verify a domain by email.

- Click on the link in the email notification.

- After being redirected to the app, and if the domain is successfully verified, navigate away from the Domains list page and back

- ~Subsequent visits to the list page display the same Snackbar~ subsequent visits should display snackbar only once.

### To Do

- [ ] Address feedback

### DIFF
--- a/cypress/tests/integration/configuration/domains/listPage.spec.js
+++ b/cypress/tests/integration/configuration/domains/listPage.spec.js
@@ -15,1470 +15,1455 @@ describe('The domains list page', () => {
     });
   });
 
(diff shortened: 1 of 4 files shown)


nstanard (approved):
Tested locally, only getting one alert. LGTM

PR #2068, opened 2021-02-09
SA-2078: Validate URL query params in report builder

kchu1337:
### What Changed

- Cleaned up query params before being saved into context
- Added a validator for filters.

### How To Test

- Go to `signals/analytics`.
- Change one of the metrics to an invalid one. It should be removed and the user alerted.
- Change the timezone to an invalid value. It should default to local timezone and the user alerted.
- Change the date to an invalid value. Might also need to change range to `custom` It should default to 7 days ad alert the user.
- Change the filter to an invalid value. It should remove an invalid filter at the top level and alert the user. This means that if you have 3 AND filters but only the first one is invalid, the ui will keep the other 2.


src/components/typeahead/TimezoneTypeahead.js
```diff
@@ -33,6 +33,14 @@ export const options = moment.tz
 
 options.unshift(UTC_OPTION);
 
+export const timezoneMap = options.reduce(
```
kchu1337, line 36: Should I move this and `options` into a constant file?

src/config/metrics.js
```diff
@@ -531,10 +531,13 @@ export const REJECTIONS_BY_DOMAIN_METRIC_KEYS = ['count_clicked', 'count_raw_cli
 
 export const BOUNCE_BY_DOMAIN_METRIC_KEYS = ['count_bounce'];
 
-export const map = list.reduce((accumulator = {}, metric) => ({
-  ...accumulator,
-  [metric.key]: metric,
-}));
+export const map = list.reduce(
```
kchu1337, line 534: For some reason, the map had keys like `type` and `description`. Decided to remove them.

src/pages/reportBuilder/helpers/index.js
```diff
@@ -373,3 +373,48 @@ export function getHasDuplicateFilters(filters) {
...
+}, new Map());
+/**
+ * Returns a new array of only valid filters. It only removes at the top level. If any part of
+ * the filter is invalid, it will remove the entire top level filter.
+ *
+ * @param {array} filters - iterable array of filter objects derived from `hydrateFilters`
+ *
+ * @returns {array} - Array of only valid filters.
+ */
+export function getValidFilters(filters) {
```
kchu1337, line 393: Please leave feedback on this. I feel like it works, but let me know if anything is confusing or doesn't work.

src/components/typeahead/TimezoneTypeahead.js
nstanard: I would vote for that. I would also move the metrics.js list to it's own file and import/re-export.

image

600 lines is a lot for a single file.

src/components/typeahead/TimezoneTypeahead.js
```diff
@@ -33,6 +33,14 @@ export const options = moment.tz
...
+  (accumulator = {}, timezone) => ({
```
nstanard, line 37: I don't think you need to assigned `accumulator = {}` with the reduce default of `{}`

src/pages/reportBuilder/context/ReportBuilderContext.js
```diff
@@ -1,18 +1,25 @@
 import React, { useCallback, useContext, useMemo, useReducer, createContext } from 'react';
 import moment from 'moment';
```
nstanard, line 2: Do we have a ticket to remove moment from this area? Just curious considering the conversations that have been happening around moment being deprecated.

src/config/metrics.js
```diff
@@ -649,10 +649,13 @@ export const REJECTIONS_BY_DOMAIN_METRIC_KEYS = ['count_clicked', 'count_raw_cli
...
+  (accumulator = {}, metric) => ({
```
nstanard, line 653: `accumulator = {}` - same thing. Unless I'm missing something, reduce defaults it already.

src/pages/reportBuilder/context/ReportBuilderContext.js
```diff
@@ -40,27 +47,67 @@ const reducer = (state, action) => {
...
+          message: `Invalid Timezone`,
+        });
+        update.timezone = getLocalTimezone();
+      }
+
       if (!update.metrics) {
         update.metrics = config.reportBuilder.defaultMetrics;
+      } else {
+        if (!update.metrics.every(metric => metricsMap[metric])) {
+          dispatchAlert({
```
nstanard, line 69: Do we need to be concerned about the dispatch being called from within an every call? Does this have the potential to flood the screen with alert?

src/pages/reportBuilder/helpers/index.js
```diff
@@ -391,14 +392,61 @@ export function getHasDuplicateFilters(filters) {
...
 }
 
+/**
+ * Hashmap of valid filter types
+ */
+const FILTERS_KEY_MAP = Object.values(REPORT_BUILDER_FILTER_KEY_MAP || {}).reduce((map, key) => {
+  //The `|| {}` is needed to fix some breaking unit tests.
+  map.set(key, true);
+  return map;
+}, new Map());
```
nstanard, line 402: Just a note - if you use `new Map()` you'll get a size property by default. If you just use `{}` you won't. We might use an object literal instead to prevent extra properties from being created.

nstanard (commented):
need to pause the review, submitting these comments for now and picking back up in a few

src/pages/reportBuilder/context/ReportBuilderContext.js
```diff
@@ -1,18 +1,25 @@
...
+import { connect, useDispatch } from 'react-redux';
+import { getLocalTimezone, getRelativeDates, formatToTimezone } from 'src/helpers/date';
+import {
+  getMetricsFromKeys,
+  getRollupPrecision as getPrecision,
+  getValidDateRange,
+} from 'src/helpers/metrics';
 import { stringifyTypeaheadfilter } from 'src/helpers/string';
 import config from 'src/config';
+import { map as metricsMap } from 'src/config/metrics';
```
nstanard, line 12: maybe format this alias as `METRICS_MAP` instead to keep in line with what you did for `import { list as METRICS_LIST } from 'src/config/metrics';`

src/config/metrics.js
nstanard: These were properties coming from the first item in the list:

image

I'm pretty certain you were right to remove them from the map call.

src/pages/reportBuilder/helpers/index.js
```diff
@@ -391,14 +392,61 @@ export function getHasDuplicateFilters(filters) {
...
+ * Returns a new array of only valid filters. It only removes at the top level. If any part of
+ * the filter is invalid, it will remove the entire top level filter.
+ *
+ * @param {array} filters - iterable array of filter objects derived from `hydrateFilters`
+ *
+ * @returns {array} - Array of only valid filters.
+ */
+export function getValidFilters(filters) {
+  //remaps top level comparators to true/false
+  //EX: [{AND:{...}},{OR:{...}}] => [true ,false] where true = valid filter and false = invalid filter
```
nstanard, line 413: Do we have prettier/linter hooked up to fix comments on save to have the space at the start? If not I'd like to add that on a future PR.

src/pages/reportBuilder/helpers/index.js
```diff
@@ -391,14 +392,61 @@ export function getHasDuplicateFilters(filters) {
...
+
+    const filterType = grouping[groupingType];
+    //Checks that every filter is valid
+    return Object.keys(filterType).every(filter => {
+      if (!FILTERS_KEY_MAP.has(filter)) {
+        return false;
+      }
+      //Checks that every comparator is allowed
+      return Object.keys(filterType[filter]).every(comparator => {
+        return ['eq', 'noteq', 'like', 'notlike'].includes(comparator.toLowerCase());
```
nstanard, line 430: Should we store the valid value (`['eq', 'noteq', 'like', 'notlike']`) in the config file and import it here too?

src/pages/reportBuilder/context/ReportBuilderContext.js
```diff
@@ -40,27 +47,67 @@ const reducer = (state, action) => {
...
+        if (!update.metrics.every(metric => metricsMap[metric])) {
+          dispatchAlert({
+            type: 'error',
+            message: `Invalid Metric`,
+          });
+          update.metrics = update.metrics.filter(metric => metricsMap[metric]);
+        }
       }
 
       if (payload.filters) {
```
nstanard, line 77: It might be good to move this code block up further. It might be good to keep this close to the start of the reducer block.

src/pages/reportBuilder/context/ReportBuilderContext.js
```diff
@@ -40,27 +47,67 @@ const reducer = (state, action) => {
...
+
+        if (validatedFilters.length !== rawFilters.length) {
+          dispatchAlert({
+            type: 'error',
+            message: `Invalid Filter`,
+          });
+        }
+        update.filters = validatedFilters;
       }
+      const updatePrecision = update.precision || 'hour'; // Default to hour since it's the recommended rollup precision for 7 days
```
nstanard, line 89: I know it's just a single string, but it might be good to use a constant instead of a string literal ('hour')

nstanard:
Let me know your thoughts on the comments. I don't want these comments to hold anything up. When you have time we can jump on a call and talk through it if needed. I didn't find any side effects from these changes and the page doesn't error when I modify the URL. We should be able to move this along when you are ready.

src/pages/reportBuilder/context/ReportBuilderContext.js
kchu1337: No, the loop gets finished in the if statement and just returns a true or false.

src/pages/reportBuilder/helpers/index.js
kchu1337: It's actually the other way around. Objects have default default keys because of prototype and stuff (Although I've never found it to be an issue). 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#objects_vs._maps

kchu1337: Oh I see what you kinda mean. It might have a size prop but not a size key though. That's why I only use the Map in this file. Might get confusing if exporting and you don't know whether to expect a Map or a regular object.

src/pages/reportBuilder/context/ReportBuilderContext.js
kchu1337: It's probably fine for now since it's being used everywhere. It could be a good later global refactor to use a variable instead.

src/pages/reportBuilder/context/ReportBuilderContext.js
kchu1337: Why does it matter? No query param should be more important than another #woke.

src/components/typeahead/TimezoneTypeahead.js
kchu1337: I tried and lot's of stuff broke. I added a comment to warn others not to try it lol.

src/pages/reportBuilder/helpers/index.js
```diff
@@ -391,14 +393,70 @@ export function getHasDuplicateFilters(filters) {
...
+ *
+ * @param {array} filters - iterable array of filter objects derived from `hydrateFilters`
+ *
+ * @returns {array} - Array of only valid filters.
+ */
+export function getValidFilters(filters) {
+  //remaps top level comparators to true/false
+  //EX: [{AND:{...}},{OR:{...}}] => [true ,false] where true = valid filter and false = invalid filter
+  const validFiltersArray = filters.map(grouping => {
+    const groupingType = Object.keys(grouping)[0];
```
nstanard, line 425: I'd feel a little better about this if we made sure `groupingType` is truthy before using it. (since we are grabbing the 0 index of the result of the Object.keys())

src/pages/reportBuilder/context/ReportBuilderContext.js
nstanard: I just figured it might be good to keep the validation close to the start of the function, but not a big deal. Your call!

nstanard (approved):
I'd still like to have those safety conditions in place as a precaution. But I don't see any issues and everything tested out as expected.

nstanard (approved):
Latest changes still lgtm.

PR #2072, opened 2021-02-11
AC-1909 - Dedicate Ip section should use count from billing/subscription instead of /sending-ips

shrirupa:
AC-1909 - Dedicate Ip section should use count from billing/subscription instead of /sending-ips

### What Changed

- On AddIps form and DedicatedIps replaced sendingip.length with quantity of dedicated ips from GET /billing/subscription
- Also removed getSendingIps action from addDedicatedIps in src/actions/billing
- updated cypress tests

### How To Test

- Check if AddIps form and DedicatedIps section looks correct and now use quantity from GET /billing/subscription
- Check if there are remnants of getSendingIps on billing page

### To Do

- [ ] Address feedback


PatrickSison (approved):
👍

PR #2076, opened 2021-02-17
AC-1904 - Upgrades Matchbox to ~4.8.0

nicklemmon:
[AC-1904](https://sparkpost.atlassian.net/browse/AC-1904)

### What Changed

- Upgrades Matchbox to 4.8.0
- Replaces use of `component` prop with `as` where appropriate
- Updates `ExternalLink` to simplify icon positioning. For Panel actions, the positioning wasn't rendering correctly. Compare the "Learn More" icon position vs. the "Enable SSO" button (the latter is correct):
Screen Shot 2021-02-17 at 12 58 36 PM

### How To Test

- Verify the app works as expected
- Verify positioning of elements on `/signals/analytics` and look for padding problems introduced by the Matchbox update
- Verify components that were using the `component` prop continue to work with the `as` prop in its place
- Verify external links and ensure icon positioning is handled appropriately
- Verify instances of the `component` prop were properly replaced with `as` for `Button` and `UnstyledLink` components

### To Do

- [X] Replace use of the deprecated `component` prop with `as`
- [ ] Incorporate feedback


src/components/links/ExternalLink.js
```diff
@@ -2,35 +2,43 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
 import { OpenInNew } from '@sparkpost/matchbox-icons';
-import { Button, UnstyledLink } from 'src/components/matchbox';
+import { Box, Button, UnstyledLink } from 'src/components/matchbox';
 
-const StyledIcon = styled('div')`
-  margin: ${props => props.iconMargin};
+// The element is immediately overwritten by the value of the `as` prop
+const StyledLink = styled('div')`
```
shrirupa, line 8: should this be UnStyledLink instead of div ?

shrirupa: ` The element is immediately overwritten by the value of the `as` prop` is this happening because of styled components or matchbox or something inherent in html ?

shrirupa:
Noticed that Expandable titles on Report Builder have different font weights, is it intentional or a bug ?

nicklemmon (commented):
> Noticed that Expandable titles on Report Builder have different font weights, is it intentional or a bug ?

It looks like Matchbox updated the design on their end

src/components/links/ExternalLink.js
nicklemmon: You know you're right - what I'm doing is more confusing than just doing that.

nicklemmon:
@shrirupa Ready for another pass when you have time.

shrirupa (approved):
looks good to me!

PR #2077, opened 2021-02-17
AC-1850 - ChoosePlanPage flashes before redirecting to dashboard when adding a paid plan

shrirupa:
AC-1850 - ChoosePlanPage flashes before redirecting to dashboard and adding a paid plan

### What Changed
_Note-_ 
_Cause of this bug seemed to be coming from the actions emitted by the redux form, the form would emit a redux/@Destroy action after a successful submit which would trigger rerender of the form before the redirect to dashboard._

- moved the redirect and alert to useEffect instead of then.

### How To Test

- Sign up for a new account.
- On Choose plan page, choose a paid plan
- check if the flasing of the page is gone and you get redirected to dashboard smoothly

### To Do

- [ ] Address feedback

### DIFF
--- a/src/actions/billingCreate.js
+++ b/src/actions/billingCreate.js
@@ -13,7 +13,6 @@ import {
 export default function billingCreate(values) {
   return (dispatch, getState) => {
     const { corsData, billingData } = formatDataForCors(values);
-
     // action creator wrappers for chaining as callbacks
     const corsCreateBilling = ({ meta }) =>
       cors({ meta, context: 'create-account', data: corsData });
@@ -42,6 +41,10 @@ export default function billingCreate(values) {
         consumePromoCode({ meta, promoCode: values.promoCode, billingId: billingData.billingId });
       actions.push(consumePromo);
     }
-    return dispatch(chainActions(...actions)());
+    dispatch({ type: 'BILLING_CREATE_PENDING' });
+
+    return dispatch(chainActions(...actions)())
+      .then(() => dispatch({ type: 'BILLING_CREATE_SUCCESS' }))
+      .catch(() => dispatch({ type: 'BILLING_CREATE__ERROR' }));
   };
 }
--- a/src/actions/billingCreate.js
+++ b/src/actions/billingCreate.js
@@ -45,6 +45,6 @@ export default function billingCreate(values) {
 
     return dispatch(chainActions(...actions)())
       .then(() => dispatch({ type: 'BILLING_CREATE_SUCCESS' }))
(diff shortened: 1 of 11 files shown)


src/selectors/onboarding.js
```diff
@@ -17,8 +17,10 @@ export const choosePlanMSTP = formName =>
         state.account.loading ||
           state.billing.countriesLoading ||
           state.billing.bundlesLoading ||
-          state.billing.bundlePlansLoading,
+          state.billing.bundlePlansLoading ||
+          state.billingCreate.loading,
```
kchu1337, line 21: It seems like this is why it's going to the loading spinner.

src/reducers/billingCreate.js
```diff
@@ -0,0 +1,35 @@
...
+  loading: null,
+  success: null,
+  error: false,
+};
+
+export default (state = initialState, { type }) => {
+  switch (type) {
+    case 'BILLING_CREATE_SUCCESS':
+      return {
+        ...state,
```
kchu1337, line 11: I don't think you need to destructure state since you are already setting the 3 keys.

src/actions/billingCreate.js
```diff
@@ -42,6 +41,10 @@ export default function billingCreate(values) {
         consumePromoCode({ meta, promoCode: values.promoCode, billingId: billingData.billingId });
       actions.push(consumePromo);
     }
-    return dispatch(chainActions(...actions)());
+    dispatch({ type: 'BILLING_CREATE_PENDING' });
+
+    return dispatch(chainActions(...actions)())
+      .then(() => dispatch({ type: 'BILLING_CREATE_SUCCESS' }))
+      .catch(() => dispatch({ type: 'BILLING_CREATE__ERROR' }));
```
kchu1337, line 48: You have an extra "_"

kchu1337 (changes requested):

src/actions/billingCreate.js
shrirupa: good catch!

src/selectors/onboarding.js
```diff
@@ -19,6 +19,8 @@ export const choosePlanMSTP = formName =>
           state.billing.bundlesLoading ||
           state.billing.bundlePlansLoading,
       ),
+    billingCreateLoading: state => state.billingCreate.loading,
+    billingCreateSuccess: state => state.billingCreate.success,
```
kchu1337, line 23: Nothing really happens when error is true. Do you want to add an alert?

kchu1337 (approved):

PR #2078, opened 2021-02-18
SA-2139: Add proper truncation for the compare by aggregate rows

PatrickSison:
### What Changed

- Added text truncation component w/ tooltip
- Added text truncation to compare by aggregate row for user provided values

### How To Test

- Add compare by filters with long string names (I've found some decent length ones in templates)
- Decrease view port to small view
- Verify that text is properly truncated
- Verify that tooltip exists for truncated text
- Verify that tooltip doesn't exist for non truncated text

### To Do

- [ ] Add test ? Not sure how to easily test this


src/pages/reportBuilder/components/GroupByTable/tests/components.test.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
```
kchu1337, line 1: What is this file for?

src/components/reportBuilder/CompareByAggregatedRow.js
```diff
@@ -43,7 +44,9 @@ export default function CompareByAggregatedRow({ comparison, reportOptions, hasD
           <LabelValue dark>
             <LabelValue.Label>{comparison.type}</LabelValue.Label>
 
-            <LabelValue.Value>{comparison.value}</LabelValue.Value>
+            <LabelValue.Value>
+              <TextTooltip>{comparison.value}</TextTooltip>
```
kchu1337, line 48: We have a `shrinkToFit` helper in src/helpers/string.js if that helps.

src/components/TextTooltip/TextTooltip.js
```diff
@@ -0,0 +1,36 @@
...
+  text-overflow: ellipsis;
+  white-space: nowrap;
+`;
+
+const NoWrapText = styled.div`
+  white-space: nowrap;
+  width: fit-content;
+`;
+
+export default function TextTooltip({ children, id = 'text-tooltip-popover' }) {
```
kchu1337, line 16: This doesn't seem to work if you wrap the string in a `<div>`. Maybe validate that children is a string using Proptype?

kchu1337 (dismissed):

src/components/reportBuilder/CompareByAggregatedRow.js
PatrickSison: Was trying not to do any JS truncation and do it through straight CSS

PatrickSison: Especially if we got variable width. Seems like it would be useful for other cases

src/components/TextTooltip/TextTooltip.js
PatrickSison: Yeah, I suppose for this iteration, we should lock it down and if we find a use case for more complex children, we can implement a solution later on

src/pages/reportBuilder/components/GroupByTable/tests/components.test.js
PatrickSison: I've no idea how this ended up here lol. I think this was committed to the wrong branch

cypress/tests/integration/signals-analytics/analytics-report/compareBy.spec.js
```diff
@@ -178,8 +178,21 @@ describe('Analytics Report - Compare By', () => {
...
+          .eq(1)
+          .should('be.visible'); //Tooltip
+
+        cy.findAllByText('Fake Subaccount 3 (ID 103)')
+          .eq(0)
+          .should('be.visible')
+          .trigger('mouseover');
+        cy.findAllByText('Fake Subaccount 3 (ID 103)')
+          .eq(1)
+          .should('be.visible'); //Tooltip
```
bkemper, line 195: Shouldn't we add a test for this?

bkemper: We can use `cy.viewport` — https://docs.cypress.io/api/commands/viewport.html#Syntax

src/components/TextTooltip/TextTooltip.js
```diff
@@ -0,0 +1,44 @@
...
+      id={id}
+      bg="white"
+      color="gray.800"
+      width="auto"
+      as="div"
+      content={<NoWrapText>{children}</NoWrapText>}
+      disabled={!showTooltip}
+    >
+      <TruncatedText ref={ref} showTooltip={showTooltip}>
+        <span ref={textRef}>{children}</span>
```
bkemper, line 34: Like `src/components/tooltip`, I think we need the `aria-controls={id}` on the `span`.

bkemper (commented):
It works for me!  I added two comments that should be addressed for both adding a quick test and adding the a11y prop.

src/components/TextTooltip/TextTooltip.js
```diff
@@ -0,0 +1,47 @@
+import React, { useRef } from 'react';
```
nicklemmon, line 1: New global component means some unit tests are needed

nicklemmon (changes requested):
@PatrickSison Just missing some unit tests for the new component you wrote - otherwise looks good.

src/components/TextTooltip/TextTooltip.js
```diff
@@ -0,0 +1,47 @@
...
+  const [ref, entry] = useResizeObserver();
+  const textRef = useRef(null);
+  const showTooltip = textRef?.current?.offsetWidth > entry.contentRect.width;
+
+  return (
+    <Tooltip
+      id={id}
+      portalId="tooltip-portal"
+      bg="white"
+      color="gray.800"
```
nicklemmon, line 28: I would think these should be controlled via props?

nicklemmon (approved):
@PatrickSison Looks good, let's do it. 👍

PR #2079, opened 2021-02-18
SA-2042: Fixed issue with dates not formatting to timezone 

PatrickSison:
### What Changed

-Dates account for timezone in  report builder

### How To Test

- Go to report report builder
- Set a time that is on the edge of a new day (previous or forward)
- Change timezone to move day to next day (or previous day)
- Verify that date is correct in aggregate data (relative to timezone)

### To Do

- [ ] Add testing around this

### DIFF
--- a/src/helpers/date.js
+++ b/src/helpers/date.js
@@ -174,7 +174,7 @@ export function getFormattedDateRangeForAggregateData(from, to, timezone = getLo
   if (!from || !to) {
     return '';
   }
-  return `${formatToTimezone(new Date(from), "MMM do',' yyyy", timezone)} - ${formatToTimezone(
+  return `${formatToTimezone(new Date(from), 'MMM do', timezone)} - ${formatToTimezone(
     new Date(to),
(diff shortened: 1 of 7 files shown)


src/helpers/date.js
```diff
@@ -170,8 +170,15 @@ export function formatDateTimeWithoutYear(datetime) {
   return formatDateTime(datetime, `${config.dateFormatWithoutYear}, ${config.timeFormat}`);
 }
 
-export function getFormattedDateRangeForAggregateData(from, to) {
-  return `${formatDate(from, 'MMM Do')} - ${formatDate(to, 'MMM Do, YYYY')}`;
+export function getFormattedDateRangeForAggregateData(from, to, timezone = getLocalTimezone()) {
+  if (!from || !to) {
```
PatrickSison, line 174: It's a bit ugly. Old date-fns would accept any input and attempt to spit out value, but it's much more strict now. There's cases where from/to are not ready yet

nicklemmon (approved):

PR #2081, opened 2021-02-23
TR-3040 - Resolves (UI) Templates validator in AB testing possibly too strict

nicklemmon:
[TR-3040](https://sparkpost.atlassian.net/browse/TR-3040)

### What Changed

- Updates validation function to use `Array.prototype.some()` instead of `.includes()` to check for matching `id` values for the selected template vs. the list of templates

### How To Test

- Go to `/ab-testing/create` and try to create an A/B testing
- Verify that you cannot create an A/B test if it is not available to that particular subaccount
- Verify that you cannot create an A/B test if the template is not available to the primary account when no subaccount is selected
- Verify that you _can_ create a template when a matching ID is present in the filtered list of templates passed to the validation helper

### To Do

- [ ] Incorporate feedback


bkemper (approved):
I was trying to reproduce this bug again. I disabled the request just in case the /templates request was slow and the validation happen before it was complete, but the field is disabled until the request is complete. I created different combinations of subaccounts and templates.  LGTM!

nicklemmon:
> I was trying to reproduce this bug again. I disabled the request just in case the /templates request was slow and the validation happen before it was complete, but the field is disabled until the request is complete. I created different combinations of subaccounts and templates. LGTM!

Looking at the helper and the supporting tests I assume this was introduced when the structure of the data somehow changed unexpectedly. It really was expecting a flat array previously.

PR #2082, opened 2021-02-23
SA-2161: Allow users to search for filter and compare values with a single character in Report Builder

bkemper:
### What Changed

- Added a "Add Filter" option to the "Filter" and "Compare" value typeaheads to allow users to enter a value not found in the typeahead options

Screen Shot 2021-02-23 at 12 16 16 PM

- The "Add Filter" option will not be displayed if the input value matches a typeahead option

Screen Shot 2021-02-23 at 12 16 29 PM

- The "Add Filter" option will not be displayed if the input value matches a selected item

Screen Shot 2021-02-23 at 12 16 47 PM
 

### How To Test

- Good question

### To Do

- [ ] specs

### Follow-Up

The following are a list of potential follow-up tickets.  Let's discuss before cutting them.

 - Typeahead Consolidation - we have several implementations which caused duplication
 - Compare onBlur behavior - the user experience is a little clunky to me, if you select something, remove, blur, then it comes back?  Is this what we want?
 - Additional messaging - we discussed in refinement meeting adding additional text letting the user know about the 3 character limit, but the filter typeahead is not flexible enough to handle this addition.  This also includes loading spinner.  It is a subtle indicator, but necessary and unable to be displayed with current Typeahead.
 - Subaccount Validation - if a non-numeric value for subaccounts is entered, the api requests will fail and display three snackbars
 - Duplicate Compare By Values - [use to remove selected values from typeahead options](https://github.com/SparkPost/2web2ui/commit/d512664599aba4f31e4b1fbe36ed3a950b1a83da#diff-8a5e18a455d72ddc4962dc85c…c075aae1505d5f81989cb660d1L202), should we add this back in?
-


src/pages/reportBuilder/helpers/index.js
```diff
@@ -273,15 +273,15 @@ export function hydrateFilters(groupings, { subaccounts } = {}) {
...
-                    subaccounts.find(subaccount => subaccount.id === Number.parseInt(value)) || {};
-                  const { name, id } = subaccount;
+                  const id = typeof rawValue === 'object' ? rawValue.id : rawValue;
+                  const subaccount = subaccounts.find(
+                    subaccount => subaccount.id === Number.parseInt(id),
+                  );
 
                   return {
-                    value: name ? `${name} (ID ${id})` : value,
+                    value: subaccount ? `${subaccount.name} (ID ${id})` : id,
```
bkemper, line 284: Small, but big change.  Before the id would be set to `undefined` and would break.  Now, we proceed custom subaccount values.

src/pages/reportBuilder/components/ReportOptions.js
```diff
@@ -132,7 +141,9 @@ export function ReportOptions(props) {
       </Panel.Section>
 
       <Drawer {...getDrawerProps()} portalId="drawer-portal">
-        <Drawer.Header showCloseButton />
+        <Drawer.Header showCloseButton>
+          <ScreenReaderOnly>Report Options</ScreenReaderOnly>
```
bkemper, line 145: Extra credit

src/pages/reportBuilder/components/Typeahead.js
```diff
@@ -57,10 +49,10 @@ function Typeahead(props) {
       id={id}
       onChange={onFilterChange}
       onInputChange={setInputValue}
-      itemToString={item => (item ? getItemToStr({ filterType, item }) : '')}
       value={value}
       results={filteredResults}
       loading={status === 'loading'}
+      customItemHelpText="Add Filter"
```
nicklemmon, line 55: Are we still sure we want to do this? I believe the option would be read to screen reader users as:

> "Add 'Fa' Add Filter"

It doesn't really make sense?

src/pages/reportBuilder/helpers/index.js
nicklemmon: Looks like there are some unit tests that support this function - should we add a test for the scenario you're describing?

src/pages/reportBuilder/components/Typeahead.js
```diff
@@ -57,10 +49,10 @@ function Typeahead(props) {
       id={id}
       onChange={onFilterChange}
       onInputChange={setInputValue}
-      itemToString={item => (item ? getItemToStr({ filterType, item }) : '')}
       value={value}
       results={filteredResults}
       loading={status === 'loading'}
```
nicklemmon, line 54: The loading state isn't rendering for me anymore? I don't see any loading indicator when the API is being hit.

It is rendering in the `CompareByForm` component but not the `FiltersForm`.

src/pages/reportBuilder/components/CompareByForm/Typeahead.js
```diff
@@ -121,41 +122,58 @@ function TypeSelect({
...
+        helpText: customOptionHelpText,
+        isVisible: !matches.some(({ id, value }) =>
+          id ? String(id) == customItem.id : value === customItem.value,
+        ),
+        item: customItem,
+      },
+      ...matches.map(item => ({
+        content: itemToString(item),
+        isVisible: true,
+        item,
```
nicklemmon, line 143: Should this be spread? The final structure is a little funny to me:
```json
{
  "content": "80-campaign",
  "isVisible": true,
  "item": {
    "type": "campaigns",
    "value": "80-campaign"
  }
}
```

src/pages/reportBuilder/components/CompareByForm/Typeahead.js
```diff
@@ -121,41 +122,58 @@ function TypeSelect({
...
+
     return (
       <div>
         <TypeaheadWrapper>
           <Box maxWidth={maxWidth} position="relative">
             {/* hack, can't apply getMenuProps to ActionList because ref prop is not supported */}
             <div {...getMenuProps()}>
               <PopoverActionList open={isMenuOpen} maxHeight={maxHeight}>
-                {loading ? (
+                {items.length ? (
```
nicklemmon, line 163: For type safety, might be best to wrap this in a `Boolean()`

cypress/tests/integration/signals-analytics/analytics-report/filtersForm.spec.js
```diff
@@ -631,27 +684,52 @@ function getFilterByIndex(index) {
...
-      cy.findByLabelText(COMPARE_BY_LABEL).select('contains');
-      cy.findByLabelText('Campaign (ID)').type('hello{enter}');
+      cy.findByLabelText(TYPE_LABEL).select('Subaccount');
+      cy.findByLabelText('Subaccount').type(text);
+
+      if (text.length >= 3) {
+        cy.wait('@getSubaccounts');
+      }
+
+      cy.findByRole('option', { name: new RegExp(text) })
```
nicklemmon, line 697: Looks like using `new RegExp()` here is introducing some test failures - the `name` option already accepts both strings and regular expressions so you should be able to just whichever as you need them rather than parsing all passed-in `text` as a regular expression:
https://testing-library.com/docs/queries/byrole/

nicklemmon (changes requested):
@bkemper First pass complete - I left a few comments for you.

You have a few test failures I _think_ are caused by the test vs. the actual implementation. I left a comment that explains why I think they are happening. The IP pools test failure is a known flaky test that I brought up to Shrirupa though it deserves a ticket at this point I believe. I will add it to my list to cut one.

src/pages/reportBuilder/components/Typeahead.js
bkemper: Discussed with UX.  Decision was to drop "Add" from the beginning.
— https://msys.slack.com/archives/GLLAW6UAX/p1614635731046900

src/pages/reportBuilder/components/Typeahead.js
bkemper: This is the problem with having two variations of typeaheads.  I would need to rebuild the filter typeahead to get the loading spinner to work.  This work would probably be best done in a follow-up ticket.  The reason is the filter typeahead uses ComboBoxMenu to render the list of options and the `emptyMessage` prop to render the spinner.  With always having a custom option available, the emptyMessage/loader is never displayed.  I should have called this out in the description.  I thought about hacking around it, but I think that will make things worse.  We need to circle back with UX, clean up this interaction and refactor these typeaheads.

src/pages/reportBuilder/components/CompareByForm/Typeahead.js
bkemper: I'm open for suggestions.  I guess it is the duplication with "content" and "item.value" that is funny?  In the initial construction of this array, I was trying to prevent attribute logic by rendering a common structure that could then be easily filtered and enriched with `getItemProps`.

src/pages/reportBuilder/components/CompareByForm/Typeahead.js
bkemper: Is that safer or should I change this to `items.lenght > 0`?

cypress/tests/integration/signals-analytics/analytics-report/filtersForm.spec.js
bkemper: 😊

src/pages/reportBuilder/components/ReportOptions.js
bkemper: I fixed this and remembered that `Drawer.Header` renders an h2.

cypress/tests/integration/signals-analytics/analytics-report/filtersForm.spec.js
bkemper: Actually, this is catching a bug.

Screen Shot 2021-03-03 at 6 59 50 AM

bkemper: I added some additional logic to improve matching on real items.

nicklemmon (approved):
@bkemper I added a few comments in spots, but nothing that really holds this one back. 👍

cypress/tests/integration/signals-analytics/analytics-report/filtersForm.spec.js
```diff
@@ -530,9 +530,10 @@ describe('Analytics Report filters form', () => {
         .should('have.focus');
     });
 
-    it('does not render after the user clears invalid, duplicated filters', () => {
+    it('does not render after the user clear invalid, duplicated filters', () => {
```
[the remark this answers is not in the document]
bkemper: 😊

cypress/tests/integration/signals-analytics/analytics-report/filtersForm.spec.js
```diff
@@ -499,25 +554,26 @@ describe('Analytics Report filters form', () => {
...
+
       cy.withinDrawer(() => {
         cy.findByRole('alert').should('exist');
       });
 
-      // Fix the error by changing the "Compare By" on one of the offending filters
       getGroupingByIndex(0).within(() => {
         getFilterByIndex(2).within(() => {
-          cy.findByLabelText(COMPARE_BY_LABEL).select('is equal to');
+          cy.findByRole('button', { name: 'Remove Filter' }).click();
```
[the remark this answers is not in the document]
bkemper: My `causeDuplicationError` change to use subaccount forced me.  This was changing the compare select from contains to equals.  Subaccount only supports equals, so I decided to just remove it instead of changing the compare to trigger the reevaluation.

PR #2083, opened 2021-02-24
SA-2194 - Resolves (UI) Compare by aggregated UI should be responsive

nicklemmon:
[SA-2194](https://sparkpost.atlassian.net/browse/SA-2139)

### What Changed

- Updated styles to collapse on certain breakpoints

### How To Test

- Resize the browser window on both `/signals/analytics` and `/dashboard` with comparisons active on a report (Click "Add Comparisons" if none are present or swap to a saved report with active comparisons)
- Verify the content within the aggregates UI is responsive:

Screen Shot 2021-02-24 at 9 12 44 AM

### To Do

- [ ] Incorporate feedback

### DIFF
--- a/src/components/reportBuilder/CompareByAggregatedRow.js
+++ b/src/components/reportBuilder/CompareByAggregatedRow.js
@@ -13,7 +13,15 @@ const MetricsGrid = styled.div`
   display: inline-grid;
   width: 100%;
   grid-gap: ${props => props.theme.space['200']};
-  grid-template-columns: repeat(4, 1fr);
+  grid-template-columns: 1fr;
+
+  @media (min-width: ${props => props.theme.breakpoints[0]}) {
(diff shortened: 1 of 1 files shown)


src/components/reportBuilder/CompareByAggregatedRow.js
```diff
@@ -13,7 +13,15 @@ const MetricsGrid = styled.div`
   display: inline-grid;
   width: 100%;
   grid-gap: ${props => props.theme.space['200']};
-  grid-template-columns: repeat(4, 1fr);
+  grid-template-columns: 1fr;
```
PatrickSison, line 16: Didn't realize this change came in, but were these supposed to be gridded in the first place? I remember a while ago (but like at the beginning), this was inline.

PatrickSison (approved):

PR #2084, opened 2021-02-24
SA-2164 - Resolves (UI) Cannot read property 'undefined' of undefined

nicklemmon:
[SA-2164](https://sparkpost.atlassian.net/browse/SA-2164)

### What Changed

Added conditional check for `groupBy` within the `useCompareByGroupByTable` hook. Removed unnecessary optional chaining as a result.

### How To Test

1. Go to `/signals/analytics`
2. Click on "Add Comparison"
3. In the two fields in the drawer, search for a value via the typeahead and select the _exact same result in both fields_
4. Click "Compare"
5. Reopen the drawer, add one more compare by item
6. Click "Compare"
5. The app should not crash!

If you want to see the crash in action, swap to another branch and repeat the above steps

### To Do

- [ ] Incorporate feedback
- [ ] Chat with Brian about re-introducing filtering by currently active comparisons/filters within `Typeahead` components in the feature


bkemper (approved):

bkemper:
@nicklemmon I'll be sure to add test coverage for this in #2082

PR #2087, opened 2021-02-26
AC-1905 - Track /join to /onboarding/plan in segment

nstanard:
### What Changed
- Removed the check for access control in the segment page call

### How To Test
- Create a new account and make sure segment is getting the /onboarding/join data after account creation

### To Do
- [ ] Double check things and test that removing the access control won't be a problem


shrirupa (approved):
looks good to me!

PR #2088, opened 2021-03-01
SA-1993: Industry Benchmark Modal

kchu1337:
### What Changed

- Added industry benchmark modal
- Added data to reportBuilderContext
- Added constants for the modal

### How To Test

- Go to /signals/analytics
- Enable deliverability product
- Add inbox folder rate as a metric
- There should be a industry benchmark button on the top-right of the charts
- Click it and a modal should appear.

### To Do

- [x] Write tests
- [ ] Integrate with SA-1997


src/pages/reportBuilder/components/IndustryBenchmarkModal.js
```diff
@@ -0,0 +1,115 @@
...
+                />
+              </Box>
+            </Inline>
+          </Stack>
+        </Modal.Content>
+        <Modal.Footer>
+          <Button
+            variant="primary"
+            onClick={() => {
+              addIndustryBenchmark(state);
```
PatrickSison, line 102: If we change this to a form, we can change the button type to submit to allow for keyboard use of the enter key

src/helpers/reports.js
```diff
@@ -163,6 +165,17 @@ export function parseSearchNew(search) {
     ret.timezone = timezone;
   }
 
+  if (industryBenchmarkMetric) {
+    ret.industryBenchmarkMetric = industryBenchmarkMetric;
+    if (industryBenchmarkFilters) {
+      try {
+        ret.industryBenchmarkFilters = JSON.parse(decodeURI(industryBenchmarkFilters));
```
PatrickSison, line 172: I don't think we explicitly need this `decodeURI`. I think qs automatically does the decode URI (as well as encode URI). Although I don't think it hurts it either, so 🤷

src/pages/reportBuilder/components/ReportOptions.js
```diff
@@ -49,6 +51,13 @@ export function ReportOptions(props) {
         update.from = null;
       }
 
+      if (update.industryBenchmarkMetric && update.industryBenchmarkFilters) {
```
PatrickSison, line 54: Nice

PatrickSison (commented):
I think this looks good. Only thing is on SA-1997, I've moved some of the constants from `src/pages/reportBuider/constants` to `src/constants` since I've had to use them and they show up in the dashboard use of the charts. 

Other leftover comment is in regards to removing the actual relevant metric (if we have the industry benchmark metric, but not the actual metric, what's our behavior?)

src/pages/reportBuilder/context/ReportBuilderContext.js
```diff
@@ -66,6 +66,15 @@ const reducer = (state, action) => {
           });
         }
         update.metrics = filteredMetrics;
+
+        //Removes the industry benchmark query params when the metric is removed
+        if (
+          update.industryBenchmarkMetric &&
```
PatrickSison, line 72: I think one last integration test around this should wrap this up

PatrickSison:
@kchu1337 I think one last section of integration test and this should be good to roll out

src/pages/reportBuilder/context/ReportBuilderContext.js
PatrickSison: Nevermind, missed that it's there

PatrickSison (approved):

PR #2096, opened 2021-03-02
[FE-1383] Resolve code scanning warning for 2web2ui - src/pages/templates/*

shrirupa:
[FE-1383] Resolve code scanning warning for 2web2ui - src/pages/templates/*

### What Changed

- Remove useless condition from src/pages/templates/hooks/useEditorContent.js

### How To Test

- Check if condition change doesn't break anything in templates (editing content)  

### To Do

- [ ] Address feedback

[FE-1383]: https://sparkpost.atlassian.net/browse/FE-1383


bkemper (approved):
I see, the `isPublishedMode` guard clause is truthy, so this falsy condition isn't adding any value. 

LGTM!

PR #2097, opened 2021-03-02
SA-2220 - (UI) Migrate from the `moment-timezone` package to `timezone-support` for listing timezones

nicklemmon:
[SA-2220](https://sparkpost.atlassian.net/browse/SA-2220)

### What Changed

- Installs `timezone-support` to generate timezone options for the `TimezoneTypeahead` component
- Unfortunately, uninstalling `moment-timezone` was causing a significant bug in the Analytics Report feature - will cut a follow-up to refactor and remove it
- Refactors `TimezoneTypeahead` to allow passed in options as a prop

### How To Test

- Go to `signals/analytics` and verify the timezone typeahead works the network requests made to the metrics endpoints
- Go to `/signals/schedule/<id>` (both the create and edit flows) and verify the timezone typeahead works with network requests

### To Do

- [ ] Incorporate feedback


cypress/tests/integration/signals-analytics/analytics-report/dateAndTimeSection.spec.js
```diff
@@ -63,13 +63,10 @@ describe('Date Time Section on Summary Report & Report Builder', () => {
     cy.findByLabelText('From Date').should('have.value', '2020-01-29');
     cy.findByLabelText('From Time').should('have.value', '2:00pm');
     //Re-stub so I can get the url params from the second call
-    cy.stubRequest({
-      url: '/api/v1/metrics/deliverability/time-series**',
-      fixture: 'metrics/deliverability/time-series/200.get.json',
-      requestAlias: 'getTimeSeries2',
-    });
+    stubTimeSeries('getTimeSeriesAgain');
```
nicklemmon, line 66: This should have been used in the first place, but the request changed structure as a result of the timezone being changed from `UTC` to `Etc/UTC`

cypress/tests/integration/signals-analytics/analytics-report/dateAndTimeSection.spec.js
```diff
@@ -40,7 +40,7 @@ describe('Date Time Section on Summary Report & Report Builder', () => {
 
   it('changing date picker values changes the precision correctly', () => {
     cy.visit(PAGE_URL);
-    cy.wait(['@getSubaccounts']);
+    cy.wait('@getSubaccounts');
```
nicklemmon, line 43: Brackets weren't necessary

src/helpers/date.js
```diff
@@ -223,3 +247,68 @@ export function getDateTicks({ to, from }) {
...
+/**
+ * @description get a list of Typeahead options configuration derived from a static list of timezones
+ * @param {Array} timezones array of [IANA timezones](https://www.iana.org/time-zones) - defaults to list retrieved from `timezone-support`
+ */
+export function getTimezoneOptions(timezones = TIMEZONES) {
+  const timezoneOptions = timezones
+    .filter(isStandardTimezone)
+    .map(timezoneStr => getTimezoneWithOffset({ timezoneStr }))
+    .sort(sortByOffset)
+    .map(getTimezoneOption);
```
nicklemmon, line 306: Is this more legible or less legible than having all of the callbacks right here? I could see either way 🤔

src/components/typeahead/TimezoneTypeahead.js
```diff
@@ -1,48 +1,26 @@
...
+    options = TIMEZONE_OPTIONS,
     onChange: parentOnChange,
     isForcedUTC,
     disabledAndUTCOnly,
     ...rest
   } = props;
   const [selected, setSelected] = useState(options[0]);
 
-  const findOptionInList = useCallback(value => options.find(option => option.value === value), []);
+  const findOptionInList = useCallback(value => options.find(option => option.value === value), [
```
nicklemmon, line 21: Just a formatting change here

src/constants/index.js
```diff
@@ -299,3 +299,24 @@ export const ONBOARDING_STEP = {
   START_SENDING: 'startSending',
   ANALYTICS_REPORT_PROMO: 'analyticsReportPromo',
 };
+
+// <option> data structure for the UTC timezone
+export const UTC_TYPEAHEAD_OPTION = {
+  value: 'Etc/UTC',
```
PatrickSison, line 305: Out of curiousity, why the change of value from UTC to Etc/UTC?

src/pages/reportBuilder/components/DateTimeSection/DateTimeSection.js
```diff
@@ -62,6 +67,7 @@ export const DateTimeSection = ({
             content="Day, week, and month precision only support UTC."
           >
             <TimezoneTypeahead
```
PatrickSison, line 69: Maybe outside of the scope of this ticket, but this typeahead has always been pretty slow. You think this is just due to the large number of options we got going on here? Maybe there's some memoization we can get on the options since it's a pretty static list

src/helpers/tests/date.test.js
```diff
@@ -345,6 +347,18 @@ describe('Date helpers', () => {
...
 
+  describe('parseDateTimeTz', () => {
+    it('returns a valid Moment date time according to the passed in date and timezone', () => {
+      const result = parseDateTimeTz({
+        timezone: 'America/Belize',
+        date: '2018-01-01',
+        time: '1:00am',
+      });
+
+      expect(result).toBeValidMomentDate();
```
PatrickSison, line 358: Hopefully we can fully detach from moment soon and then just fully be using javascript dates. It's a hassle to keep bouncing between and not sure when to expect which version

cypress/tests/integration/signals-analytics/analytics-report/dateAndTimeSection.spec.js
```diff
@@ -1,4 +1,4 @@
-import { commonBeforeSteps } from './helpers';
+import { commonBeforeSteps, stubTimeSeries } from './helpers';
```
PatrickSison, line 1: Are there any timezones that are no longer in the list, but are still supported? I imagine that should be a very small list and we just prevent it from being in the dropdown, but not sanitizing it down

src/helpers/date.js
```diff
@@ -182,15 +187,34 @@ export function getFormattedDateRangeForAggregateData(from, to, timezone = getLo
...
+
 export const parseTime = str => moment(str, FORMATS.MOMENT.INPUT_TIMES, true);
+
 export const parseDatetime = (...args) =>
   moment(args.join(' '), FORMATS.MOMENT.INPUT_DATETIMES, true);
-export const parseDateTimeTz = (timezone, ...args) =>
-  moment.tz(args.join(' '), FORMATS.MOMENT.INPUT_DATETIMES, true, timezone);
+
+/**
+ * @name parseDateTimeTz
```
PatrickSison, line 205: Sorry, I created this function without unit testing so there's even less context for this.

The purpose of this function is to combine a date string and time string and generate an actual date object. Issue is that without the timezone, it interprets it to local timezone. Report builder provides a contextual timezone so it would make sense that those date and time strings are converted to date/time objects relative to that timezone rather than local timezone.

src/helpers/date.js
```diff
@@ -223,3 +247,68 @@ export function getDateTicks({ to, from }) {
...
+    offset,
+    formattedOffset,
+  };
+}
+
+/**
+ * @param {Object} timezone timezone object as derived from `getTimezoneWithOffset` helper
+ * @description reformat list to UI-friendly/formatted list consumable by the Typeahead
+ */
+function getTimezoneOption(timezone) {
```
PatrickSison, line 271: maybe a bit more clear with the name `formatTimezoneOption` or `formatTimezoneToOption`? Function by itself is clear, but in the mapping/filtering function below, reads a bit clearer to me I think as the map is the one applying some function to the array rather than retrieving from it

PatrickSison (commented):
A few things. I also wonder if there's any specific behaviors we should be checking against timezone changes once we get to daylight savings?

cypress/tests/integration/signals-analytics/analytics-report/dateAndTimeSection.spec.js
nicklemmon: Basically the reports API is migrating to support this list vs. the old list.

The metrics API, though, is using a different list 😬  - I guess this could cause problems with some saved URLs/reports.

src/constants/index.js
nicklemmon: That's how UTC is listed in this package - and it still appears to work with the metrics and reports endpoints this way.

src/pages/reportBuilder/components/DateTimeSection/DateTimeSection.js
nicklemmon: Let me mess with that! Great idea.

nicklemmon: I ended up memoizing the function to retrieve the timezone options. The good news is the component performs just fine with the production build - it just chugs a bit when running things locally.

nicklemmon:
@PatrickSison Ready for another pass when you have time 👍

src/helpers/date.js
```diff
@@ -182,15 +188,34 @@ export function getFormattedDateRangeForAggregateData(from, to, timezone = getLo
...
+
+/**
+ * @name parseDateTimeTz
+ * @description combines a date string and time string and to generate a date object
+ * @param {string} timezone
+ * @param {Date} date
+ * @param {Date} time
+ * @returns Moment date
+ */
+export const parseDateTimeTz = ({ timezone, date, time }) => {
```
PatrickSison, line 213: I also double checked this and it doesn't seem to be working as expected anymore. It seems like it's parsing to a different timezone? Or I'm not sure what. If you enter in a time in the manual entry form, I think there are 2 issues happening:
1. The timezone is no longer accounted for (as a result of this change)
2. The times automatically get formatted to the precision (outside of this change), so it's a bit confusing on if it should be doing that. 

For the first thing, it could be that the manual entry form is the one that's incorrect, but I'm not sure. It needs to spit out the format string again, so maybe the moment is correct now (and the other helpers for splitting the moment back into a date string and time string are now incorrect). I would look more into the DatePickerV2 and ManualEntryFormV2 in it's behavior.

For the second thing, I think it's just confusing UX. I enter in a time and then it rounds to a precision, so it feels like it didn't understand my input. Current behavior in prod as well.

PatrickSison (commented):
Looks good overall for the most part. I think there's one critical thing that needs updating with the removal of `moment-timezone`, and then I think we should be good

src/helpers/date.js
nicklemmon: I _think_ I fixed it - I was passing the `timezone` to `moment()` and that wasn't doing anything at all. Silly me.

So what I did was used `timezone-support` to reformat the date string and then pass that back to `moment()`. 

You're right - the UX is very unclear and it never accepts any date I type in to it. I don't really understand what the intention is, so it's very hard to test as a result.

src/helpers/date.js
```diff
@@ -182,15 +194,38 @@ export function getFormattedDateRangeForAggregateData(from, to, timezone = getLo
...
+
+/**
+ * @name parseDateTimeTz
+ * @description combines a date string and time string and to generate a date object
+ * @param {string} timezone timezone string derived from [timezone-support](https://github.com/prantlf/timezone-support/blob/HEAD/docs/API.md)
+ * @param {string} date formatted YYYY-MM-DD, e.g. "2021-02-09"
+ * @param {string} time formatted h:mma, e.g. "9:29am"
+ * @returns Moment date
+ */
+export const parseDateTimeTz = ({ timezone, date, time }) => {
```
PatrickSison, line 219: Unfortunately, this is still having some issues. If you are in local timezone, this works as expected, but if you change the timezone, this doesn't work as expected. For example, setting the timezone to UTC, and then putting in 12:00pm returns 10:00am (and consistently 2 hours behind for that timezone). What might be better is to convert the timezone into a regular ISO offset, then just do a regular parse of the date string

PatrickSison (approved):
Looks good. Behavior seems right. We'll just need to follow up to remove the rest moment-timezone later

PR #2101, opened 2021-03-03
SA-2027: Fix parsing 20+ metrics

kchu1337:
### What Changed

- Added option to qs.parse so it would always parse to array

### How To Test

- go to signals/analytics.
- Select 20+ metrics
- Verify that everything still works correctly
### To Do


src/helpers/reports.js
```diff
@@ -4,7 +4,7 @@ import queryString from 'query-string'; //TODO: Deprecate
 import { getRelativeDates, relativeDateOptions } from 'src/helpers/date';
 import { stringifyTypeaheadfilter } from 'src/helpers/string';
 import { REPORT_BUILDER_FILTER_KEY_MAP } from 'src/constants';
-
+import { list as metricsList } from 'src/config/metrics';
```
nstanard, line 7: Can we do `METRICS_LIST` again for the alias?

nstanard: Wondering if we need to (on a diff PR) do this same thing for other spots we use `qa.parse`

image

nstanard: Doesn't report builder use `usePageFilters`?
image

nstanard: Applied all 42 visible metric checkboxes:
image


With these changes, it loaded the UI. 

I then undid the arrayLimit option param and reloaded the page and got: 
image

nstanard: Nitpick: There are only 42 visible metrics in the drawer and 57 items in the list. I don't _think_ we need the `+ 1` part of the object assignment but it also can't hurt.

nstanard (approved):
Fixes the issue, couple small nitpick type comments though. Your call - just let me know what you want to do and when you want to roll this out.

src/helpers/reports.js
nstanard: Random thought before I sign off for the weekend - if the page still breaks after applying a bunch of metrics, comparisons, and filters - we might take a different approach for this: 

something like:
image

nstanard (approved):
Still LGTM. Down the road we may have to increase the array limit still, but this does fix the issue currently.

PR #2103, opened 2021-03-04
SA-2253: Remove timezone validation

kchu1337:
### What Changed

- Removed timezone validation

### How To Test

- Go to signals/analytics&timezone=Europe%2FLondon
- Also check signals/analytics&timezone=UTC


nstanard (approved):
lgtm, tests out locally with Europe/London - and all other UTC+0 options are available in the typeahead now.

That's 31 options being added back in:
image

PR #2104, opened 2021-03-05
AC-1917 - (UI) Incorporate valid HTML form elements in the app

shrirupa:
AC-1917 - (UI) Incorporate valid HTML form elements in the app

### What Changed

**On Ab testing page**
- [ ] EditMode.js - Reschedule and Finalize test button should be a submit button inside a `<form>`

**On Domain Pages**
- [ ] In SetupForSending : Re-verify Domain Panel action should be type="submit" instead of having onClick, as they are already inside a `<form>`
- [ ] In SetupBounceDomainSection - Re-verify Domain Panel action should be type="submit" instead of having onClick, as they are already inside a `<form>`
- [ ] In SendingAndBounceDomainSection - Re-verify Domain Panel action should be type="submit" instead of having onClick, as they are already inside a `<form>`
- [ ] In TrackingDnsSection - Re-verify Domain Panel action should be type="submit" instead of having onClick, as they are already inside a `<form>`
- [ ] In VerifyEmailSection | MailboxVerificationModal - should be wrapped in `<form>`

**On Report Builder**
- [ ] MetricsDrawer - this should be wrapped in a `<form>`

**On Snippet**
- [ ] CreatePage and EditPage Save Snippet and Create Snippet should have associated form id and type submit.

**On JoinForm**
- [ ] JoinForm submits on form submit instead of button click.

### How To Test

### To Do

- [ ] Address feedback
- [ ] Check if all form have submit buttons.

### DIFF
--- a/src/pages/abTesting/EditMode.js
+++ b/src/pages/abTesting/EditMode.js
@@ -4,7 +4,7 @@ import { connect } from 'react-redux';
 import { reduxForm, getFormValues } from 'redux-form';
 import { withRouter } from 'react-router-dom';
 import { list as listSubaccounts } from 'src/actions/subaccounts';
-
+import { Form } from 'src/components/form';
 import { showAlert } from 'src/actions/globalAlert';
 import {
   updateDraft,
@@ -77,12 +77,12 @@ export class EditMode extends Component {
   };
 
   getPrimaryAction = () => {
-    const { handleSubmit, test, rescheduling, rescheduleLoading } = this.props;
+    const { test, rescheduling, rescheduleLoading } = this.props;
 
     if (rescheduling) {
       return {
         content: 'Finalize and Reschedule Test',
-        onClick: handleSubmit(this.handleReschedule),
+        type: 'submit',
         disabled: rescheduleLoading,
       };
     }
@@ -90,14 +90,28 @@ export class EditMode extends Component {
     if (test.status === 'draft') {
       return {
         content: 'Finalize and Schedule Test',
-        onClick: handleSubmit(this.handleSchedule),
+        type: 'submit',
       };
     }
 
     return {
       content: 'Update Test',
-      onClick: handleSubmit(this.handleUpdateScheduled),
+      type: 'submit',
     };
+  };
+
+  getSubmitAction = () => {
+    const { test, rescheduling } = this.props;
+
+    if (rescheduling) {
+      return this.handleReschedule;
+    }
+
+    if (test.status === 'draft') {
+      return this.handleSchedule;
+    }
+
+    return this.handleUpdateScheduled;
   };
 
   getSecondaryActions = () => {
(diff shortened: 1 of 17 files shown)


src/pages/abTesting/EditMode.js
```diff
@@ -4,7 +4,7 @@ import { connect } from 'react-redux';
 import { reduxForm, getFormValues } from 'redux-form';
```
nstanard, line 4: General Note: We need to be on the lookout for how our `<Form` instances are used in relation to the submit events. 

For instance:

ManualEntryForm.js doesn't have an `onSubmit` event and I dont' think it will trigger the `handleSubmit` event in our new `src/componnets/Form` component. 

```
...
      <Form
        onKeyDown={this.handleEnter}
        className={styles.DateFields}
        id="datepicker-manualentry-form"
      >
...
```

I think these (date pickers) are the only instance that don't have the `onSubmit` prop, so I think this is the only spot that has the issue but it's something we need to be mindful of and look out for.

nicklemmon: Hey, just a side not - you can submit a form with a `<button>` that is outside of the wrapping `<form/>` using the `form` attribute on that button. Something I learned from Kevin a little while ago.

shrirupa: ok, I checked all the Forms in the app, fortunately there are only six places that have this problem where submit event is not fired - 

src/pages/profile/components/DisableTfaModal.js
src/pages/profile/components/BackupCodesModal.js
src/components/enableTfaForm/EnableTfaForm.js
src/components/datePicker/ManualEntryFormV2.js
src/components/datePicker/ManualEntryForm.js

shrirupa: I am trying to find a way this can be fixed easily in this PR, if not I will cut a separate ticket for this.

nstanard: Since it's more than just the DatePicker component I'd say we cut a new ticket. I'm reviewing this now and we can fix the submit event in the following tickets where we validate each forms fields have their needed attributes.

src/pages/join/components/JoinForm.js
```diff
@@ -70,7 +73,7 @@ export class JoinForm extends Component {
     const pending = loading || submitting || !reCaptchaReady;
 
     return (
-      <Form id="sign-up-form">
+      <Form id="sign-up-form" onSubmit={this.executeRecaptcha}>
```
nstanard, line 76: Note to self: Test that recaptcha still works when moving this along.

src/pages/domains/components/TrackingDnsSection.js
```diff
@@ -72,7 +72,7 @@ export default function TrackingDnsSection({ domain, isSectionVisible, title })
                   </Text>
                   <TranslatableText>record for this domain at your DNS provider.</TranslatableText>
                 </p>
-                <Panel.Action onClick={onSubmit}>
+                <Panel.Action type="submit">
```
nstanard, line 75: I'm not working with a real/valid domain so perhaps that's the issue but I'm not seeing this fire the `<Form` submit event.

src/pages/domains/components/SetupBounceDomainSection.js
```diff
@@ -110,7 +110,7 @@ export default function SetupBounceDomainSection({
                   record for the Hostname and Value for this domain at your DNS provider.
                 </TranslatableText>
               </p>
-              <Panel.Action onClick={onSubmit}>
+              <Panel.Action type="submit" form="verify-bounce-form">
```
nstanard, line 113: We have 4 re-verify buttons in the domains area but I only see three on this PR: 
image

I think we should break this PR into 2 parts. The domain area files can be it's own PR and we can move the rest along. I didn't see anything off with the other files. (and testing/verifying  will be easier if it's broken up into smaller pieces)

nstanard: I see now that it's already hooked up - and I think that 4th file that's not on the PR is working correctly with the `<Form` submit 👍

src/pages/domains/components/SetupForSending.js
```diff
@@ -106,7 +106,7 @@ export default function SetupForSending({ domain, isSectionVisible }) {
                     record for the Hostname and DKIM value of this domain.
                   </TranslatableText>
                 </p>
-                <Panel.Action onClick={onSubmit}>
+                <Panel.Action type="submit">
```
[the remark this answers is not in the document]
shrirupa: I checked the DOM it does get applied.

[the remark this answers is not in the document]
nstanard: Was able to verify this just now too 👍

src/pages/domains/components/SetupBounceDomainSection.js
```diff
@@ -110,7 +110,7 @@ export default function SetupBounceDomainSection({
                   record for the Hostname and Value for this domain at your DNS provider.
                 </TranslatableText>
               </p>
-              <Panel.Action onClick={onSubmit}>
```
nstanard, line 113: Verified this location submits the `<Form` submit event. 🟢

src/pages/domains/components/VerifyEmailSection.js
```diff
@@ -71,7 +71,7 @@ export default function VerifyEmailSection({ domain, isSectionVisible }) {
 
 function VerifyButton({ onClick, variant = 'primary', submitting }) {
   return (
-    <Button variant={variant} loading={submitting} onClick={onClick}>
+    <Button variant={variant} loading={submitting} onClick={onClick} type="submit">
```
nstanard, line 74: This submit event only has a button in the submit event - there is no input being passed through to the `<Form` submit event.

shrirupa:
closing in favor of
https://github.com/SparkPost/2web2ui/pull/2118
https://github.com/SparkPost/2web2ui/pull/2119

PR #2107, opened 2021-03-08
AC-1911: Unable to re-enter webui when using SSO and Safari

shrirupa:
AC-1911: Unable to re-enter webui when using SSO and Safari

Note - Reason for the below change - https://sparkpost.atlassian.net/browse/AC-1911?focusedCommentId=204739
Related to https://bugs.webkit.org/show_bug.cgi?id=77538

### What Changed

- Added an extra check on DefaultRedirect to set the authCookie, if it is not already set

### How To Test

- Not sure of a way to test this locally, Can be easily tested in staging though using the credentials we have in 1pass.

### To Do

- [ ] Address feedback
- [ ] Add a test to check this functionality for non safari browsers


nicklemmon (approved):
Conditional approval to test in staging - will revert shortly.

PR #2080, opened 2021-02-22
AC-1863 - (UI) Create a global <Form/> component and configure existing form fields to be easily trackable

shrirupa:
AC-1863

### What Changed

- Create a global `<Form>` component.
- Replace all `<form>` with <Form>
- added id to all `<Form>`
- added eslint rule to throw a waning when `<form>` is used
- Update lot of snapshots 😅 

### How To Test

- Check no active pages have html `<form>` and are replaced by `<Form>` with ids

### To Do

- [ ] Address feedback

PR #2111, opened 2021-03-09
AC-1934 - Valid DOM Attributes for form fields

nstanard:
https://sparkpost.atlassian.net/browse/AC-1864

### What Changed

### How To Test
- Describe any special dev environment setup required.

### To Do

#### Resources
https://www.w3schools.com/tags/att_input_name.asp
https://reactjs.org/docs/dom-elements.html
https://www.w3.org/WAI/tutorials/forms/labels/

#### Previous Related Tasks:
https://github.com/SparkPost/2web2ui/pull/2080
https://github.com/SparkPost/2web2ui/pull/2104

### DIFF
--- a/src/components/datePicker/ManualEntryForm.js
+++ b/src/components/datePicker/ManualEntryForm.js
@@ -153,6 +153,7 @@ export class ManualEntryFormClassComponent extends Component {
           <Grid.Column>
             <TextField
               id="fromDate"
+              name="fromDate"
               label="From Date"
               labelHidden
               placeholder={DATE_PLACEHOLDER}
@@ -164,6 +165,7 @@ export class ManualEntryFormClassComponent extends Component {
           <Grid.Column>
             <TextField
               id="fromTime"
+              name="fromTime"
               label="From Time"
               labelHidden
               placeholder={TIME_PLACEHOLDER}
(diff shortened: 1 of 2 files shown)


src/components/datePicker/ManualEntryForm.js
```diff
@@ -153,6 +153,7 @@ export class ManualEntryFormClassComponent extends Component {
           <Grid.Column>
             <TextField
               id="fromDate"
+              name="fromDate"
```
nstanard, line 156: Location to test: http://localhost:3100/lists/suppressions

src/components/datePicker/ManualEntryForm.js
```diff
@@ -153,7 +153,8 @@ export class ManualEntryFormClassComponent extends Component {
           <Grid.Column>
```
nstanard, line 153: Currently, the date picker form submit is not firing the `<Form` handler event. For AC-1864 I am going to add to the AC that we add logging to the submit event handler and test the form field data sent through to the handler. So we can evaluate what the data will actually look like before we go to pass it off to the tracking method.

nstanard: https://github.com/SparkPost/2web2ui/pull/2104/files#r590750075

nstanard:
Taking a different approach

PR #2118, opened 2021-03-11
[AC-1917-without-domain-changes] All other form updates other than domains

shrirupa:
This is split branch from AC-1917 and has all other changes other than domain

### What Changed

**On Ab testing page**
- [ ] EditMode.js - Reschedule and Finalize test button should be a submit button inside a `<form>`

**On Report Builder**
- [ ] MetricsDrawer - this should be wrapped in a `<form>`

**On Snippet**
- [ ] CreatePage and EditPage Save Snippet and Create Snippet should have associated form id and type submit.

**On JoinForm**
- [ ] JoinForm submits on form submit instead of button click.

### How To Test

### To Do

- [ ] Address feedback
- [ ] Check if all form have submit buttons.

### DIFF
--- a/src/pages/abTesting/EditMode.js
+++ b/src/pages/abTesting/EditMode.js
@@ -4,7 +4,7 @@ import { connect } from 'react-redux';
 import { reduxForm, getFormValues } from 'redux-form';
 import { withRouter } from 'react-router-dom';
 import { list as listSubaccounts } from 'src/actions/subaccounts';
-
+import { Form } from 'src/components/form';
 import { showAlert } from 'src/actions/globalAlert';
 import {
   updateDraft,
@@ -77,12 +77,12 @@ export class EditMode extends Component {
   };
 
   getPrimaryAction = () => {
-    const { handleSubmit, test, rescheduling, rescheduleLoading } = this.props;
+    const { test, rescheduling, rescheduleLoading } = this.props;
 
     if (rescheduling) {
       return {
         content: 'Finalize and Reschedule Test',
-        onClick: handleSubmit(this.handleReschedule),
+        type: 'submit',
         disabled: rescheduleLoading,
       };
     }
@@ -90,14 +90,28 @@ export class EditMode extends Component {
     if (test.status === 'draft') {
       return {
         content: 'Finalize and Schedule Test',
-        onClick: handleSubmit(this.handleSchedule),
+        type: 'submit',
       };
     }
 
     return {
       content: 'Update Test',
-      onClick: handleSubmit(this.handleUpdateScheduled),
+      type: 'submit',
     };
+  };
+
+  getSubmitAction = () => {
+    const { test, rescheduling } = this.props;
+
+    if (rescheduling) {
+      return this.handleReschedule;
+    }
+
+    if (test.status === 'draft') {
+      return this.handleSchedule;
+    }
+
+    return this.handleUpdateScheduled;
   };
 
   getSecondaryActions = () => {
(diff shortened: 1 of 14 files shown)


src/pages/abTesting/EditMode.js
```diff
@@ -4,7 +4,7 @@ import { connect } from 'react-redux';
 import { reduxForm, getFormValues } from 'redux-form';
```
nstanard, line 4: 🟢  - submit handler event fires, lgtm
image

src/pages/join/components/JoinForm.js
```diff
@@ -54,7 +54,10 @@ export class JoinForm extends Component {
     this.setState({ recaptcha_response: response }, this.props.handleSubmit(this.onSubmit));
```
nstanard, line 54: 🟢  - submit handler event fires, lgtm
image

nstanard: Although... for some reason I can't actually create an account right now. I don't think it's related to this PR but i'll double check this tomorrow morning.

src/pages/join/components/JoinForm.js
```diff
@@ -54,7 +54,10 @@ export class JoinForm extends Component {
...
   };
 
-  executeRecaptcha = () => this.state.reCaptchaInstance.execute();
+  executeRecaptcha = event => {
+    event.preventDefault();
+    this.state.reCaptchaInstance.execute();
```
nstanard, line 59: You might want to add a `return` so this method returns `this.state.reCaptchaInstance.execute();` like the arrow functionw as doing already. (lowers risk of unintended consequences)

src/pages/snippets/EditPage.js
```diff
@@ -95,18 +95,18 @@ export default class EditPage extends React.Component {
     }
 
     return (
-      <Page
```
nstanard, line 98: 🟢  - submit handler event fires, lgtm (for create and edit)

image

src/pages/reportBuilder/components/MetricsForm/MetricsForm.js
```diff
@@ -97,7 +99,7 @@ export default function MetricsForm(props) {
   //Needs this for current tests as hibana is not enabled for tests but is required for the Drawer component
   const { DrawerFooter = Drawer.Footer } = props;
   return (
-    <>
+    <Form id="reportbuilder-metrics-form" onSubmit={handleApply}>
```
nstanard, line 102: 🟢  - submit handler event fires, lgtm

image

src/pages/abTesting/EditMode.js
```diff
@@ -77,29 +77,43 @@ export class EditMode extends Component {
...
 
     return {
       content: 'Update Test',
-      onClick: handleSubmit(this.handleUpdateScheduled),
+      type: 'submit',
     };
   };
 
+  getSubmitAction = () => {
+    const { test, rescheduling } = this.props;
```
nstanard, line 104: note to self: Technically we want to test/verify all three action types when we move this along

src/pages/join/components/JoinForm.js
nstanard: 🤔  ... now I'm not sure sure actually. Probably doesn't make a difference. But the arrow function did return the execute method call.

src/pages/join/components/JoinForm.js
nstanard: Whew, finally got the create to go through. (it was just sitting there loading the first couple times I tried .. probably VPN)
image

nstanard (approved):
One small note/comment but this LGTM!

src/pages/join/components/JoinForm.js
shrirupa: Good catch!

PR #2119, opened 2021-03-11
AC-1917 - with Domain changes only

shrirupa:
### What Changed

**On Domain Pages**
- [ ] In SetupForSending : Re-verify Domain Panel action should be type="submit" instead of having onClick, as they are already inside a `<form>`
- [ ] In SetupBounceDomainSection - Re-verify Domain Panel action should be type="submit" instead of having onClick, as they are already inside a `<form>`
- [ ] In SendingAndBounceDomainSection - Re-verify Domain Panel action should be type="submit"
- [ ] In TrackingDnsSection - Re-verify Domain Panel action should be type="submit" instead of having onClick, as they are already inside a `<form>`
- [ ] In VerifyEmailSection | MailboxVerificationModal - should be wrapped in `<form>`

### How To Test

- Easier approach to testing this section would be take a existing verified sending/bounce domain and navigate to 
     - /domains/details/:domainName/verify-sending-bounce
     - domains/details/:domainName/verify-bounce
     - domains/details/:domainName/verify-sending
     for tracking domain
    - domains/details/:domainName/verify-tracking
- For email verification Modal
   - For testing MailboxVerificationModal open src/config/env/dev-config.js change allow_anyone_at_verification = false and allow_mailbox_verification = true, no spc/spceu/staging account(we can't test for enterprise account in staging) has this feature flag disabled, so the default AllowAnyoneVerificationModal currently is the most used one. Testing the MailboxVerificationModal locally should be fine
   - For testing AllowAnyoneVerificationModal open src/config/env/dev-config.js change allow_anyone_at_verification = true and allow_mailbox_verification = true

### To Do

- [ ] Address feedback
- [ ] Check if Re-Verify works correctly.

### DIFF
--- a/src/pages/domains/components/VerifyEmailSection.js
+++ b/src/pages/domains/components/VerifyEmailSection.js
@@ -71,7 +71,7 @@ export default function VerifyEmailSection({ domain, isSectionVisible }) {
 
 function VerifyButton({ onClick, variant = 'primary', submitting }) {
   return (
-    <Button variant={variant} loading={submitting} onClick={onClick}>
+    <Button variant={variant} loading={submitting} onClick={onClick} type="submit">
       Send Email
     </Button>
   );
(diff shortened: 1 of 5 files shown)


shrirupa:
@nstanard I have also added the change for SendingAndBounceDomainSection

src/pages/domains/components/VerifyEmailSection.js
```diff
@@ -71,7 +71,7 @@ export default function VerifyEmailSection({ domain, isSectionVisible }) {
```
nstanard, line 71: I'm not sure the <Form on line 99 of this file is correctly setup with the modal:

image

image

The form instance is outside the modal content and has no children. Can we double check this too for this PR? I want to make sure that's hooked up correctly as well.

shrirupa: I think this is happening because the Modal is rendered in a portal outside DOM. I can move the form inside it.

nstanard: latest pull / commit also tested out - this form now submits to the tracking form submit event correctly. 👍 
image

src/pages/domains/components/VerifyEmailSection.js
```diff
@@ -169,36 +168,38 @@ function MailboxVerificationModal(props) {
...
-            </Grid.Column>
-            <Grid.Column xs={6}>
-              <VerifyButton
-                onClick={verifyWithAbuse}
-                variant="secondary"
-                submitting={verifyEmailLoading}
-              />
-            </Grid.Column>
-          </Grid>
+          <Form id="domain-email-verification-postmaster">
```
nstanard, line 171: getting a full page refresh here for these two forms - doing some more testing to make sure that's not an issue

nstanard: Had to change to `onSubmit` with `e.preventDefault()` to get this to not refresh the whole page. 

image

shrirupa: Thanks!

src/pages/domains/components/VerifyEmailSection.js
```diff
@@ -169,36 +168,50 @@ function MailboxVerificationModal(props) {
...
+          >
+            <Grid middle="xs">
+              <Grid.Column xs={6}>
+                <p>
+                  <strong>{`abuse@${id}`}</strong>
+                </p>
+              </Grid.Column>
+              <Grid.Column xs={6}>
+                <VerifyButton
+                  onClick={verifyWithAbuse}
```
nstanard, line 208: are these onclicks still supposed to be here?

nstanard (approved):
Latest changes LGTM. Lets let the tests finish and send this out after lunch.

PR #2122, opened 2021-03-12
SA-2248: Deliverability metrics changes

PatrickSison:
### What Changed

- Changed industry benchmark modal button to "Apply Settings"
- Changed calculation for moved to inbox and moved to spam rates
- Added tooltips for deliverability metrics

### How To Test
- Use an account with d12y metrics enabled (appteam)
- Open metrics modal
- Verify d12y metrics have tooltip (and content is correct)
- Verify Accepted metrics removed from this d12y section
- Add "Inbox Folder Rate" as metric,
-  then above chart, click "Industry Benchmark"
- Verify text has changed to "Apply Settings"


src/helpers/metrics.js
```diff
@@ -483,6 +483,20 @@ export function splitDeliverabilityMetric(metric, dataSource) {
             data[`count_inbox_${targetString}`] + data[`count_spam_${targetString}`],
           ),
       };
+    case 'count_moved_to_inbox':
```
PatrickSison, line 486: I'm going to add some unit and integration testing around this. There's a lot of logic here

nicklemmon: @PatrickSison Should I hold off on the review until you add these tests or would that be for another ticket down the road?

PatrickSison: Forgot to push it. Just pushed some up

src/hooks/reportBuilder/useIndustryBenchmark.js
```diff
@@ -18,7 +18,12 @@ function useIndustryBenchmark(reportOptions) {
   }, [metrics]);
 
   const formattedOptions = useMemo(() => {
-    return getQueryFromOptions({ ...reportOptions, metrics: formattedMetrics });
+    const { from, to } = roundBoundaries({
+      from: reportOptions.from,
+      to: reportOptions.to,
+      precision: 'day',
+    });
+    return getQueryFromOptions({ ...reportOptions, metrics: formattedMetrics, from, to });
```
nicklemmon, line 26: Just curious, why would did you introduce this?

nicklemmon (commented):
@PatrickSison Just had a question about why a certain function was used - let me know why - seems a little funky to me?

src/hooks/reportBuilder/useIndustryBenchmark.js
PatrickSison: Industry benchmark wasn't working as intended. It always works with day precision, and in order to match rounding/behavior of existing charts, we match the same rounding property. So if we select middle of day, it rounds to the beginning of the day. This can be removed if in the future, we change how rounding works in the front end

nicklemmon (approved):
@PatrickSison  👍 Looks good, sir

PR #2123, opened 2021-03-16
AC-1944 - rename LargeForm and Form to LargeMain and Main

nstanard:
### What Changed

- LargeForm.js renamed to LargeMain.js
- Form.js renamed to Main.js
- Tests and imports updated

### How To Test

- Form -> Main
- LargeForm -> LargeMain

- Enable2FA- https://github.com/SparkPost/2web2ui/pull/2123/files#diff-ec169d4f5084fe885b50954b2758d16ef2c2d166b8a58b1eb6a273861648e05dR112
- RV Onboarding- https://github.com/SparkPost/2web2ui/pull/2123/files#diff-ec169d4f5084fe885b50954b2758d16ef2c2d166b8a58b1eb6a273861648e05dR614

- We need to load up those routes and make sure there's no issues or differences. It should look the same as before. (nothing was changed, file was just renamed and import was updated.)

### To Do

- [ ] Feedback


src/components/layout/Main.js
```diff
@@ -4,7 +4,7 @@ import ScrollToTop from './components/ScrollToTop';
 import OGStyles from './Layout.module.scss';
 import HibanaStyles from './LayoutHibana.module.scss';
 
-const Form = ({ children }) => {
+const Main = ({ children }) => {
```
kchu1337, line 7: Could replace both files with a single generator like 
```js
const layoutGenerator = (size = 'default') => ({ children }) => {
  const styles = useHibanaOverride(OGStyles, HibanaStyles);
  const style = size === 'large' ? styles.form : styles.largeForm;
  return (
    <div className={`${styles.wrapper} ${style}`}>
      <main role="main" className={styles.content}>
        <div className={styles.container}>{children}</div>
      </main>
      <ScrollToTop/>
    </div>
  );
};

export const Main = layoutGenerator();
export const LargeMain = layoutGenerator('large');
```

kchu1337: Oh and I guess it would be good to restrict proptypes for `size` to be an enum of `default` or `large`

kchu1337 (changes requested):
Up to you to make that change. I like pushing for less code duplication, but it can make code more confusing. If you think this is a case where the code becomes more confusing, feel free to ignore.

src/components/layout/Main.js
nstanard: I like that feedback a lot! My only hesitation is managing risk for doing that and testing/verifying it. This current change carries low risk. I might take that snippet and write it up as a different task. That feedback is great and I think we should 100% make that change at some point.

nstanard: https://sparkpost.atlassian.net/browse/FE-1415 - Great feedback. I can't decide if I want to push this PR along or close it in favor of that idea down the road.

nstanard:
Closing this one out in favor of: https://sparkpost.atlassian.net/browse/FE-1415

Good proposal @kchu1337

PR #2125, opened 2021-03-16
AC-1950 - Webhook Form Fields

nstanard:
https://sparkpost.atlassian.net/browse/AC-1864

### What Changed

- Added ids to the webhook form fields.

### How To Test

- Test the webhook form submit event calls the `src/components/tracking/form` tracking handler.
- Test create AND edit for webhooks.
- Double check that all fields have the right attributes so we can send that info through to segment in a future ticket.

See original ticket for understanding elements/attributes: https://sparkpost.atlassian.net/browse/AC-1864

### To Do

- [ ] Feedback

### Important Notes:
 - `src/components/tracking/form` component has logging in place to verify the submit event fires: 
 
image

Create Submit:
image

Edit Submit:
image


src/pages/webhooks/components/Fields.js
```diff
@@ -102,11 +107,12 @@ const AuthDropDown = ({ disabled }) => (
     ]}
     helpText={
       <span>
-        Select "None" if your target URL has no authentication scheme.{' '}
+        <TranslatableText>
+          Select "None" if your target URL has no authentication scheme.
+        </TranslatableText>
         <ExternalLink to="https://support.sparkpost.com/customer/portal/articles/2112385">
-          More information
+          More information.
```
nicklemmon, line 114: It's weird from a JSX standpoint, but the period should be outside of the link - the text is clickable not the punctuation in the sentence.

src/pages/webhooks/components/Fields.js
```diff
@@ -102,11 +107,12 @@ const AuthDropDown = ({ disabled }) => (
     ]}
     helpText={
       <span>
-        Select "None" if your target URL has no authentication scheme.{' '}
+        <TranslatableText>
```
nicklemmon, line 110: 👍

src/pages/webhooks/components/tests/__snapshots__/Fields.test.js.snap
```diff
@@ -5,6 +5,7 @@ exports[`should render ActiveField 1`] = `
   component={[Function]}
```
nicklemmon, line 5: Snapshot test should just be deleted at this point - this is pretty useless. Even slowed you down on implementing this ticket!

nicklemmon (changes requested):
Left a few comments.

Also, it would be worth changing these `id` values - they are getting passed in a way that looks confusing because of the use of the `.` and if you needed to use any DOM manipulation browser APIs, it would cause some problems:
Screen Shot 2021-03-16 at 12 52 11 PM

For example, `document.querySelector('#events.delay')` does not grab the element correctly because it thinks the `.delay` refers to the `class` attribute rather than the `id`.

src/pages/webhooks/components/Fields.js
nicklemmon: https://ux.stackexchange.com/questions/17331/should-a-full-sentence-html-link-include-the-period-in-the-linked-text#:~:text=Use%20a%20period%20(or%20other,don't%20link%20the%20period.

src/pages/webhooks/components/tests/__snapshots__/Fields.test.js.snap
nstanard: I actually don't mind the snapshots. Might delete here later when I have time.

nstanard:
> Left a few comments.
> 
> Also, it would be worth changing these `id` values - they are getting passed in a way that looks confusing because of the use of the `.` and if you needed to use any DOM manipulation browser APIs, it would cause some problems:
> Screen Shot 2021-03-16 at 12 52 11 PM
> 
> For example, `document.querySelector('#events.delay')` does not grab the element correctly because it thinks the `.delay` refers to the `class` attribute rather than the `id`.

I'd actually prefer we write these locations up and limit the changes we make for the scope of the label/id work. For this PR I did add an id override but I don't think I'll do that for all of them. If the form fields have the right attributes and labels I'll leave them alone.

src/pages/webhooks/components/WebhookForm.js
```diff
@@ -34,6 +34,7 @@ export function EventCheckBoxes({ events, disabled }) {
         <div key={key}>
           <Tooltip dark content={description}>
             <Field
+              id={key}
```
nstanard, line 37: We should probably decide if we want to enforce `id` attributes on `<Field` components or let redux-form set them via the `name=` attribute. Turns out that I don't actually have to add these id attributes - the IDs are already there.

image

nstanard: So basically, the changes on this PR aren't really needed. (but has helped me better understand the actual scope of work needed to make sure we are tracking the form fields properly)

nstanard:
This form actually has all the ids/names and labels as far as I can tell. They may not be formatted in an ideal way but they are making it through to the tracking submit event so I'm closing this and using it as a base of what to look for as I make my way through all the Form instances.

PR #2126, opened 2021-03-16
AC-1932 - Track all pageviews 

shrirupa:
AC-1932 - Track all pageviews 

### What Changed

- Added tracking for page redirects and also added supporting cypress tests for the same
- Added tracking for url changes and also added supporting cypress tests for the same

### How To Test

- For checking redirects 
    - Enter /reports/summary, you should be redirected to signals/analytics
        - Expectation: reports/summary should be tracked in segment along with signal/analytics
- For checking url changes
    - Best places to test this is domains and signal analytics page, try changing filters on these pages.
        - Expectation: all query param changes should be tracked in segment
------------------------------------------------------------------------------------------------
- For testing the utm change, we can use the url below
   - /templates?utm_campaign=fake
### To Do

- [ ] Address feedback


nstanard (approved):
This looks good, I'm looking at the events show up in segment and the tracking is working well.

My only concern is that we might want to check with our current billing/price since we are already over our limit. This change will most likely increase the overage charges since we'll be tracking all filter/url query param changes.
image

src/helpers/segment.js
```diff
@@ -97,10 +97,12 @@ const getUTMsFromURL = () => {
   }, {});
 };
 
-export const segmentPage = () => {
+export const segmentPage = url => {
   if (window.analytics && window.analytics.page) {
     const UTMs = getUTMsFromURL();
-    if (UTMs) {
+    if (url) {
+      window.analytics.page(url);
```
nstanard, line 104: I approved the PR but I think we should test this if block to make sure we are still tracking UTMs if needed. This block has the possibility of preventing the UTM condition from ever being true.

src/components/segment/SegmentPage.js
```diff
@@ -5,13 +5,18 @@ import { segmentPage } from 'src/helpers/segment';
 export const SegmentPage = () => {
   const history = useHistory();
   const prevPathname = useRef(history.location.pathname);
+  const prevSearchParams = useRef(history.location.search);
 
   // On URL path changes, inform segment via "PAGE" event
   useEffect(() => {
```
nstanard, line 11: Testing now but I think we can condense these two useEffect blocks to one: 
image

nstanard (approved):
Everything is testing out locally and I really like how you put tests in place for the segmentPage calls. LGTM!

PR #2130, opened 2021-03-18
SA-2047: Changed expandable behavior

PatrickSison:
### What Changed

- Changed default state for expandables

### How To Test

- Verify that nav is smaller
- Verify metrics expandable is open
- Verify other expandables are closed
- Verify that the buttons in expandables are smaller

### To Do

- [ ] Update testing


nicklemmon (approved):
@PatrickSison Looks good to me - nice work adding those re-usable helpers for integration tests. Good call.

PR #2131, opened 2021-03-18
FE-1414 - Download CSV button

nstanard:
### What Changed

- Added Download CSV button to domain tabs via the Pagination component.

`http://localhost:3100/domains/list/sending`
`http://localhost:3100/domains/list/bounce`
`http://localhost:3100/domains/list/tracking`

### How To Test

- Visit sending, bounce, and tracking domain tabs with and without pagination visible.
- Filter and sort the table results and test the CSV Download for each tab.

### To Do

- [ ] Feedback


src/pages/domains/components/styles.js
```diff
@@ -62,6 +62,15 @@ export const Chevron = styled(ChevronRight)`
 
 export const DomainsListPanel = styled(Panel)``;
 
+export const PaginationDownloadButtonWrapper = styled.div`
```
nstanard, line 65: @jonambas Do you have an alternate way to do this with Matchbox components?

nstanard: Scratch that, used built in pagination save as by mapping the table instance rows.

src/pages/domains/components/TrackingDomainsTab.js
```diff
@@ -192,6 +192,7 @@ export default function TrackingDomainsTab() {
 
       return;
     }
+
```
nicklemmon, line 195: 👾

src/pages/domains/components/SendingDomainsTab.js
```diff
@@ -383,10 +383,10 @@ export default function SendingDomainsTab({ renderBounceOnly = false }) {
       </DomainsListPanel>
 
       <Pagination
-        data={rows}
+        data={tableInstance.rows.map(i => i.original)}
```
nicklemmon, line 386: Mind using a more descriptive variable name than `i` here? Maybe `row`? Or you could always destructure it:
```js
tableInstance.rows.map(({ original }) => original));
```

nicklemmon (changes requested):
@nstanard Your changes work - I left one comment, re: variable naming.

Adding automated testing behind this is also needed. It looks like there is a Cypress test that was avoiding testing a downloaded file due to a missing feature in some of the billing page specs (`cypress/tests/integration/accounts/billing/billingPage.spec.js`).

The good news is our current version of Cypress now supports file downloading. They even have a testing recipe to validate file content if you want to go that far:
https://github.com/cypress-io/cypress-example-recipes/blob/master/examples/testing-dom__download/cypress/integration/local-download-spec.js

At the very least, testing that the button is there for each tab is important.

Thanks!

nstanard:
> @nstanard Your changes work - I left one comment, re: variable naming.
> At the very least, testing that the button is there for each tab is important.
> 
> Thanks!

I did what I could with the time I had available but the actual download test will have to wait for another day. We should probably write that up to add some cypress helpers/tasks/commands to make checking downloaded CSV files easier.

cypress/plugins/index.js
```diff
@@ -34,8 +36,19 @@ module.exports = (on, config) => {
...
+      return new Promise((resolve, reject) => {
+        rmdir(folderName, { maxRetries: 10, recursive: true }, err => {
+          if (err) {
+            error(err);
+            return reject(err);
+          }
+          resolve(null);
+        });
+      });
+    },
```
nicklemmon, line 51: This is cool!

nicklemmon (approved):
Looks good to me!

PR #2112, opened 2021-03-10
AC-1865 - (UI) Incorporate global web analytics tracking for Modal component open and close events

shrirupa:
AC-1865 - (UI) Incorporate global web analytics tracking for <Modal /> component open and close events

### What Changed

- Added segment tracking for when Modal is opened and closed
- Also added sending Modal name along with these events
      - For Legacy Modals added title prop to them.
      - For new Modals this name is parsed from Modal.Header.
- Added cypress tests and units test coverage for the change.

### How To Test
 Check if Opening and Closing of Modals send correct event and name to Segment,  it is not necessary to check each and every Modal in the app. Covering a few Modal of each type, namely a few Legacy Modal and a few new Modals should suffice.
- Check if All Legacy Modals have a title.
      - For the Legacy Modal, I have noted with each file, how to get to these locations easily. Check comments below on each file.
- Check when opening a new Modal, name gets correctly parsed from the header.
       - For new Modal, these are the only location it is present at, it would suffice to check one on the dashboard which open up by clicking `Change Report`. Worst that could happen in this case is Modal.Header is not set by developer when they add a new Modal, in that case, this should show warning in console, and name won't be tracked but I have also added the location to the segment event so that data isn't completely useless and we know on which location was this Modal opened.
Screen Shot 2021-03-15 at 5 10 57 PM

### To Do

- [ ] Address feedback
- [x] Add test for name parsing in the Modal and Legacy components.
- [x] Add supporting unit test to TrackModal.

PR #2135, opened 2021-03-19
SA-1998: Added Form tracking component to industry benchmark modal

PatrickSison:
### What Changed

- Added `<Form />` component to industry benchmark modal

### How To Test

- Add Inbox Folder Rate as metric in report builder
- Click Industry Benchmark above chart
- Verify that the industry benchmark form still works

### To Do

- [ ] Address feedback


nstanard (commented):
image
Modal looks good.


image
Wondering how we are going to identify the selected option in the form submit. I'm not seeing the selected attribute on the option I picked. Testing more to see if I am missing something.

nstanard:
> Wondering how we are going to identify the selected option in the form submit. I'm not seeing the selected attribute on the option I picked. Testing more to see if I am missing something.

image

It seems that if we need to have the <option /> that's selected indicated in the form submit we need Matchbox to set the selected attribute. I'm not seeing that happen right now.

PatrickSison:
@nstanard I think that would be an issue with all forms now then right? We're doing most forms as having inputs externally controlled in react state, so not sure if those values would ever get passed through the submit handler.

nstanard (approved):
Alright - I got the info I needed....

We can look at the el.nodeName and as @nicklemmon helped point out in Slack we can look at the el.value for select dropdowns to get the value of the option they picked. Obviously that's not important for this PR but will come into play over the next few days/weeks.

image

LGTM!

src/pages/reportBuilder/components/IndustryBenchmarkModal.js
```diff
@@ -38,11 +41,19 @@ export const IndustryBenchmarkModal = ({ metrics, isModalOpen, closeModal }) =>
...
+    setIndustryBenchmark(state);
+    closeModal();
+  };
+
   return (
-    <>
-      <Modal open={isModalOpen} onClose={closeModal} showCloseButton>
-        <Modal.Header>Industry Benchmark</Modal.Header>
-        <Modal.Content>
+    <Modal open={isModalOpen} onClose={closeModal} showCloseButton>
```
nstanard, line 53: I could be wrong but at a glance I think we may also need a title attribute for the modals if this PR doesn't already handle it - https://github.com/SparkPost/2web2ui/pull/2112/files

nstanard: Nvm - only legacy instances need title 🥇

PR #2138, opened 2021-03-22
FE-1380 - Removes use of `javascript` URLs in the app

nicklemmon:
[FE-1380](https://sparkpost.atlassian.net/browse/FE-1380)

### What Changed

- Removes OG Navigation which leveraged `javascript` URLs and was no longer in use
- Removes remaining `javascript` URLs in the app, updating tests accordingly

### How To Test

- Verify UI components are still interactive using a keyboard
- Verify the changes did not introduce a regression, particularly on the signup tabs

### To Do

- [ ] Incorporate feedback


shrirupa (approved):
looks good to me!

PR #2142, opened 2021-03-25
FE-1398: Addressed inconsistent state update

PatrickSison:
### What Changed

- Changed setState functions to use callback variation of the function

### How To Test

- Verify that affected pages still work correctly
- [ ] Any page with date picker (Like message events)
- [ ] Webhooks test tab
- [ ] Alerts mute toggling
- [ ] Polling (not sure if RV still uses this or anything does)

### To Do

- [ ] Address feedback


config/paths.js
```diff
@@ -33,9 +28,8 @@ const getPublicUrl = appPackageJson =>
 // like /todos/42/static/js/bundle.7289d.js. We have to know the root.
 function getServedPath(appPackageJson) {
   const publicUrl = getPublicUrl(appPackageJson);
-  const servedUrl =
-    envPublicUrl || (publicUrl ? url.parse(publicUrl).pathname : '/');
-  return ensureSlash(servedUrl, true);
+  const servedUrl = envPublicUrl || (publicUrl ? url.parse(publicUrl).pathname : '/');
+  return ensureSlash(servedUrl);
```
PatrickSison, line 32: Only place this function ever gets used, so removed unnecessary param

src/pages/ipPools/components/PoolForm.js
```diff
@@ -115,7 +115,7 @@ const mapStateToProps = (state, props) => {
   const pool = selectCurrentPool(state, props);
   return {
     pool,
-    pools: getIpPools(state, props),
+    pools: getIpPools(state),
```
PatrickSison, line 118: If you look at this function, it only accepts and uses 1 param so removed extra param

bkemper (approved):
LGTM!

PR #2144, opened 2021-03-25
AC-1979 - Global tracking for all forms in dashboardPageV2 and reportBuilder

shrirupa:
AC-1979 - Global tracking for all forms in dashboardPageV2 and reportBuilder

### What Changed

- Allowlisted all dashboard and report builder pages
- On Dashboard had to adjust ids for Radio button so that they are valid for querySelector(valid ids shouldn't start with numbers). This change doesn't effect current dashboard functioning.
- Also on Dashboard added a small snippet of code so on Form submission the radio button state is not reset. Left some comments in the code around this.
- Added input type=hidden for TypeAhead, this helps tracking this data when form is submitted
- Added cypress tests around these forms

### How To Test

- Check if all information from dashboard is sent correctly to segment
    1. Click on Change Report and submit the Form. Check if data sent to segment looks correct

### To Do


cypress/tests/integration/dashboard/dashboardPage.spec.js
```diff
@@ -91,13 +84,37 @@ describe('the dashboard page', () => {
...
+        }
+        if (event === 'Modal Closed') {
+          modalClosedProperties = property;
+        }
+        if (event === 'Form Submitted') {
+          formSubmittedProperties = property;
+        }
+        return event;
+      });
+      expect(events).to.include.members(['Form Submitted', 'Modal Closed', 'Modal Opened']);
```
nstanard, line 104: It's not a huge issue but locally I'm getting a failure for this test and the event array only has: `["Modal Opened", "Form Submitted"]`. We can probably ignore that since it's passing on circle but thought I'd mention it.

src/components/matchbox/ComboBoxTextField.js
```diff
@@ -15,6 +15,13 @@ function ComboBoxTextField(props) {
     return (
       <Box maxWidth={maxWidth ? maxWidth : '1200'}>
         <HibanaComboBoxTextField {...rest} />
+        <input
+          type="hidden"
+          label={props.label}
+          id={props.id}
+          name={props.name}
+          value={props.selectedItems.map(x => props.itemToString(x)).join(',')}
```
nstanard, line 23: Is there ever a scenario we need to be concerned that `props.selectedItems` is undefined?

nstanard: I tested this in the Templates -> Send Test modal and saw it was always an empty array to begin with. No worries here.

nstanard (approved):
This all LGTM. I wish there were a different way to handle the change report modal form submit but I'm not sure it's worth putting time into.

PR #2146, opened 2021-03-26
SA-2305: Sets custom values as being lowercase

PatrickSison:
### What Changed

- Custom values for input value in typeaheads now get set to lowercase

### How To Test

- Go to report builder
- Go to filters or compare form
- Enter custom value with uppercase letters
- Verify that string changed to lowercase
- Verify that uppercase strings can't be passed in

### To Do

- [ ] Address feedback


shrirupa (commented):
Looks like these integration tests needs to be updated too

- does not request filter options from server until user enters three or more characters
- allows any value to be added as a custom filter
- does not allow duplicate custom filters to be selected

nicklemmon:
Closing this for now - I spoke with Haley, re: https://sparkpost.atlassian.net/browse/ESCSP-4670 and it looks like our bug ticket doesn't quite address the feedback the user provided. In fact, these changes may introduce more confusion.

I'm going to add some notes to the escalation and chat with the API team to see about handling case insensitivity for Metrics endpoints.

PR #2148, opened 2021-03-28
AC-1978: Global tracking for all forms in recipientValidation, register and messageEvents

shrirupa:
AC-1978: Global tracking for all forms in recipientValidation, register and messageEvents

### What Changed

- Added hasError key with form submission data (Jon's feedback)
- Allowlisted forms of recipientValidation, register and messageEvents
- Added cypress tests for the above forms
- Added unit tests for redux forms 

### How To Test

- Check for form submissions on recipientValidation and check if all the sensitive fields are blocklisted
- Check for form submissions on register page.
    > Invite a user and click on the link in the invitation email
- Check for form submissions on Events page 

### To Do

- [ ] Address feedback


cypress/tests/integration/recipients/recipient-validation/list-page-routes/singleRoute.spec.js
```diff
@@ -115,6 +115,31 @@ describe('The recipient validation /single route', () => {
...
+      cy.wrap(event).should('eq', 'Form Submitted');
+      cy.wrap(properties).should('deep.equal', {
+        address: 'fake-email@sparkpost.com',
+        'billingAddress.country': '<filtered>',
+        'billingAddress.state': '<filtered>',
+        'billingAddress.zip': '<filtered>',
+        'card.expCombined': '<filtered>',
+        'card.name': '<filtered>',
+        'card.number': '<filtered>',
+        'card.securityCode': '<filtered>',
```
shrirupa, line 129: These are ids instead of labels because the page redirects on successful submission and the dom no longer has the label fields.

src/components/tracking/constants.js
```diff
@@ -18,9 +20,29 @@ export const FIELD_IDENTIFIER_BLOCKLIST = [
...
+  'Cardholder Name',
+  'card.name',
+  'Expiration Date',
+  'card.expCombined',
+  'Security Code',
+  'card.securityCode',
 ];
 
 export const FORM_IDS_ALLOWLIST = [
+  'mockuserform', //Added for tests
```
nstanard, line 45: When the tests run, the NODE_ENV is 'test' in case you wanted to remove`mockuserform` from the allow list and always let node_env test pass the condition:

image
image

src/components/tracking/helpers.js
```diff
@@ -82,13 +82,22 @@ export function isFieldToBeIgnored(type, node, value) {
  * @param {HTMLElement} fieldEl
  */
 export function isError(fieldEl) {
-  const { 'aria-invalid': ariaInvalid } = getElAttributes(fieldEl);
+  const { 'aria-invalid': ariaInvalid, type } = getElAttributes(fieldEl);
 
-  return ariaInvalid === 'true';
+  return ariaInvalid === 'true' || type === 'file';
```
nstanard, line 87: This was added so that getError does the actual check for file upload errors right? At a glance it reads like file types always are errors and I'm trying to think about how we would write that differently.

nstanard: One idea I'm thinking we could go with is: 

pass the result of `getElAttributes` into `isError` and check the type value in the main `getFormData` method. 
image

nstanard:
These checkbox key values are pretty long, I wonder if we need to shorten those a little. 

image

src/components/tracking/constants.js
shrirupa: sounds good!

src/components/tracking/helpers.js
shrirupa: I had it exactly the way you mentioned earlier https://github.com/SparkPost/2web2ui/pull/2154/commits/9c4bcb83e6b56ffac4e71bc2bf53057d7ead86d9 , it looked less cleaner though. I can try changing the isError a bit and add comments to clarify this.

shrirupa:
@nstanard Hi, for the label values being too long I rearranged the JSX a bit for EventTypeFilters.js. This is ready for another pass.

nstanard (approved):
This all LGTM! I'll double check the form submit blocklist as we roll this out too.

PR #2150, opened 2021-03-29
AC-1973: Global tracking for all forms in webhooks, join, onboarding and passwordReset

shrirupa:
AC-1973: Global tracking for all forms in webhooks, join, onboarding and passwordReset

### What Changed

- Allowlisted forms for webhooks, join, onboarding and passwordReset
- Blocklisted the sensitive fields from these form
- Added cypress tests for these forms

### How To Test

- Check for form submissions on webhooks and check if all the sensitive fields are blocklisted
- Check for form submissions on join and onboarding ChoosePlan page.
- Check for form submissions on password reset page 

### To Do

- [ ] Address feedback
- [ ] Removed a few unit tests from PlanPicker (they didn't seem very useful), may be they should be replaced with some new tests.


cypress/tests/integration/configuration/webhooks/createPage.spec.js
```diff
@@ -0,0 +1,52 @@
...
+    cy.findByLabelText('Webhook Name').type('new webhook');
+    cy.findByLabelText('Target URL').type('http://ajw.fakewebhook.com/test');
+    cy.findByLabelText('Primary account only').check({ force: true });
+    cy.findByLabelText('All events').check({ force: true });
+    cy.findByRole('button', { name: 'Create Webhook' }).click();
+    cy.wait('@trackEvents').then(xhr => {
+      const { event, properties } = xhr.request.body;
+      cy.wrap(event).should('eq', 'Form Submitted');
+      cy.wrap(properties).should('deep.equal', {
+        'All events': true,
```
bkemper, line 32: This is great!  How hard would it be to group the event radio values into an array like errors?  Would that make it easier to filter mixpanel?

bkemper (approved):

cypress/tests/integration/configuration/webhooks/createPage.spec.js
shrirupa: Jon mentioned it is difficult to parse the errors array in mixpanel, so may be we should avoid doing the same for radio values.

PR #2154, opened 2021-03-30
AC-1972 - Global tracking for all forms in data privacy and users

shrirupa:
AC-1972 - Global tracking for all forms in data privacy and users

### What Changed

- Added tracking for file upload errors
- Added cypress tests for data privacy and users
- Allowlisted forms for data privacy and users
- Blocklisted sensitive fields
- Fixed a small bug where hasError wasn't set correctly when the error happened only on the fields types that are to be ignored.

### How To Test

- Check if all information from data privacy is sent correctly to segment
- Check if all information from users is sent correctly to segment

### To Do

- [ ] Address feedback


nstanard (approved):
LGTM!

PR #2156, opened 2021-03-31
AC-1975 - Add Global tracking for all forms in templates

shrirupa:
AC-1975 - Add Global tracking for all forms in templates

### What Changed

- Added cypress tests for templates forms

 Note - 
 Duplicate template form and Send Test Email forms - on successful submission both don't have label values in form data in-spite of having correct label and ids in fields, that is because of the way how we handle Loading in these Modals, possible solutions could be either we change we handle loading in Modals or leave the formdata to have ids instead of labels.

### How To Test

- Test all the form submission data on all forms (there are 5 forms ) on templates page. Check if all these forms are allowlisted and all sensitive fields are allowlisted

### To Do

- [ ] Address feedback


nstanard (approved):
LGTM!

PR #2167, opened 2021-04-06
SA-2070 - (UI) Primary account can select which subaccount seedlist s/he wants to view

nicklemmon:
[SA-2070](https://sparkpost.atlassian.net/browse/SA-2070)

**NOTE:** This is based on the branch `SA-2062` and should not be merged first

### What Changed

- Adds a new subaccount selection `Panel` to the redesigned seed list page allowing admin request a seedlist specific to a particular subaccount

### How To Test

- TBD

### To Do

- [ ] Add supporting integration tests
- [ ] Test the page using a non-admin account
- [ ] Follow-up with UX, re: content that needs to be updated, missing `href` to support docs


nicklemmon:
Put this back in to "draft" - some UX tweaks are needed before we can move forward.

PR #2172, opened 2021-04-09
AC-2030

shrirupa:

### What Changed

- Filter out all field values temporarily
- skip Form submitted cypress test for now

### How To Test

- verify that all field values no matter what gets filtered (except checkboxes and radio they will be just true and false)

### To Do

- [ ] Address feedback


nstanard (approved):
lgtm!

nstanard (approved):
still lgtm

PR #2174, opened 2021-04-12
FE-1425 - fix empty state segment call going out

nstanard:
https://sparkpost.atlassian.net/browse/FE-1425

### What Changed

- Added a pre pending boolean to prevent setting empty state before the data call has been made.

### How To Test

- Load up this branch and stage side by side. 
- Visit the domains list pages while looking at segment. 
- Hard refresh the page to see if the `Empty State Loaded` event goes through.
- The `Empty State Loaded` event should not show up when there are domains returned from the api call. 

### To Do

- [ ] Feedback


src/pages/domains/components/Container.js
```diff
@@ -42,10 +42,12 @@ function mapStateToProps(state) {
     sendingDomains: selectSendingDomainsRows(state),
     sendingDomainsListError: state.sendingDomains.listError,
     bounceDomains: selectBounceDomainsRows(state),
+    sendingDomainsPrePending: state.sendingDomains.listLoading === undefined,
+    trackingDomainsPrePending: state.trackingDomains.listLoading === undefined,
```
nstanard, line 46: This works for sending domains call because `listLoading` is undefined until the call goes out but this doesn't work for tracking domains because `listLoading` defaults to false. (and that's the issue)

